PR #239, opened 2014-08-05
morebits: Ignore file/category embeds when removing links

### LINKED ISSUE #231: `Morebits.wikitext` doesn't know that [[File:]] [[Category:]] syntax are special
See https://test.wikipedia.org/w/index.php?title=Tto_page_532&diff=prev&oldid=203091. I ran unlink with only "backlinks" selected for this particular page (as can be seen from the edit summary). But `Morebits.wikitext` doesn't know that the [[File:]] syntax is a file inclusion (which is handled separately in unlink and in `Morebits.wikitext`), not a link, so it removed the file inclusion as well as correctly removing the [[:File:]] link.

Perhaps unbind File: and Category: links, or alter the regex to ignore them.

mc10:
Fixes #231. @atlight Could I get a second pair of eyes on that regex? I'm pretty sure it's correct but I might have made a simple mistake.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -3024,8 +3024,14 @@ Morebits.wikitext.page.prototype = {
 	removeLink: function( link_target ) {
 		var first_char = link_target.substr( 0, 1 );
 		var link_re_string = "[" + first_char.toUpperCase() + first_char.toLowerCase() + ']' + RegExp.escape( link_target.substr( 1 ), true );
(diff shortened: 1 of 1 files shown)


QEDK:
You could ask Magioalditis or even Reaper Eternal (since RE made many edit filters with regexes).

mc10:
I've finally come back to this with a simpler solution (no lookaheads!).

PR #249, opened 2015-01-07
Remove moment.js from repo; redundant to core

### LINKED ISSUE #248: Use moment.js from core
The moment.js library is [in core](https://gerrit.wikimedia.org/r/#/c/95555/) and need not be loaded separately by a gadget. Loading it again is causing [i18n issues](https://phabricator.wikimedia.org/T78701) with another extension.

mc10:
Fixes #248. I don't think we need some sort of `require` as the code's in core, but correct me if I'm wrong.


atlight:
We might need to add a `dependencies` directive to our gadget definition. I'll check it out.

Krinkle:
Indeed. Without adding the MediaWiki-provided module to the `dependencies`, it will not be loaded.

Please prioritise this as it is clashing in the javascript global namespace with the one that is provided by core (which is initialised for the proper user language and is also a slightly newer version of the library).

mc10:
Looks like [Amalthea already made the proper change on wiki](https://en.wikipedia.org/w/index.php?title=MediaWiki%3AGadgets-definition&diff=641771608&oldid=641604583), so I'll merge this in.

Issue #129, opened 2012-10-06
Allow internationalization of Twinkle


Huji:
I would like to suggest Twinkle to be internationalized. This will allow easier translation and updating of Twinle on non-English projects.

All we need at this point, is to store the messages in a separate file (or even in the same file, but in a separate array).

Huji:
More than 2.5 years have passed and this issue is still unresolved. Moreover, Twinkle has been modularized since then, and Gadgets 2.0 has not been released on Wikipedia. I think it is time to get serious with this.

I would like to take this bug myself. Here is what I plan to do:

1. Add a new module to Twinkle, called twinklelang.js in which I will create an object that contains the messages for each language, as well as a set of methods for that object that can retrieve the messages (or fallback to English when no translation is found).

2. Replace all hardcoded messages throughout Twinkle with references to the messages defined in twinklelang.js (although for the time being, I will NOT make console log messages internationalized; only content that is shown in the browser window (messages, link texts, etc) will be internationalized.

Specific suggestions about the above are welcome. If someone assigns this issue to me that would be great too (although you don't have to).

mc10:
If you're willing to take this on, then definitely props to you, but be warned that this will require quite a bit of effort. Style-wise I'm okay with a WordPress-style `__()` (double underscore) function to wrap around all interface strings, and let that function determine which string to display.

atlight:
Okay, a couple of suggestions:

* Start with a simple module first (e.g. twinkleshared) and show your work.
* One of the real challenges is that, to localise Twinkle, it is not only the message strings that need to be translated. The actual templates and options themselves (for example, the welcome templates, shared IP templates, list of CSD criteria, etc), with their attendant subgroups, parameters, etc. need to be localised. That is why this work hasn't been completed yet - it is a truly enormous undertaking that essentially amounts to "rewriting Twinkle".

If you are going to do this, I would suggest you have a look at the prior attempts at this: Siddhartha Ghai's "TWG" or "TW Global" (probably stored on enwiki somewhere, you may have to hunt for this one) and my initial work on this (now in the branch "twinklelocal" of this repository). It's a big challenge and it will be interesting to see your approach.

If you do end up "rewriting Twinkle", it would be nice to take the opportunity to ditch XML and switch to JSON :)

Huji:
Thanks for both comments!

@mc10 I am aware of how much work it is going to be; that is the primary reason I hesitate to get to this for over two years. I have some free time at hand now, so might as well get this done!

[...]

PR #251, opened 2015-02-15
Hide certain functions when the content model is non-normal

zhaofengli:
Basic API functions like edit, protect and delete won't work on pages
with special content models like Flow discussion boards. So let's
hide functions requiring them on such pages.

Signed-off-by: Zhaofeng Li hello@zhaofeng.li

### DIFF
--- a/modules/twinklediff.js
+++ b/modules/twinklediff.js
@@ -14,7 +14,8 @@
  */
 
 Twinkle.diff = function twinklediff() {
-	if( mw.config.get('wgNamespaceNumber') < 0 || !mw.config.get('wgArticleId') ) {
+	var ns = mw.config.get('wgNamespaceNumber');
+	if( ns < 0 || ns == 2600 || ns == 90 || !mw.config.get('wgArticleId') ) {
 		return;
 	}
 
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -16,7 +16,8 @@
 // Note: a lot of code in this module is re-used/called by batchprotect.
 
 Twinkle.protect = function twinkleprotect() {
-	if ( mw.config.get('wgNamespaceNumber') < 0 ) {
+	var ns = mw.config.get('wgNamespaceNumber');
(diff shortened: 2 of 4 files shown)


atlight:
Thanks for your contribution.

FYI, the content models currently in use on enwiki are NULL, wikitext, css, javascript, Scribunto, flow-board, and json.

As far as I can tell, Twinkle's basic functionality (unlink, RPP, diff) should mostly work fine on pages of all content models. Obviously most of what Twinkle does relies on action=edit being available and {{transclusion}} functioning normally, so I guess we have to identify how Twinkle can best interact with content models where this is not the case.

In any case, I think disabling these Twinkle actions on all non-traditional content models (particularly Scribunto, where it has always been active) is a step backwards. We should consider each module and content model individually:
* Flow doesn't appear to have traditional diffs, so it is right to disable the diff module on Flow boards. But it should be available on all other pages.
* RPP should be available on all pages, except possibly those in the MediaWiki namespace. The only parts that might need to be altered for Flow pages are (a) protection tagging - I don't know how the enwiki community has decided to handle that on Flow boards - and (b) action=protect for sysops - I don't know whether this works in the same way for Flow boards as it does for regular pages.
* CSD... well, I tried this in both user and sysop modes on a Flow board and it indeed appears not to work. However, note bug https://phabricator.wikimedia.org/T63730, which will mean that deletion should be possible in the future.
* XFD could easily be made to work by modifying the page header, which is a separate API action.

Also, if you are going to do this, the function should go in morebits.js under the Morebits object.

zhaofengli:
Okay, I've updated the commit. The namespace checks are now hard-coded into the modules themselves. I've disabled CSD/XfD for Flow and LQT as normal tagging and/or the delete API does not work there.

quiddity-wp:
@atlight re: namespace 90, it's LQT's "Thread:" per https://www.mediawiki.org/wiki/Extension_default_namespaces

Amorymeltzer:
I think there's some validity to this long-term, especially inre #129, but should probably be dealt with on a broader basis.

PR #253, opened 2015-02-17
Add GPLv3 license

### LINKED ISSUE #252: Stick a license in the GitHub repo
The repo doesn't have a license. Suggest putting one in? CC-0 or CC-BY-SA 3.0 or MIT or...whatever.

mc10:
Fixes #252.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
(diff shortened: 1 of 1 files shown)


atlight:
No, see my comment on the issue.

mc10:
Neither of the two licenses that Wikipedia uses [really apply to software](http://choosealicense.com/). In addition, there are certain benefits to dual-licensing the project as GPLv3.

atlight:
Fair enough, but if you want to dual-license it (really triple license with CC-BY-SA and GFDL - neither of which are appropriate to code, but that is irrelevant) you should explicitly state that, and also confirm that this combination is allowed (I admit to knowing very little about multi-licensing so I can't speak for that).

PR #254, opened 2015-02-21
TfD: don't enter a newline after the TfD template

martijnhoekstra:
Newlines are significant and may break the functioning of the template


atlight:
I would be happier just not inserting a newline at all.

martijnhoekstra:
done

atlight:
I'll merge this when you submit only a single commit with a nice commit message :) At the moment there are two commits on top of one another.

Thanks for noticing this and taking the time to submit a pull request.

PR #257, opened 2015-02-22
Remove |static=true ; bring back the bouncy ball!

QEDK:
TTO, you seem to have added the template under much duress. But, PamD still wants the bouncy ball. I guess, we'll give it. xD

### DIFF
--- a/modules/friendlywelcome.js
+++ b/modules/friendlywelcome.js
@@ -354,7 +354,7 @@ Twinkle.welcome.templates = {
 	"welcome-screen": {
 		description: "welcome message with clear, annotated table of 10 links",
 		linkedArticle: false,
-		syntax: "$HEADER$ {{subst:welcome-screen|static=true}}"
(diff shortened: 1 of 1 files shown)


atlight:
Well, I'm not going to merge this, that's for sure...

QEDK:
Hahaha. This was foreseeable.

Technical-13:
I'm not sure adding it here is the best thing (despite the fact that I like the bouncy ball and apply that version myself with my WikiLove ACC helper script) because then all of the people who are use to it applying without the bouncy ball will show up to complain that there shouldn't be any animation and it will simply be a back and forth mess.  I would instead suggest that those people that want it add it as a custom welcome template and be done with it.

QEDK:
There wasn't any discussion to support either of the viewpoints. So.....

atlight:
The discussion petered out, and if I remember correctly, PamD was happy with my proposed solution of using the custom welcome template feature.

PR #260, opened 2015-03-02
Block module for administrators

MusikAnimal:
Twinkle block module. Blocks, reblocks, issues block templates, but unblocking not yet supported.

The data for block templates was moved from the warn module to the block module, but the warn module is used to actually issue the warning, and also previewing the templates. Also cleaned up some of the code in the warn module, most of it solely for block module integration.

The templates are now grouped together, grouped by popularity / type and sorted alphabetically. Also updated warn module with the durationless blocking templates, and although they are not uw- prefixed or conform to the user warnings wikiproject, they are essential for the blocking module.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -1,0 +1,1065 @@
+//<nowiki>
+
+
+(function($){
+
+var api = new mw.Api(), relevantUserName;
+
+/*
+ ****************************************
+ *** twinkleblock.js: Block module
+ ****************************************
+ * Mode of invocation:     Tab ("Block")
+ * Active on:              any page with relevant user name (userspace, contribs, etc.)
+ * Config directives in:   [soon to be TwinkleConfig]
+ */
+
+Twinkle.block = function twinkleblock() {
+	// should show on Contributions pages, anywhere there's a relevant user
+	if ( Morebits.userIsInGroup('sysop') && mw.config.get('wgRelevantUserName') ) {
+		Twinkle.addPortletLink(Twinkle.block.callback, 'Block', 'tw-block', 'Block relevant user' );
+	}
+};
+
+Twinkle.block.callback = function twinkleblockCallback() {
+	if( mw.config.get('wgRelevantUserName') === mw.config.get('wgUserName') &&
+			!confirm( 'You are about to block yourself! Are you sure you want to proceed?' ) ) {
+		return;
+	}
+
+	var Window = new Morebits.simpleWindow( 650, 530 );
+	// need to be verbose about who we're blocking
+	Window.setTitle( 'Block or issue block template to ' + mw.config.get('wgRelevantUserName') );
+	Window.setScriptName( 'Twinkle' );
+	Window.addFooterLink( 'Block templates', 'Template:Uw-block/doc/Block_templates' );
+	Window.addFooterLink( 'Block policy', 'WP:BLOCK' );
+	Window.addFooterLink( 'Twinkle help', 'WP:TW/DOC#block' );
+
+	// clean up preset data (defaults, etc.), done exactly once, must be before Twinkle.block.callback.change_action is called
+	Twinkle.block.transformBlockPresets();
+
+	Twinkle.block.currentBlockInfo = undefined;
+	Twinkle.block.field_block_options = {};
+	Twinkle.block.field_template_options = {};
+
+	var form = new Morebits.quickForm( Twinkle.block.callback.evaluate );
+	var actionfield = form.append( {
+			type: 'field',
+			label: 'Type of action'
+		} );
+	actionfield.append({
+			type: 'checkbox',
+			name: 'actiontype',
+			event: Twinkle.block.callback.change_action,
+			list: [
+				{
+					label: 'Block user',
+					value: 'block',
+					tooltip: 'Block the relevant user with given options.',
+					checked: true
+				},
+				{
+					label: 'Add block template to user talk page',
+					value: 'template',
+					tooltip: 'If the blocking admin forgot to issue a block template, or you have just blocked the user without templating them, you can use this to issue the appropriate template.',
+					checked: true
+				}
+			]
+		});
+
+	form.append({ type: 'field', label: 'Preset', name: 'field_preset' });
+	form.append({ type: 'field', label: 'Template options', name: 'field_template_options' });
+	form.append({ type: 'field', label: 'Block options', name: 'field_block_options' });
+
+	form.append( { type:'submit' } );
+
+	var result = form.render();
+	Window.setContent( result );
+	Window.display();
+	result.root = result;
+
+	Twinkle.block.fetchUserInfo(function() {
+		// init the controls after user and block info have been fetched
+		var evt = document.createEvent( 'Event' );
+		evt.initEvent( 'change', true, true );
+		result.actiontype[0].dispatchEvent( evt );
+	});
+};
+
+Twinkle.block.fetchUserInfo = function twinkleblockFetchUserInfo(fn) {
+
+	api.get({
+		format: 'json',
+		action: 'query',
+		list: 'blocks|users|logevents',
+		letype: 'block',
+		lelimit: 1,
+		bkusers: mw.config.get('wgRelevantUserName'),
+		ususers: mw.config.get('wgRelevantUserName'),
+		letitle: 'User:' + mw.config.get('wgRelevantUserName')
+	})
+	.then(function(data){
+		var blockinfo = data.query.blocks[0],
+			userinfo = data.query.users[0];
+
+		Twinkle.block.isRegistered = !!userinfo.userid;
+		relevantUserName = Twinkle.block.isRegistered ? 'User:' + mw.config.get('wgRelevantUserName') : mw.config.get('wgRelevantUserName');
+
+		if (blockinfo) {
+			// handle frustrating system of inverted boolean values
+			blockinfo.disabletalk = blockinfo.allowusertalk === undefined;
+			blockinfo.hardblock = blockinfo.anononly === undefined;
+			Twinkle.block.currentBlockInfo = blockinfo;
+		}
+
+		Twinkle.block.hasBlockLog = !!data.query.logevents.length;
+
+		if (typeof fn === 'function') return fn();
+	}, function(msg) {
+		Morebits.status.init($('div[name="currentblock"] span').last()[0]);
+		Morebits.status.warn('Error fetching user info', msg);
+	});
+};
+
+Twinkle.block.callback.saveFieldset = function twinkleblockCallbacksaveFieldset(fieldset) {
+	Twinkle.block[$(fieldset).prop('name')] = {};
+	$(fieldset).serializeArray().forEach(function(el) {
+		Twinkle.block[$(fieldset).prop('name')][el.name] = el.value;
+	});
+};
+
+Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction(e) {
+	var field_preset, field_template_options, field_block_options, $form = $(e.target.form);
+
+	Twinkle.block.callback.saveFieldset($('[name=field_block_options]'));
+	Twinkle.block.callback.saveFieldset($('[name=field_template_options]'));
+
+	if ($form.find('[name=actiontype][value=block]').is(':checked')) {
+		field_preset = new Morebits.quickForm.element({ type: 'field', label: 'Preset', name: 'field_preset' });
+		field_preset.append({
+				type: 'select',
+				name: 'preset',
+				label: 'Choose a preset:',
+				event: Twinkle.block.callback.change_preset,
+				list: [{
+					label: '',
+					value: '',
+					type: 'option'
+				}].concat(Twinkle.block.callback.filtered_block_groups())
+			});
+
+		field_block_options = new Morebits.quickForm.element({ type: 'field', label: 'Block options', name: 'field_block_options' });
+		field_block_options.append({ type: 'div', name: 'hasblocklog', label: ' ' });
+		field_block_options.append({ type: 'div', name: 'currentblock', label: ' ' });
+		field_block_options.append({
+				type: 'select',
+				name: 'expiry_preset',
+				label: 'Expiry:',
+				event: Twinkle.block.callback.change_expiry,
+				list: [
+					{ label: 'custom', value: 'custom', selected: true },
+					{ label: 'indefinite', value: 'infinity' },
+					{ label: '3 hours', value: '3 hours' },
+					{ label: '12 hours', value: '12 hours' },
+					{ label: '24 hours', value: '24 hours' },
+					{ label: '31 hours', value: '31 hours' },
+					{ label: '36 hours', value: '36 hours' },
+					{ label: '48 hours', value: '48 hours' },
+					{ label: '60 hours', value: '60 hours' },
+					{ label: '72 hours', value: '72 hours' },
+					{ label: '1 week', value: '1 week' },
+					{ label: '2 weeks', value: '2 weeks' },
+					{ label: '1 month', value: '1 month' },
+					{ label: '3 months', value: '3 months' },
+					{ label: '6 months', value: '6 months' },
+					{ label: '1 year', value: '1 year' },
+					{ label: '2 years', value: '2 years' },
+					{ label: '3 years', value: '3 years' }
+				]
+			});
+			field_block_options.append({
+					type: 'input',
+					name: 'expiry',
+					label: 'Custom expiry',
+					tooltip: 'You can use relative times, like "1 minute" or "19 days", or absolute timestamps, "yyyymmddhhmm" (e.g. "200602011405" is Feb 1, 2006, at 14:05 UTC).',
+					value: Twinkle.block.field_block_options.expiry || Twinkle.block.field_template_options.template_expiry
+				});
+		var blockoptions = [
+				{
+					checked: Twinkle.block.field_block_options.nocreate,
+					label: 'Block account creation',
+					name: 'nocreate',
+					value: '1'
+				},
+				{
+					checked: Twinkle.block.field_block_options.noemail,
+					label: 'Block user from sending email',
+					name: 'noemail',
+					value: '1'
+				},
+				{
+					checked: Twinkle.block.field_block_options.disabletalk,
+					label: 'Prevent this user from editing their own talk page while blocked',
+					name: 'disabletalk',
+					value: '1'
+				}
+			];
+
+		if (Twinkle.block.isRegistered) {
+			blockoptions.push({
+					checked: Twinkle.block.field_block_options.autoblock,
+					label: 'Autoblock any IP addresses used (hardblock)',
+					name: 'autoblock',
+					value: '1'
+				});
+		} else {
+			blockoptions.push({
+					checked: Twinkle.block.field_block_options.hardblock,
+					label: 'Prevent logged-in users from editing from this IP address (hardblock)',
+					name: 'hardblock',
+					value: '1'
+				});
+		}
+
+		blockoptions.push({
+				checked: Twinkle.block.field_block_options.watchuser,
+				label: 'Watch user and user talk pages',
+				name: 'watchuser',
+				value: '1'
+			});
+
+		field_block_options.append({
+				type: 'checkbox',
+				name: 'blockoptions',
+				list: blockoptions
+			});
+		field_block_options.append({
+				type: 'textarea',
+				label: 'Reason (for block log):',
+				name: 'reason',
+				value: Twinkle.block.field_block_options.reason
+			});
+
+		if (Twinkle.block.currentBlockInfo) {
+			field_block_options.append( { type: 'hidden', name: 'reblock', value: '1' } );
+		}
+	}
+
+	if ($form.find('[name=actiontype][value=template]').is(':checked')) {
+		field_template_options = new Morebits.quickForm.element({ type: 'field', label: 'Template options', name: 'field_template_options' });
+		field_template_options.append( {
+				type: 'select',
+				name: 'template',
+				label: 'Choose talk page template:',
+				event: Twinkle.block.callback.change_template,
+				list: Twinkle.block.callback.filtered_block_groups(true),
+				value: Twinkle.block.field_template_options.template
+			} );
+		field_template_options.append( {
+				type: 'input',
+				name: 'article',
+				display: 'none',
+				label: 'Linked article',
+				value: Twinkle.block.field_template_options.article || ( Morebits.queryString.exists( 'vanarticle' ) ? Morebits.queryString.get( 'vanarticle' ) : '' ),
+				tooltip: 'An article can be linked within the notice, perhaps if it was the primary target of disruption. Leave empty for no article to be linked.'
+			} );
+		if (!$form.find('[name=actiontype][value=block]').is(':checked')) {
+			field_template_options.append( {
+				type: 'input',
+				name: 'template_expiry',
+				display: 'none',
+				label: 'Period of blocking: ',
+				value: Twinkle.block.field_template_options.template_expiry || Twinkle.block.field_block_options.expiry || $form.find('[name=template_expiry]').val() || '',
+				tooltip: 'The period the blocking is due for, for example 24 hours, 2 weeks, indefinite etc...'
+			} );
+		}
+		field_template_options.append( {
+			type: 'input',
+			name: 'block_reason',
+			label: '"You have been blocked for ..." ',
+			display: 'none',
+			tooltip: 'An optional reason, to replace the default generic reason. Only available for the generic block templates.',
+			value: Twinkle.block.field_template_options.block_reason
+		} );
+		var $previewlink = $( '<a id="twinkleblock-preivew-link">Preview</a>' );
+		$previewlink.off('click').on('click', function(){
+			Twinkle.block.callback.preview($form[0]);
+		});
+		$previewlink.css({cursor: 'pointer'});
+		field_template_options.append( { type: 'div', id: 'blockpreview', label: [ $previewlink[0] ] } );
+		field_template_options.append( { type: 'div', id: 'twinkleblock-previewbox', style: 'display: none' } );
+	}
+
+	var oldfield;
+	if (field_preset) {
+		oldfield = $form.find('fieldset[name="field_preset"]')[0];
+		oldfield.parentNode.replaceChild(field_preset.render(), oldfield);
+	} else {
+		$form.find('fieldset[name="field_preset"]').hide();
+	}
+	if (field_block_options) {
+		oldfield = $form.find('fieldset[name="field_block_options"]')[0];
+		oldfield.parentNode.replaceChild(field_block_options.render(), oldfield);
+	} else {
+		$form.find('fieldset[name="field_block_options"]').hide();
+	}
+	if (field_template_options) {
+		oldfield = $form.find('fieldset[name="field_template_options"]')[0];
+		oldfield.parentNode.replaceChild(field_template_options.render(), oldfield);
+		e.target.form.root.previewer = new Morebits.wiki.preview($(e.target.form.root).find('#twinkleblock-previewbox').last()[0]);
+	} else {
+		$form.find('fieldset[name="field_template_options"]').hide();
+	}
+
+	if (Twinkle.block.hasBlockLog) {
+		var $blockloglink = $( '<a target="_blank" href="' + mw.util.getUrl('Special:Log', {action: 'view', page: mw.config.get('wgRelevantUserName'), type: 'block'}) + '">block log</a>)' );
+
+		Morebits.status.init($('div[name="hasblocklog"] span').last()[0]);
+		Morebits.status.warn('This user has been blocked in the past', $blockloglink[0]);
+	}
+
+	if (Twinkle.block.currentBlockInfo) {
+		Morebits.status.init($('div[name="currentblock"] span').last()[0]);
+		Morebits.status.warn(relevantUserName + ' is already blocked', 'Submit query to reblock with supplied options');
+		Twinkle.block.callback.update_form(e, Twinkle.block.currentBlockInfo);
+	}
+};
+
+/*
+ * Keep alphabetized by key name, Twinkle.block.blockGroups establishes
+ *    the order they will appear in the interface
+ *
+ * Block preset format, all keys accept only 'true' (omit for false) except where noted:
+ * <title of block template> : {
+ *   autoblock: <autoblock any IP addresses used (for registered users only)>
+ *   disabletalk: <disable user from editing their own talk page while blocked>
+ *   expiry: <string - expiry timestamp, can include relative times like "5 months", "2 weeks" etc, use "infinity" for indefinite>
+ *   forAnonOnly: <show block option in the interface only if the relevant user is an IP>
+ *   forRegisteredOnly: <show block option in the interface only if the relevant user is registered>
+ *   label: <string - label for the option of the dropdown in the interface (keep brief)>
+ *   noemail: prevent the user from sending email through Special:Emailuser
+ *   pageParam: <set if the associated block template accepts a page parameter>
+ *   prependReason: <string - prepends the value of 'reason' to the end of the existing reason, namely for when revoking talk page access>
+ *   nocreate: <block account creation from the user's IP (for anonymous users only)>
+ *   reason: <string - block rationale, as would appear in the block log,
+ *            and the edit summary for when adding block template, unless 'summary' is set>
+ *   reasonParam: <set if the associated block template accepts a reason parameter>
+ *   sig: <string - set to ~~~~ if block template does not accept "true" as the value, or set null to omit sig param altogether>
+ *   summary: <string - edit summary for when adding block template to user's talk page, if not set, 'reason' is used>
+ *   suppressArticleInSummary: <set to suppress showing the article name in the edit summary, as with attack pages>
+ *   templateName: <string - name of template to use (instead of key name), entry will be omitted from the Templates list.
+ *                  (e.g. use another template but with different block options)>
+ *   useInitialOptions: <when preset is chosen, only change given block options, leave others as they were>
+ *
+ * WARNING: 'anononly' and 'allowusertalk' are enabled by default.
+ *   To disable, set 'hardblock' and 'disabletalk', respectively
+ */
+Twinkle.block.blockPresetsInfo = {
+	'anonblock' : {
+		expiry: '31 hours',
+		forAnonOnly: true,
+		nocreate: true,
+		reason: '{{anonblock}}',
+		sig: '~~~~'
+	},
+	'anonblock - school' : {
+		expiry: '36 hours',
+		forAnonOnly: true,
+		nocreate: true,
+		reason: '{{anonblock}} <!-- Likely a school based on behavioral evidence -->',
+		templateName: 'anonblock',
+		sig: '~~~~'
+	},
+	'blocked proxy' : {
+		expiry: '1 year',
+		forAnonOnly: true,
+		nocreate: true,
+		reason: '{{blocked proxy}}',
+		sig: null
+	},
+	'blocked talk-revoked-notice' : {
+		disabletalk: true,
+		reason: 'Revoking talk page access: inappropriate use of user talk page while blocked',
+		prependReason: true,
+		summary: 'Your user talk page access has been disabled',
+		useInitialOptions: true
+	},
+	'CheckUser block' : {
+		reason: '{{CheckUser block}}',
+		sig: '~~~~'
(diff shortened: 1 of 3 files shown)


atlight:
A good start :)

> Template definitions are duplicated in the block module, to some extent, that we can look into merging for better maintainability.

Yes, let's not have these in two places. Personally I would take them out of the warn module altogether and require admins to access them from the block module. It's not good to have things accessible from two places like that.

MusikAnimal:
Agreed, having it live in the block module sounds the most fitting. I'll add some comments for clarity of what all the key-value pairs are, as you'll have `summary` `reason` and `label`, possibly two separate labels

modules/twinkleblock.js
```diff
@@ -0,0 +1,879 @@
...
+// Contains the current block info in an object
+// Once filled, it will look something like:
+// { expiry: <some date>, anononly: true, ... }
+Twinkle.block.currentBlockInfo = undefined;
+
+Twinkle.block.callback = function twinkleblockCallback() {
+
+  var Window = new Morebits.simpleWindow( 650, 530 );
+  // need to be verbose about who we're blocking
+  Window.setTitle( "Block or template " + relevantUser );
```
atlight: I don't like using "template" as a verb, as it's not very clear. I'd prefer a simple title like "Block user or issue a block template"

modules/twinkleblock.js
```diff
@@ -0,0 +1,879 @@
...
+Twinkle.block.fetchUserInfo = function twinkleblockFetchUserInfo(fn) {
+
+  api.get({
+    format: 'json',
+    action: 'query',
+    list: 'blocks|users',
+    bkusers: mw.config.get('wgRelevantUserName'),
+    ususers: mw.config.get('wgRelevantUserName')
+  })
+  .then(function(data){
```
atlight: I'm not familiar with `then`. Does the callback still run even if the API query fails (e.g. network connection is down)? If so, this is going to cause JavaScript errors, because `data.query` will be undefined, and trying to access `data.query.blocks` will cause a ReferenceError.

I think this code should only run when the API query succeeds. Isn't there a `done` method, or similar?

modules/twinkleblock.js
```diff
@@ -0,0 +1,879 @@
...
+
+    field_block_options = new Morebits.quickForm.element({ type: 'field', label: 'Block options', name: 'field_block_options' });
+    field_block_options.append({ type: 'div', name: 'currentblock', label: ' ' });
+    field_block_options.append({
+        type: 'select',
+        name: 'expiry_preset',
+        label: 'Expiry:',
+        event: Twinkle.block.callback.change_expiry,
+        list: [
+          { label: 'custom', value: 'custom', selected: true },
```
atlight: Why is "custom" selected by default?

modules/twinkleblock.js
```diff
@@ -0,0 +1,879 @@
...
+        type: 'input',
+        name: 'expiry',
+        label: 'Custom expiry',
+        tooltip: 'You can use relative times, like "1 minute" or "19 days", or absolute timestamps, "yyyymmddhhmm" (e.g. "200602011405" is Feb 1, 2006, at 14:05 UTC).'
+      });
+    var blockoptions = [
+        {
+          label: 'Block account creation',
+          name: 'nocreate',
+          value: '1'
```
atlight: I don't think these `value`s are required.

modules/twinkleblock.js
```diff
@@ -0,0 +1,879 @@
...
+    if (Twinkle.block.isRegistered) {
+      blockoptions.push({
+          label: 'Autoblock any IP addresses used',
+          name: 'autoblock',
+          value: '1'
+        });
+    } else {
+      blockoptions.push({
+          label: 'Prevent logged-in users from editing from this IP address',
+          name: 'hardblock',
```
atlight: Feel free to use WP jargon in the label if you think it would be clearer. I don't deal much in blocking, but if everyone thinks of this as a "hard block", then the label in Twinkle could be "Hard block (prevent logged-in users from editing from this IP address)". Up to you.

modules/twinkleblock.js
```diff
@@ -0,0 +1,879 @@
...
+          value: '1'
+        });
+    }
+
+    blockoptions.push({
+        label: 'Watch user/IP user and talk pages',
```
atlight: What is "user/IP user"? Shouldn't "Watch user and user talk pages" be enough?

modules/twinkleblock.js
```diff
@@ -0,0 +1,879 @@
...
+ *    To disbale, set 'hardblock' and 'disabletalk', respectively
+ */
+Twinkle.block.blockPresetsInfo = {
+  'anonblock' : {
+    expiry: '31 hours',
+    forAnonOnly: true,
+    nocreate: true,
+    reason: '{{anonblock}}'
+  },
+  'anonblock – school' : {
```
atlight: Let's not use funky characters like – in Twinkle code, especially not in object keys. If you ever need to use – in strings, use the Unicode escape sequence.

modules/twinkleblock.js
MusikAnimal: I guess at Special:Block indefinite is default, but I always thought that was rather dangerous. Better to have it error out and say choose an expiry than to block indefinitely by accident, in my opinion, but I suppose it's better to stay consistent with the native block module.

modules/twinkleblock.js
```diff
@@ -0,0 +1,879 @@
...
+      { label: 'school block', value: 'school block' },
+      // { label: 'rangeblock', value: 'rangeblock' },
+      { label: 'tor', value: 'tor' },
+      { label: 'webhostblock', value: 'webhostblock' }
+    ]
+  }
+];
+
+Twinkle.block.callback.filtered_block_groups = function twinkleblockCallbackFilteredBlockGroups(show_template) {
+  return $.map(Twinkle.block.blockGroups, function(value, key) {
```
atlight: Paucity of comments and unclear parameter names...

It appears that you're  not using the keys so you could get rid of them.

modules/twinkleblock.js
```diff
@@ -0,0 +1,879 @@
...
+
+      // boolean-flipped options
+      blockoptions.anononly = blockoptions.hardblock ? undefined : true;
+      blockoptions.allowusertalk = blockoptions.disabletalk ? undefined : true;
+
+      // temporary fix for bug with block API, see [[phab:T68646]]
+      if (blockoptions.expiry === 'infinity') blockoptions.expiry = 'infinite';
+
+      // execute block
+      api.postWithToken("block", blockoptions).then(function(data) {
```
atlight: Please use Morebits.wiki.api for critical API requests, instead of hacking up your own error handling...

modules/twinkleblock.js
MusikAnimal: Same situation with the above note, I just took the copy from Special:Block. For this option though I think your suggested copy is more clear, so I'll go with that

modules/twinkleblock.js
MusikAnimal: Sorry I think my last push made all your comments outdated even though I didn't touch this line... Anyway, with `then` the first callback function is for success, the second for failure. For the latter it prints the error. It should only execute the first function when it gets a 200 response. Sounds like there's a Morebits API utility that could handle this, I'll figure out how to use it if that is preferred

modules/twinkleblock.js
```diff
@@ -0,0 +1,879 @@
+//<nowiki>
+
+
+(function($){
+
+var api = new mw.Api(), result, relevantUser = mw.config.get('wgRelevantUserName'), relevantUserName, formCleared = false;
```
atlight: I don't think any of these are required, or at least shouldn't be. e.g. no need to cache relevantUser like this.

twinkle.js
```diff
@@ -438,6 +438,7 @@ Twinkle.load = function () {
 		Twinkle.batchdelete();
 		Twinkle.batchprotect();
 		Twinkle.batchundelete();
+		Twinkle.block();
```
atlight: Block is unusual in that it is a module admins are likely to want to use frequently. The existing 6 admin-only modules are rather specialised or little-used (I don't think anyone ever uses batch undelete), hence why they are at the bottom. I would suggest to group block with the user-related modules (inside the necessary if statement).

modules/twinklewarn.js
```diff
@@ -1644,16 +1475,18 @@ Twinkle.warn.callbacks = {
...
 	preview: function(form) {
-		var templatename = form.sub_group.value;
-		var linkedarticle = form.article.value;
-		var templatetext;
-
-		if (templatename in Twinkle.warn.messages.block) {
-			templatetext = Twinkle.warn.callbacks.getBlockNoticeWikitext(templatename, linkedarticle, form.block_timer.value,
-				form.block_reason.value, Twinkle.warn.messages.block[templatename].indefinite);
+		var templatename = form.sub_group.value,
+				linkedarticle = form.article.value,
```
atlight: Why the double tab? Just one is fine.

modules/twinklewarn.js
```diff
@@ -1644,16 +1475,18 @@ Twinkle.warn.callbacks = {
...
+				templatetext;
+
+		if (templatename in Twinkle.block.blockPresetsInfo) {
+			// form.expiry.value is for block module, where expiry is the key name that the API uses
```
atlight: This is a little confusing, at first glance. Why do we need two similarly named but distinct keys?

modules/twinklewarn.js
```diff
@@ -1452,115 +1268,130 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
...
-		} else if( !e.target.form.block_reason.disabled ) {
+			$(form.block_reason).prop('readonly', false);
+		} else if( !form.block_reason.readonly ) {
 			if(Twinkle.warn.prev_block_reason === null) {
-				Twinkle.warn.prev_block_reason = e.target.form.block_reason.value;
+				Twinkle.warn.prev_block_reason = form.block_reason.value;
 			}
-			e.target.form.block_reason.disabled = true;
-			e.target.form.block_reason.value = '';
+			$(form.block_reason).prop('readonly', true);
```
atlight: Why change this?

modules/twinklewarn.js
```diff
@@ -1418,14 +1213,35 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
...
 
-	if( value === "singlenotice" || value === "singlewarn" || value === "block" ) {
+	if( value === "singlenotice" || value === "singlewarn" ) {
 		// no categories, just create the options right away
 		createEntries( Twinkle.warn.messages[ value ], sub_group, true );
 	} else if( value === "custom" ) {
 		createEntries( Twinkle.getPref("customWarningList"), sub_group, true );
 	} else {
+		var optionGroups = {};
+
```
atlight: Tabs not spaces here.

modules/twinklewarn.js
```diff
@@ -1418,14 +1213,35 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
...
+    // do a nasty transform of data from block module to format needed for warn module
+    if ( value === "block" ) {
+      for (var i in Twinkle.block.blockGroups) {
```
atlight: Don't use for .. in. Use $.each() or [].forEach() instead.

modules/twinklewarn.js
```diff
@@ -1409,7 +1204,7 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 
 			var elem = new Morebits.quickForm.element( {
 				type: 'option',
-				label: "{{" + key + "}}: " + itemProperties.label,
+				label: "{{" + key + "}}" + (!!itemProperties.label ? ": " + itemProperties.label : ""),
```
atlight: No need for the !! here.

modules/twinkleblock.js
```diff
@@ -0,0 +1,870 @@
...
+  var params = {
+    reason: form.block_reason.value || form.block_reason,
+    main_group: 'block',
+    sub_group: form.sub_group.value || form.sub_group,
+    article: form.article.value || form.article,
+    block_timer: form.block_timer.value || form.block_timer,
+    messageData: Twinkle.block.blockPresetsInfo[form.sub_group.value || form.sub_group]
+  };
+
+  if (!formCleared) {
```
atlight: I haven't closely followed the chain of logic related to `formCleared` yet, but no other Twinkle modules require this, so I am a bit sceptical of it.

modules/twinkleblock.js
```diff
@@ -0,0 +1,870 @@
...
+  Twinkle.warn.callback.change_subcategory(e);
+  if (prev_block_timer) form.expiry.value = prev_block_timer;
+};
+
+Twinkle.block.callback.evaluate = function twinkleblockCallbackEvaluate(e) {
+  var $form = $(e.target),
+      toBlock = $form.find('[name=actiontype][value=block]').is(':checked'),
+      toWarn = $form.find('[name=actiontype][value=template]').is(':checked');
+
+  if (!toBlock && !toWarn) return alert("Please give Twinkle something to do!");
```
atlight: Move this to an else after line 835

modules/twinkleblock.js
```diff
@@ -0,0 +1,870 @@
...
+
+  if (toBlock) {
+    var blockoptions = $form.find('[name=field_block_options]').serializeObject(),
```
atlight: I've never seen serializeObject before; it's apparently a non-standard jQuery extension. What is its purpose, and why is it needed here?

modules/twinkleblock.js
```diff
@@ -0,0 +1,870 @@
...
+        warnoptions = $form.find('[name=field_template_options]').serializeObject();
+    delete blockoptions.expiry_preset; // remove extraneous
+
+    if (!blockoptions.expiry) return alert('Please provide an expiry!');
+    if (blockoptions.reason || (!blockoptions.reason && confirm('No block reason was provided. Continue?'))) {
```
atlight: Block reason should be obligatory. We require delete reasons etc, even though they are technically optional in MW software.

modules/twinkleblock.js
```diff
@@ -0,0 +1,870 @@
...
+    // don't override original options if useInitialOptions is set
+    if (data.useInitialOptions && data[el.name] === undefined) return;
+    var check = data[el.name] === '' || !!data[el.name];
+    $(el).prop('checked', check);
+  });
+  $form.find('[name=reason]').val(data.reason);
+};
+
+Twinkle.block.callback.change_subcategory = function twinkleblockcallbackChangeSubcategory(e) {
+  var form = e.target.form, prev_block_timer = Twinkle.warn.prev_block_timer;
```
atlight: `prev_block_timer = Twinkle.warn.prev_block_timer`

A waste of characters!

modules/twinkleblock.js
```diff
@@ -0,0 +1,870 @@
...
+  $form.find('[name=sub_group]').val(key);
+  var preset = Twinkle.block.blockPresetsInfo[key];
+  Twinkle.block.callback.update_form(e, preset);
+  Twinkle.block.callback.change_subcategory(e);
+};
+
+Twinkle.block.callback.change_expiry = function twinkleblockCallbackChangeExpiry(e) {
+  var $expiry = $(e.target.form).find('[name=expiry]');
+  if ($(e.target).val() === 'custom') {
+    $expiry.parent().show();
```
atlight: Consider using the utility method `Morebits.quickForm.setElementVisibility` instead.

modules/twinkleblock.js
```diff
@@ -0,0 +1,870 @@
...
+        list: list
+      };
+  });
+};
+
+Twinkle.block.callback.change_preset = function twinkleblockCallbackChangePreset(e) {
+  var $form = $(e.target.form), key = $form.find('[name=category]').val();
+  if (!key) return;
+
+  $form.find('[name=sub_group]').val(key);
```
atlight: Is this some sort of hack to "integrate" block and warn? I don't really like it. 

As a general comment, I think creating a messy "web of interdependence" between warn and block is the wrong way to go. Blocks are special-cased in warn half the time anyway; I really think you should just reimplement the warning logic in the block module, to avoid calling back and forth and passing hackish hybrid data objects around.

I'm open to being convinced that the templating logic has enough in common with the other stuff in warn that it can stay there, but you'll have to persuade me.

modules/twinkleblock.js
```diff
@@ -0,0 +1,870 @@
...
+
+/* NOTE: warn module references these, some properties are only used by
+ *    that module, others only used by block module
+ * Keep alphabetized by key name, Twinkle.block.blockGroups establishes
+ *    the order they will appear in the interface
+ *
+ * WARNING: 'anonblock' and 'allowusertalk' are enabled by default.
+ *    To disbale, set 'hardblock' and 'disabletalk', respectively
+ */
+Twinkle.block.blockPresetsInfo = {
```
atlight: This needs a comment documenting ALL the available keys, and what values they can take. It's true that some parts of the warn and protect modules are lacking such a comment, but that is because the maintainers (over the past few years, mainly me) know what's available. On the other hand, this is something new and complex. See the top of twinkleconfig.js for an example.

modules/twinkleblock.js
atlight: Why not have something immediately useful, like "31 hours", as default? In fact I thought that was default (not being an admin myself)...

atlight: I should also note that the intention with protect (and other sysop modules, like CSD) was to *improve* on the built-in functionality, not simply to replicate it. I don't know if it succeeds in that goal, but that was the idea. So don't feel obliged to precisely replicate core functionality into Twinkle.

modules/twinklewarn.js
MusikAnimal: I actually don't know what I was trying to say here in the comment. Warn wants `block_timer`,
where with Block we use `expiry` as that's what the API wants. However we also want `block_timer` because you can also use Block to warn (issue block template). They are separate because I wanted to give the ability to provide any duration string value for the template (e.g. "temporarily", "a fortnight", etc.), however I thought it was weird to have two duration input fields in the form. On second thought maybe they should both just be visible.

modules/twinklewarn.js
MusikAnimal: Values of disabled fields aren't submitted with the form, but they are with readonly and we can still disable user input. The idea is the checkboxes for the block options directly map to API parameters.

modules/twinklewarn.js
MusikAnimal: When I make my next push the tabs will fix themselves. The tab settings in my text editor made them look the same, anyway I'm converting all leading spaces to tabs as it appears tabs is the convention with the Twinkle codebase.

modules/twinklewarn.js
atlight: I haven't tested this yet, but with readonly, don't you lose the UI affordance (i.e. graying out of the form field)?

In any case, isn't it a moot point, since we're not actually "submitting" the form to a GET or POST endpoint (a la plain old HTML), but we're simply reading off the values in JS?

modules/twinkleblock.js
```diff
@@ -0,0 +1,870 @@
...
+    reason: '{{blocked proxy}}'
+  },
+  'CheckUser block' : {
+    reason: '{{CheckUser block}}'
+  },
+  'checkuserblock-account' : {
+    reason: '{{checkuserblock-account}}'
+  },
+  'checkuserblock-wide' : {
+    reason: '{{checkuserblock-wide}}'
```
atlight: Is this a stylistic variant of {{checkuser block}}? I'm sceptical of including variant templates like this. I did a big prune of welcome templates a while ago, and it was nice to weed out some substantially similar templates that differed only in appearance.

modules/twinklewarn.js
MusikAnimal: Yes and yes, "submit" was not the right word. I'm using $.serializeObject to get the form data and hand it off to the API (with a few transformations), and I guess it ignores the disabled fields. We can style the input fields with CSS if we want, or I guess I could enable the fields right before I call $.serializedObject. Probably go with the latter

modules/twinkleblock.js
MusikAnimal: Explained above, the hope was to give the block option checkboxes the same names as the API parameters, call this function and just hand over the data. Obviously some transformations had to be made, namely the annoying boolean inversion with `anononly` and `allowusertalk`. I actually didn't realize it wasn't part of core jQuery. I could rewrite this to map out the values with `serializeArray`

modules/twinkleblock.js
MusikAnimal: Just habit of mine, here it indeed doesn't really simplify anything

modules/twinkleblock.js
MusikAnimal: Well, not quite. With Block we have separate fieldsets for the warn and block options, so we keep the same field names in the warn fieldset so we can just pass it to warn to do the actual warning. All that's happening on this line is if a preset is chosen it will choose the appropriate template for you along with the associated block options. I'm not sure if that clears it up or if your concern goes beyond that. The "nasty transform" of the `blockGroups` in Warn (line 1224) is because unfortunately Warn does not utilize the `.append({ type: 'select', list: [...array of option objects...] })` functionality Morebits provides, and instead builds the `<optgroup>`s and `<option>`s from it's own data structure format, one that's not easy to manage given how many properties we have on the block templates. Hopefully that all makes sense. Essentially my foremost concern with implementation was to avoid repetition and share functionality where possible.

modules/twinkleblock.js
MusikAnimal: The block options themselves and acceptable values are the same names as those used in the API, except for the two frustratingly inverted-value parameters mentioned in the comment that had to require special handling. Beyond that documentation is certainly lacking.

modules/twinkleblock.js
MusikAnimal: I don't believe so, all of the "templated reasons" were taken from [[MediaWiki:Ipbreason-dropdown]]. They do not actually conform to the WikiProject Warnings stewardship which is probably intentional.

modules/twinkleblock.js
MusikAnimal: Oh we're definitely making an improvement :) The concern I had was you don't want to be too quick to block. Not supplying a default value helps prevent mistakes, which do happen with the built-in interface. However if you chose a _preset_ it will provide the options for you, such as the standard 31 hours for an anonblock, 24 for edit warring, etc. So for everyday admin use, we're looking at one/two more clicks to chose your preset and/or expiry than if we supplied a default, but then we're saving about 10 or so clicks plus page loads times after that when factoring in issueing warnings. In the end we're still saving oceans of time.

modules/twinkleblock.js
```diff
@@ -0,0 +1,870 @@
...
+      formCleared = true;
+      blockoptions.action = 'block';
+      blockoptions.user = relevantUser;
+
+      // boolean-flipped options
+      blockoptions.anononly = blockoptions.hardblock ? undefined : true;
+      blockoptions.allowusertalk = blockoptions.disabletalk ? undefined : true;
+
+      // temporary fix for bug with block API, see [[phab:T68646]]
+      if (blockoptions.expiry === 'infinity') blockoptions.expiry = 'infinite';
```
atlight: Task T68646 has just been marked resolved.

MusikAnimal: From what I've been told it will be deployed with 1.25wmf22, due to released on all wikis 25 March https://www.mediawiki.org/wiki/MediaWiki_1.25/Roadmap

modules/twinkleblock.js
```diff
@@ -0,0 +1,883 @@
...
+		expiry: 'infinity',
+		forRegisteredOnly: true,
+		nocreate: true,
+		pageParam: true,
+		reason: '[[WP:Vandalism-only account|Vandalism-only account]]',
+		summary: 'You have been indefinitely blocked from editing because your account is being [[WP:VOA|used only for vandalism]]'
+	}
+};
+// supply default edit summaries where necessary
+$.each(Twinkle.block.blockPresetsInfo, function(preset, settings) {
```
atlight: This code is going to run every time anyone (including non-admins) loads Twinkle, whether or not they are going to use the block module. Why can't this just be done on the fly in the place where this data is actually used?

atlight:
Looks very good - I'm most impressed. Just a couple of minor things before I actually try it out for myself:

* There are a couple of bits of commented-out code. What are they for? It's OK to leave commented-out code in Twinkle if you need to (e.g. if you are waiting for a MediaWiki feature to become available), but you should at least explain why it is commented.
* The blankTalkpageOnIndefBlock pref should be moved to a new "Block" section in twinkleconfig.js. I hope modifying that file is fairly self-explanatory.
* File needs to end with a single blank line to make Git happy :)

MusikAnimal:
All done! The module should be fully working on test-wiki. Thanks

atlight:
Thanks, just need to remove "Blocking (admin only)" from `defaultWarningGroup` in twinkleconfig. Also mark the new Block section as "adminOnly" (no sections are marked "adminOnly" currently, but the functionality is there and looks like it should work).

Sorry I've been so frustrating to deal with! I can see you've done some really good and dedicated work, and it's now up to me to push the button. I promise I will test this as soon as I have time!

MusikAnimal:
Thank YOU TTO! Again there was a fundamental misunderstanding from the beginning. I thought we wanted to share functionality between Block and Warn, when you had actually said the opposite. I simply misread. This caused things to take a little longer than expected, but I'm not all upset about it and all along the way it has been a pretty fun project to work on.

Note my last commit (I'll squash when we're ready). I got rid of the `indefinite` parameters for the block templates that were used in Warn. I don't think we need to know if it's an indefinite template or not (`uw-voablock`

MusikAnimal:
Thank YOU TTO! Again there was a fundamental misunderstanding from the beginning. I thought we wanted to share functionality between Block and Warn, when you had actually said the opposite. I simply misread. This caused things to take a little longer than expected, but I'm not all upset about it and all along the way it has been a pretty fun project to work on.

Note my last commit (I'll squash when we're ready). I got rid of the `indefinite` parameters for the block templates that were used in Warn. I don't think we need to know if it's an indefinite template or not (`uw-voablock` vs `uw-vblock`) as for those if a `time` param is supplied it is simply ignored. So you won't end up with things like `you have been blocked for infinity from editing...`. For the non-indef templates the modifications I made will look for any valid indefinite-like expiry and have it add the `indef=yes` param.

atlight:
This keeps coming into my head in the most useless places, such as when I'm on the bus or (like now) when I'm about to go to sleep. Hopefully I'll get to it very soon.

modules/twinkleblock.js
```diff
@@ -0,0 +1,1064 @@
...
+};
+
+Twinkle.block.blockGroups = [
+	{
+		label: 'Common block reasons',
+		list: [
+			{ label: 'anonblock', value: 'anonblock' },
+			{ label: 'anonblock - likely a school', value: 'anonblock - school' },
+			{ label: 'school block', value: 'school block' },
+			{ label: 'Generic block', value: 'uw-block' },
```
atlight: Maybe append `(custom reason)` to the description of the "Generic block" options to remind admins to type a custom block reason.

modules/twinkleblock.js
```diff
@@ -0,0 +1,1064 @@
...
+	{
+		label: 'Templated reasons',
+		list: [
+			{ label: 'blocked proxy', value: 'blocked proxy' },
+			{ label: 'CheckUser block', value: 'CheckUser block' },
+			{ label: 'checkuserblock-account', value: 'checkuserblock-account' },
+			{ label: 'checkuserblock-wide', value: 'checkuserblock-wide' },
+			{ label: 'colocationwebhost', value: 'colocationwebhost' },
+			{ label: 'oversightblock', value: 'oversightblock' },
+			// { label: 'rangeblock', value: 'rangeblock' },
```
atlight: Why is this here?

modules/twinkleblock.js
```diff
@@ -0,0 +1,1064 @@
...
+				list: Twinkle.block.callback.filtered_block_groups(true),
+				value: Twinkle.block.field_template_options.template
+			} );
+		field_template_options.append( {
+				type: 'input',
+				name: 'article',
+				display: 'none',
+				label: 'Linked article',
+				value: Twinkle.block.field_template_options.article || ( Morebits.queryString.exists( 'vanarticle' ) ? Morebits.queryString.get( 'vanarticle' ) : '' ),
+				tooltip: 'An article can be linked within the notice, perhaps because it was a revert to said article that dispatched this notice. Leave empty for no article to be linked.'
```
atlight: This tooltip could be rewritten to be more specific to the block module.

modules/twinkleblock.js
```diff
@@ -0,0 +1,1064 @@
...
+	},
+	'colocationwebhost' : {
+		expiry: '1 year',
+		forAnonOnly: true,
+		reason: '{{colocationwebhost}}',
+		sig: null
+	},
+	'oversightblock' : {
+		reason: '{{OversightBlock}}',
+		sig: '~~~~'
```
atlight: It should never be necessary to pass signatures as a parameter to the template anymore. A `sig=yes` parameter is fine, but `sig=~~~~` is anachronistic. If the templates do in fact still require this, please take the time to fix the templates.

modules/twinkleblock.js
```diff
@@ -0,0 +1,1064 @@
...
+			{ label: 'Username represents a non-profit, soft block', value: 'uw-causeblock' },
+			{ label: 'Username violation, vandalism-only account', value: 'uw-vaublock' }
+		]
+	},
+	{
+		label: 'Templated reasons',
+		list: [
+			{ label: 'blocked proxy', value: 'blocked proxy' },
+			{ label: 'CheckUser block', value: 'CheckUser block' },
+			{ label: 'checkuserblock-account', value: 'checkuserblock-account' },
```
atlight: Why don't these have proper descriptions?

atlight:
When I tested the module just now, I got the following Status output:

```
Info: Executing block
Info: Blocked User:TTO's apostrophe
User talk page modification: completed (User talk:TTO's apostrophe)
Info: Will create a new level 2 heading for the date, as none was found for this month
Warning complete, loading talk page in a few seconds: completed
```

Instead of the first two lines, there should be a line in blue, something like "Executing block: processing..." which changes to "Executing block: completed" in green when done. Friendly used to work this way, spitting out green status lines like this, but it's a bit confusing and shouldn't be done.

The "Warning complete" bit at the bottom is inaccurate and should probably be changed to something like "Actions complete".

Also consider changing "talk page" to "user talk page"

atlight:
It also doesn't seem possible to unblock users using this module. I suppose unblocking isn't really done that often, but it does seem like an obvious omission.  Was this an intentional choice?

atlight:
And also, please squash your commits before I merge it.

modules/twinkleblock.js
MusikAnimal: We can remove it you'd prefer. It's a placeholder for if and when we implement the ability to do rangeblocks. Of all the block templates this is the only one not included.

atlight: Maybe add an explanatory comment that states this in the code itself.

modules/twinkleblock.js
MusikAnimal: This just matches [MediaWiki:Ipbreason-dropdown] (https://en.wikipedia.org/wiki/MediaWiki:Ipbreason-dropdown) and is what the admins are used to seeing. For procedural blocks like these it's common to use `{{block template}}` as the block reason and edit summary.

atlight: Fair enough. Still, I like to see Twinkle improve wherever it can on what 
the default MediaWiki interface offers, not simply copy it, so it would be 
nice to give a more friendly description here.

modules/twinkleblock.js
MusikAnimal: These block templates serving as "templated reasons" are the only ones that don't conform to the stewardship of the user warnings WikiProject. I could have them accept `sig=yes` while also still allowing for `sig=~~~~`, but I wonder the value given all the other standard parameters aren't supported either.

MusikAnimal:
I could try to quickly add unblock... How would you envision it in the interface? Maybe have it as a third checkbox that when clicked, disables the other two checkboxes and hides the Preset, Template options, and Block options fieldsets, instead providing a single textarea for the unblock reason. How does that sound?

There's other ideas I have planned for a future update as well. Namely, there should be an option to place the template blocks (namely `{{anonblock}}` and `{{school block}}`) at the top of the IP talk page, without a heading, as this is fairly common.

modules/twinkleblock.js
MusikAnimal: The ipbreason-dropdown is editable by the admins, so I believe the appearance is intentional. While the other blocks have actual descriptions, these are nested under the heading "templated reasons" and appears as code, e.g. `{{anonblock}} <!-- likely a school based on behaviourial evidence -->` when obviously that code won't be rendered in the block log. I don't really know why it is done this way, it just is :) We as admins know what they do, and the checkusers in particular seemingly always use templated block reasons.

MusikAnimal:
@atlight Disregard my idea about how to display the unblock option. A better solution would be to first check if the user is blocked, and if they are, show the two radio buttons "Re-block user" and "Unblock user". Perhaps also a checkbox for add block template indented under "Re-block user", unchecked by default, as I think we usually wouldn't add another template if they are already blocked.

Either way this is a lot more work, and if you don't mind I'd like to save it for a future version. Most unblocking is done when responding to an unblock request, and [Template:Unblock](https://en.wikipedia.org/wiki/Template:Unblock) has a link to [[Special:Unblock/username]].

What we have now is still an immensely useful script that I am eager to have other admins enjoy. I've squashed the commits and addressed all other concerns. Thanks!

atlight:
Alright, let's do this! Once again, I'm very sorry that this has taken so long. I'll try to be more responsive to any future updates or improvements you may wish to add. Or alternatively I could ask for you to be added as a repository collaborator.

Thanks for all your hard work and I'm sure administrators will appreciate the ease of use your new module will provide.

Are you comfortable with running Perl scripts? If so, you are welcome to use our `sync.pl` script to synchronise the on-wiki gadget with the new script. I've never used it myself, but there are some instructions (possibly outdated) in the README.md file. If you plan to use it, it might be a good idea to read through the script to see if it still works as advertised.

MusikAnimal:
Yayyy!!! Thank you TTO! It's been a pleasure working with you. Becoming a collaborator sounds very inviting. As I've said, I have more updates planned not only for Block but for Protect and it would be cool to help out with feature/bug requests, and to just be a part of some as integral to the project as Twinkle. I'd run all my work by you first however, as I think this experience has shown you are quite a proficient code reviewer :)

Anyway, as for deployment: I take it you don't use the "alltwinkle.js" solution anymore, since ResourceLoader does the concatenation for us. There's only three files involved for my PR, do you think it would be okay to manually update? I've never worked with Perl, but I believe I could figure out how to use it if needed. I also see that the source needs to be updated to sync the new "twinkleblock.js". Finally, I don't see a way to set the version number. Where is that being tracked? Would this be considered a major release? Thanks!

atlight:
No, we don't use alltwinkle.js anymore - it arguably made deployment simpler, but eventually we started to cause WMF servers to run out of memory, because the JavaScript minifier couldn't deal with such a long file...

You missed one very important thing - actually calling the module initialisation function! I handled this in d3e0f94cc4ed9f66aafc7db6c34e3dc77d790381.

We don't have a version number as far as I know. "Version 2.0" simply refers to Twinkle post-June 2011 after the big rewrite that replaced screen-scraping with full usage of the API. The Version 2.0 development effort was why this repository was created to begin with.

Manually updating should be OK. You'll need to update our entry in MediaWiki:Gadgets-description to include `twinkleblock.js` though (it should be reasonably obvious what to do there). If I were you, I'd post on WP:AN and WT:TW after you've made the relevant changes, just for a bit of added publicity (and also to let admins know where the "blocking" section of the Warn module has gone).

If you'd like to become a collaborator, I'm sure @AzaToth will be willing to push the button. Be sure to carefully test all changes before committing them to the master branch, though. We always aim to keep master in a state that is ready to be synchronised with the on-wiki gadget files at any minute.

MusikAnimal:
Doh! I had "twinkle.js" updated at some point, not sure what happened! I even remember you code reviewing it. Thank you for catching that.

I'm did the manual update, but the Block link isn't showing up on user and user talk pages, only on the contributions page. The block module is definitely there, as I can use `Twinkle.block()` in the JS console to get it to appear. The conditional in that file is `Morebits.userIsInGroup('sysop') && mw.config.get('wgRelevantUserName')` which evaluates to true on user and user talk pages... so I'm guessing maybe there's a load time issue where `mw.config.get('wgRelevantUserName')` isn't returning the right thing, or maybe even Morebits hasn't fully loaded?

MusikAnimal:
Looks like it was just me... inexplicably. Going to figure that out tonight, along with fixed a bug and whatever other concerns the admins discovered. So updates soon to come! Thank you TTO!

Issue #267, opened 2015-05-28
Uncaught TypeError on "Rollback failed" page


MrStradivarius:
I'm getting an uncaught TypeError when native rollback fails and I am taken to the "Rollback failed" page. This was after I tried to rollback one of my own rollbacks in my sandbox.

URL: https://en.wikipedia.org/w/index.php?title=User:Mr._Stradivarius/Sandbox&action=rollback&from=Mr.+Stradivarius&token=(token)

(That specific link won't work, as the token will fail.)

Stack trace:

> Uncaught TypeError: Cannot read property 'indexOf' of undefinedtwinklewarn @  VM793:421
> Twinkle.load @ VM793:398jQuery.Callbacks.fire @ load.php?debug=false&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=vector&version=20150520T2…:45
> jQuery.Callbacks.self.add @ load.php?debug=false&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=vector&version=20150520T2…:45
> jQuery.fn.ready @ load.php?debug=false&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=vector&version=20150520T2…:49
> jQuery.fn.init @ load.php?debug=false&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=vector&version=20150520T2…:41
> jQuery @ load.php?debug=false&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=vector&version=20150520T2…:1
> (anonymous function) @ VM793:398jQuery.Callbacks.fire @ load.php?debug=false&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=vector&version=20150520T2…:45
> jQuery.Callbacks.self.fireWith @ load.php?debug=false&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=vector&version=20150520T2…:46
> done @ load.php?debug=false&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=vector&version=20150520T2…:135
> jQuery.ajaxTransport.send.callback @ load.php?debug=false&lang=en&modules=jquery%2Cmediawiki&only=scripts&skin=vector&version=20150520T2…:141

It looks like the problem is [in twinklewarn.js](https://github.com/azatoth/twinkle/blob/master/modules/twinklewarn.js#L23); the jQuery selector "#mw-rollback-success" is assumed to exist, but it doesn't appear on rollback failure pages.

[...]

PR #268, opened 2015-05-28
Check for $vandalTalkLink existence

MrStradivarius:
On "Rollback failed" pages, the selector $vandalTalkLink doesn't match any elements, causing an error when we do `href.indexOf("?")` on line 30. See issue #267.


mc10:
Thanks!

PR #270, opened 2015-06-02
Remove old RFPP code

MrStradivarius:
Remove code that is no longer needed now that RFPP has switched to the new format.


atlight:
Thanks!

PR #271, opened 2015-06-10
Block modue bug fixes/requests from feedback, update config

MusikAnimal:
This is in response to feedback given at [WP:AN](https://en.wikipedia.org/wiki/Wikipedia:Administrators%27_noticeboard#New_Twinkle_block_module.21). Changes include:

* Don't autofill the article field with the value of the `vanarticle` URL parameter, as this may be less relevant for a block than when issuing a normal warning
* Do not show fields for duration, article, etc. for templates such as `{{anonblock}}` that do not conform to standards outlined by WikiProject User Warnings (as these params will likely not work)
* When issuing only a template, Leave the duration field blank and not supplied with expiry of an existing block
* When issuing only a template, add option to add `notalk=yes` to template
* When blocking and templating, add option to exclude the duration from the template altogether so that it reads "you have been blocked from editing temporarily..."
* When blocking and templating, make it so that `notalk=yes` is added to the block template if "Prevent this user from editing their own talk page while blocked" is checked
* Remove one last remaining reference to the block option in the Warn module from twinkleconfig.js

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -350,6 +350,11 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 		Morebits.status.warn(relevantUserName + ' is already blocked', 'Submit query to reblock with supplied options');
 		Twinkle.block.callback.update_form(e, Twinkle.block.currentBlockInfo);
 	}
+
+	if ($form.find('[name=actiontype][value=template]').is(':checked')) {
+		// make sure right fields are hidden based on default template
+		Twinkle.block.callback.change_template(e);
+	}
 };
 
 /*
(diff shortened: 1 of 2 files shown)


MusikAnimal:
@atlight Mind looking this over?

atlight:
hm, I thought I left a comment. Please split this into piecemeal commits with clearer commit messages that explain what is being fixed in each commit. Use the commit message format `block: Xyz xyz xyz` (or e.g. `config: Xyz xyz xyz` if fixing something in the config module etc  etc).

MusikAnimal:
@atlight done!

MusikAnimal:
Not done, actually.

atlight:
Let me know when I can do this.

MusikAnimal:
@atlight Okay I think I've got everything fixed, and also addressed more concerns based on feedback. I hope the commits are individualized enough, it's difficult to redo them without breaking something. Thanks!

atlight:
I'll merge without testing; I trust you :)

PR #272, opened 2015-06-11
protect: add preset for unsourced content and disruptive editing

MusikAnimal:
Next to persistent vandalism/sockpuppetry, these are the two most common semi-protection reasons, frustratingly missing from Twinkle. Surely other admins have gotten tired of having to either manually type in this rationale or use the native interface, where these options are available.

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -641,6 +641,7 @@ Twinkle.protect.protectionTypes = [
 			{ label: 'Generic (semi)', value: 'pp-semi-protected' },
 			{ label: 'Persistent vandalism (semi)', selected: true, value: 'pp-semi-vandalism' },
 			{ label: 'Persistent disruptive editing (semi)', value: 'pp-semi-disruptive' },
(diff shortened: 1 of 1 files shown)


modules/twinkleprotect.js
```diff
@@ -640,6 +640,8 @@ Twinkle.protect.protectionTypes = [
 		list: [
 			{ label: 'Generic (semi)', value: 'pp-semi-protected' },
 			{ label: 'Persistent vandalism (semi)', selected: true, value: 'pp-semi-vandalism' },
+			{ label: 'Persistent disruptive editing (semi)', value: 'pp-semi-disruptive' },
+			{ label: 'Persistent addition of unsourced content (semi)', value: 'pp-semi-unsourced' },
```
atlight: Keep it short. "Disruptive editing" and "Adding unsourced content" should be enough. The admin will see the actual protection log reason for themselves when they scroll down to have a look.

MusikAnimal:
@atlight done!

MusikAnimal:
Thank you! :)

PR #273, opened 2015-06-11
protect: fix bug using stabilizeValues that prevented salting of pages

MusikAnimal:
Looks like `form.pclevel` doesn't exist outside `stabalizeIt` when attempting to salt a page. It errors out at `form.pclevel.value`. Since `stablizeValues` is only used within `stabalizeIt` I just moved it there and all seems to work fine. I don't know why this bug affected only salting of pages (protecting nonexistent/deleted pages), and not protecting of existing pages.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -141,11 +141,7 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 				name: 'preset',
 				label: 'Choose a preset:',
 				event: Twinkle.block.callback.change_preset,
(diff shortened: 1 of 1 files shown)


atlight:
Can't test right now (I'm on the move) but I think this will fail if applying both classic protection and PC protection at once. See my recent change to protect, which probably introfuced the bug you're trying to fix here, while fixing another.

MusikAnimal:
Closing. I can take a stab at this later but sounds like I'm not the best fit for the job anyway

PR #276, opened 2015-06-17
More updates to block module based on feedback

MusikAnimal:
* Make generic block the default preset when module is first loaded
* Moved disruptive editing to the list of common block reasons
* Add support for page parameter to edit warring block template
* Fixed typo

Sorry they're not separate commits, I keep forgetting!

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -141,11 +141,7 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 				name: 'preset',
 				label: 'Choose a preset:',
 				event: Twinkle.block.callback.change_preset,
-				list: [{
-					label: '',
-					value: '',
-					type: 'option'
-				}].concat(Twinkle.block.callback.filtered_block_groups())
+				list: Twinkle.block.callback.filtered_block_groups()
 			});
 
 		field_block_options = new Morebits.quickForm.element({ type: 'field', label: 'Block options', name: 'field_block_options' });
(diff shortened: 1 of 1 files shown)


atlight:
Please improve the commit message, per my comment somewhere else:

> We don't mind long commit messages; see ecfff3b for an example. (That's an extreme example but you get the picture, I think.)

It's OK to have a few minor changes in the one commit - after all I make commits sometimes that are called "Miscellaneous ameliorations" - but please take the time to explain them in a multiline commit message. That will help us in the future when we look back and wonder, it will help people like Jimmy Xu who are downstream from us, and also it is useful if bugs are found in the commit later.

MusikAnimal:
@atlight done! I hope :)

atlight:
Are you aware that you can make multi-line commit messages?

I think it would help if I wrote an example commit message for this commit:

```
block: Various minor updates           [or something similar. But try to stay under 80 chars]
                                       [blank line here]
* Make "generic block" the default preset when the block dialog is loaded
* Move "disruptive editing" to the common block reasons section, by popular request at WT:TW
                                       [adding a reason why you make the change can be helpful]
* Add support for page parameter to {{uw-ewblock}}
* Fix typo in edit summary of {{uw-aoablock}}
```

Bear in mind that that is only an example. I just want to demonstrate how it could be done.

Sorry for making such a fuss over what may seem like a trivial detail; it takes a little while to get used to writing good commit messages, but once you get it, it should start to come more naturally to you. I promise I won't keep going on about it!

atlight:
The code itself is fine, so to avoid unnecessary delay, I'll merge in spite of my previous comment. But please bear what I said in mind for future commits.

MusikAnimal:
Sure, thank you! I will be sure to be very descriptive moving forward. The thing is the long commit messages differ from what I'm used to doing professionally. That's not to say that my employer's system is better, I just do what I'm told :) Cheers

PR #279, opened 2015-06-23
block: better handling when blocking IPs, move attack-only account block to personal attacks block

MusikAnimal:
block: add |anon=yes to template when blocking IPs

uw-aoablock (attack-only account) is now uw-pablock (personal attacks block) and allowed for anonymous users

When blocking an IP, use 31 hour duration on presets that are indefinite for accounts


MusikAnimal:
going to add another commit or to

PR #281, opened 2015-06-28
Fixes for indefinite duration presets, only set anon=yes for soft blocks

MusikAnimal:
A rather **urgent pull request**... a careless mistake with the last pull request caused indefinite presets to have default durations (e.g. vandalism-only preset sets block duration to 31 hours). This is easily solved by not transforming the block presets data until after we've got the current block information, if present.

Also omitting the `anon=yes` parameter for hard blocks of IPs. Despite what the documentation says, the copy that the template generates makes it clear this should be used only for soft blocks.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -35,9 +35,6 @@ Twinkle.block.callback = function twinkleblockCallback() {
 	Window.addFooterLink( 'Block policy', 'WP:BLOCK' );
 	Window.addFooterLink( 'Twinkle help', 'WP:TW/DOC#block' );
 
-	// clean up preset data (defaults, etc.), done exactly once, must be before Twinkle.block.callback.change_action is called
-	Twinkle.block.transformBlockPresets();
-
 	Twinkle.block.currentBlockInfo = undefined;
 	Twinkle.block.field_block_options = {};
 	Twinkle.block.field_template_options = {};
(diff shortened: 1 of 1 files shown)


modules/twinkleblock.js
```diff
@@ -982,13 +982,14 @@ Twinkle.block.prev_reason = null;
...
-		template: form.template.value,
 		article: form.article.value,
-		expiry: form.template_expiry ? form.template_expiry.value : form.expiry.value,
-		indefinite: !!(/indef|infinity|never|\*|max/).exec( form.template_expiry ? form.template_expiry.value : form.expiry.value ),
 		blank_duration: form.blank_duration ? form.blank_duration.checked : false,
 		disabletalk: form.disabletalk.checked || (form.notalk ? form.notalk.checked : false),
-		reason: form.block_reason.value
+		expiry: form.template_expiry ? form.template_expiry.value : form.expiry.value,
+		hardblock: Twinkle.block.isRegistered ? form.autoblock.checked : form.hardblock.checked,
+		indefinite: !!(/indef|infinity|never|\*|max/).exec( form.template_expiry ? form.template_expiry.value : form.expiry.value ),
```
mc10: If you want just `true`/`false` you could just use `.test` instead and omit the `!!`.

MusikAnimal: thank you, I've made this change

atlight:
Seems legit :)

I emailed AzaToth to see if he would add you as a repository collaborator. Just a word of warning though - if you do make commits directly to the repository, be very careful not to introduce any syntax errors. If there are syntax errors in one module, it can break all of Twinkle (because the modules are concatenated). Before I merge your pull requests (or anyone's for that matter) or make a commit myself, I always copy-paste the modified file(s) into a web browser console and execute it to make sure there are no syntax errors.

MusikAnimal:
Thank you! @atlight I use jshint with Vim so I'm told about syntax errors immediately after they are made :)

mc10:
I thought we no longer concatenated the Twinkle modules as the concatenated file was growing too large, so one module failing shouldn't cause all of them to fail. (Having morebits or one of the core files fail would still cause everything to not load, however.)

atlight:
What you say is correct, but I think ResourceLoader ends up concatenating the minified script files for delivery to the user.

mc10:
Ahh that makes sense, thanks.

PR #287, opened 2015-08-02
protect: force default-enabled <noinclude> wrapping of prot template …

MusikAnimal:
When protecting a template, make sure the option to wrap the protection template with `<noinclude>` is always checked by default, and not just when performing template-protection.

We semi-protect templates a lot, and many of us, including myself (sad face) forget to tick the `<noinclude>` checkbox, and then bad things happen.

The `checked: (mw.config.get('wgNamespaceNumber') === 10)` on line 500 gets overridden in the `changePreset` event handler on line 933

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -929,7 +929,7 @@ Twinkle.protect.callback.changePreset = function twinkleprotectCallbackChangePre
 			if( /template/.test( form.category.value ) ) {
 				form.noinclude.checked = true;
 				form.editexpiry.value = form.moveexpiry.value = form.pcexpiry.value = "indefinite";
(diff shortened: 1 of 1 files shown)


MusikAnimal:
@atlight look right to you? I tried it out on testwiki and it seemed fine

atlight:
Seems good to me, although I didn't test it.

It would be nice to fix the problem described at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_36#Protection_templates and https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_36#Don.27t_add_pp-icon_unneeded - I might file an issue for that

PR #291, opened 2015-09-15
Prevent jumping to next section when target section is empty

jcgoble3:
Multiple reports have been made to WT:TW regarding protection requests being posted in the wrong section. The common thread between all reported cases is that the target section was empty at the time.

According to testing I did on a copy of the RFPP page in Notepad++, the problem was that the regex was consuming both the target section and the following section when the target section was empty. I'm honestly not sure how this fixes the problem, but Notepad++ testing says that it does. I'll just call it black magic and not argue with it. :P

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -1403,9 +1403,9 @@ Twinkle.protect.callbacks = {
 
 		var reg;
 		if ( increase ) {
-			reg = /(\n==\s*Current requests for increase in protection level\s*==\s*\n\s*\{\{[^\}\}]+\}\}\s*\n)([\s\S]*?)\s*(\n==[^=])/;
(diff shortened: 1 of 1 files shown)


atlight:
Yeah, this is a known issue. It's quite a fickle regex, I'll have to do some more testing before I merge this.

jcgoble3:
Given the complexity of the regex, have you considered a more verbose definition? JS doesn't *technically* support verbose regexes, but it can be [faked with string concatenation](http://stackoverflow.com/questions/3949991/javascript-regex-compared-to-perl-regex#comment20611186_3950684) and normal comments. That might be worth considering here, so you can more precisely document exactly what each token is intended to match.

jcgoble3:
Any update on this?

jcgoble3:
Obsoleted by 6b02198e90cd4e84ab55cd2644d904b0932e6914. Closing request and will be deleting my fork.

PR #292, opened 2015-09-21
This pull request is not expected to be approved!

Huji:
This pull request can address #129 but I don't expect it to be approved because:

* it needs cleanup in terms of indentation
* the actual localization module should be empty (unless we decide that even EN WP stuff should be considered localization, rather than defaults)
* we should decide if we want to pull now (which will be functional but incomplete) or if we want to i18nize and l10nize all modules first and then pull


jquery-i18n.js
```diff
@@ -0,0 +1,120 @@
...
+	 * @param	string str					 : The string to translate .
+	 * @param	property_list params : Params for using printf() on the string.
+	 *
+	 * @return	element							: Chained and translated element(s).
+	*/
+	$.fn._t = function( str, params ) {
+		return $( this ).html( i18n._.apply( i18n, arguments ) );
+	};
+
+	$.i18n = i18n;
```
atlight: MediaWiki implements its own module called "jQuery.i18n", written by Santhosh Thottingal. See https://git.wikimedia.org/tree/mediawiki%2Fcore/HEAD/resources%2Flib%2Fjquery.i18n. Including this module in Twinkle will probably cause a conflict of some kind unless the symbol is renamed. (Sorry for not noticing this earlier...)

modules/friendlyshared.js
```diff
@@ -16,15 +16,18 @@
...
 
 Twinkle.shared.callback = function friendlysharedCallback( uid ) {
+	console.log(Twinkle.shared.standardList);
 	var Window = new Morebits.simpleWindow( 600, 420 );
-	Window.setTitle( "Shared IP address tagging" );
-	Window.setScriptName( "Twinkle" );
-	Window.addFooterLink( "Twinkle help", "WP:TW/DOC#shared" );
+	Window.setTitle( $.i18n._('tw-sharedip-window-title') );
+	Window.setScriptName( $.i18n._('tw-core-script-name') );
+	Window.addFooterLink( $.i18n._('tw-core-help'), $.i18n._('tw-core-help-page') );
```
atlight: Wikis should have the opportunity to set their own array of footer links.

modules/friendlyshared.js
```diff
@@ -75,49 +78,53 @@ Twinkle.shared.callback = function friendlysharedCallback( uid ) {
...
-		value: 'ISP'
-	},
-	{
-		label: '{{Mobile IP}}: shared IP address template modified for mobile phone companies and their customers',
-		value: 'Mobile IP'
-	},
-	{
-		label: '{{Whois}}: template for IP addresses in need of monitoring, but unknown whether static, dynamic or shared',
-		value: 'Whois'
+Twinkle.shared.initialize = function() {
```
atlight: I'm not sure what this is supposed to accomplish. I say, get rid of it; enwiki can have its config moved out into an appropriate l10n module itself.

modules/friendlyshared.js
```diff
@@ -154,7 +161,7 @@ Twinkle.shared.callbacks = {
 		}
 		text += '}}\n\n';
 
-		var summaryText = 'Added {{[[Template:' + params.value + '|' + params.value + ']]}} template.';
+		var summaryText = $.i18n._('tw-sharedip-added', params.value, params.value);
```
atlight: ???? Can't the i18n string reuse the same value more than once?

modules/friendlyshared.js
```diff
@@ -166,14 +173,14 @@ Twinkle.shared.callbacks = {
...
-		alert( 'You must select a shared IP address template to use!' );
+		alert( $.i18n._('tw-sharedip-alert-template') );
 		return;
 	}
 
 	var value = shared[0];
 
 	if( e.target.organization.value === '') {
-		alert( 'You must input an organization for the {{' + value + '}} template!' );
+		alert( $.i18n._('tw-sharedip-alert-organization'), value );
```
atlight: `)` in wrong place

modules/twinklei18n.js
```diff
@@ -0,0 +1,53 @@
+TwinkleMessages = {
```
atlight: Wrap this in 
`//<nowiki>`
and
`//</nowiki>`
like the other Twinkle files.

modules/twinklei18n.js
```diff
@@ -0,0 +1,53 @@
...
+	'en': {
+		'tw-core-portlet-name': 'TW',
+		'tw-core-summary': ' ([[WP:TW|TW]])',
+		'tw-core-summary-del': ' ([[WP:TW|TW]])',
+		'tw-core-summary-prot': ' ([[WP:TW|TW]])',
+		'tw-core-script-name': 'Twinkle',
+		'tw-core-help': 'Twinkle Help',
+		'tw-core-help-page': 'WP:TW/DOC#shared',
+		'tw-morebits-info': 'Info',
+		'tw-sharedip-label': 'Shared IP',
```
atlight: This file absolutely must have comments. That seems pretty important to me. Otherwise we will end up with lots of wrong translations.

Also separate it into sections, e.g. above this line put a blank line and a big comment like `// ********** Shared IP (tw-sharedip) **********`

Actually this makes me wonder, why did you use the prefix `tw-sharedip-` instead of `tw-shared`? The internal name of the Shared IP module is `shared`. Let's not introduce any more internal names...

modules/friendlyshared.js
```diff
@@ -16,15 +16,18 @@
 Twinkle.shared = function friendlyshared() {
 	if( mw.config.get('wgNamespaceNumber') === 3 && Morebits.isIPAddress(mw.config.get('wgTitle')) ) {
 		var username = mw.config.get('wgTitle').split( '/' )[0].replace( /\"/, "\\\""); // only first part before any slashes
-		Twinkle.addPortletLink( function(){ Twinkle.shared.callback(username); }, "Shared IP", "friendly-shared", "Shared IP tagging" );
+		Twinkle.addPortletLink( function(){ Twinkle.shared.callback(username); }, $.i18n._('tw-sharedip-label'), "friendly-shared", 'tw-sharedip-label-hover' );
```
atlight: Maybe implement a function that looks like
`Twinkle.addPortletLink_i18n( function() { ... }, 'tw-sharedip' /* i18n prefix */, 'friendly-shared' /* CSS class */ );`

modules/friendlyshared.js
```diff
@@ -16,15 +16,18 @@
...
 	}
 };
 
 Twinkle.shared.callback = function friendlysharedCallback( uid ) {
```
atlight: As far as I can tell, `uid` is unused. Not an i18n issue specifically, but may as well fix it anyway.

jquery-i18n.js
Huji: Yes I am aware. And the i18n module that MW implements actually handles the '{{' characters as internal parameters which makes it hard/impossible to have messages where '{{' is supposed not to be parsed by the i18n module. Hence the use of a separate module for i18n in TW.

I will rename the module though so it doesn't conflict though. Thanks for bringing it to my attention.

modules/friendlyshared.js
Huji: Unsure what you mean. If the purpose is to let them replace the footer link with something else, it is already possible (they need to override the values for `tw-core-help` and `tw-core-help-page`). If you mean they should be able to have more than one footer link if they deem necessary, then that is an issue beyond the scope of this change. I am trying not to add or remove any functionality, but simply make them i18nized and l10inzed exactly as they are.

modules/friendlyshared.js
Huji: That answers one of my questions on the #129 :) Seems like you want even EnWP settings to be treated as a localization of TW. I will make it that way.

modules/friendlyshared.js
Huji: Unsure what you mean. Which value should have been reused?

modules/twinklei18n.js
Huji: Will do as advised. I will also add a 'qqq' group of translations to explain what the messages are when necessary. This can be then used on TranslateWiki to help boost TW translations!

Huji:
Closing for now. Will start over shortly/

PR #295, opened 2015-10-05
Reporting if page has been protected before! And minor morebits.css change

MusikAnimal:
Fairly sizable update to the protect module that could use another set of eyes.

When protecting a page, admins must first check the protection log to make an informed decision on the duration. Just as with the block module, I've made it report if there were prior protection periods, and link to the protection log if there were. Here we also have to do a separate API call to fetch the pending changes log.

Next, I've included the log notice and current protection notice to be visible for all users, not just admins. We often get requests for semi-protection when the page is semi-protected, and the log info is useful for them to decide what duration to request and to explain their rationale. We also want to use meaningful colours to grab the user's attention. For instance, here again, if a page is already protected we want that to be clear so they don't request the same level of protection. So "no protection" appears as green and if there is a current protection the message will appear in red. The log messages are also in red as that is crucial information for the admins. This puts the log/current state colours in line with the block module.

Finally, a minor CSS change. To conditionally include one or two links to the logs (protect or pending changes), I need to use a `<span>` wrapper so I could pass it to Morebits.status. There might be a better way of doing this, not sure, but I noticed with each child span the alignment gets increasingly off. Turns out we've got a `vertical-align:middle` on the text, which I think we only need for inline images which appears to only include the ui-icon. I tested on various modules and multiple browsers and everything looks good.

@atlight mind taking a look? Or any other Twinkle followers? :)

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -151,24 +151,30 @@ Twinkle.protect.fetchProtectionLevel = function twinkleprotectFetchProtectionLev
 };
 
 Twinkle.protect.callback.showLogAndCurrentProtectInfo = function twinkleprotectCallbackShowLogAndCurrentProtectInfo() {
+	var currentlyProtected = !$.isEmptyObject(Twinkle.protect.currentProtectionLevels);
+
 	if (Twinkle.protect.hasProtectLog || Twinkle.protect.hasStableLog) {
 		var $linkMarkup = $("<span>");
 
 		if (Twinkle.protect.hasProtectLog)
-			$linkMarkup.append($( '<a target="_blank" href="' + mw.util.getUrl('Special:Log', {action: 'view', page: mw.config.get('wgPageName'), type: 'protect'}) + '">protection log</a>' ), '&nbsp;&nbsp;');
+			$linkMarkup.append(
+				$( '<a target="_blank" href="' + mw.util.getUrl('Special:Log', {action: 'view', page: mw.config.get('wgPageName'), type: 'protect'}) + '">protection log</a>' ),
+				Twinkle.protect.hasStableLog ? $("<span> &bull; </span>") : null
+			);
 		if (Twinkle.protect.hasStableLog)
-			$linkMarkup.append($( '<a target="_blank" href="' + mw.util.getUrl('Special:Log', {action: 'view', page: mw.config.get('wgPageName'), type: 'stable'}) + '">pending changes log</a>)' ), '&nbsp;&nbsp;');
+			$linkMarkup.append($( '<a target="_blank" href="' + mw.util.getUrl('Special:Log', {action: 'view', page: mw.config.get('wgPageName'), type: 'stable'}) + '">pending changes log</a>)' ));
 
 		Morebits.status.init($('div[name="hasprotectlog"] span')[0]);
-		Morebits.status.warn('This page has been protected in the past', $linkMarkup[0]);
+		Morebits.status.warn(
+			currentlyProtected ? 'Previous protections' : 'This page has been protected in the past',
+			$linkMarkup[0]
+		);
 	}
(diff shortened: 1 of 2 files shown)


atlight:
> I tested on various modules and multiple browsers and everything looks good.

Multiple skins as well? 

The rule in question has been in place since day dot (2007) and Twinkle's UI has come a long way since then, so it's probably not needed. However, it is worthwhile looking at this in all browsers, especially IE9, which we apparently still support (although I don't much care if the appearance is broken in IE9, it's worth giving it a quick check-over anyway). So if you've done that, I'm prepared to trust your judgement :)

The code looks good, I assume you tested it thoroughly as a non-admin?

MusikAnimal:
Tried it out in all skins, and even on IE9 (I have a VM at my disposal) and all checked out visually. Honestly the CSS could use a lot of cleanup, if not just formatting of the code itself. I'll save that for rainy day. I did also test admin/non-admin side by side during development. Cheers

atlight:
Couple of UI observations:

* On the Main Page:
```
Current protection level: Edit: sysop (indefinite) (cascading) Move: sysop (indefinite) (cascading)
This page has been protected in the past: protection log  
```

Taken alone, the sentence "This page has been protected in the past" would be interpreted as saying that the page was protected but is not currently. I think you might need to rethink the way these two lines are phrased when they appear together.

* Barack Obama:

```
Current protection level: Edit: autoconfirmed (indefinite) Move: sysop (indefinite)
This page has been protected in the past: protection log  pending changes log  
```

Put a space, U+2022, space between the protection log and pending changes log.

Other than that I really like it, and it seems to work well. Thanks for taking the initiative!

atlight:
One reason why the vertical-align: middle is there is to centre checkboxes with respect to the line of text in which they lie. I think that needs to be addressed before I'm prepared to merge this.

MusikAnimal:
Alrighty, I've got the bullet between the log links, and changed the copy to read "Previous protections" if we're always reporting a current protection. How does that sound?

As for vertically aligning the checkboxes/radio buttons, this has long been a frustration by web developers as there isn't really a rock solid cross-browser/platform solution. For instance, everything looks on point with the updated CSS on OSX Chrome. The old CSS with vertical-align:middle looked off for me. Meanwhile the old CSS looked better in Firefox, and is still slightly off with the new CSS. It's true more people probably use Windows, so maybe we could tweak it a little with some relative positioning. Another option is to implement our own custom checkboxes and radio buttons using CSS, giving us more control. Thoughts?

Also, I have to ask, how are you testing this on enwiki? Look at your contribs you haven't attempted to override any of the Twinkle assets.

atlight:
> Also, I have to ask, how are you testing this on enwiki? Look at your contribs you haven't attempted to override any of the Twinkle assets.

In Firefox, by using the Style Editor to add the new rule and the Inspector to turn off the old one :) That's the way I test everything, no need to waste time mucking around in editing.

MusikAnimal:
I meant the JS, not the CSS! I don't even develop CSS on the filesystem, instead do everything in the inspector then copy it over =P

atlight:
I just copy and paste the raw JS module into the console, press Enter, and run `Twinkle.protect.callback()`.

MusikAnimal:
Ah, nice. Anyway, I've done a lot of tweaking, and I think I've found a cross-platform cross-browser happy medium for the form elements. It's not perfect, but possibly better than what we had. Mind taking a look at various modules on your Windows machine (assuming that's what you have) in Firefox/Chrome? I briefly tried it by my Windows VM likes to randomly shut itself down, making it a rather painstaking task.

MusikAnimal:
@atlight barring objections I'm going to move forward with this. Many thanks for your time

atlight:
Yes, please go ahead.

PR #297, opened 2015-10-26
Ask if user has contacted protecting admin before requesting unprotection

MusikAnimal:
It is procedural that users ask the protecting admin about unprotection before requesting it at [[WP:RFPP]]. More often than not this does not happen, so per request we're going to try to have Twinkle remind them.

In short: When the protection module is opened, you'll see the current protections as you do normally, but with the name of the protecting admin next to each entry, which links to their talk page. When you attempt to request unprotection you are prompted (with `confirm`) "Have you attempted to contact the protecting admins (MusikAnimal, NeilN) first?"

Some notes:
- If an admin changes _any_ protection settings they appear in the log as the admin responsible for all protection settings. E.g. if NeilN move-protects the page and I semi-protect it, it is not feasible to infer that NeilN did the move protection and not me. I don't think this is a big concern, as the re-protecting admin will have enough insight into why the other protections are there.
- Protection settings can be _moved_ when the page is moved. If the script detects the current protection was moved from a previous page, it will recursively go through the logs to find the original protecting admin. ~~I believe there to be a rock-solid base case and graceful handling of failures so that the recursion won't lock up.~~ Still working on this! Fixed... it now won't check logs that have already been checked. Otherwise it could go into an infinite loop when you protect a page, move it somewhere, then move it back to the original location.
- Pending changes protection have their own log entry, but if moved from a previous page, it does _not_ appear in the logs. This is a MediaWiki bug as far as I can tell, and the workaround to recurse through the page moves to find a pending changes log entry seems initially not worth the effort, but we may revisit it.
- The interface shows the admin who applied the protections, as with "Current protections: Move: (sysop) ... by [link to MusikAnimal's talk page]". We have the sysop name next to each edit/move/create protect entry, even though they will be the same. This is purely for aesthetics, since it looks weird to have it just at the end. I also separated the entries by a bullet for better readability
  - A side note: What about a separate line for each entry, to make it even more readable? Or one line for edit/move/create protections and another for pending changes?
- You are only prompted if you contacted the protecting admin if the script successfully figured out who the protecting admin was

Thoughts? I think this change is less than perfect but still effective, and will be well-received.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -4,6 +4,7 @@
 	"unused": true,
 	"smarttabs": true,
 	"eqeqeq": true,
+	"curly": true,
 	"noarg": true,
 	"globals": {
 		"mw": false,
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -98,6 +98,8 @@ Twinkle.protect.callback = function twinkleprotectCallback() {
 // { edit: { level: "sysop", expiry: <some date>, cascade: true }, ... }
 Twinkle.protect.currentProtectionLevels = {};
 
+// returns a jQuery Deferred object, usage:
+//   Twinkle.protect.fetchProtectingAdmin(apiObject, pageName).done(function(admin_username) { ...code... });
 Twinkle.protect.fetchProtectingAdmin = function twinkleprotectFetchProtectingAdmin(api, pageName, logIds) {
 	logIds = logIds || [];
 
@@ -109,7 +111,7 @@ Twinkle.protect.fetchProtectingAdmin = function twinkleprotectFetchProtectingAdm
 		letype: 'protect'
 	}).then(function( data ) {
 		// don't check log entries that have already been checked (e.g. don't go into an infinite loop!)
-		var event = $.grep(data.query.logevents, function(le) { return $.inArray(le.logid, logIds); })[0];
+		var event = data.query ? $.grep(data.query.logevents, function(le) { return $.inArray(le.logid, logIds); })[0] : null;
 		if (!event) {
 			// fail gracefully
 			return null;
@@ -156,7 +158,9 @@ Twinkle.protect.fetchProtectionLevel = function twinkleprotectFetchProtectionLev
 					cascade: protection.cascade === ''
 				};
 				// logs report last admin who made changes to either edit/move/create protection, regardless if they only modified one of them
-				if(!adminEditDeferred) adminEditDeferred = Twinkle.protect.fetchProtectingAdmin(api, mw.config.get('wgPageName'));
+				if (!adminEditDeferred) {
+					adminEditDeferred = Twinkle.protect.fetchProtectingAdmin(api, mw.config.get('wgPageName'));
+				}
 			}
 		});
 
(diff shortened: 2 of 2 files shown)


MusikAnimal:
I should clarify that I'm not actually done with this, just asking for a preliminary review and input in general. There's a bug or two I still need to fix. I'll report back when I'm confident the script is stable.

MusikAnimal:
Alright, after those last 2 commits I'm confident with what I have. I guess in my testing I got a little confused... turns out the logging API does report who actually applied pending changes protection, it just doesn't indicate it was transferred from a page move. As for the MediaWiki implementation this is probably the better way to do it, I guess (who cares whether it was moved).

Meanwhile tracking down who edit/move/create protected a page after it has been moved has long been a headache for users to figure out, so I guess that's a major bonus with this change to Twinkle. That is, one could open up the protect module only to see who applied the original protection. Pretty cool!

MusikAnimal:
@atlight made the recommended changes, and also took the liberty of adding `curly` to the `jshintrc`. Thank you for your time!

atlight:
Sorry, I'm very busy at present. I'd like to test this before merging, hopefully I will find time to do it soon.

Amorymeltzer:
@MusikAnimal Any interest in reviving this?  Seems useful.

MusikAnimal:
It's been a long time! I can't believe the only conflict is with .jhhintrc! I'll try to take another look at this soon. Hopefully it still works :)

Amorymeltzer:
@MusikAnimal 	I had initially checked this out at test.wiki, and my recollection is that it failed with no warnings, but I've been thinking on it and I bet it just crapped out 'cause of the stabilize log.  I just ran a quick test on en.wiki, and it worked.  I'd like to do some more investigating but is there any chance this is... good?  Two points until then:

1. The `twinkleprotect.js` module on test.wiki incorporates this pr, but I assume it never worked properly, do you recall?
2. I think probably better to toss the .jshintrc mod and leave that to #500

Amorymeltzer:
@MusikAnimal 	Tested this a bunch, seems good on my end.  I've put it up on test wiki with my edits in #551 merged in, so it should be available to test if you like.  If you like what you see and undo the `.jshintrc` change, it should be ready to go!

Amorymeltzer:
@MusikAnimal Any opposition to me just going ahead with this?

MusikAnimal:
Sorry I somehow missed your first two pings! I trust your testing so let me take out those .jshintrc changes and resubmit. Amazing that's the only conflict after all these years!

MusikAnimal:
@Amorymeltzer shit shit shit... I just forced pushed to origin/master instead of my remote! Could you please force push back the latest?

MusikAnimal:
> @Amorymeltzer shit shit shit... I just forced pushed to origin/master instead of my remote! Could you please force push back the latest?

I think I fixed it.. thanks to `git reflog`. Please double check master is correct! Why I used my own `master` branch instead of a named branch back in 2015, I don't know. But that's what caused this mishap. Sorry about that!

I will abandon this and create a new PR with a different branch name.

Amorymeltzer:
`reflog` has saved my ass a few times for sure, but I didn't even know it was possible to open a pull from a `master` branch!

Amorymeltzer:
#586

PR #300, opened 2015-11-14
Admins delete under multiple rationale, use template-generated delete reasons

MusikAnimal:
This is **incomplete**. Still needs testing, and there's at least one bug I haven't been able to fix

* Allow admins to delete under multiple rationale
* Allow admins to use the subgroups to enter in values such as the URL for G12 deletions
* Use the corresponding template to generate a deletion summary
* Move custom rationale to top with it's own section, and supply the rationale generated by the CSD template already on the page, if it exists

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -110,6 +110,9 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
 							// enable/disable delete multiple
 							cForm.delmultiple.disabled = cChecked;
 							cForm.delmultiple.checked = false;
+							// enable/disable open talk page checkbox
+							cForm.openusertalk.disabled = cChecked;
+							cForm.openusertalk.checked = false;
 
 							// enable/disable notify checkbox
 							cForm.notify.disabled = !cChecked;
@@ -175,6 +178,18 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
 				}
 			]
 		} );
+		form.append( {
+				type: 'checkbox',
+				list: [
+					{
+						label: 'Open user talk page on submit',
+						value: 'openusertalk',
+						name: 'openusertalk',
+						tooltip: 'Use this is override your preferences for the selected rationale',
+						checked : false
+					}
+				]
+			} );
 		form.append( { type: 'header', label: 'Tag-related options' } );
 	}
 
@@ -432,8 +447,21 @@ Twinkle.speedy.generateCsdList = function twinklespeedyGenerateCsdList(list, mod
 					}
 				];
 			}
+			// FIXME: does this do anything?
 			criterion.event = openSubgroupHandler;
 		}
+
+		criterion.event = function(e) {
+			if (multiple) return;
+			if( isSysop ) {
+				var normalizedCriterion = Twinkle.speedy.normalizeHash[e.target.value];
+				$('[name=openusertalk]').prop('checked',
+						Twinkle.getPref('openUserTalkPageOnSpeedyDelete').indexOf(normalizedCriterion) !== -1
+					);
+			} else {
+				$('[name=openusertalk]').prop('checked', false);
+			}
+		};
 
 		return criterion;
 	});
@@ -1133,7 +1161,7 @@ Twinkle.speedy.callbacks = {
 
 			// look up initial contributor. If prompting user for deletion reason, just display a link.
 			// Otherwise open the talk page directly
-			if( params.openusertalk ) {
+			if( params.openUserTalk ) {
 				thispage.setCallbackParameters( params );
 				thispage.lookupCreator( Twinkle.speedy.callbacks.sysop.openUserTalkPage );
 			}
@@ -1816,7 +1844,7 @@ Twinkle.speedy.callback.evaluateSysop = function twinklespeedyCallbackEvaluateSy
 	});
 
 	// analyse each criterion to determine whether to watch the page, prompt for summary, or open user talk page
-	var watchPage, promptForSummary, openUserTalk;
+	var watchPage, promptForSummary;
 	normalizeds.forEach(function(norm) {
 		if (Twinkle.getPref("watchSpeedyPages").indexOf(norm) !== -1) {
 			watchPage = true;
@@ -1824,18 +1852,15 @@ Twinkle.speedy.callback.evaluateSysop = function twinklespeedyCallbackEvaluateSy
 		if (Twinkle.getPref("promptForSpeedyDeletionSummary").indexOf(norm) !== -1) {
 			promptForSummary = true;
 		}
-		if (Twinkle.getPref('openUserTalkPageOnSpeedyDelete').indexOf(norm) !== -1) {
-			openUserTalk = true;
-		}
 	});
 
 	var params = {
 		values: values,
 		normalizeds: normalizeds,
 		watch: watchPage,
-		openusertalk: openUserTalk,
 		deleteTalkPage: form.talkpage && form.talkpage.checked,
 		deleteRedirects: form.redirects.checked,
+		openUserTalk: form.openusertalk.checked,
 		promptForSummary: promptForSummary,
 		templateParams: Twinkle.speedy.getParameters( form, values )
 	};
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -185,7 +185,7 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
 						label: 'Open user talk page on submit',
 						value: 'openusertalk',
 						name: 'openusertalk',
-						tooltip: 'Use this is override your preferences for the selected rationale',
+						tooltip: 'This defaults to your open-talk-page preferences when deleting pages under the currently selected rationale. It is left unchanged if you choose to delete under multiple criteria.',
 						checked : false
 					}
 				]
@@ -1120,7 +1120,7 @@ Twinkle.speedy.callbacks = {
 		};
 
 		var statusIndicator = new Morebits.status( 'Building deletion summary' );
-		var api = new Morebits.wiki.api( 'Fetching parsing deletion template', query, function(apiObj) {
+		var api = new Morebits.wiki.api( 'Parsing deletion template', query, function(apiObj) {
 				statusIndicator.info( 'complete' );
 				callback(apiObj);
 			},  statusIndicator);
@@ -1864,6 +1864,9 @@ Twinkle.speedy.callback.evaluateSysop = function twinklespeedyCallbackEvaluateSy
 		promptForSummary: promptForSummary,
 		templateParams: Twinkle.speedy.getParameters( form, values )
 	};
+	if(!params.templateParams) {
+		return;
+	}
 
 	Morebits.simpleWindow.setButtonsEnabled( false );
 	Morebits.status.init( form );
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -451,17 +451,18 @@ Twinkle.speedy.generateCsdList = function twinklespeedyGenerateCsdList(list, mod
 			criterion.event = openSubgroupHandler;
 		}
 
-		criterion.event = function(e) {
-			if (multiple) return;
-			if( isSysop ) {
+		if ( isSysop ) {
+			var originalEvent = criterion.event;
+			criterion.event = function(e) {
+				if (multiple) return originalEvent(e);
+
 				var normalizedCriterion = Twinkle.speedy.normalizeHash[e.target.value];
 				$('[name=openusertalk]').prop('checked',
 						Twinkle.getPref('openUserTalkPageOnSpeedyDelete').indexOf(normalizedCriterion) !== -1
 					);
-			} else {
-				$('[name=openusertalk]').prop('checked', false);
-			}
-		};
+				return originalEvent(e);
+			};
+		}
 
 		return criterion;
 	});
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -460,7 +460,9 @@ Twinkle.speedy.generateCsdList = function twinklespeedyGenerateCsdList(list, mod
 				$('[name=openusertalk]').prop('checked',
 						Twinkle.getPref('openUserTalkPageOnSpeedyDelete').indexOf(normalizedCriterion) !== -1
 					);
-				return originalEvent(e);
+				if ( originalEvent ) {
+					return originalEvent(e);
+				}
 			};
 		}
 
@@ -820,7 +822,8 @@ Twinkle.speedy.generalList = [
 			label: 'Page where the deletion discussion took place: ',
 			tooltip: 'Must start with "Wikipedia:"',
 			size: 60
-		}
+		},
+		hideSubgroupWhenMultiple: true
 	},
 	{
 		label: 'G5: Banned or blocked user',
@@ -920,7 +923,8 @@ Twinkle.speedy.generalList = [
 			label: 'Optional explanation: ',
 			tooltip: 'Perhaps linking to where the author requested this deletion.',
 			size: 60
-		}
+		},
+		hideSubgroupWhenSysop: true
 	},
 	{
 		label: 'G8: Pages dependent on a non-existent or deleted page',
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1121,9 +1121,14 @@ Twinkle.speedy.callbacks = {
 
 		var statusIndicator = new Morebits.status( 'Building deletion summary' );
 		var api = new Morebits.wiki.api( 'Parsing deletion template', query, function(apiObj) {
-				statusIndicator.info( 'complete' );
-				callback(apiObj);
-			},  statusIndicator);
+				var reason = decodeURIComponent($(apiObj.getXML().querySelector('text').childNodes[0].nodeValue).find('#delete-reason').text()).replace(/\+/g, ' ');
+				if (!reason) {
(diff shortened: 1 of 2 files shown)


MusikAnimal:
@atlight When there's already a CSD tag on the page, the module has "custom rationale" selected with the template-generated rationale in the subgroup input field. Problem is I can't get that subgroup to hide when I click on another radio button. If I go back and forth twice it will work as it should.

Basically I'm having trouble understanding what hides the subgroups. I put a breakpoint in `openSubgroupHandler` within `twinklespeedyGenerateCsdList`, and it never gets hit, no matter what I click on.

modules/twinklespeedy.js
```diff
@@ -41,21 +41,24 @@ Twinkle.speedy.dialog = null;
 // The speedy criteria list can be in one of several modes
 Twinkle.speedy.mode = {
 	sysopSubmit: 1,  // radio buttons, no subgroups, submit when "Submit" button is clicked
-	sysopRadioClick: 2,  // radio buttons, no subgroups, submit when a radio button is clicked
-	userMultipleSubmit: 3,  // check boxes, subgroups, "Submit" button already pressent
-	userMultipleRadioClick: 4,  // check boxes, subgroups, need to add a "Submit" button
-	userSingleSubmit: 5,  // radio buttons, subgroups, submit when "Submit" button is clicked
-	userSingleRadioClick: 6,  // radio buttons, subgroups, submit when a radio button is clicked
+	sysopMultipleSubmit: 2, // check boxes, subgroups, "Submit" button already present
```
atlight: The mode constants were very carefully laid out in an odd-even pattern to satisfy the code at line 247. Please test the code again with the "When to go ahead and tag/delete the page" Twinkle preference set to "As soon as I click an option" (i.e. speedySelectionStyle = radioClick) and fix the code as required. 

I also wonder whether there need to be two sysop modes. The reason why tagging has separate single and multiple modes is that some CSD templates (e.g. db-hoax and the A7 subcategories) are not properly supported by db-multiple, so we have to prevent a different list to the user. (Arguably they could be combined into one, but that's a separate issue for another day.) For actual page deletion, is there any reason why we couldn't just get rid of the radio-button list and *always* show checkboxes?

modules/twinklespeedy.js
```diff
@@ -392,6 +413,27 @@ Twinkle.speedy.generateCsdList = function twinklespeedyGenerateCsdList(list, mod
 	});
 };
 
+Twinkle.speedy.existingCSDRationale = function() {
+	return Morebits.userIsInGroup('sysop') && ($("#delete-reason").length ? decodeURIComponent($("#delete-reason").text()).replace(/\+/g, ' ') : '');
+};
+
+Twinkle.speedy.customRational = [
```
atlight: Misspelt.

modules/twinklespeedy.js
```diff
@@ -248,6 +270,10 @@ Twinkle.speedy.callback.modeChanged = function twinklespeedyCallbackModeChanged(
 
 	var radioOrCheckbox = (Twinkle.speedy.mode.isMultiple(mode) ? 'checkbox' : 'radio');
 
+	work_area.append( { type: 'header', label: 'Custom rationale' } );
+
+	work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.customRational, mode) } );
```
atlight: Doesn't need to be under a separate header for non-admins. For tagging, simply merge the customRationale array with the generalList array and add them to the general header, to preserve current behaviour. Otherwise you are just making the list longer for no good  reason.

modules/twinklespeedy.js
```diff
@@ -1806,20 +1784,40 @@ Twinkle.speedy.callback.evaluateSysop = function twinklespeedyCallbackEvaluateSy
...
 
-	var value = Twinkle.speedy.resolveCsdValues(e)[0];
-	if (!value) {
+	var values = Twinkle.speedy.resolveCsdValues(e);
+	if (!values) {
 		return;
 	}
-	var normalized = Twinkle.speedy.normalizeHash[ value ];
+	var normalizeds = [];
+	$.each(values, function(index, val) {
```
atlight: Can just use values.map here, I think.

modules/twinklespeedy.js
```diff
@@ -1295,35 +1297,11 @@ Twinkle.speedy.callbacks = {
...
-					code += "|help=off";
-				}
-				code += "}}";
-				params.utparams = Twinkle.speedy.getUserTalkParameters(params.normalizeds[0], parameters);
-			}
+			// given the params, builds the template and also adds the user talk page parameters to the params that were passed in
+			// returns => [<string> wikitext, <object> newParams]
+			var buildData = Twinkle.speedy.callbacks.getTemplateCodeAndParams(params),
+				code = buildData[0];
+			params = buildData[1];
```
atlight: Since it's only modifying params.utparams  within the params object, just make it return [wikitext, utparams]. Better encapsulation that way, I think.

modules/twinklespeedy.js
```diff
@@ -998,84 +1027,55 @@ Twinkle.speedy.normalizeHash = {
...
+		var query = {
+			action: "parse",
+			prop: "text",
+			pst: "true",
+			text: wikitext,
+			title: mw.config.get("wgPageName")
+		};
+		var api = new Morebits.wiki.api( 'Fetching deletion summary from template', query, callback);
+		api.post({
+			async: false
```
atlight: Yikes! Don't use non-async queries.

MusikAnimal: Yeah, I wouldn't normally do async either but I saw it was being done here on [L1647](https://github.com/MusikAnimal/twinkle/blob/speedy-admins/modules/twinklespeedy.js#L1647). It's just so tough to work without deferred/promises! I'll do a series of callbacks instead

modules/twinklespeedy.js
```diff
@@ -40,22 +40,27 @@ Twinkle.speedy.dialog = null;
...
-	userMultipleSubmit: 3,  // check boxes, subgroups, "Submit" button already pressent
-	userMultipleRadioClick: 4,  // check boxes, subgroups, need to add a "Submit" button
-	userSingleSubmit: 5,  // radio buttons, subgroups, submit when "Submit" button is clicked
-	userSingleRadioClick: 6,  // radio buttons, subgroups, submit when a radio button is clicked
+	sysopMultipleSubmit: 3, // check boxes, subgroups, "Submit" button already present
+	sysopMultipleRadioClick: 4, // check boxes, subgroups, need to add a "Submit" button
+	userMultipleSubmit: 5,  // check boxes, subgroups, "Submit" button already pressent
+	userMultipleRadioClick: 6,  // check boxes, subgroups, need to add a "Submit" button
+	userSingleSubmit: 7,  // radio buttons, subgroups, submit when "Submit" button is clicked
+	userSingleRadioClick: 8,  // radio buttons, subgroups, submit when a radio button is clicked
```
MusikAnimal: @atlight Fixed this bug. As for having only multi-select for admins: I can see admins (myself included) wanting to pick a specific A7 criteria. Secondly I think it would be an odd user experience to have the "custom rationale" as a checkbox, that would have to be disabled should you check other criteria

modules/twinklespeedy.js
```diff
@@ -40,22 +40,27 @@ Twinkle.speedy.dialog = null;
...
 
 	// are we in "delete page" mode?
 	// (sysops can access both "delete page" [sysop] and "tag page only" [user] modes)
 	isSysop: function twinklespeedyModeIsSysop(mode) {
-		return mode === Twinkle.speedy.mode.sysopSubmit ||
+		return mode === Twinkle.speedy.mode.sysopSingleSubmit ||
+			mode === Twinkle.speedy.mode.sysopMultipleSubmit ||
 			mode === Twinkle.speedy.mode.sysopRadioClick;
```
atlight: Four sysop modes now..?

MusikAnimal:
Two outstanding bugs, as far as I can tell:
1. As described above, when there's already a CSD tag, I can't get the custom rationale subgroup (input field) to hide when another radio button is selected.
2. When in `sysopMultipleRadioClick` mode with `speedySelectionStyle` as radio, the module deletes as soon as a rationale checkbox is selected, rather than allowing you to select multiple ones then a hit submit under "_When finished choosing criteria, click_"

Any insight on these two issues? Thank you very much for your help and time in reviewing this!

modules/twinklespeedy.js
MusikAnimal: to match the four user modes, right? similar scenarios, just need to delete instead of tag

MusikAnimal:
@atlight just making sure you saw my note on the two outstanding bugs :)

atlight:
I'll have a look...

atlight:
For number 1, try adding `event({ target: subnode });` after line 388 in morebits.js. See if that fixes it.

MusikAnimal:
@atlight That did not work :( The subnode does not have an associated form, which I see is a timing issue, as it errors out before the module fully loads. If I remove that code and let the module load, all the elements do have an associated form. So not sure what to do about that... but frankly while undesirable it is not a crucial bug. #2 described above is a big problem. Do you think that is a Morebits issue as well?

Perhaps we could remove the delete-on-click functionality and only allow it for tagging? It seems to me deleting like that would be super error-prone, as one could easily click the wrong thing.

MusikAnimal:
Maybe for number 1 we could instead let the module load, then check if there is a CSD already on the page, and if so check custom rationale and add supply the reason. I will try that

MusikAnimal:
Deleting last two comments... fixed but number 1! Now for number 2, which is the bigger issue

MusikAnimal:
@atlight Alright! Going to move forward with testing in production, with the help our friend Callanecc :)

Let me know if you see any issues code-wise. Thanks!

MusikAnimal:
@atlight Made some changes to the default Twinkle preferences. First, don't prompt for deletion summary under _any_ criterion. This is not needed so much anymore with the advent of being able to pick multiple ratione and supply params, etc. Delete talk page by default, which we almost always want to do (except for user talk pages, which are excluded).

Finally, open user talk pages in a new _tab_ and not a window. This is unrelated to the speedy update, but I figure this has been long overdue. Most browsers will block popups, so you have to explicitly allow Twinkle to do it. Most people prefer tabs anyway, I am willing to bet. Any input on this change is appreciated, otherwise I'm ready to merge in the speedy update. I've got a number of admins waiting for it! :)

modules/twinklespeedy.js
```diff
@@ -153,6 +163,33 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
...
+			]
+		} );
+		form.append( {
+				type: 'checkbox',
+				list: [
+					{
+						label: 'Open user talk page on submit',
+						value: 'openusertalk',
+						name: 'openusertalk',
+						tooltip: 'Use this is override your preferences for the selected rationale',
```
atlight: Word "is" doesn't make sense here.

Also, if I were an admin, I would have trouble figuring out how this checkbox interacts with my preferences. This could do with some more explanation in the tooltip.

atlight:
I tried deleting a page under G4, typing garbage characters into the subgroup textbox. I was notified that my input was invalid (which is true), but then there was a JavaScript error: `TypeError: params.templateParams is null` and the CSD dialog sat there in an unusable state.

modules/twinklespeedy.js
```diff
@@ -998,126 +1075,96 @@ Twinkle.speedy.normalizeHash = {
...
+		var query = {
+			action: "parse",
+			prop: "text",
+			pst: "true",
+			text: wikitext,
+			title: mw.config.get("wgPageName")
+		};
+
+		var statusIndicator = new Morebits.status( 'Building deletion summary' );
+		var api = new Morebits.wiki.api( 'Fetching parsing deletion template', query, function(apiObj) {
```
atlight: This isn't grammatical.

modules/twinklespeedy.js
atlight: Also, this never seems to get checked in the case of multiple criteria

modules/twinklespeedy.js
MusikAnimal: Haha, this and the comment above about `is` are a result of working on this during wee hours of the night

modules/twinklespeedy.js
MusikAnimal: You could loop through the selected rationale and see if your preferences are all the same for those criteria, but I don't really think that is that advantageous. In the case of multiple we'll just let the admin decide what to do. Maybe it should default to checked, though

MusikAnimal: Correction, it leaves the checkbox state as-is once you choose to delete under multiple rationale. I feel this is best, as I don't want to make any assumptions or go about the crazy logic I mentioned above.

MusikAnimal:
Fixed the bug you found, and made some copy edits. Not sure about the "open user talk page" when deleting under multiple criteria. I don't want to default to checked as some will find it annoying. If it gets unchecked when they didn't want it to they can view the deleted edits and click on the user talk page link there. I don't think we'll get many complaints.

atlight:
I'll just test under "radioClick" and other modes, and then am happy to merge this.

The checkboxes and controls at the top of the list for admins are starting to get a bit out of hand. It would be helpful if, rather than disabling irrelevant controls (i.e. the "tag-related options" when in delete mode, and vice versa), the irrelevant checkboxes were actually hidden altogether. This would save vertical real estate. I won't insist on it for this change though; in fact, it would probably be better to address this in a separate pull request just to keep this PR manageable.

MusikAnimal:
Yes I was thinking the same. I was even considering making it checkbox-based like the block module. Probably easier said than done.

MusikAnimal:
^ @atlight one more commit. Figured it's better to at least say a deletion summary couldn't be generated by the template rather than assume the admin didn't enter one. In production, it will most certainly be the latter and not some bug with the templates, but nonetheless good to be informative.

atlight:
A bug: Use Twinkle as a non-sysop in `radioClick` mode. Open CSD and click "G6: Housekeeping". You are never prompted for a rationale.

I am really wondering whether it is worthwhile to keep supporting `radioClick`. Only [18 users](https://en.wikipedia.org/w/index.php?title=Special%3ASearch&profile=all&search=radioClick+intitle%3Atwinkleoptions&fulltext=Search) use it... But there are 9 admins among these 18 (and also 2 socks!), and I imagine some of them would be a bit irate if we got rid of it. If it is not too much effort, we should try to keep supporting `radioClick`, but I can see the code could easily get out of hand.

MusikAnimal:
@atlight Alright, fixed I believe. I can try to reach out to those admins in particular and ask how dependent they are on the `radioClick` mode.

MusikAnimal:
@atlight sorry to bug you again... how do we feel about this?

atlight:
I tested it out on a non-admin account, but pretending to be a sysop. The deletion encountered an error (obviously), but the error message immediately disappeared, being replaced by "Opening user talk page edit form for This, that and the other: complete". Although this specific situation is contrived, it is a problem if deletion error messages aren't hanging around to let people read and copy them.

atlight:
I tried deleting as a sysop under the G7 criterion, but the rationale I entered into the form disappeared into the ether...

atlight:
Also, a bunch of the form inputs aren't used to construct the deletion summary when deleting with multiple criteria. I think you should make more liberal use of the `hideWhenMultiple` and `hideWhenSysop` options in the subgroup configuration. Maybe  we even need a `hideWhenSysopMultiple` option for a select few?

MusikAnimal:
@atlight fixed the last 2 bugs you mentioned, I believe (some of those subgroups didn't do anything when requesting speedy in multi-mode). How do I pretend to be a sysop?

atlight:
> How do I pretend to be a sysop?

You are a sysop, so you don't need to pretend; I, on the other hand, am not a sysop, so I run `wgUserGroups.push('sysop')` in the browser console so Twinkle thinks I am one.

MusikAnimal:
FYI if you're going through and reviewing stuff, this one still needs a little work. I've just been sidetracked with other matters.

MusikAnimal:
@atlight Just now getting back to this... I tried deleting as a non-admin but got `Deleting page: Failed to delete the page: Permission denied` when I submitted. Are you able to reproduce?

Also, just a general question, how do you normally do development/testing of Twinkle (or any MediaWiki JavaScript project)? I noticed you had very little Twinkle related-edits on testwiki. Are you using another wiki or just pasting in the code in the JS console on enwiki?

MusikAnimal:
Actually maybe I fixed that bug with [this](https://github.com/MusikAnimal/twinkle/commit/7191d4cb24887b58909033ff7b58df3de5e8e616). Were there any other outstanding bugs you encountered?

I've been using it regularly on enwiki without issues. In a future update I'd like to clean up the interface some as you suggest. I also had the idea that maybe we could make post-delete templates to put on user talk pages, and automate adding those. Right now many admins tag pages first, then deleting them, so that the informative user talk notices are sent out. We just need those templates in the past-tense. That's something I'll need to discuss on enwiki, but anyway if we could do that the interface would closely mimic the Block module.

MusikAnimal:
@atlight Going to merge this if there is no opposition. I've been using it for a while without issues, and will be sure to swiftly address any concerns users bring up once it's rolled out. Many thanks

atlight:
Feel free to merge it, but make sure you're around to fix any reported bugs :)

MusikAnimal:
Will do. Going to be cautious and deploy this tomorrow morning (EST) when I'm going to be in front of a computer all day.

Also... still curious how you do Twinkle development. I didn't know if you knew of an easier way? I write code in my text editor, copy/paste into the relevant MediaWiki namespace file on testwiki, reload whatever page, and do testing. This is tedious. Copy/pasting into the JS console would be initially quicker, but with a refresh I'd have to copy/paste again. Chrome has some sort of "work off of filesystem" that maybe I should look into.

PR #301, opened 2015-11-18
protect: have move protect settings match edit protect settings by de…

MusikAnimal:
…fault

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -605,6 +605,14 @@ Twinkle.protect.formevents = {
 		e.target.form.editexpiry.disabled = (e.target.value === 'all');
 	},
 	movemodify: function twinkleprotectFormMovemodifyEvent(e) {
+		// sync move settings with edit settings if applicable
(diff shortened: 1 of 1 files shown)


MusikAnimal:
This is technically a "workaround" to achieve the desired behaviour. Ideally the presets would work like they do with Block, where there's a big JS object mapping of what the presets should apply when they are selected. So for example here you can go to the "Dispute/move warring (move)" preset, then go to "Content dispute/edit warring (full)" and you end up with edit protect at 2 days and move still at indefinite. The same is true if you go back and forth between other presets though, so for now I think this will do.

Note I just opted to go with MediaWiki default settings in making the move protection level semi by default if the edit protection level is semi, even though unconfirmed users can't move pages. I think this is specific to enwiki, so we're set up to support other wiki configurations. In practice we'll rarely semi-protect and also move-protect at the same time, I think... @Callanecc any thoughts regarding this?

Callanecc:
Sounds good to me!

PR #306, opened 2015-12-21
MoreBits.js: always assert the user is logged in when making a POST

MusikAnimal:
I think this will do it. There might be a better way, I'm not familiar with how Morebits.js works

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1398,6 +1398,7 @@ Morebits.wiki.api = function( currentAction, query, onSuccess, statusElement, on
 	this.currentAction = currentAction;
 	this.query = query;
 	this.query.format = 'xml';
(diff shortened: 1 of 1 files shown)


morebits.js
```diff
@@ -1398,6 +1398,7 @@ Morebits.wiki.api = function( currentAction, query, onSuccess, statusElement, on
 	this.currentAction = currentAction;
 	this.query = query;
 	this.query.format = 'xml';
+	this.query.assert = 'user';
```
MusikAnimal: @atlight That worked. I thought I could put it here but didn't take the time to try =P

atlight:
Guess I should look at your other PRs, eh?...

MusikAnimal:
Thanks!

PR #310, opened 2016-01-18
protect: adjust regex at RFPP to look for section after the one we wa…

MusikAnimal:
…nt to post to

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -1403,12 +1403,13 @@ Twinkle.protect.callbacks = {
 
 		var reg;
 		if ( increase ) {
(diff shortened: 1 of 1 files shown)


MusikAnimal:
@atlight what do you think of this implementation? Since we are now posting reports at the bottom of the section, it seems easier to use regex against the subsequent heading, rather than attempt to capture the whole section. We'll need to update [Wikipedia:Twinkle/Fixing RPP](https://en.wikipedia.org/wiki/Wikipedia:Twinkle/Fixing_RPP) accordingly

atlight:
Makes more sense. I reckon, go ahead and merge. I don't think anything could be much worse than what we had :)

PR #319, opened 2016-03-07
Fixes #318 : Makefile should be updated since delimages is deleted

### LINKED ISSUE #318: Makefile should be updated since delimages is deleted
del-images module was removed in a209b75f82deba8010dea639d4950aff9d2b2e86 but Makefile has not been updated accordingly. I will submit a pull request shortly.

Huji:


atlight:
Do you use the Makefile? If no-one's using it I don't think we really need it...

Huji:
I do. That's my primary method of syncing.

PR #320, opened 2016-03-08
i18n and l10n of Twinkle

Huji:
Please use http://fawp.wmflabs.org/wiki/User_talk:1.2.3.4 and http://fawp.wmflabs.org/wiki/User_talk:1.2.3.4?uselang=de as test environment.

This is NOT a final pull request. It is WIP to induce comments. It aims to address issue #129


modules/friendlyshared.js
```diff
@@ -16,15 +16,17 @@
 Twinkle.shared = function friendlyshared() {
 	if( mw.config.get('wgNamespaceNumber') === 3 && Morebits.isIPAddress(mw.config.get('wgTitle')) ) {
 		var username = mw.config.get('wgTitle').split( '/' )[0].replace( /\"/, "\\\""); // only first part before any slashes
-		Twinkle.addPortletLink( function(){ Twinkle.shared.callback(username); }, "Shared IP", "friendly-shared", "Shared IP tagging" );
+		Twinkle.addPortletLink( function(){ Twinkle.shared.callback(username); }, $.i18n._('tw-sharedip-label'), "friendly-shared", 'tw-sharedip-label-hover' );
```
atlight: Aren't you supposed to pass `_` as an additional parameter to the wrapping anonymous function, thereby allowing it to be called as just `_(...)`?

modules/friendlyshared.js
```diff
@@ -75,49 +77,53 @@ Twinkle.shared.callback = function friendlysharedCallback() {
...
-	{
-		label: '{{Mobile IP}}: shared IP address template modified for mobile phone companies and their customers',
-		value: 'Mobile IP'
-	},
-	{
-		label: '{{Whois}}: template for IP addresses in need of monitoring, but unknown whether static, dynamic or shared',
-		value: 'Whois'
+Twinkle.shared.initialize = function() {
+	if( typeof Twinkle.shared.standardList === 'undefined' ) {
+		Twinkle.shared.standardList = [
```
atlight: I'm unlikely to accept a patch that uses this approach to localising the template list. The whole variable `Twinkle.shared.standardList` needs to be able to be set for the specific wiki, allowing the local template set to be added.

Obviously the logic that relates to individual templates is going to have to be retained for now (I think we agreed not to localise the callbacks yet), so some of these strings will need to be kept.

modules/friendlyshared.js
```diff
@@ -144,7 +150,7 @@ Twinkle.shared.callbacks = {
 			return;
 		}
 
-		Morebits.status.info( 'Info', 'Will add the shared IP address template to the top of the user\'s talk page.' );
+		Morebits.status.info( $.i18n._('tw-core-morebits-info'), $.i18n._('tw-sharedip-will-add') );
```
atlight: Not `tw-core-morebits`. Just `morebits` will do. morebits.js is not Twinkle-specific.

modules/friendlyshared.js
Huji: Will do. I was hoping to get more comments on the bigger picture though, rather than function naming. Specifically: the "logic" of the tool is embedded into the code and no matter how much we separate the message texts (i18n) and references to them (i10n), we still will have lines of code that are EN WP logic and may not apply to another wiki. How do we want to proceed with them?

In other words, if we are fine with EN WP logic being in the code, and the only limiting issue is the name of _ function, I'll happily fix that!

modules/friendlyshared.js
Huji: Now this is really useful. How do you envision it to be done in a perfect world though?

We could change it such that it takes a symbolic list of what to show, and each item in the symbolic list is then tied to appropriate messages. But then we cannot have logic associated with the them items.

modules/friendlyshared.js
Huji: Will fix.

modules/friendlyshared.js
atlight: The bigger picture looks mostly OK, which is why I didn't comment on it. Ideally we'd get rid of all the enwiki-specific logic, but it's best to go partway to begin with and get it right, I reckon.

I guess my only concern is still the separation into i18n and l10n. I think we need to remember that different wikis in the same language might have different names for the same process, so the distinction between i18n and l10n is quite blurry.

modules/friendlyshared.js
atlight: Something like: the data contains an entry which exactly matches the format of this array, e.g. 
```
[
    {
        label: '{{ergojduhd}}: lorem ipsum dolor sit amet',
        value: 'ergojduhd',
        tooltip: 'consectetur adipiscing elit'
    },
    ...
]
```
Then these are accompanied by strings like
`'tw-sharedip-edu-template-name': 'Shared IP edu'`
which would be used below in place of the hard-coded template name.

Obviously this is deficient. Most wikis will not need this semi-hard-coded opt-out. In the end, the enable/disable callback and form validation code would be able to be localised as well. Whether or not you want to do that at this point is up to you, I guess.

modules/friendlyshared.js
```diff
@@ -144,7 +150,7 @@ Twinkle.shared.callbacks = {
...
 		text += params.value + '|' + params.organization;
 		if( params.value === 'Shared IP edu' && params.contact !== '') {
```
atlight: Evil hard-coded template names. See my comment above.

modules/friendlyshared.js
Huji: Okay, I don't know how to do that.

Try this after the page loads:

`_ = $.i18n._`

It won't allow you to use _('foobar'), because that function depends on the entire object.

Can you elaborate how I can do what I asked me to?

modules/friendlyshared.js
Huji: But bear in mind: right now the SharedIP doesn't even sepcify a single message. All messages are defined in the twinklei18n and twinklel10n files.

Is your concern that the l10n file is defining the messages that matter to EN WP and calls them "en" group of messages? And that other English projects may want a different set of options shown? Well, can't they simply modify the l10n file? Wouldn't that be _much easier_ than what they do now, which is the edit 10 different files in 100 different places?

modules/friendlyshared.js
Huji: Yes, but what is the alternative? To move all callbacks into the l10n file? I am not saying that is doable. I am saying that will require a LOT of time, and not something I can do alone.

modules/friendlyshared.js
atlight: That is my concern. More precisely, my concern is that other wikis are likely to have a completely different template setup from enwiki, so these predefined slots are most likely not going to make sense to them. What if (unlikely, but bear with me) a wiki had 15 different shared IP templates? Because there aren't 15 different slots, their only recourse is to modify the Twinkle code, which is precisely what we are trying to avoid.

Is that what you're asking? I'm not really sure I've addressed your concern.

modules/friendlyshared.js
atlight: Hmm, I'll have a look later on.

modules/friendlyshared.js
Huji: The change I have done already supports that!

Try the link I sent to you before: http://fawp.wmflabs.org/w/index.php?title=User_talk:1.2.3.4&uselang=de

make sure you have created account and enabled Twinkle. In the DE interface, it'll show you 2 shared IP templates? Why, because the l10n only lists 2. If they want 15, they can list 15.

modules/friendlyshared.js
atlight: As I said above, I'm not asking for a perfect solution - leaving the callbacks in is, for now, OK. However, I really have a hard time swallowing the idea of keeping some hard-coded template names in. 

> Besides, this is something I have raised over and over again: that EN WP logic is embedded in Twinkle. And that is bad, we all know, but it is not l10n/i18n's job to fix it. 

In principle, I disagree. To me, localising Twinkle is precisely removing all enwiki-specific strings, objects, logic etc. from the core code and into a separate localisation file. However, it certainly does not have to be done all at once, and as you say, it is a Herculean effort (which is why it hasn't been done until now). Some localisation is better than none, so I'm prepared to accept a partially-completed transition.

I know I frustrate you by my constant conflation of i18n and l10n, but from a Twinkle perspective the overlap is much greater than for most other projects. If I talk about i18n as the translation of project-independent strings and l10n as adaptation to an individual project, then in fact a lot of string translation will fall under the l10n heading.

modules/friendlyshared.js
atlight: Yes, but it's a hack. Why do we even need to keep this partly-localised Twinkle.shared.sharedList in this file at all? What good does it serve? Every project is going to override with their own version of it.

Huji: You are absolutely correct! I don't know why I didn't push my change about that. I've already fixed it in an unpushed change!

Let me create a new pull request. :) I apologize for confusion.

Huji:
New one is being created.

modules/friendlyshared.js
Huji: I don't mind the frustration as long as it leads to progress :)

modules/friendlyshared.js
MusikAnimal: Not that it's a big deal, but devs who see `_('something')` might think it is [Underscore.js](underscorejs.org). How about `t('something')` or `_t('something')`?

PR #329, opened 2016-03-14
Bring i18n and l10n features to Twinkle

### LINKED ISSUE #129: Allow internationalization of Twinkle

Huji:
This one partially fixes #129 by removing the Twinkle.shared.standardList entirely from the module, moving it to the localization.


jquery-i18n.js
```diff
@@ -0,0 +1,120 @@
...
+	 * @param	property_list params : Params for using printf() on the string.
+	 *
+	 * @return	element							: Chained and translated element(s).
+	*/
+	$.fn._t = function( str, params ) {
+		return $( this ).html( i18n._.apply( i18n, arguments ) );
+	};
+
+	$.i18n = i18n;
+})(jQuery);
```
Huji: @atlight I know you prefer _(...) over $.i18n._(...) but I don't know how to achieve that. Trying the following after the page loads:

`_ = $.i18n._`

And then calling `_('tw-core-portlet-name')` throws an error while `$.i18n._('tw-core-portlet-name')` of course works.

I cannot think of a way to get around the dependency of `_()` function on the entire `$.i18n` object.

atlight: Yes, I saw that. It's rather odd. I'm sure there is a way around it using `apply()` or one of those weird functions. As I said, I'll look later.

modules/twinklel10n.js
```diff
@@ -0,0 +1,84 @@
...
+
+
+// Adding translations is easy, for example:
+TwinkleMessages.de = {
+	'tw-core-portlet-name': 'DE TW',
+};
+
+// Localizing twinkle is also easy, for example:
+switch( mw.config.get('wgUserLanguage') ) {
+	case 'en':
```
Huji: @atlight This is an important part of the change that I somehow had managed not to push (accidentally put it in a separate branch).

As you can see, the default set of templates for 'en' is what EN WP uses. We can decide if we want to include any of the other translations (I vote no), but for now I am keeping a mock-translation for DE, just so you can try the demo link and see how it works (only two templates will be listed in DE interface).

modules/friendlyshared.js
```diff
@@ -16,16 +16,16 @@
 Twinkle.shared = function friendlyshared() {
 	if( mw.config.get('wgNamespaceNumber') === 3 && Morebits.isIPAddress(mw.config.get('wgTitle')) ) {
 		var username = mw.config.get('wgTitle').split( '/' )[0].replace( /\"/, "\\\""); // only first part before any slashes
-		Twinkle.addPortletLink( function(){ Twinkle.shared.callback(username); }, "Shared IP", "friendly-shared", "Shared IP tagging" );
+		Twinkle.addPortletLink( function(){ Twinkle.shared.callback(username); }, $.i18n._('tw-sharedip-label'), "friendly-shared", 'tw-sharedip-label-hover' );
 	}
 };
 
 Twinkle.shared.callback = function friendlysharedCallback() {
```
Huji: @atlight In future, we can move this into the l10n file as well. Just like how we move the Twinkle.shared.standardLIst to there.

jquery-i18n.js
Huji: I have given you write access to my fork, so feel free to experiment there.

MusikAnimal: Sorry, I accidentally deleted @Huji's comment (meant to delete my own). It read: `I have given you write access to my fork, so feel free to experiment there.`

Huji:
@atlight sorry for being away from this for a while.

I cleaned up the code again. I also changed `$.i18n` to `$._` which is closer to what you liked. What are the next steps?

siddharthvp:
This PR has been superseded by other work as y'all are aware. Closing to avoid any confusion.

Huji:
Totally agree.

PR #331, opened 2016-03-19
'Fixes #330: Reword the message for 'New unreviewed article'

### LINKED ISSUE #330: Template:New unreviewed article
The description of {{New unreviewed article}} in Twinkle is a bit misleading, and leads to the sort of abuse that caused the template to be nominated at TfD. The template documentation says "It should never be placed on someone else's article under any circumstances", but the Twinkle description just says "mark article for later review >", which strongly implies that it is for marking other editor's articles that you plan on reviewing later.

The easiest solution is just to remove this template from Twinkle altogether. This template is mostly designed to be used by the Article Wizard, but users can also place it on their own articles. However, an editor who is advanced enough to be using Twinkle is very unlikely to every need to use this tag, as they are also likely experienced enough not to need to tag their new article for review.

If the template is kept, is Twinkle smart enough to only place it on articles created by the user? If not, the description should be changed to read "mark your own article for later review >" and/or pop up an "are you sure?"-type warning.

JadeMaveric:
Implements the simplest suggestion mentioned in #330.
I plan on implementing the others if possible.
Plus, it'll be a nice way to get to know twinkle

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -465,7 +465,7 @@ Twinkle.tag.article.tags = {
 	"merge to": "article should be merged into another given article",
 	"metricate": "article exclusively uses non-SI units of measurement",
 	"more footnotes": "article has some references, but insufficient in-text citations",
-	"new unreviewed article": "mark (your own) article for later review",
+	"new unreviewed article": "mark your own article for later review",
 	"news release": "article reads like a news release",
 	"no footnotes": "article has references, but no in-text citations",
 	"non-free": "article may contain excessive or improper use of copyrighted materials",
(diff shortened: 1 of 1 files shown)


modules/friendlytag.js
```diff
@@ -465,7 +465,7 @@ Twinkle.tag.article.tags = {
 	"merge to": "article should be merged into another given article",
 	"metricate": "article exclusively uses non-SI units of measurement",
 	"more footnotes": "article has some references, but insufficient in-text citations",
-	"new unreviewed article": "mark article for later review",
+	"new unreviewed article": "mark (your own) article for later review",
```
atlight: No need for the parentheses.

atlight:
Good, but I don't think any of the other suggestions in #330 are worth bothering with. Particularly the "are you sure" warning - I don't think that is needed at all. The "your own" wording ought to act as a strong enough deterrent.

JadeMaveric:
Okay then. Should I remove the brackets?

JadeMaveric:
Okay, I'll make the changes and update

JadeMaveric:
Done!

JadeMaveric:
@atlight anything else I should do?

MusikAnimal:
Going to assume everything is fine, looks good to me at least.

PR #333, opened 2016-04-03
speedy: Fix delete reason fetching on non-wikitext pages

jimmyxu:
Currently `parseWikitext()` would bork on `.js`/`.css` pages for that `parse`'d give out something like `<pre>{{db-spam}}</pre>` instead of expanding the template.

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1172,6 +1172,7 @@ Twinkle.speedy.callbacks = {
 			prop: "text",
 			pst: "true",
 			text: wikitext,
+			contentmodel: "wikitext",
 			title: mw.config.get("wgPageName")
 		};
 


MusikAnimal:
@jimmyxu Have you tested this? I tried it out and Twinkle still doesn't seem to parse it as wikitext

jimmyxu:
@MusikAnimal Merged into my tree and it [worked fine](https://zh.wikipedia.org/wiki/User:9shi/vector.js). Haven't tested on enwiki though, would try in a bit.

[before](https://en.wikipedia.org/w/api.php?action=parse&prop=text&pst=true&title=User:Example/common.js&text={{db-spam}}) & [after](https://en.wikipedia.org/w/api.php?action=parse&prop=text&pst=true&title=User:Example/common.js&text={{db-spam}}&contentmodel=wikitext)

MusikAnimal:
I guess I'm confused what this is supposed to fix. When you browse to a .js or .css page, the markup of the `{{db}}` template isn't there, so Twinkle would never notice in the first place, since it does this via a jQuery selector (see [#L266](https://github.com/azatoth/twinkle/blob/master/modules/twinklespeedy.js#L266)). Right?

jimmyxu:
This is for when you ask Twinkle to speedy delete a page. Since now building the deletion summary requires parsing wikitext, trying to delete any .js or .css pages would give out an error:

> Building deletion summary: Unable to generate summary from deletion template
> Asking for reason: you didn't give one. I don't know... what with admins and their apathetic antics... I give up...

Try this by submitting a U1 or anything on a .js or .css page.

MusikAnimal:
Ah yes of course! Sorry, I thought we were talking about auto-filling in the reason based on what was already on the page. Works on my end. Merging

PR #334, opened 2016-04-06
protect: add new 30/500 protection

MusikAnimal:
This is subject to misuse, so including a confirmation popup indicating it is (currently) strictly for Arbitration purposes on qualifying subjects.

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -250,6 +250,11 @@ Twinkle.protect.callback.changeAction = function twinkleprotectCallbackChangeAct
 						label: 'Autoconfirmed',
 						value: 'autoconfirmed'
 					});
+				editlevel.append({
+						type: 'option',
+						label: 'Extended confirmed',
+						value: 'extendedconfirmed'
+					});
 				if (isTemplate) {
 					editlevel.append({
 							type: 'option',
@@ -272,6 +277,7 @@ Twinkle.protect.callback.changeAction = function twinkleprotectCallbackChangeAct
 								Twinkle.protect.doCustomExpiry(e.target);
 							}
 						},
+						// default expiry selection is conditionally set in Twinkle.protect.callback.changePreset
 						list: [
 							{ label: '1 hour', value: '1 hour' },
 							{ label: '2 hours', value: '2 hours' },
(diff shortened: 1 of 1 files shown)


MusikAnimal:
@atlight Do you care to review this? Only thing that worries me is the protection weight stuff. I tested it out. E.g. the page was semi'd, I tried extendend confirmed and it counts as an increase. Meanwhile if it's already extended confirmed, requesting PC or semi counts as a decrease.

modules/twinkleprotect.js
```diff
@@ -718,6 +736,7 @@ Twinkle.protect.protectionWeight = {
 	sysop: 30,
 	templateeditor: 20,
 	flaggedrevs_review: 15,  // Pending Changes level 2 protection alone
+	extendedconfirmed: 12,
```
atlight: Shouldn't be 12. This is for semiprotected + PC1 (see comment above).  Maybe it is time to increase sysop to 40, TE to 30, then put 30/500 at 20. Unless PC2 alone (15) and semiprotected + PC2 (17) should really be considered "higher" than 30/500?

modules/twinkleprotect.js
```diff
@@ -915,13 +941,23 @@ Twinkle.protect.callback.changePreset = function twinkleprotectCallbackChangePre
...
+			isArbEnforcement = item.edit === 'extendedconfirmed';
+
 		if (mw.config.get('wgArticleId')) {
 			if (item.edit) {
 				form.editmodify.checked = true;
 				Twinkle.protect.formevents.editmodify({ target: form.editmodify });
 				form.editlevel.value = item.edit;
 				Twinkle.protect.formevents.editlevel({ target: form.editlevel });
+
+				// NOT: Currently extendedconfirmed protection is only for Arbitration enforcement,
```
atlight: `NOTE`?

modules/twinkleprotect.js
```diff
@@ -1021,6 +1068,14 @@ Twinkle.protect.callback.evaluate = function twinkleprotectCallbackEvaluate(e) {
 		case 'protect':
 			// protect the page
 
+			if (form.editlevel.value === 'extendedconfirmed') {
+				if (!confirm('Extended confirmed protection is only for Arbitration enforcement. ' +
+						'Please confirm that the subject of this page is a qualifying topic (see WP:30/500)'
```
atlight: Full stop needed at the end of the message. And you can use && to merge the two if statements into one and avoid crazily deep indent levels.

modules/twinkleprotect.js
```diff
@@ -1129,6 +1184,15 @@ Twinkle.protect.callback.evaluate = function twinkleprotectCallbackEvaluate(e) {
 
 		case 'request':
 			// file request at RPP
+
+			if (form.category.value === 'pp-30-500') {
+				if (!confirm('Extended confirmed protection is only for Arbitration enforcement. ' +
+						'Please confirm that the subject of this page is a qualifying topic (see WP:30/500)'
```
atlight: Same as above.

atlight:
Duly reviewed.

MusikAnimal:
@atlight those weight changes worked. Also cleaned up the confirm popups. Thanks for the help, I'll deploy once I get a final OK

MusikAnimal:
Merging!

PR #342, opened 2016-06-01
Adjust {{uw-3rr}} description

stickee-wp:
A discussion on [WT:TWINKLE](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#.22Violating.22_the_three_revert_rule) has resulted in a consensus that the edit summary left by Twinkle shouldn't imply the user has broken the 3RR. The warning itself {{[uw-3rr](https://en.wikipedia.org/wiki/Template:Uw-3rr)}} doesn't say the user has broken 3RR, so neither should the edit summary.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1012,8 +1012,8 @@ Twinkle.warn.messages = {
 
 	singlewarn: {
 		"uw-3rr": {
-			label: "Violating the three-revert rule; see also uw-ew",
-			summary: "Warning: Violating the three-revert rule"
+			label: "Potential three-revert rule violation; see also uw-ew",
+			summary: "Warning: Three-revert rule"
 		},
 		"uw-affiliate": {
 			label: "Affiliate marketing",


modules/twinklewarn.js
```diff
@@ -1012,8 +1012,8 @@ Twinkle.warn.messages = {
 
 	singlewarn: {
 		"uw-3rr": {
-			label: "Violating the three-revert rule; see also uw-ew",
-			summary: "Warning: Violating the three-revert rule"
+			label: "Three-revert rule; see also uw-ew",
```
atlight: The label is only seen by Twinkle users, so it doesn't have to change. If you think it should be changed, I'd prefer to see something slightly more descriptive than just "three-revert rule". Could two or three more words be added?

stickee-wp:
Okay @atlight, how's that? I used similar warning to that at [WP:WARN](https://en.wikipedia.org/wiki/Wikipedia:Template_messages/User_talk_namespace).

atlight:
Excellent! If you can squash the two commits into one, I'll merge.

stickee-wp:
Squashed. Ready for merge.

atlight:
Sorry for the wait.

PR #343, opened 2016-06-11
Add G6 criterion {{db-error}}

nyuszika7h:


atlight:
If the page was "obviously" created in error, why do we need a separate G6 template for it? Call me grumpy, but it just seems like a waste of time.

Amorymeltzer:
I generally think the more specific we can be about G6 the better, and likewise the more we can avoid the tyranny of "_housekeeping_" the better, but there are already four or five G6 options present (depending on mode) so I'm inclined to agree with @atlight here.

PR #302, opened 2015-11-20
warn: pruning single-issue notices

MusikAnimal:
Rationales for removal:

{{uw-2redirect}} - pretty sure a bot fixes these
{{uw-articlesig}} - button to add signature was removed from the edit toolbar when editing in mainspace
{{uw-csd}} - better you actually say why the speedy criteria they chose was not applicable
{{uw-directcat}} - surely is a rare occurrance. When it happens, just fix it
{{uw-fuir}} - I believe a bot does this, or those AWB people do. Surely doesn't happen enough to warrant inclusion here
{{uw-imageuse}} - Just going to go ahead and remove this. We have a verbose explanation shown via edit filter when they try to do this. Better we just write a personal note then send them to walls of outdated text
{{uw-incompleteAFD}} - Cyberbot issues more descriptive warnings
{{uw-nonfree}} - this is a warning, not a notice, and should happen when we nominate the image for deletion
{{uw-notaiv}} - we don't ever use this. Just explain the issue to them
{{uw-notifysd}} - this doesn't say what warnings to use
{{uw-notvand}} - tell them which edit was mislabeled
{{uw-patrolled}} - very rare, NPP is supposed to be for experienced editors, and they'll likely be using semi-automated tools that mark the page as patrolled
{{uw-probation}} - talk about being vague. We will be adding a way to issue sanction warnings in the future, otherwise tell them what the situation is
{{uw-refimprove}} - linked resources are grossly outdated, and we usually will instead issue a general notice on adding unverifiable content
{{uw-removevandalism}} - no one uses this
{{uw-repost}} - This is speedy deletion criteria and has it's own template
{{uw-samename}} - renames now happen on meta, and has got to be a pretty rare. The global renamers or stewards should let them know of the issue
{{uw-uaa}} - rarely used, and newcomers don't usually report to UAA
{{uw-upincat}} - a bot removes or "deactivates" these categories, or should
{{uw-uploadfirst}} - same with {{uw-imageuse}}
{{uw-userspacenoindex}} - we draftify user pages like this

PR #344, opened 2016-06-13
Add {{uw-uploadfirst}}

nyuszika7h:


MusikAnimal:
This was removed with #302 because an [edit filter](https://en.wikipedia.org/wiki/Special:AbuseFilter/220) already warns the user when they attempt to add external images. The issue here is there are a _lot_ of single-issue templates, so within Twinkle we wanted to limit it to those that are used frequently. I'm going to be bold and close this (PR is also almost 2 years old)... but of course I'm not in charge here so please re-open if you disagree.

Note you can add any custom warnings in your [Twinkle preferences](https://en.wikipedia.org/wiki/Wikipedia:Twinkle/Preferences), look for "Custom warning templates to display" under "Warn user".

PR #377, opened 2017-03-24
Add 'Template:Cleanup rewrite' to tagging module.

bardiharborow:
Would be helpful in situations where the issue is too extensive to explain in a {{cleanup}} tag.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -431,6 +431,7 @@ Twinkle.tag.article.tags = {
 	"BLP unsourced": "BLP article has no sources at all (use BLP PROD instead for new articles)",
 	"citation style": "article has unclear or inconsistent inline citations",
 	"cleanup": "article may require cleanup",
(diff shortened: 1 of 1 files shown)


siddharthvp:
I'd support adding this as it has over [4400 transclusions](https://tools.wmflabs.org/templatecount/index.php?lang=en&namespace=10&name=Cleanup_rewrite#bottom), which is significantly more than what many of the other templates included in the tag module have. For instance, {{cleanup reorganize}}, {{condense}}, {{lead missing}, {{lead rewrite}}, {{lead too long}}, {{sections}} all have <1000 or barely 1000+, and these are all from the same category "Structure, formatting, and lead section
" into which {{cleanup rewrite}} should go.  @Amorymeltzer @MusikAnimal

Amorymeltzer:
@bardiharborow Shouldn't this have the input box for the optional `reason` parameter?  See the entry for `{{copy edit}}` for an example.

Amorymeltzer:
Looked over a few dozen from the past few months, basically nobody uses the `1` parameter.  LGTM

PR #383, opened 2017-05-02
prod: Remove superfluous "move to Commons" tags

XXN:
Remove superfluous "move to Commons" tags when PRODing files. Deletion-tagged files cannot be moved to Commons

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -185,7 +185,11 @@ Twinkle.prod.callbacks = {
 
 		// Remove tags that become superfluous with this action
 		text = text.replace(/\{\{\s*([Nn]ew unreviewed article|[Uu]nreviewed|[Uu]serspace draft)\s*(\|(?:\{\{[^{}]*\}\}|[^{}])*)?\}\}\s*/g, "");
-
+		if (mw.config.get('wgNamespaceNumber') === 6) {
+			// remove "move to Commons" tag - deletion-tagged files cannot be moved to Commons
+			text = text.replace(/\{\{(mtc|(copy |move )?to ?commons|move to wikimedia commons|copy to wikimedia commons)[^}]*\}\}/gi, "");
+		}
+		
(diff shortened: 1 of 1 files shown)


atlight:
Can we have a better commit message, please?

XXN:
@atlight OK, edited.

modules/twinkleprod.js
```diff
@@ -185,7 +185,11 @@ Twinkle.prod.callbacks = {
 
 		// Remove tags that become superfluous with this action
 		text = text.replace(/\{\{\s*([Nn]ew unreviewed article|[Uu]nreviewed|[Uu]serspace draft)\s*(\|(?:\{\{[^{}]*\}\}|[^{}])*)?\}\}\s*/g, "");
-
+		if (mw.config.get('wgNamespaceNumber') === 6) {
```
MusikAnimal: At the top of this file the variable `namespace` is set to `article` or `file` (and eventually `book`). This is because we needed the noun itself to put in messages, so to keep it consistent I've been using `namespace === 'file'`, etc., when doing these checks. This is also helpful because you don't need to remember what namespace `6` is.

modules/twinkleprod.js
```diff
@@ -185,7 +185,11 @@ Twinkle.prod.callbacks = {
...
+			// remove "move to Commons" tag - deletion-tagged files cannot be moved to Commons
+			text = text.replace(/\{\{(mtc|(copy |move )?to ?commons|move to wikimedia commons|copy to wikimedia commons)[^}]*\}\}/gi, "");
```
MusikAnimal: One thing this misses is if there are spaces after `{{`. The regex otherwise seems fine but you might consider keeping the structure consistent with line 187. Frankly you could even merge this into line 187 entirely and remove the namespace check, as if for some reason there is a `{{move to commons}}` on an article then it obviously should also be removed :)

XXN: done

PR #390, opened 2017-08-18
Update {{uw-spam1}} label and edit summary

JJMC89:
"spam links" → "inappropriate external links"
Requested at [Wikipedia talk:Twinkle#Edit summary for uw-spam1](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=796122026#Edit_summary_for_uw-spam1)

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -243,8 +243,8 @@ Twinkle.warn.messages = {
 				summary: "General note: Not adhering to neutral point of view"
 			},
 			"uw-spam1": {
-				label: "Adding spam links",
-				summary: "General note: Adding spam links"
+				label: "Adding inappropriate external links",
+				summary: "General note: Adding inappropriate external links"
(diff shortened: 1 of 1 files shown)


rchard2scout:
It looks like the edit summaries have been pulled from [WP:UTM](https://en.wikipedia.org/wiki/Wikipedia:Template_messages/User_talk_namespace), so maybe it should be changed there as well?

MusikAnimal:
What about uw-spam2, 3, 4, 4im?

I otherwise see nothing wrong with this change, and yes I would update WP:UTM, too.

JJMC89:
@MusikAnimal: uw-spam1 doesn't mention spam, only that the links are inappropriate. The rest of the series does. WP:UTM only has the one line for the series, so an single one cannot be changed there.

mc10:
Seems reasonable to me.

PR #395, opened 2017-11-09
Minor tweaks

KylieStarzz:


atlight:
Thanks for your input, but I'd say that most of those tweaks aren't especially beneficial. Unless you can give a clear and reasonable rationale for each change I would not be prepared to merge them.

mc10:
Closing for now; if you'd like to split off the constructive changes, feel free to make another PR.

PR #397, opened 2017-11-09
Update twinklewarn.js

KylieStarzz:


Dinoguy1000:
Warning templates don't have a level 5, and you'd need to establish a consensus for adding one on Wikipedia before it would be added to Twinkle.

PR #400, opened 2017-11-11
Lint JS files with ESLint

mc10:
This is the beginning of some lint-based cleanup; we can enable more rules as we go on.
@azatoth: can you enable Travis builds on https://travis-ci.org/azatoth/twinkle?

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -1,6 +1,3 @@
-//<nowiki>
-
-
 (function($){
 
 
@@ -185,7 +182,7 @@ Twinkle.prod.callbacks = {
 
 		// Remove tags that become superfluous with this action
 		text = text.replace(/\{\{\s*(new unreviewed article|unreviewed|userspace draft|mtc|(copy|move) to wikimedia commons|(copy |move )?to ?commons)\s*(\|(?:\{\{[^{}]*\}\}|[^{}])*)?\}\}\s*/gi, "");
-		var prod_re = /\{\{\s*(?:dated prod|dated files|dated prod blp|Prod blp\/dated|Proposed deletion\/dated)\s*\|(?:\{\{[^\{\}]*\}\}|[^\}\{])*\}\}/i;
+		var prod_re = /{{\s*(?:dated prod|dated files|dated prod blp|Prod blp\/dated|Proposed deletion\/dated)\s*\|(?:{{[^{}]*}}|[^{}])*}}/i;
 		var summaryText;
 		if( !prod_re.test( text ) ) {
 			// Notification to first contributor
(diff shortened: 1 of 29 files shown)


WMDE-Fisch:
:+1:  this codebase could need some more structure :-)

azatoth:
@mc10 have activated travis now

modules/twinkleprod.js
```diff
@@ -185,7 +182,7 @@ Twinkle.prod.callbacks = {
 
 		// Remove tags that become superfluous with this action
 		text = text.replace(/\{\{\s*(new unreviewed article|unreviewed|userspace draft|mtc|(copy|move) to wikimedia commons|(copy |move )?to ?commons)\s*(\|(?:\{\{[^{}]*\}\}|[^{}])*)?\}\}\s*/gi, "");
```
azatoth: should change above as well then I assume

modules/twinkleprotect.js
```diff
@@ -1405,7 +1402,7 @@ Twinkle.protect.callbacks = {
 		var text = rppPage.getPageText();
 		var statusElement = rppPage.getStatusElement();
 
-		var rppRe = new RegExp( '===\\s*(\\[\\[)?\s*:?\s*' + RegExp.escape( Morebits.pageNameNorm, true ) + '\s*(\\]\\])?\\s*===', 'm' );
+		var rppRe = new RegExp( '===\\s*(\\[\\[)?\\s*:?\\s*' + RegExp.escape( Morebits.pageNameNorm, true ) + '\\s*(\\]\\])?\\s*===', 'm' );
```
azatoth: one more \\s

mc10: I think this one is okay; is there a specific `\s` you're referring to?

azatoth: there's a \\s at the end of the regex as well

azatoth: I read the diff backwards, I though you removed escapes :|

MusikAnimal:
The removal of `<nowiki>` causes some code to get substituted when syncing to the wiki. Could you restore those? Surely you can do this and not upset the linter?

MusikAnimal:
I do seem to recall someone saying <nowiki> is no longer needed for JS pages? Or maybe I'm making that up. Either way it definitely was substituting when I tried to sync, so I've created #413 and am syncing those versions. The code for #411 needs to go out today.

mc10:
My bad, for some reason I thought it wasn't needed either. I think it's included for backwards compatibility purposes but can be disabled on a wiki if desired. The relevant bug [T34858](https://phabricator.wikimedia.org/T34858) seems like a won't-fix in the near future.

PR #403, opened 2017-12-18
Modules to TfD

JJMC89:
[Wikipedia talk:Twinkle#Modules at TfD instead of MfD](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=815907267#Modules_at_TfD_instead_of_MfD)


atlight:
It would be good to squash these, but for now let's just go with it.

Issue #360, opened 2016-10-04
Twinkle JS does not reload after RevisionSlider is used to switch diffs


addshore:
Twinkle should listen to the JS hooks in core and reinit itself.

https://www.mediawiki.org/wiki/Extension:RevisionSlider#JavaScript_Hooks

terrorist96:
Sadly I'll have to disable RevisionSlider due to this issue. I'd love to be able to use RS and TW together. No news on this being fixed?

atlight:
Twinkle was written well before JS hooks existed in MediaWiki, so that's our excuse. To make matters worse there seems to be no documentation about JS hooks on mediawiki.org, apart from a tiny mention at https://www.mediawiki.org/wiki/ResourceLoader/Core_modules#mediaWiki.hook, so I don't know how we were meant to know that we were able to take advantage of this feature.

Whingeing aside, it doesn't look like that ought to be too troublesome to implement.

hartman:
@atlight 

hooks are 'dom events' to which you can add handlers. These handlers are then guaranteed to run whenever the event is triggered. It functions a bit like document.ready, but also works for JS loaded blobs and are in wide use by wikEd, VE, live preview, ajax preview etc. 

wikipage.content is triggered whenever content is added to the page (either initially for page views, on page preview, or on JS live preview). similarly wikipage.diff is triggered when a 'diff' is inserted in a page. [There are also hooks](https://doc.wikimedia.org/mediawiki-core/master/js/#!/api/mw.hook) for categories, collapsible content, the editform, postedit and a few more.

The contents being added to the page, are passed as a parameter (so you don't query the document, but the blob).

Twinkle can attach to these and reinitialise it's UI.

atlight:
Looks like RevisionSlider is about to be enabled by default: https://lists.wikimedia.org/pipermail/wikitech-l/2017-May/088199.html

So we ought to fix this. I don't have time right now unfortunately.

PR #406, opened 2018-01-04
Reload Twinklefluff interface when diff view changes

WMDE-Fisch:
This patch will hook Twinklefluff initialization to the mw.hook responsible for pages containing diff view.

Now for example the RevisionSlider extension that fires that hook will re-initiate the Twinkle links added to the page when moving through the revisions. This should fix https://github.com/azatoth/twinkle/issues/360

The hook gets fired everytime a diff page loads / reloads. To avoid having the links twice the original call got replaced.

Further modules that might need to be re-initiated by hook could be added in a follow up.

### DIFF
--- a/twinkle.js
+++ b/twinkle.js
@@ -19,7 +19,7 @@
 
 //<nowiki>
 
-( function ( window, document, $, undefined ) { // Wrap with anonymous function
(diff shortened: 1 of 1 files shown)


WMDE-Fisch:
Abandoned in favor of #407 - sorry, this PR was a quick and dirty approach using Github only ;-).

PR #407, opened 2018-01-05
Reload Twinklefluff interface when diff view changes

WMDE-Fisch:
This patch will hook Twinklefluff initialization to the mw.hook
responsible for pages containing a diff. Hooking is only done when
the methods for diff pages are loaded.

Now for example the RevisionSlider extension that fires that hook
will re-initiate the Twinkle links added to the page when moving
through the revisions.

This should fix https://github.com/azatoth/twinkle/issues/360

The hook gets fired everytime a diff page loads / reloads. To
avoid having the diff links twice the original method got re-
factored and the part addressing diff pages reacts only to the
hook now.

Further modules that might need to be re-initiated by hook
could be added in a follow up.


WMDE-Fisch:
The diff is a bit ugly. But I separated what looked like stuff affecting diff pages from stuff affecting other pages in Twinklefluff. Just put it into own methods and then applied the hooking on the part ( I assume ) loaded for diff pages.

WMDE-Fisch:
*poke*
@azatoth or anyone :-)

MusikAnimal:
Here's the diff with [whitespace removed](https://github.com/blog/967-github-secrets), makes it a bit easier to review: https://github.com/azatoth/twinkle/pull/407/files?diff=unified&w=1

I'm going to try to test this today.

modules/twinklefluff.js
```diff
@@ -29,173 +36,174 @@ Twinkle.fluff = {
...
-				revertNode.appendChild( vandNode );
+			ntitle.insertBefore( revertNode, ntitle.firstChild );
+		}
+	},
 
-				ntitle.insertBefore( revertNode, ntitle.firstChild );
-			}
+	normal: function() {
+		if( mw.config.get('wgNamespaceNumber') === -1 && mw.config.get('wgCanonicalSpecialPageName') === "Contributions" ) {
+			this.contributions();
```
atlight: Did you test this? I'm pretty sure this isn't called from an object context, so `this` will be undefined.

Amorymeltzer: Replacing `this` with `Twinkle.fluff` here and below should be sufficient, no?  A quick test suggests this works, @atlight @MusikAnimal were there any other issues either of you recall?

If so, I think we can handle the merge conflict on-wiki if @WMDE-Fisch isn't around

Amorymeltzer:
I want to get #511 merged first, since I think that will create some nasty conflicts, but once that's done I'll merge changes into this and throw it up on test.wiki; my initial tests looked good!

modules/twinklefluff.js
```diff
@@ -17,6 +17,13 @@
  Twinklefluff revert and antivandalism utility
  */
 
+var spanTag = function( color, content ) {
+  var span = document.createElement( 'span' );
```
Amorymeltzer:
```suggestion
	var span = document.createElement( 'span' );
```

modules/twinklefluff.js
Amorymeltzer: Yeah, this works - enabled on test.wiki

Amorymeltzer:
@WMDE-Fisch 	Finally got around to this — I've dealt with the merge conflicts here — but as I can't commit to your branch, if you could commit the two suggestions I've made (one necessary, one cosmetic) this looks good to go AFAICT.

WMDE-Fisch:
> @WMDE-Fisch Finally got around to this — I've dealt with the merge conflicts here — but as I can't commit to your branch, if you could commit the two suggestions I've made (one necessary, one cosmetic) this looks good to go AFAICT.

Thanks for the suggestions, I haven't looked into this for some time now and do not really have the time to test this atm. But if you're confident that its working feel free to merge.

modules/twinklefluff.js
```diff
@@ -75,126 +75,136 @@ Twinkle.fluff = {
...
+	},
 
-				ntitle.insertBefore( revertNode, ntitle.firstChild );
-			}
+	normal: function() {
+		if( mw.config.get('wgNamespaceNumber') === -1 && mw.config.get('wgCanonicalSpecialPageName') === "Contributions" ) {
+			Twinkle.fluff.contributions();
+		} else {
+			mw.hook( 'wikipage.diff' ).add( function () {
+				this.diff();
```
Amorymeltzer:
```suggestion
				Twinkle.fluff.diff();
```
@WMDE-Fisch 	This too

PR #411, opened 2018-03-08
Suppress edit summary pings by prefixing links with a colon (T189067)

MusikAnimal:
See https://phabricator.wikimedia.org/T189067

Soon any and every link to a userpage will ping that user. Some of the colons added with this commit probably aren't necessary but I think this is still okay, just as a matter of good practice.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -972,7 +972,7 @@ Twinkle.tag.callbacks = {
 				}
 			}
 
-			summaryText += ' {{[[';
+			summaryText += ' {{[[:';
 			if( tagName === 'globalize' ) {
 				summaryText += "Template:" + params.tagParameters.globalize + '|' + params.tagParameters.globalize;
 			} else {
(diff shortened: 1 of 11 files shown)


MusikAnimal:
I'm creating this PR in case other maintainers know of something I missed. I'm doing some thorough testing now. Please don't merge yet :)

MusikAnimal:
As far as I can tell this is doing what it's supposed to, but I have not tested every bit of code that has been changed. I think this is low risk... because we're only adding something to a string, not logic or syntax changes.

Edit summary pings goes live at 00:00, March 9 (UTC), so a few hours from now. I'm going to try to do a bit more testing, then merge and deploy before that time. I ran some queries and it's especially common for user pages to be tagged for speedy deletion/MfD, and UAA reports. These can issue multiple pings to the same user, assuming they were also given a talk page notification.

Sorry we weren't notified of this sooner!

mc10:
I took a brief glance at the changes and they look okay to me. I wonder if it'd make sense though to extract this into a function like the following:
```js
function wikiLink(page, name=null) {
  const pipedName = (name !== null ? `|${name}` : '');
  return `[[:${page}${pipedName}]]`;
}
```
just so we don't forget to add the `:` for all of our links. This is probably out-of-scope for this PR, but I think it's worth considering.

MusikAnimal:
> I wonder if it'd make sense though to extract this into a function

Totally. Refactoring is always a good thing (to me)! But indeed, that might be a bit too much for this PR, because it'd require a lot more testing and is more prone to error. I'll look into this soon though, because it is a great idea :)

With your "okay" I'm going to go ahead and merge and deploy, and monitor Twinkle edits closely via recent changes. I want some wiggle room between now and when the ping feature goes live, in case something goes wrong.

MusikAnimal:
The release of edit summary pings has been postponed till Monday. This gives me a lot more time to test, and maybe take on that refactoring, too :) Not going to deploy these Twinkle changes just yet, probably will tomorrow though.

Issue #278, opened 2015-06-22
Add uw-plotsum1 and uw-plotsum2


siphedious:
Could the above templates be added to the User Warnings/Behavior in articles dropdown? I've just added them to https://en.wikipedia.org/wiki/Wikipedia:Template_messages/User_talk_namespace after support at WT:FILM. I initiated a discussion at Wikipedia talk:Template messages/User talk namespace where nobody objected to the templates after 5 days.

Thanks!

atlight:
In order to avoid a excessively long list of templates in the warn dialog box, we only include a selection of the most common user warning template series. Little-used template series, such as uw-icsX, are omitted from Twinkle. (We never got around to pruning the single-issue lists, so they are as long as ever...)

As such I am a little hesitant to add these. However, I am also wondering whether we need to rethink our warning list - whether we should include more template series in the list...

siphedious:
Thanks for your response! I know at WT:FILM there's a fair amount of interest in these, and they would presumably be of interest to other projects as well, but I admittedly can't speak to the overarching amount of interest. I use them quite a bit myself.

Thanks for your consideration, either way!

Amorymeltzer:
Agree with the above, and better discussed in #412

PR #412, opened 2018-03-11
Reorganize user talk multi-level warnings

### LINKED ISSUE #402: Add uw-paid1 to uw-paid4
Could you please add the uw-paid warning templates to the UW module? 

TIA
SoWhy

rchard2scout:
The lists of warning in modules/twinklewarn.js were a mess, with many recent additions missing. I've completely revamped the list of multi-level warnings, and it's now completely up to date with [WP:UTM](https://en.wikipedia.org/wiki/Wikipedia:Template_messages/User_talk_namespace).

I've left the "Common warnings" as they were. After that, I've followed the categorization as on WP:UTM, and I've included templates in the subsections that are also in "Common warnings". This means that some templates are included twice, but I don't think that's a big issue. The only case where I haven't followed WP:UTM is that I've merged "Vandalism" and "General disruptive editing" into "Vandalism and disruptive editing", because those only contain 2 and 1 templates respectively. "Vandalism and disruptive editing" disappear for level 3 and higher, because they're completely redundant with "Common warnings".

This also closes #402.

I've only just noticed the discussion on #278, over 2 years ago, which said that little-used templates are excluded. I've decided to ignore that for now, I think that especially with this categorization, there aren't any excessively long lists.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -382,34 +382,44 @@ Twinkle.warn.messages = {
 	level2: {
 		"Common warnings": {
 			"uw-vandalism2": {
-				label: "Vandalism",
-				summary: "Caution: Unconstructive editing"
+				label: "Blatant vandalism",
+				summary: "Caution: Blatant vandalism"
 			},
 			"uw-disruptive2": {
 				label: "Disruptive editing",
-				summary: "Caution: Unconstructive editing"
+				summary: "Caution: Disruptive editing"
 			},
 			"uw-test2": {
 				label: "Editing tests",
 				summary: "Caution: Editing tests"
 			},
 			"uw-delete2": {
-				label: "Removal of content, blanking",
-				summary: "Caution: Removal of content, blanking"
+				label: "Removal of content without adequate explanation",
+				summary: "Caution: Removal of content without adequate explanation"
+			}
+		},
+		"Vandalism and disruptive editing": {
+			"uw-vandalism2": {
+				label: "Blatant vandalism",
+				summary: "Caution: Blatant vandalism"
+			},
+			"uw-subtle2": {
+				label: "Subtle vandalism",
+				summary: "Caution: Subtle vandalism"
+			},
+			"uw-disruptive2": {
+				label: "Disruptive editing",
+				summary: "Caution: Disruptive editing"
 			}
 		},
 		"Behavior in articles": {
-			"uw-biog2": {
-				label: "Adding unreferenced controversial information about living persons",
-				summary: "Caution: Adding unreferenced controversial information about living persons"
+			"uw-blank2": {
+				label: "Page blanking",
+				summary: "Caution: Page blanking"
 			},
-			"uw-defamatory2": {
-				label: "Addition of defamatory content",
-				summary: "Caution: Addition of defamatory content"
-			},
-			"uw-error2": {
-				label: "Introducing deliberate factual errors",
-				summary: "Caution: Introducing factual errors"
+			"uw-delete2": {
+				label: "Removal of content without adequate explanation",
+				summary: "Caution: Removal of content without adequate explanation"
 			},
 			"uw-genre2": {
 				label: "Frequent or mass changes to genres without consensus or references",
@@ -423,10 +433,6 @@ Twinkle.warn.messages = {
 				label: "Using improper humor in articles",
 				summary: "Caution: Using improper humor in articles"
 			},
-			"uw-nor2": {
-				label: "Adding original research, including unpublished syntheses of sources",
-				summary: "Caution: Adding original research, including unpublished syntheses of sources"
-			},
 			"uw-notcensored2": {
 				label: "Censorship of material",
 				summary: "Caution: Censorship of material"
@@ -435,27 +441,25 @@ Twinkle.warn.messages = {
 				label: "Ownership of articles",
 				summary: "Caution: Ownership of articles"
 			},
+			"uw-plotsum2": {
+				label: "Inappropriate expansion of plot summaries",
+				summary: "Caution: Inappropriate expansion of plot summaries"
+			},
+			"uw-taxonomy2": {
+				label: "Disrupting the taxonomy templates",
+				summary: "Caution: Disrupting the taxonomy templates"
+			},
 			"uw-tdel2": {
 				label: "Removal of maintenance templates",
 				summary: "Caution: Removal of maintenance templates"
 			},
-			"uw-unsourced2": {
-				label: "Addition of unsourced or improperly cited material",
-				summary: "Caution: Addition of unsourced or improperly cited material"
-			}
-		},
-		"Promotions and spam": {
-			"uw-advert2": {
-				label: "Using Wikipedia for advertising or promotion",
-				summary: "Caution: Using Wikipedia for advertising or promotion"
+			"uw-thumb2": {
+				label: "Using thumbnails in Infoboxes",
+				summary: "Caution: Using thumbnails in Infoboxes"
 			},
-			"uw-npov2": {
-				label: "Not adhering to neutral point of view",
-				summary: "Caution: Not adhering to neutral point of view"
-			},
-			"uw-spam2": {
-				label: "Adding spam links",
-				summary: "Caution: Adding spam links"
+			"uw-upload2": {
+				label: "Uploading unencyclopedic images",
+				summary: "Caution: Uploading unencyclopedic images"
 			}
 		},
 		"Behavior towards other editors": {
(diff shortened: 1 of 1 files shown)


modules/twinklewarn.js
```diff
@@ -537,74 +649,126 @@ Twinkle.warn.messages = {
...
 			"uw-spam3": {
 				label: "Adding spam links",
 				summary: "Warning: Adding spam links"
 			}
 		},
-		"Behavior towards other users": {
-			"uw-agf3": {
-				label: "Not assuming good faith",
-				summary: "Warning: Not assuming good faith"
+		"Inserting factual inaccuriacies and libel": {
```
mc10: Ditto.

modules/twinklewarn.js
```diff
@@ -168,34 +168,44 @@ Twinkle.warn.messages = {
 	level1: {
 		"Common warnings": {
 			"uw-vandalism1": {
-				label: "Vandalism",
-				summary: "General note: Unconstructive editing"
+				label: "Blatant vandalism",
+				summary: "General note: Blatant vandalism"
```
mc10: I think this is a bit strong for [uw-vandalism1](https://en.wikipedia.org/wiki/Template:Uw-vandalism1); I'd prefer "Unconstructive editing" for this level, but it seems okay for later levels.

modules/twinklewarn.js
```diff
@@ -168,34 +168,44 @@ Twinkle.warn.messages = {
...
 			}
 		},
-		"Behavior in articles": {
-			"uw-biog1": {
-				label: "Adding unreferenced controversial information about living persons",
-				summary: "General note: Adding unreferenced controversial information about living persons"
+		"Vandalism and disruptive editing": {
+			"uw-vandalism1": {
+				label: "Blatant vandalism",
+				summary: "General note: Blatant vandalism"
```
mc10: Same comment as above.

modules/twinklewarn.js
```diff
@@ -816,17 +1006,21 @@ Twinkle.warn.messages = {
 				summary: "Only warning: Adding spam links"
 			}
 		},
-		"Behavior towards other editors": {
-			"uw-harass4im": {
-				label: "Harassment of other users",
-				summary: "Only warning: Harassment of other users"
+		"Inserting factual inaccuriacies and libel": {
```
mc10: Ditto.

modules/twinklewarn.js
```diff
@@ -262,33 +270,95 @@ Twinkle.warn.messages = {
...
+			"uw-paid1": {
+				label: "Paid editing without disclosure under the Wikimedia Terms of Use",
+				summary: "General note: Paid editing without disclosure under the Wikimedia Terms of Use"
+			},
+			"uw-spam1": {
+				label: "Adding inappropriate external links",
+				summary: "General note: Adding inappropriate external links"
+			}
+		},
+		"Inserting factual inaccuriacies and libel": {
```
mc10: s/inaccuriacies/inaccuracies

modules/twinklewarn.js
```diff
@@ -420,37 +488,99 @@ Twinkle.warn.messages = {
...
+			"uw-paid2": {
+				label: "Paid editing without disclosure under the Wikimedia Terms of Use",
+				summary: "Caution: Paid editing without disclosure under the Wikimedia Terms of Use"
+			},
+			"uw-spam2": {
+				label: "Adding spam links",
+				summary: "Caution: Adding spam links"
+			}
+		},
+		"Inserting factual inaccuriacies and libel": {
```
mc10: Ditto.

modules/twinklewarn.js
```diff
@@ -697,52 +853,82 @@ Twinkle.warn.messages = {
...
 			"uw-spam4": {
 				label: "Adding spam links",
 				summary: "Final warning: Adding spam links"
 			}
 		},
-		"Behavior towards other editors": {
-			"uw-harass4": {
-				label: "Harassment of other users",
-				summary: "Final warning: Harassment of other users"
+		"Inserting factual inaccuriacies and libel": {
```
mc10: Ditto.

modules/twinklewarn.js
rchard2scout: I took the description straight from WP:UTM, but I agree that this wording is too strong for uw-vandalism1.

modules/twinklewarn.js
```diff
@@ -221,27 +227,29 @@ Twinkle.warn.messages = {
...
 				summary: "General note: Ownership of articles"
 			},
+			"uw-plotsum1": {
+				label: "Inappropriate expansion of plot summaries",
+				summary: "General note: Inappropriate expansion of plot summaries"
+			},
+			"uw-taxonomy1": {
+				label: "Disrupting the taxonomy templates",
+				summary: "General note: Disrupting the taxonomy templates"
+			},
```
MusikAnimal: This is certainly a rarity (broadly speaking), and doesn't need to be cluttering the Twinkle interface.

modules/twinklewarn.js
```diff
@@ -221,27 +227,29 @@ Twinkle.warn.messages = {
...
+			"uw-test1": {
+				label: "Editing tests",
+				summary: "General note: Editing tests"
 			},
-			"uw-npov1": {
-				label: "Not adhering to neutral point of view",
-				summary: "General note: Not adhering to neutral point of view"
+			"uw-thumb1": {
+				label: "Using thumbnails in Infoboxes",
+				summary: "General note: Using thumbnails in Infoboxes"
```
MusikAnimal: Questionable, also pretty rare?

modules/twinklewarn.js
```diff
@@ -262,33 +270,95 @@ Twinkle.warn.messages = {
...
-		},
-		"Other": {
 			"uw-chat1": {
 				label: "Using talk page as forum",
 				summary: "General note: Using talk page as forum"
 			},
+			"uw-color1": {
+				label: "Adding non-compliant colours",
+				summary: "General note: Adding non-compliant colours"
+			},
```
MusikAnimal: Not for Twinkle, I don't think... how many people know when to use this?

modules/twinklewarn.js
```diff
@@ -297,63 +367,63 @@ Twinkle.warn.messages = {
...
-			"uw-upload1": {
-				label: "Uploading unencyclopedic images",
-				summary: "General note: Uploading unencyclopedic images"
-			}
-		}/*,
-		"To be removed from Twinkle": {
 			"uw-redirect1": {
 				label: "Creating malicious redirects",
 				summary: "General note: Creating malicious redirects"
 			},
```
MusikAnimal: I agree with the original code that this probably doesn't belong in Twinkle. How often does this happen? Wouldn't "uw-vandalism1" or the like work just as well, if not the speedy deletion notice?

SoWhy:
What's the status on this? Handling of #402 is pretty important imho since this is a problem more and more

MusikAnimal:
I don't oppose reorganization but I do oppose adding templates that are rarely used. This is just my opinion of course.

SoWhy:
@MusikAnimal I agree. @rchard2scout Would you be willing to make the Changes MusikAnimal requested? That way, the uncontroversial changes can be merged and the rest can be discussed separately without holding up necessary additions like uw-paid[1-4]

ToBeFree:
Current changes: https://github.com/azatoth/twinkle/pull/412/files

@MusikAnimal: These seem to be okay, except for the commented parts, which I agree with. Can we merge this pull request and then remove the unnecessary lines afterwards, so that there is no further action needed by rchard2scout?

This closes #423 as well

modules/twinklewarn.js
```diff
@@ -168,34 +168,44 @@ Twinkle.warn.messages = {
...
+			"uw-vandalism1": {
+				label: "Unconstructive editing",
+				summary: "General note: Unconstructive editing"
 			},
-			"uw-defamatory1": {
-				label: "Addition of defamatory content",
-				summary: "General note: Addition of defamatory content"
+			"uw-subtle1": {
+				label: "Subtle vandalism",
+				summary: "General note: Subtle vandalism"
```
ToBeFree: While okay in the Twinkle menu, "vandalism" should not appear in the edit summary for level 1

ToBeFree: I'd fix this after merging instead of waiting for a fix here.

modules/twinklewarn.js
```diff
@@ -297,63 +367,63 @@ Twinkle.warn.messages = {
...
 	},
 
 
 	level2: {
 		"Common warnings": {
 			"uw-vandalism2": {
-				label: "Vandalism",
-				summary: "Caution: Unconstructive editing"
+				label: "Blatant vandalism",
+				summary: "Caution: Blatant vandalism"
```
ToBeFree: I think that this is still too strong for level 2, and belongs to level 3.

I'd fix this after merging instead of waiting for a fix here.

modules/twinklewarn.js
```diff
@@ -459,33 +589,27 @@ Twinkle.warn.messages = {
...
 	},
 
 
 	level3: {
 		"Common warnings": {
 			"uw-vandalism3": {
-				label: "Vandalism",
-				summary: "Warning: Vandalism"
+				label: "Blatant vandalism",
+				summary: "Warning: Blatant vandalism"
```
ToBeFree: "Blatant" is used as information and clarification in the Twinkle interface. It does not belong into the public edit summary.

modules/twinklewarn.js
```diff
@@ -537,74 +649,126 @@ Twinkle.warn.messages = {
...
+				label: "Not assuming good faith",
+				summary: "Warning: Not assuming good faith"
+			},
+			"uw-harass3": {
+				label: "Harassment of other users",
+				summary: "Warning: Harassment of other users"
+			},
+			"uw-npa3": {
+				label: "Personal attack directed at a specific editor",
+				summary: "Warning: Personal attack directed at a specific editor"
```
ToBeFree: Almost sounds inviting: Next time, please attack multiple editors instead. 😉 So this, too, is not useful in the edit summary but rather only in the Twinkle interface.

modules/twinklewarn.js
ToBeFree: Feel free to merge unmodified, I'll fix this shortly afterwards. It won't hurt and seems to be the easiest and fastest solution. 😄

modules/twinklewarn.js
ToBeFree: Feel free to merge unmodified, I'll fix this shortly afterwards. It won't hurt and seems to be the easiest and fastest solution. 😄

ToBeFree:
There is a "squash and merge" button, by the way -- no need to wait for a squash by the proposer

atlight:
I'm not a fan of this pull request, I have to say. I'd rather start from scratch. My main concerns are: why are things being shuffled from category to category (this is rather user-hostile), and why are lots of rarely-used templates being added? A lot of careful thought went into curating this list about 7 years ago, and while I'm certainly not opposed to adding uw-paid (since paid editing has emerged as a key issue since then) I would prefer to see rationales for the other templates. Certainly I cannot merge a pull request that adds templates only of use to a few editors in narrow situations, like uw-taxonomy1.

ToBeFree:
@atlight I completely agree, and I would like to remove these unnecessary templates from the list after the request is merged. They won't do harm when existing there for a few hours/days, maybe even without being copied over to Wikipedia yet. It's just that we have conflicting pull requests right now, and this would easily fix the problem.

ToBeFree:
General restructuring of the list seems to be okay to me. I personally would merge the request and remove all the unnecessary templates afterwards, to fix the current problem of two conflicting pull requests.

atlight:
There's nothing inherently wrong with two conflicting pull requests; I wouldn't stress about that if I were you. 

@rchard2scout: If you are interested in addressing the issues in my comment above and the review comments, please let us know. Otherwise we can close it and work on this issue elsewhere.

Amorymeltzer:
I've opened #533 to be a better, centralized location for the discussion around inclusion of warning templates cc @atlight @MusikAnimal @ToBeFree @kevinji @SoWhy @rchard2scout

PR #428, opened 2018-08-03
unlink: restrict access to extendedconfirmed users

MusikAnimal:


MusikAnimal:
For context, see https://en.wikipedia.org/wiki/Special:Permalink/853306352#Restricting_the_use_of_the_Unlink_function

This won't stop them from accessing the tool via the JavaScript console, but I think that's fine... if they're that tech-savvy they'll find a way around any additional checks we add.

PR #432, opened 2018-08-09
twinklewarn - Treat userlink properly for uw-aiv

### LINKED ISSUE #404: Edit summary bug
Easy fix I think... just need these edit summaries to use [[User:]] instead of just a regular wikilink: `https://en.wikipedia.org/w/index.php?title=User_talk:FULBERT&diff=prev&oldid=816923288`

EDIT: Account verified as Coffee@enwiki here: `https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&diff=prev&oldid=816924130`

Amorymeltzer:
Fixes #404
The function to change the input label for certain warnings (change_subcategory) is only called when changing items in the submenu, so since uw-aiv is the first item, it won't show unless you change submenu items rather than the main category menu.  This is obviously less-than-ideal, but still an improvement over the current situation.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1215,9 +1215,10 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	// clear overridden label on article textbox
 	Morebits.quickForm.setElementTooltipVisibility(e.target.root.article, true);
 	Morebits.quickForm.resetElementLabel(e.target.root.article);
-
 	// hide the big red notice
 	$("#tw-warn-red-notice").remove();
+	// add custom label.redWarning
+	Twinkle.warn.callback.change_subcategory(e);
 };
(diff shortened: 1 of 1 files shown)


MusikAnimal:
I haven't tested, but would it work to call `Twinkle.warn.callback.change_subcategory(e)` at the [bottom of the `createEntries` function](https://github.com/azatoth/twinkle/blob/e573e07956a49e705568c1596db07afe6ebd73c0/modules/twinklewarn.js#L1173)? From the looks of it, `e.target.form` would be same so `twinklewarnCallbackChangeSubcategory` won't care. Still kind of sloppy, I suppose.

Amorymeltzer:
Putting it at the end of `Twinkle.warn.callback.change_category` did the trick; there're some cleanup functions there.

PR #433, opened 2018-08-10
protect - A few fixes for tagging redirect pages

### LINKED ISSUE #408: Protection template tagging breaks redirects
This has been happening for ages, e.g. https://en.wikipedia.org/wiki/Special:Diff/825759276

Amorymeltzer:
- Place tag underneath the #REDIRECT header (don't break the redirect, remove need for `<noinclude>`)
- Detect {{Rcat shell (and sundry redirects) and do nothing if present

Closes #408

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -1382,10 +1382,16 @@ Twinkle.protect.callbacks = {
 		if( params.tag === 'none' ) {
 			summary = 'Removing protection template' + Twinkle.getPref('summaryAd');
 		} else {
-			if( params.noinclude ) {
+			if( Morebits.wiki.isPageRedirect() ) {
+				//Only tag if no {{rcat shell}} is found
+				if (!text.match(/{{(?:redr|this is a redirect|r(?:edirect)?(?:.?cat.*)?[ _]?sh)/i)) {
+					text = text.replace(/#REDIRECT ?(\[\[.*?\]\])(.*)/i, "#REDIRECT $1$2\n\n{{" + tag + "}}");
+				} else {
+					Morebits.status.info("Redirect category shell present", "nothing to do");
+					return;
+				}
+			} else if( params.noinclude ) {
 				text = "<noinclude>{{" + tag + "}}</noinclude>" + text;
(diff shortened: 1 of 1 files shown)


MusikAnimal:
Beautiful! I love the "Redirect category shell present, nothing to do" message.

I did notice an unrelated bug -- it's not reporting that the redirect is currently or was previously protected. This is at https://test.wikipedia.org/w/index.php?title=Redirect&redirect=no. I'm thinking this might be because testwiki doesn't have pending changes enabled. I see the same issue with the current twinkleprotect code (without your patch), so I think we're safe to merge :)

Another question for you, out of curiosity -- how do you go about Twinkle development? What I've been doing is copying/pasting the whole module (twinkleprotect.js in this case) into the JS console on testwiki, and testing it that way. Far quicker than updating the MediaWiki JS page, and having to wait for ResourceLoader to catch up. I wasn't sure if you had another tactic that I haven't tried.

Amorymeltzer:
Thanks, and yeah, that sounds right.  I would guess it's [L101](https://github.com/azatoth/twinkle/blob/master/modules/twinkleprotect.js#L101) that's returning undefined?  Checking if `wgFlaggedRevsParams` exists ahead of time should work, IIRC.  Honestly, I can never remember when to use PC, flagged, or stabilize.

And yup, that's what I've been doing, except on enWiki.  Convenient enough, but still kind of a pain.

Amorymeltzer:
Could do this:
````javascript
if (mw.config.exists('wgFlaggedRevsParams')) {
	var stableDeferred = api.get({
		format: 'json',
		action: 'query',
		list: 'logevents',
		letype: 'stable',
		letitle: mw.config.get('wgPageName')
	});
} else { //Fake some data
	var stableDeferred = {};
	stableDeferred[0] = {query: {logevents: ''}};
}
````
Ugly.  Instead of the `else` block faking data, maybe it'd be neater to wrap the assignation in a `typeof`:
````javascript
if (typeof stableData !== 'undefined') {
	Twinkle.protect.hasStableLog = !!stableData[0].query.logevents.length;
}
````

PR #436, opened 2018-09-02
speedy: A few tweaks to CSD logging

Amorymeltzer:
Two different behaviors fixed:
- A few edge cases exist where twinkle won't notify the original creator, but it would still erroneously say you had notified the user in your log.
- Avoid using the title of a page nominated for G10.  The name of G10 noms was removed from the notification edit summary in e296afc roughly six years ago but it was still present in the edit summary for the log.  That's been fixed.  This also removes the title from displaying in the log itself, hiding it behind a piped wikilink.

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1470,14 +1470,17 @@ Twinkle.speedy.callbacks = {
 					// disallow warning yourself
 					if (initialContrib === mw.config.get('wgUserName')) {
 						Morebits.status.warn("You (" + initialContrib + ") created this page; skipping user notification");
+						initialContrib = null;
 
 					// don't notify users when their user talk page is nominated
 					} else if (initialContrib === mw.config.get('wgTitle') && mw.config.get('wgNamespaceNumber') === 3) {
 						Morebits.status.warn("Notifying initial contributor: this user created their own user talk page; skipping notification");
+						initialContrib = null;
 
 					// quick hack to prevent excessive unwanted notifications, per request. Should actually be configurable on recipient page...
 					} else if ((initialContrib === "Cyberbot I" || initialContrib === "SoxBot") && params.normalizeds[0] === "f2") {
 						Morebits.status.warn("Notifying initial contributor: page created procedurally by bot; skipping notification");
+						initialContrib = null;
 
 					} else {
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
While I think hiding the name/link of a G10 nom in the log's edit summary should be uncontroversial, I did go out on a limb in removing the display of the page in the log.  Theoretically, those things don't get viewed much unless someone is at RfA or has had a rash of poor taggings, but likewise it doesn't cost much to hide the title.  The link should remain for the sake of transparency (and actually being a log!) so I think I struck a good middle ground on the abundance of caution front, but I'd appreciate at least a second on that front.  Easy to remove if undesired.

Moreover, one other potential source is the notification logging itself.  If the username itself is a problem, that will still get logged.  I thought about adding in an additional check to change the log to say "notified the `{{user|1=BadName|2=page creator}}` if:
- G10 nom
- page nominated is User:BadName
- page creator is User:BadName

But that seemed to be too much, and would catch too many false positives while missing a lot of cases.  Not sure there's a good way around it, but wanted to float it out there regardless.

MusikAnimal:
Looks great! I did some tests on testwiki and it checks out. Per usual I can't find anything wrong with your code. Merging!

modules/twinklespeedy.js
```diff
@@ -1566,7 +1569,12 @@ Twinkle.speedy.callbacks = {
 				appendText += "\n\n=== " + date.getUTCMonthName() + " " + date.getUTCFullYear() + " ===";
 			}
 
-			appendText += "\n# [[:" + Morebits.pageNameNorm + "]]: ";
+			appendText += "\n# [[:" + Morebits.pageNameNorm;
+			if (params.normalizeds.indexOf("g10") === -1) {  // no article name in log for G10 taggings
```
siddharthvp: @Amorymeltzer @MusikAnimal TypeError: Cannot read property 'indexOf' of undefined is raised on this line when trying to CSD a file under F11 rationale (and probably with other rationales also). First reported at [TW talk page](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Most_CSDs_not_being_added_to_log)

Amorymeltzer: Could've sworn I checked all the di options, but yeah, looks like I should've tested `params.normalizeds` first/as well. #460 should take care of it?  The one below this throws as well.

PR #437, opened 2018-09-03
speedy: Correct language for G6 db-disambig

Amorymeltzer:
As requested on [WT:TW](https://en.wikipedia.org/w/index.php?oldid=857841491#Db-disambig), this corrects the db-disambig tooltip to say that it applies to a page that dabs one page, not two or fewer.

Original text was introduced in f92641ebf02c60a8d635fa1e80edb93dca43389b, which was [correct at the time](https://en.wikipedia.org/w/index.php?title=Template:Db-disambig&oldid=353013319).


mc10:
Looks good to me, thanks!

Issue #215, opened 2014-03-29
Special:Contributions throws error


hartman:
https://en.wikipedia.org/w/index.php?title=Special%3AContributions

TypeError: 'undefined' is not an object (evaluating '$('#contentSub').find('a[title^="Special:Log"]').last().attr("href").replace')

PR #439, opened 2018-09-07
friendly and fluff: Make use of wgRelevantUserName

Amorymeltzer:
wgRelevantUserName was added to MW (v1.23, https://gerrit.wikimedia.org/g/mediawiki/core/+/7127c1792bb9205d7313ffabb210963c485a2ebc) in late 2013, well after the friendly modules were brought in to the repo (072860c).  Some uses were later removed as part of a morebits.js tidying (37e9ded), but wgRelevantUserName is a much cleaner way of getting the username.

The changes in twinklefluff.js are a little more involved, but amount to the same thing.  Here, the over-complicated code for getting the username has been around for even longer, but the major relevant change was in response to #215.  Basically, the code used to look for a log link and then calculate the username based off that.  While it obviously still works, wgRelevantUserName is easier.  #215 dealt with the fact that loading on an empty contribs page ('Special:Contributions/') broke things, which this still handles via mw.config.exists.  twinklefluff doesn't care if the username doesn't exist, though, but that doesn't seem to cause any problems.


MusikAnimal:
I'm curious -- how did you go about testing these changes? Fluff loads immediately on page load (adding the [rollback] links, etc.), so my usual tactic to copy/paste in the JS console broke things. I updated MediaWiki:Gadget-Twinklefluff.js with the latest and all seems fine... so I think we're okay, but it's possible ResourceLoader still hasn't caught up.

Amorymeltzer:
Don't really remember, but I think in order to check IPs I turned off Twinkle/turned on safemode on enWiki and loaded all the twinkle gadgets in.  I'll double check again tomorrow.

Amorymeltzer:
I think I actually did just plop fluff into the console — what broke for you?  I get duplicate links, and with some well-placed returns, color changes, etc., it seems to work just fine.

* https://test.wikipedia.org/w/index.php?title=Page542&diff=364076&oldid=363863
* https://en.wikipedia.org/w/index.php?title=Sidra_(name)&diff=prev&oldid=867387850

Amorymeltzer:
You need to run `Twinkle.fluff.init()` after you dump the code into the console.  That should create new links, which work: https://test.wikipedia.org/wiki/Special:Diff/364114

Did you check just on diffs?  'Cause this *should* change `fluff`'s behavior just on contribs pages.  Still need to `init()` though: turns out the old links don't break on contribs when you paste the new code in.

Amorymeltzer:
Actually, @MusikAnimal, with a slight modification to a mediawiki footer page, we can get the contribs links to show up for IP ranges (at least until/if https://phabricator.wikimedia.org/T206954 happens).  It seems to work without any console errors (http://en.wikipedia.org/wiki/Special:Diff/867489159 and http://en.wikipedia.org/wiki/Special:Diff/867472737) although for obvious reasons it doesn't load a userpage to provide a warning.  Worthwhile?

MusikAnimal:
So it would be dependent on an on-wiki configuration? I'm okay with it, if we leave a comment in the wikitext clarifying this dependency. I'd also leave a comment in the Twinkle code somewhere saying what needs to be changed on-wiki, so that other wikis who adapt our code will know how to make it work. And of course, add some sort of fail-safe if there's a chance the JS will break if the MediaWiki message isn't set up correctly.

PR #441, opened 2018-09-11
speedy: Add user-specified CSD parameters to the CSD log

### LINKED ISSUE #136: CSD log should be more verbose
It would be nice if the CSD log produced by Twinkle were more verbose.  In particular, I tag a lot of articles as CSD G5 (creation by banned user).  Twinkle prompts me for the name of the banned user, which it adds to the CSD template.  It would be nice if this information were also added to the log.  This would make reviewing the log very useful for identifying sockpuppets which recreate their own deleted articles.  Similarly, if Twinkle logged the URL prompted for CSD G12s, then reviewing the log for recreated articles would make it easier to find repeat copyvios.

I suppose that in general, whenever Twinkle prompts for further information when tagging an article for CSD, this information should be added to the log.

Amorymeltzer:
Closes #136
Improves/expands the CSD log with the various parameters that a user
may enter upon nomination (e.g., custom G7 rationale, multiple G12 urls, etc.).
Basically just abuses params.templateParams[0] (which is itself an array),
with most of the complexity/duplication coming from the fact that DI has
an entirely different data structure, G12 can have multiple options unlike
most of the others, and the custom db option is of course different as well.

Ignores:
- G13 ts
- G10 blanked
- Check boxes from DI (non_free, derivative)

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1579,11 +1579,24 @@ Twinkle.speedy.callbacks = {
 				appendText += "\n\n=== " + date.getUTCMonthName() + " " + date.getUTCFullYear() + " ===";
 			}
 
+			var formatParamLog = function(normalize, csdparam, input) {
+				return ' [' + normalize + ' ' + csdparam + ': ' + input + ']';
+			};
+
+			var extraInfo = '';
 			var editsummary = "Logging speedy deletion nomination";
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Updated version to resolve conflicts and with a slightly cleaner way of doing things.  #546, #536, and #543 should handle the unnamed parameters; this should probably wait until those are merged.

PR #442, opened 2018-09-12
twinkleprotect: Remove expiry param from tag

### LINKED ISSUE #363: Expiry param no longer needed in protection templates
The expiry of pages under protection or pending changes is now automatically detected by the protection template, so Twinkle no longer has to add it as a param when adding the protection template.
See https://en.wikipedia.org/wiki/Module:Effective_protection_expiry.

Amorymeltzer:
Closes #363


MusikAnimal:
Thanks!

PR #445, opened 2018-10-24
friendlytag: Group old and new redirect tags within Rcat shell

### LINKED ISSUE #444: Redirect categories not placed in existing redirect category shells.
Redirect category templates added using Twinkle are not being placed in redirect category shell templates that were already on the page. I don't see the point of Rcat shells, but if one is on a page, Twinkle should add new tags to them. Otherwise they look weird and probably cause errors. 

Instance:
[https://en.wikipedia.org/w/index.php?title=Template:R_from_event&oldid=861784966
](url)

Amorymeltzer:
Closes #444.  While processing new tags on a redirect page, check the page for
already-existing redirect categories.  If they exist, whether within or
without an Rcat shell, fold them and the new tags into an Rcat shell.

There is some double-processing, as the already-existing method to check for
duplication remains, but the intent of this is not to affect 'tags' so as not
to affect the edit summary.  Moreover, the old tags remain unsorted, mainly to
avoid creating a broader-scope-than-necessary variable, but is probably a good
idea anyway, to avoid the perception that those tags were added by the user.

Rcat regex courtesy of [[User:Kephir/gadgets/sagittarius.js]] and #433

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1071,7 +1071,24 @@ Twinkle.tag.callbacks = {
 		$.each(tags, addTag);
 
 		if( Twinkle.tag.mode === 'redirect' ) {
-			pageText += tagText;
+			// Check for all Rcat shell redirects (from #433)
+			if (pageText.match(/{{(?:redr|this is a redirect|r(?:edirect)?(?:.?cat.*)?[ _]?sh)/i)) {
+				// Regex courtesy [[User:Kephir/gadgets/sagittarius.js]] at [[Special:PermaLink/831402893]]
+				var oldTags = pageText.match(/(\s*{{[A-Za-z ]+\|)((?:[^|{}]*|{{[^|}]*}})+)(}})\s*/i);
+				pageText = pageText.replace(oldTags[0], oldTags[1]+tagText+oldTags[2]+oldTags[3]);
+			} else {
+				// Fold any pre-existing Rcats into taglist and under Rcatshell
+				var pageTags = pageText.match(/\n{{R(?:edirect)? .*?}}/img);
+				var oldPageTags ='';
+				if (pageTags) {
+					pageTags.forEach(function(pageTag) {
+						var pageRe = new RegExp(pageTag, 'img');
+						pageText = pageText.replace(pageRe,'');
+						oldPageTags = oldPageTags.concat(pageTag);
+					});
+				}
+				pageText += '\n{{Redirect category shell|' + tagText + oldPageTags + '\n}}';
+			}
 		} else {
 			// smartly insert the new tags after any hatnotes. Regex is a bit more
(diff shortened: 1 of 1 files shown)


siddharthvp:
Amazing work with the regexes there! 

Everything seems to be working as intended. @MusikAnimal merge maybe? This is a very helpful and belated update, considering that {{redr}} has been around for a long time now.

MusikAnimal:
As the one clicking the merge button, I should do some testing myself. There's a long list of PRs to review, some much older than this one. I'll try to get to this one soon!

I'd love to have some help from more collaborators, though. @Amorymeltzer has been contributing for a while. I'd like to have them added as a maintainer so they can at least help the non-Amory PRs, if they are interested? :)

Amorymeltzer:
Sure — happy to help out where/when I can with some reviews, etc.

MusikAnimal:
Pinging @azatoth about adding Amorymeltzer as a collaborator. I also emailed you.

Now, to review this PR...

modules/friendlytag.js
```diff
@@ -1074,7 +1074,24 @@ Twinkle.tag.callbacks = {
...
+			if (pageText.match(/{{(?:redr|this is a redirect|r(?:edirect)?(?:.?cat.*)?[ _]?sh)/i)) {
+				// Regex courtesy [[User:Kephir/gadgets/sagittarius.js]] at [[Special:PermaLink/831402893]]
+				var oldTags = pageText.match(/(\s*{{[A-Za-z ]+\|)((?:[^|{}]*|{{[^|}]*}})+)(}})\s*/i);
+				pageText = pageText.replace(oldTags[0], oldTags[1]+tagText+oldTags[2]+oldTags[3]);
+			} else {
+				// Fold any pre-existing Rcats into taglist and under Rcatshell
+				var pageTags = pageText.match(/\n{{R(?:edirect)? .*?}}/img);
+				var oldPageTags ='';
+				if (pageTags) {
+					$.each(pageTags, function(pageIndex, pageTag) {
```
MusikAnimal: Twinkle is littered with `$.each` (which we should fix), but from here on out I would use the native JavaScript [`Array.prototype.forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach). So something like `pageTags.forEach(function (pageTag) {`. This should work on IE 9+. MediaWiki itself no longer has official JavaScript support for anything below IE 11 anyway.

Amorymeltzer: I'll take care of this one, you do the rest, deal?

On a more serious note, might it be worth putting together some sort of contributing.md guide?

MusikAnimal: Yeah we'll remove all the other $.each's in a separate PR. Not urgent.

I agree a contributing guide is needed. Want to open an issue for that? We should rope in all maintainers.

PR #447, opened 2018-10-26
ARV: Don't report without either a sockmaster or one sockpuppet

### LINKED ISSUE #372: Do not file SPI cases with no sockpuppets listed
Make sure that a sockmaster and atleast one sockpuppet is listed when filing SPI cases.

Faced a problem here: https://en.wikipedia.org/wiki/Wikipedia:Sockpuppet_investigations/Ascar123

Amorymeltzer:
Closes #372
Removes the autofile suggestion, which while convenient, was likely unhelpful in the long run.

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -651,10 +651,11 @@ Twinkle.arv.callback.evaluate = function(e) {
 
 			var puppetReport = form.category.value === "puppet";
 			if (puppetReport && !(form.sockmaster.value.trim())) {
-				if (!confirm("You have not entered a sockmaster account for this puppet. Do you want to report this account as a sockpuppeteer instead?")) {
-					return;
-				}
-				puppetReport = false;
+				alert("You have not entered a sockmaster account for this puppet. Consider reporting this account as a sockpuppeteer instead.");
+				return;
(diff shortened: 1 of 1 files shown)


modules/twinklearv.js
```diff
@@ -651,10 +651,9 @@ Twinkle.arv.callback.evaluate = function(e) {
 
 			var puppetReport = form.category.value === "puppet";
 			if (puppetReport && !(form.sockmaster.value.trim())) {
-				if (!confirm("You have not entered a sockmaster account for this puppet. Do you want to report this account as a sockpuppeteer instead?")) {
-					return;
-				}
-				puppetReport = false;
+				alert("You have not entered a sockmaster account for this puppet. Consider reporting this account as a sockpuppeteer instead.");
+			} else if (!puppetReport && !(form.sockpuppet[0].value.trim())) {
+				alert("You have not entered any sockpuppet account(s) for this sockmaster. Consider reporting this account as a sockpuppet instead.");
```
MusikAnimal: Shouldn't we `return` after these alerts?

Amorymeltzer: Ugh yes, absolutely: frequently when testing I insert `return`s to slow things down, etc.  Got in the habit of removing them all!  Fixed.

Issue #217, opened 2014-04-20
Changes to default protection durations


Callanecc:
SEMI-PROTECTION PRESETS
At the moment (when selecting any semi-protection as a preset) the default move protection is an indefinite full protection. Wikipedia's default is to duplicate what is currently selected as the edit protection. Given that would it be possible for the deafultWould it be possible for the Twinkle protection module to do the same?

FULL-PROTECTION PRESETS
The same could be done for full protection presets, for example rather than the default move protection length being indefinite can it be the same as the edit protection default (which is currently 2 days)?

Thanks,
Callanecc

atlight:
@MusikAnimal Is this still an issue?

Callanecc:
Whether this is an issue to be fixed is obviously up to you. But Twinkle still has the same behaviour as when I initiated the request.

MusikAnimal:
It is true Twinkle differs from the normal form in the default duration of move protection, but I have to ask is this truly desirable? When I protect against moves, I'm usually inclined to believe it's not likely to change moving forward. Obviously not always the case, but the prime example simply being vandalism. E.g. `Google` being changed to `Google on wheels!`. Well that's it, we've reason to protect, so indef it is. Or `U.S. Open` being changed to `US Open`. That could also go to `United States Open` etc, let's put a halt to the ambiguity and force discussion once we've seen it's an issue. Page moves should be infrequent, is my thought.

So for Twinkle I just think we should default to whatever is most convenient or the most common scenario. If we feel that's a temporary duration we can make it happen, no problem. What do you think?

Callanecc:
True, however if you're protecting due to an edit war why would you want indefinite move protection?

MusikAnimal:
Ah, sorry, I should have read what you say more carefully =P With full protection against edit warring it does default to indef move protection, when page moves may have nothing to do with it. In that case yes it's quite sensible to have the durations be the same. Same if you semi-protect then check the move protection, it should only semi-protect the move for the same duration. Correct? Meanwhile, selecting move protection by itself will default to indefinite sysop. I guess that equates to default MediaWiki behaviour! We can make this happen :)

Callanecc:
Exactly my thoughts! :)

MusikAnimal:
Also I can try to make the durations sync as they do with MW, where if you change the protection level/duration of the page, the move protection settings change with it, but not the other way around

MusikAnimal:
[...]

PR #451, opened 2018-11-02
protect: Remove autoconfirmed from dropdowns when inappropriate

Amorymeltzer:
1. (auto)confirmed is needed to move any page, even in your own userspace
2. After WP:ACPERM, (auto)confirmed is needed to create an article in mainspace
3. Removed AC move protection option from batchdelete

### DIFF
--- a/modules/twinklebatchprotect.js
+++ b/modules/twinklebatchprotect.js
@@ -134,11 +134,6 @@ Twinkle.batchprotect.callback = function twinklebatchprotectCallback() {
 		});
 	movelevel.append({
 			type: 'option',
-			label: 'Autoconfirmed',
-			value: 'autoconfirmed'
-		});
-	movelevel.append({
-			type: 'option',
 			label: 'Extended confirmed',
 			value: 'extendedconfirmed'
 		});
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -324,11 +324,6 @@ Twinkle.protect.callback.changeAction = function twinkleprotectCallbackChangeAct
 					});
 				movelevel.append({
 						type: 'option',
-						label: 'Autoconfirmed',
-						value: 'autoconfirmed'
-					});
-				movelevel.append({
-						type: 'option',
 						label: 'Extended confirmed',
 						value: 'extendedconfirmed'
 					});
@@ -446,11 +441,13 @@ Twinkle.protect.callback.changeAction = function twinkleprotectCallbackChangeAct
 						label: 'All',
 						value: 'all'
 					});
-				createlevel.append({
-						type: 'option',
-						label: 'Autoconfirmed',
-						value: 'autoconfirmed'
-					});
+				if (mw.config.get("wgNamespaceNumber") !== 0) {
+					createlevel.append({
+							type: 'option',
+							label: 'Autoconfirmed',
+							value: 'autoconfirmed'
+						});
+				}
 				if (isTemplate) {
 					createlevel.append({
 							type: 'option',


Amorymeltzer:
This should probably be discussed.

It's been a de facto standard since forever to apply equivalent edit and move protection levels even if only edit was desired/required; this is certainly less common than it used to be a decade ago, likely thanks to Twinkle and better `action=protect` menus/options, but it's still a very common occurrence (myself included).  It's possible that removing the options from Twinkle will lead to some confusion — I'm not certain how many sysops are aware of either/both of these facts.  Changin existing workflows and creating a lack of parallelism (between both edit/move menus and the `action=protect` menus from mw) could be confusing.  On the other hand, preventing template errors (e.g., using only move protection at the AC level and `{{pp-move}}`) or pointless protection is a good thing.  I suppose it is a very enWiki-specific thing, but then again so is Twinkle.

Additionally, for item 2 in the commit message, there was a loophole (stopped with https://en.wikipedia.org/wiki/Special:AbuseFilter/934 and fixed in https://phabricator.wikimedia.org/T204016) although 1.33-wmf.2 has apparently yet to be successfully pushed.

MusikAnimal:
I personally think this is a sane change, but let's see if anyone else getting following this repo disagrees.

QEDK:
Seconded.

Amorymeltzer:
@MusikAnimal Good catch, thanks.  This appears to be coming from the if-else block in the `movemodify` function.  If the move form is disabled and the edit is active, it preemptively copies the edit options over to the move dropdown (or apparently tries to).  If the edit option is unchecked, it makes the move dropdown default to sysop.  Smart!  I never noticed that before, pretty helpful.  Added in #301.

Anyway, not sure what is best.  If the option for autoconfirmed move protection is indeed to be removed, I suppose we could default to sysop move if AC is selected in the edit dropdown?  I think that's basically a return to #217 but a blank menu, while arguably accurate, is a very poor user experience.

JJMC89:
> Anyway, not sure what is best. If the option for autoconfirmed move protection is indeed to be removed, I suppose we could default to sysop move if AC is selected in the edit dropdown? I think that's basically a return to #217 but a blank menu, while arguably accurate, is a very poor user experience.

IMO, when adding autoconfirmed edit protection, no move protection should be added unless explicity specified since
1. autoconfirmed move protection has no effect, and
2. adding a higher level of move protection, especially sysop, is unlikely to be what is desired.

Amorymeltzer:
Having thought about it some more, I've become somewhat more comfortable with the idea of a blank menu — I don't love it and it might be poor user interface, but like #442 it might help nudge folks who weren't aware.

Also, I've got an update that also removes autoconfirmed move protection for `batchprotect`, but will wait for this decision to pan out.  Preventing mainspace autoconfirmed create protection in `batchprotect` would be unwieldy.

Amorymeltzer:
I've just pushed the removal of ac move protection from `batchprotect`, and wanted to revive this to consider our options.  Assuming the intent is to apply semi protection:

1. AC missing: the empty menu will show up.  This is jarring, but probably the strongest sign/nudge that AC move protection is extraneous
2. AC missing, autoselect sysop.  Could erroneously suggest that edit=autoconfirmed, move=sysop is something to be encouraged, and likely to happen out of habit.
3. AC present but disabled, autoselect AC: Most similar to how things are now, this will have the least impact but is also the least change.  Unless they open the menu and change items, semiprotection will be applied to move protection as is done currently and none will be the wiser.
4. AC present but disabled, autoselect sysop: As above, could suggest such a thing is desirable.

My hope with 2 and 4 is that seeing sysop jump up would be a "Whoa!" to the sysop who will attempt to correct it, learn, etc.  My concern with 2 and 4 is that they'll just move protect at the sysop level.  Likewise, I'm not sure with 3 we'll get any change in behavior.  1 is ugly, but I think I still favor it.  cc @JJMC89 @MusikAnimal 

For mainspace creation protection, is it worth listing autoconfirmed but disabled?  I've just removed it entirely but could be worth reconsidering in light of the above.

JJMC89:
My personal preference is 1, but 3 is OK too since it matches the default MW behavior.

> My concern with 2 and 4 is that they'll just move protect at the sysop level.

Yes, this is what would mostly happen.

> For mainspace creation protection, is it worth listing autoconfirmed but disabled?

I don't think it makes much difference.

Amorymeltzer:
@MusikAnimal Do you still have concerns over this?

Issue #353, opened 2016-08-31
Edit summary wrong when using template parameters for custom tags


nyuszika7h:
It appers I can specify template parameters for custom tags, such as `Film IMDb refimprove|only=yes`, but then the edit summary shows up as `Added {{[[Template:Film IMDb refimprove|only=yes]]}} tag to article (TW)` (with the template linked and only the piped text `only=yes` displaying). I could create a wrapper template, but it might be nice to get this fixed.

PR #456, opened 2018-11-06
tag: |reason= and |talk= fields for {{Expert needed}}, edit summary fix for custom tags with parameter

siddharthvp:
1.  Added |reason= and |talk= fields to {{[Expert needed](https://en.wikipedia.org/wiki/Template:Expert_needed)}} as requested at [Wikipedia:Twinkle/to_do](https://en.wikipedia.org/wiki/Wikipedia:Twinkle/to_do) . Either of these two fields is compulsory per documentation. (Though this isn't enforced here) 

2. Per Issue #353, edit summary is buggy when a custom tag with parameter is added. Also handled this issue for the case that the tag isn't in template namespace.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -195,13 +195,26 @@ Twinkle.tag.updateSortOrder = function(e) {
 				};
 				break;
 			case "expert needed":
-				checkbox.subgroup = {
+				checkbox.subgroup = [ {
 					name: 'expertNeeded',
 					type: 'input',
 					label: 'Name of relevant WikiProject: ',
 					tooltip: 'Optionally, enter the name of a WikiProject which might be able to help recruit an expert. Don\'t include the "WikiProject" prefix.'
-				};
-				break;
+				}, 
+				{
+					name: 'expertNeededReason',
+					type: 'input',
+					label: 'Reason: ',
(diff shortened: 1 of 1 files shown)


siddharthvp:
Squashed the commits into one.

modules/friendlytag.js
```diff
@@ -1361,6 +1393,9 @@ Twinkle.tag.callback.evaluate = function friendlytagCallbackEvaluate(e) {
 				globalize: form["articleTags.globalize"] ? form["articleTags.globalize"].value : null,
 				notability: form["articleTags.notability"] ? form["articleTags.notability"].value : null
 			};
+			params.expertNeeded = form["articleTags.expertNeeded"] ? form["articleTags.expertNeeded"].value : null,
```
Amorymeltzer: What's the reason for putting these here — i.e., under `params` rather than `params.tag` (see L944-945)?  Should be removed from L1392.

siddharthvp: removed L1392 which was unused. Doesn't matter whether you put it into params.tagParameters or not, anyway in #487, I have fully gotten rid of this manual method of saving values in `params` object.

Amorymeltzer: Fair enough!

Amorymeltzer:
Thanks for tackling both of these — the code for the custom templates works like a charm!

PR #457, opened 2018-11-07
Added BOOKPROD

siddharthvp:
Added [WP:BOOKPROD](https://en.wikipedia.org/wiki/Wikipedia:Proposed_deletion_(books)) support to prod module. This has been marked as 'to do' in the source of twinkeprod.js for a long time.  Only pages in book namespace are supported as of now, though userspace books are also eligible per BOOKPROD.

PS: I am new here and this is amongst my first PRs ever so any feedback is welcome. Changes are untested. What are the suggested ways to test TW changes? I am not an admin on testwiki nor do I have a local MW installation. Is there any other way to do testing?


MusikAnimal:
Hello and welcome :)

To answer your question about testing: Most of us do it by copying/pasting the whole module (twinkleprod in this case) into the JS console. This will override the deployed code. I wouldn't test PROD'ing books on enwiki, so testwiki is the way to go. You don't need to be an admin. Let us know when you've tested your code and we'll move forward with a review. Cheers

QEDK:
I am concerned about the usability of this, it is mostly an extinct (or atleast, abandoned) namespace, so how much of an utility does it bring to the users?

siddharthvp:
> I am concerned about the usability of this, it is mostly an extinct (or atleast, abandoned) namespace, so how much of an utility does it bring to the users?

Not much. But that doesn't mean we should not have it, right? It looks like 95% of [all book MfDs](https://en.wikipedia.org/wiki/Special:AllPages?from=Miscellany+for+deletion%2FBook%3A&to=Miscellany+for+deletion%2FBp&namespace=4) tend to close as 'delete', mostly unanimously so. BOOKPROD could have been used instead in most of the cases but isn't used much because well... Twinkle doesn't support it. Consider discussions like this, [this](https://en.wikipedia.org/wiki/Wikipedia:Miscellany_for_deletion/Book:Mercury_-_A_Collection_of_Wikipedia_Articles), [this](https://en.wikipedia.org/wiki/Wikipedia:Miscellany_for_deletion/Book:Messier_objects_1_and_2), [this ](https://en.wikipedia.org/wiki/Wikipedia:Miscellany_for_deletion/Book:Mors_Wikisider), [this](https://en.wikipedia.org/wiki/Wikipedia:Miscellany_for_deletion/Book:101_Topics) and [this](https://en.wikipedia.org/wiki/Wikipedia:Miscellany_for_deletion/Book:ORF), which needn'thave taken place if BOOKPROD had been there. No one wants to take the trouble of manually placing a prod tag; rather take the easier route of MfDing it using twinkle.

MusikAnimal:
> I am concerned about the usability of this, it is mostly an extinct (or atleast, abandoned) namespace, so how much of an utility does it bring to the users?

I was also a little concerned for the same reason. I believe I'm the one who added the TODO. If I remember correctly this following a rewrite of how we fetched deletion rationale, so I think you were able to PROD books before then. At any rate, the BOOKPROD stuff isn't really obtrusive to the user, so as long as there's still a BOOKPROD process I think it's okay to move forward with it.

> pasting the code in the browser console didn't work, presumably because the main function of twinklepord.js had already been called by the time code was pasted in

Odd! I worked on this module quite a bit and I think I always tested by copying/pasting the code.

Amorymeltzer:
Don't think there's a `contributing` standard here, but is there a reason not to squash these?

MusikAnimal:
> @Amorymeltzer 
> Don't think there's a `contributing` standard here, but is there a reason not to squash these?

You can do this within GitHub; select "Squash and merge" instead of "Merge pull request" using the dropdown. We don't have a contributing guide, but we should, and I think squashing related commits should be part of it. Subject to debate! I also use GitHub to prepend the module name to the commit summary, if the author didn't include it. That I think has been standard practice for a long time, but not a big deal if it isn't followed. For now whoever is merging can squash/reword commits at their discretion.

Amorymeltzer:
You'll note I didn't have commit privileges when I wrote that. :wink:  I've got a draft issue to open on that front, but got stuck with OS cleanup duty.  Expect it tonight or tomorrow.

By means of a seamless transition @siddharthvp, I meant to get to this today as well but... cleanup duty.  I'll try and tackle it tomorrow if you're not MA; I agree that while the whole concept of Bookspace is wildly underused, if we've got the policy it's fine to introduce it here.  Who knows, maybe it would help!

modules/twinkleprod.js
```diff
@@ -142,20 +165,20 @@ Twinkle.prod.callback.prodtypechanged = function(event) {
...
-			);
+				} );
 			//temp warning, can be removed down the line once BLPPROD is more established. Amalthea, May 2010.
 			var boldtext = document.createElement('b');
 			boldtext.appendChild(document.createTextNode('Please note that only unsourced biographies of living persons are eligible for this tag, narrowly construed.'));
 			field.append({
 				type: 'div',
 				label: boldtext
 			});
+			// TO DO: check if article is in Category:Living people, warn if not
```
Amorymeltzer: Isn't this #458?  Could use a rebase, I suppose.

modules/twinkleprod.js
```diff
@@ -91,11 +119,7 @@ Twinkle.prod.callback = function twinkleprodCallback() {
 	var evt = document.createEvent( "Event" );
 	evt.initEvent( 'change', true, true );
 	result.prodtype[0].dispatchEvent( evt );
```
Amorymeltzer: This is throwing an error outside mainspace (aka files and books).  I think part of the issue is that you're trying to make it smarter about the notification (lines 93-97) but haven't refactored `Twinkle.prod.callback.prodtypechanged` (lines 125+).

siddharthvp: This was supposed to execute only on articles. Fixed now by wrapping it in a if (namespace==='article'). 
Didn't notice this earlier because the PROD nom still goes on smoothly in any namespace. There are no visible problems, barring the error shown in the console.

modules/twinkleprod.js
siddharthvp: I just removed the cmt. Isn't it enough? I am not sure how to do a rebase.

modules/twinkleprod.js
```diff
@@ -329,19 +367,21 @@ Twinkle.prod.callbacks = {
 Twinkle.prod.callback.evaluate = function twinkleprodCallbackEvaluate(e) {
 	var form = e.target;
 	var prodtype;
-
-	var prodtypes = form.prodtype;
-	for( var i = 0; i < prodtypes.length; i++ ) {
-		if( !prodtypes[i].checked ) {
-			continue;
+
```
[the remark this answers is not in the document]
siddharthvp: All fixed now, but seriously was doing this necessary?

modules/twinkleprod.js
Amorymeltzer: Eh, don't worry about it — I thought it was a comment present previously that got readded by merging in an old version, but it seems not.

Rebasing is either the greatest or worst thing ever in git: https://git-scm.com/docs/git-rebase

modules/twinkleprod.js
Amorymeltzer: I meant it more to try to understand the reason for duplicating the code in `Twinkle.prod.callback.prodtypechanged`.  AFAICT, the "Notify page creator if possible" stuff you added for files and books just repeats what's in that function.  I think it's nice to ditch the fake event if possible (maybe even the hiding of the fieldset?!) but am I missing a reason to do it this way as opposed to just falling back to the `prod` case?

siddharthvp: There can be two options:
1. Generate the prod-type fieldset on all pages, and then hide it on non-articles. This avoid duplication of the code that generates the notify checkbox and reason textarea.
2. Generate the fieldset only on articles. This mean no duplication of code. But on non-articles, this scheme is more efficient as the fieldset isn't generated in the first place.
 
It originally was (1) but I changed it to (2) in the 2nd commit. On further thought, the minor efficiency gain, that too only on non-articles, is probably not worth the code duplication. So I have reverted it back to (1), as you also seem to be suggesting. Is this fine?

Ditch the fake event? This seems pretty standard across TW modules. I don't see why or how it can be ditched.

siddharthvp:
Just noticed that the `prod_re` that checks for existing prod tag doesn't work for files. Fixed it updating the regex. Also, I simplified it and removed the useless `dated prod` and `dated files`. As prod nomination are done by substituting the `prod` or `prod-blp` or `book-prod` which respectively produce `Proposed deletion/dated` or `Proposed deletion/dated files`, `Prod blp/dated`, and `book-prod/dated`. These are the only templates we need to look for.

modules/twinkleprod.js
Amorymeltzer: Awesome, thanks.  I'll get around to a final test tomorrow — library internet miserable today — but that looks good to me.

>Ditching
I had thought that was your intention.

siddharthvp:
@Amorymeltzer Thanks for reviewing this. Pls squash commits while merging (I'd say keep one for adding bookprod, one for regex fix). I'll update the on-wiki documentation (if any) after that.

siddharthvp:
@Amorymeltzer oh i managed to mess this up. See above. Trying to learn git rebases and all, what I did was this: In order to mess up the main branch `bookprod` which is feeding this PR, I created a new branch `boookprod` (3 o's) from the `bookprod` branch . Then did the rebase upon it to squash the 3 unneeded commits. Had to fix a couple of merge conflicts. Then I pushed to `origin/bookprod` from there. It seemed that github has decided that all those old commits by others are part of the PR! 

I moved back to the `bookprod` branch and force-pushed to fix the problem. But that isn't fixing anything. 

I think that this can't be merged now, right? As all those old commits would be duplicated? Looks like the simplest thing to do is to copypaste the changes from twinkleprod.js and create a new commit in a new branch and file a new PR?

siddharthvp:
@Amorymeltzer  Great! It matches 100% with my files. Thanks.

BTW, do you mind telling me how you could move just those two commits to a new branch to create the PR?

Amorymeltzer:
You should... _checkout_ the command `git checkout`😆😆😆

For real though, it's helpful for moving around between branches, but is also prone to some issues if you don't watch what commits are where.  `rebase` is super tricky and finicky to get the hang of, but can be useful once you do.  I'm a notorious editor/tweaker, so I ended up forced to use it a lot to make up for my inadequacies!

siddharthvp:
Cool, I didn't know `checkout` can be used to checkout commits!

PR #458, opened 2018-11-11
prod: add alert if BLPPROD is applied but article is not in Category:Living people

siddharthvp:
If page is being nominated for BLPPROD, a check is done to see if article has [[Category:Living people]]. If not, user is warned about it and they can choose to either continue or abort the action.  If they choose to continue, [[Category:Living people]] is added to the article along with the prod tag. 

This feature makes TW smarter and helps prevent bios of deceased people (and non-bios) from being mistakenly tagged for blp-prod.

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -177,10 +177,21 @@ Twinkle.prod.callbacks = {
 		var text = pageobj.getPageText();
 		var params = pageobj.getCallbackParameters();
 
+		// Check for already existing deletion tags
 		var tag_re = /({{(?:db-?|delete|[aitcmrs]fd|md1)[^{}]*?\|?[^{}]*?}})/i;
 		if( tag_re.test( text ) ) {
 			statelem.warn( 'Page already tagged with a deletion template, aborting procedure' );
 			return;
+		}
+
+		// Alert if article is not in Category:Living people and BLPPROD is selected
+		if( params.blp ) {
+			var blpcheck_re = /\[\[Category:Living people\]\]/i;
+			if( !blpcheck_re.test( text ) ) {
+				if( ! confirm("Please note that the article is not in Category:Living people and hence may be ineligible for BLPPROD. Are you sure you want to continue? \n\nYou may wish to add the category if you proceed, unless the article is about a recently deceased person." ) ) {
+					return;
+				}
+			}
 		}
 
 		// Remove tags that become superfluous with this action


Amorymeltzer:
So, BLP is a little misleading, as the policy includes the [recently dead](https://en.wikipedia.org/wiki/Wikipedia:Biographies_of_living_persons#Recently_dead_or_probably_dead) while Category:Living people includes only the living.  It might be unlikely, but BLPPROD could well apply to a page not in CAT:LP.  I think it's probably okay to offer an alert in that case, but I don't think automatically adding CAT:LP is a good idea in that case.  Perhaps reword the `The category will be added...` comment to suggest the user add it if they do indeed proceed?

siddharthvp:
The BLP policy includes the recently dead as well as biographical details covered in other articles that are not bios. But the policy on [BLPPROD](https://en.wikipedia.org/wiki/Wikipedia:Proposed_deletion_of_biographies_of_living_people#Scope) does not make any mention of recently dead people. 

> Although policy on biographies of living persons covers all types of biographical material, the use of this deletion procedure applies only to articles that are biographies of living people.

So I don't think there's a problem here.

Amorymeltzer:
I checked with another sysop, who reads that as I do: in direct reference to the policy rather than the vernacular meaning.  The lines after it give some context as to what BLPPROD wouldn't cover.

siddharthvp:
@Amorymeltzer ok, reluctantly removed my precious category-adding code :)

modules/twinkleprod.js
```diff
@@ -177,12 +177,23 @@ Twinkle.prod.callbacks = {
 		var text = pageobj.getPageText();
 		var params = pageobj.getCallbackParameters();
 
+		// Check for already exisitng deletion tags
```
Amorymeltzer:
```suggestion
		// Check for already existing deletion tags
```

modules/twinkleprod.js
```diff
@@ -177,12 +177,23 @@ Twinkle.prod.callbacks = {
...
 		var tag_re = /({{(?:db-?|delete|[aitcmrs]fd|md1)[^{}]*?\|?[^{}]*?}})/i;
 		if( tag_re.test( text ) ) {
 			statelem.warn( 'Page already tagged with a deletion template, aborting procedure' );
 			return;
 		}
 
+		// Alert if article is not in Category:Living people and BLPPROD is applied
+		if( params.blp ) {
+			var blpcheck_re = /\[\[Category:Living people\]\]/i;
+			if( !blpcheck_re.test( text ) ) {
```
Amorymeltzer:
```suggestion
			if( !blpcheck_re.test( text ) ) {
```

modules/twinkleprod.js
```diff
@@ -177,12 +177,23 @@ Twinkle.prod.callbacks = {
...
 		var params = pageobj.getCallbackParameters();
 
+		// Check for already exisitng deletion tags
 		var tag_re = /({{(?:db-?|delete|[aitcmrs]fd|md1)[^{}]*?\|?[^{}]*?}})/i;
 		if( tag_re.test( text ) ) {
 			statelem.warn( 'Page already tagged with a deletion template, aborting procedure' );
 			return;
 		}
 
+		// Alert if article is not in Category:Living people and BLPPROD is applied
```
Amorymeltzer:
```suggestion
		// Alert if article is not in Category:Living people and BLPPROD is selected
```

modules/twinkleprod.js
```diff
@@ -177,12 +177,23 @@ Twinkle.prod.callbacks = {
...
+		if( params.blp ) {
+			var blpcheck_re = /\[\[Category:Living people\]\]/i;
+			if( !blpcheck_re.test( text ) ) { 			
+				if( ! confirm("Please note that article is not in Category:Living people and hence may be ineligible for BLPPROD. Are you sure you want to continue? \n\nYou may wish to add the category if you proceed, unless the article is about a recently deceased person." ) ) {
```
Amorymeltzer:
```suggestion
				if( ! confirm("Please note that the article is not in Category:Living people and hence may be ineligible for BLPPROD. Are you sure you want to continue? \n\nYou may wish to add the category if you proceed, unless the article is about a recently deceased person." ) ) {
```

Amorymeltzer:
It was nice code!

siddharthvp:
All grammar/stylistic changes implemented and resulting commits squashed.

PR #459, opened 2018-11-11
xfd: automate correct venue selection for userbox and stub templates 

siddharthvp:
Per policy, the xfd venue for stub templates ("Template:xxx-stub") is CfD, and for userboxes ("Template:User xxx") is MfD. I updated the automatic xfd venue selection process to accommodate these special cases.


Amorymeltzer:
Neat idea, this should be super helpful!  The MfD for userboxes is straightforward and works fine, but the Twinkle's handling of the CfD nomination only seems to fit for Category nominations, and runs afoul of the [procedure for stub nominations](https://en.wikipedia.org/wiki/Wikipedia:Categories_for_discussion#Stub_types).  I'm sure there are a few ways to deal with this, but that should probably be dealt with separately.

siddharthvp:
I see that you're talking about stub template deletion ({{sfd-t}}) and stub template renaming ({{sfd-r}}) not being supported. 

So how do we move on with this? Just merge in the Mfd for userboxes part? Or merge in both - ok, this 
does awkwardly produce the {{cfd}} banner on a template page, pointing to the CfD discussion, but I guess this shouldn't be too worrisome as only editors see these pages. I believe the existing practise while noming a stub template is to open TW and manually change the venue to CfD (most editors being unaware of the sfd-t and sfd-r templates). This PR eliminates the latter step.

Amorymeltzer:
Not just {{sfd-t}} — indeed an issue — but the actual nomination incorrectly treats the template as if it were a category.

siddharthvp:
I see. So for now, I've removed the stub types part of the edit.

modules/twinklexfd.js
```diff
@@ -60,28 +60,30 @@ Twinkle.xfd.callback = function twinklexfdCallback() {
...
 			label: 'FfD (Files for discussion)',
-			selected: mw.config.get('wgNamespaceNumber') === 6,  // File namespace
+			selected: namespace === 6,  // File namespace
 			value: 'ffd'
 		} );
 	categories.append( {
 			type: 'option',
 			label: 'CfD (Categories for discussion)',
-			selected: mw.config.get('wgNamespaceNumber') === 14,  // Category namespace
+			selected: namespace === 14,
```
```suggestion
			selected: namespace === 14,  // Category namespace
```

Amorymeltzer: I think if you allow collaborators to edit your PRs, I can resolve these minor things myself

PR #460, opened 2018-11-12
speedy: Fix broken logging on incoming di request

Amorymeltzer:
See https://en.wikipedia.org/wiki/Special:Permalink/868528808#Most_CSDs_not_being_added_to_log and #436 

Check for `params.normalizeds` before checking if it's a g10 page.  I only changed two in #436, but put the check in for all three.


Amorymeltzer:
CC @siddharthvp and @MusikAnimal

siddharthvp:
oops, didn't see this while opening #461 . I think my version is slightly better as I've also reduced the total number of condition checks by pushing the existing `editsummary+=` lines into the existing if-else structures, rather than create new ones.

Amorymeltzer:
Closed in favor of #461 (haven't tested that but looks fine)

PR #462, opened 2018-11-21
twinkle: improve TW menu button in vector skin

siddharthvp:
Per https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=868832371&diffmode=source#Twinkle_dropdown_in_Vector, added a `<input type="checkbox" class="vectorMenuCheckBox" aria-labelledby="p-twinkle-label">` to the relevant div element. And added `id="p-twinkle-label"` to the "TW" heading text. 

This provides TW button with the same special effects of the More button. 

I have been unable to test this by any of the usual methods. Pasting code to console or loading it from userspace js page lead to a bizarre "Twinkle.arv() is not a function" error. Looks like the only way to test would be to apply these changes to https://test.wikipedia.org/wiki/MediaWiki:Gadget-Twinkle.js .

### DIFF
--- a/twinkle.js
+++ b/twinkle.js
@@ -290,7 +290,19 @@ Twinkle.addPortlet = function( navigation, id, text, type, nextnodeid )
 		root.appendChild( outerDiv );
 	}
 
+	if( outerDivClass === "vectorMenu" ) {	
+		// add invisible checkbox to make menu keyboard accessible
+		// similar to the p-cactions ("More") menu
+		var chkbox = document.createElement( "input" );
+		chkbox.className = "vectorMenuCheckbox";
+		chkbox.setAttribute( "type","checkbox" );
+		chkbox.setAttribute( "aria-labelledby", "p-twinkle-label" );
+		outerDiv.appendChild( chkbox );
+	}
 	var h5 = document.createElement( "h3" );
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
You can test it by:
- Loading in safe mode
- Importing ``morebits``, ``twinkle.css``, and the vector pagestyles CSS
- Remove all the module functions from `twinkle.js`, paste that in your console
- Import some modules, then load them manually

It seems to work for me, although I don't see something similar in MoreMenu.  @MusikAnimal How'd you do it there?

MusikAnimal:
> I don't see something similar in MoreMenu. @MusikAnimal How'd you do it there?

I did not add any code to add keyboard shortcuts, that I recall. I'm assuming @siddharthvp is referring to the normal "More" menu. I really couldn't have chosen a more confusing name for that gadget! But I like the play on words ;)

Amorymeltzer:
Ahh, gotcha, now this makes sense to me.  @siddharthvp 	am I right that by "special effects" you mean that if you click the menu, it stays open?  Pretty neat!  I'm a Modern evangelist, not a Vector expert, is there anything else I should be looking for?

Amorymeltzer:
I've merged this, but @siddharthvp 	lmk if I've missed anything

PR #463, opened 2018-11-21
add new module for requesting oversight

siddharthvp:
Per the strong consensus for this at https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle#Children's_personal_info. 

This is an interface to send emails to User:Oversight. This creates a tab "OS" that creates a morebits simplewindow with email subject and body fields pre-populated on basis of current page name and reason selected from the dropdown. 

Further improvements planned:  done
~~1. Changes to default email content.~~
~~2. Direct addition of link to diff when invoking the module while on a diff page.~~

### DIFF
--- a/modules/twinkleos.js
+++ b/modules/twinkleos.js
@@ -191,31 +191,25 @@ Twinkle.os.evaluate = function evaluate(e) {
 	token_api.parent = params;
 	token_api.post();
 	Morebits.status.init(form);
-	//Morebits.wiki.actionCompleted.notice = "Email sent";
+	Morebits.wiki.actionCompleted.notice = "Email";
+	Morebits.wiki.actionCompleted.postfix = "Sent successfully";
 };
 
 Twinkle.os.sendEmail = function sendemail(apiobj) {
 	var params = apiobj.parent;
 	var token = $(apiobj.responseXML).find("tokens").attr("csrftoken");
-	var mail_api = new Morebits.wiki.api( 'Email', {
+	var mail_api = new Morebits.wiki.api( 'Sending email', {
 		'action': 'emailuser',
-		'target': 'SD0001',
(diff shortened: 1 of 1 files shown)


siddharthvp:
To test the module add `importScript('User:SD0001/twinkleos.js');` to [your common.js](https://en.wikipedia.org/wiki/special:mypage/common.js). That version differs slightly from the one here in that emails are sent to yourself (not to Oversight), and there lots of `console.log`s out there :)

Amorymeltzer:
I left a broader comment on-wiki, but I wanted to suggest two things here that are more design-specific:
- A clear disclaimer summarizing the OS policy on PII would be good to discourage frivolous reports.
- Might be just me, but I think a blank message box that requires entered text in order to be sent would help avoid form emails.

siddharthvp:
@Amorymeltzer Is is okay to omit the protocol ("https://") in email links? Since emails are in plain text, it'd would make them a bit more readable. 

I've made some changes. When invoked on a diff page or history page, the script fills out the link to the diff or oldid page. On history pages, you get the additional option of adding ranges of revisions by selecting them using the native radio buttons on the history page, then clicking on a button within the tool. 

Also added a warning pointing to the  OS policy.

Amorymeltzer:
@siddharthvp What do you mean, like the body of the email would say "OS this plz: en.wikipedia.org/wiki/..."?  I think it'd be good to retain that, you never know how people's email clients or browsers will interpret links.

siddharthvp:
> @siddharthvp What do you mean, like the body of the email would say "OS this plz: en.wikipedia.org/wiki/..."? I think it'd be good to retain that, you never know how people's email clients or browsers will interpret links.

Ok.

PR #467, opened 2018-12-08
Fix rollback link on contribs page

Amorymeltzer:
Fix rollback link on contribs page
    
Previous discussion: https://github.com/azatoth/twinkle/commit/9b61d36a364cd1a9fce2e72b0b7ed13aa291fc00#comments
    
Starting around Nov 10, there were multiple reports of failed rollbacks from the contribs page at WT:TW: [here](https://en.wikipedia.org/w/index.php?oldid=872508505#Problem_when_clicking_on_%22vandalism%22_link_while_examining_user_contributions) and [here](https://en.wikipedia.org/w/index.php?oldid=872508505#Rollback_malfunction)
    
[T205581](https://phabricator.wikimedia.org/T205581) introduced a number of css changes to allow for full skinnability.  [One of these changes](https://gerrit.wikimedia.org/r/c/mediawiki/core/+/461237) put the diff and hist links into their own span, thus hiding the a elements as children of that span rather than span.mw-uctop.  This finds the link with .mw-changeslist-diff directly, on the off chance the order is changed.

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -61,7 +61,7 @@ Twinkle.fluff = {
 				revVandNode.appendChild(revVandLink);
 
 				list.each(function(key, current) {
-					var href = $(current).children("a:eq(1)").attr("href");
+					var href = $(current).find(".mw-changeslist-diff").attr("href");
 					current.appendChild( document.createTextNode(' ') );
 					var tmpNode = revNode.cloneNode( true );
 					tmpNode.firstChild.setAttribute( 'href', href + '&' + Morebits.queryString.create( { 'twinklerevert': 'norm' } ) );


Amorymeltzer:
Ping @MusikAnimal and @mc10

Putting this up as while I've done a fair bit of testing, it likely needs some more since I'm flying a bit blind not knowing what it was _before_.  Seems to work fine on testwiki, and with some minimal testing looks like it'll be fine on enwiki.  The `.first()` is required on enwiki but not test since there's a second (invisible?) repeat of the class on `oldid` pages afaict.

I'd love to know what the pages looked like before this broke.  What were the links like on contribs previously?  I'm guessing that the `href` was built off the `oldid` link as it works, but it's not based off anything I know.  Did `#mw-diff-ntitle2` really exist on `oldid` non-diff pages?  That goes back to https://github.com/azatoth/twinkle/commit/045af9e20fb3e25a3e394c6013ccb2cfb6606d91.

What's bugging me is _where_ these came from.  I don't see anything obvious in either of the two [most](https://www.mediawiki.org/wiki/MediaWiki_1.33/wmf.3) [proximate](https://www.mediawiki.org/wiki/MediaWiki_1.33/wmf.2); the most relevant thing I found is [T205581](https://phabricator.wikimedia.org/T205581) but that's about CSS and mobile.

PR #468, opened 2018-12-08
speedy: add R4, merge db-blankdraft to G13

### LINKED ISSUE #466: Add new deletion-tagging template {{Db-r4}}
A discussion <https://en.wikipedia.org/wiki/Wikipedia_talk:Criteria_for_speedy_deletion#Redirects_in_the_File:_namespace_G6_→_R4> found consensus to use new deletion tag {{db-r4}}. It should only be accesible in File: namespace.

siddharthvp:
Added newly-introduced CSD R4 per [discussion](https://en.wikipedia.org/wiki/Special:PermanentLink/872246394#Redirects_in_the_File:_namespace_G6_%E2%86%92_R4). Merged db-blankdraft to db-g13 per [discussion](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Criteria_for_speedy_deletion&oldid=872669362#Proposal_to_merge_db-blankdraft_into_G13)

Closes #466.

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -797,9 +797,9 @@ Twinkle.speedy.userNonRedirectList = [
 		hideWhenMultiple: true
 	},
 	{
-		label: 'G13: Old, abandoned AfC submission or blank draft',
+		label: 'G13: AfC draft submission or a blank draft, stale by over 6 months',
 		value: 'afc',
-		tooltip: 'Any rejected or unsubmitted AfC submission, or a blank draft, unedited in over 6 months,',
+		tooltip: 'Any rejected or unsubmitted AfC draft submission or a blank draft, that has not been edited in over 6 months (excluding bot edits).',
 		hideWhenMultiple: true
 	}
 ];
@@ -1037,9 +1037,9 @@ Twinkle.speedy.generalList = [
 		]
 	},
 	{
-		label: 'G13: Old, abandoned Articles for Creation submissions',
+		label: 'G13: Page in draft namespace or userspace AfC submission, stale by over 6 months',
 		value: 'afc',
-		tooltip: 'Any rejected or unsubmitted AfC submission that has not been edited for more than 6 months.',
+		tooltip: 'Any rejected or unsubmitted AfC submission in userspace or any page in draft namespace, that has not been edited for more than 6 months. Blank drafts in either namespace are also included.',
 		hideWhenRedirect: true
 	}
 ];
@@ -1088,7 +1088,6 @@ Twinkle.speedy.normalizeHash = {
 	'disambig': 'g6',
 	'movedab': 'g6',
 	'copypaste': 'g6',
-	'blankdraft': 'g6',
 	'g6': 'g6',
 	'author': 'g7',
 	'g8': 'g8',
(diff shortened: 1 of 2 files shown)


modules/twinklespeedy.js
```diff
@@ -1055,6 +1055,11 @@ Twinkle.speedy.redirectList = [
 		value: 'redirtypo',
 		tooltip: 'However, redirects from common misspellings or misnomers are generally useful, as are redirects in other languages'
 	},
+	{
```
Amorymeltzer: This should really be in the `fileList` no?

modules/twinklespeedy.js
```diff
@@ -790,17 +790,17 @@ Twinkle.speedy.userNonRedirectList = [
...
-		hideWhenMultiple: true
-	},
 	{
 		label: 'G11: Promotional user page under a promotional user name',
 		value: 'spamuser',
 		tooltip: 'A promotional user page, with a username that promotes or implies affiliation with the thing being promoted. Note that simply having a page on a company or product in one\'s userspace does not qualify it for deletion. If a user page is spammy but the username is not, then consider tagging with regular G11 instead.',
 		hideWhenMultiple: true
+	},
+	{
+		label: 'G13: Old, abandoned AfC submission or blank draft',
```
QEDK: Drafts unedited in 6 months are also included, not only blank ones, modify the label as well. Change value to g13 to preserve uniformity with the other rationales imo.

modules/twinklespeedy.js
```diff
@@ -790,17 +790,17 @@ Twinkle.speedy.userNonRedirectList = [
...
+		value: 'afc',
+		tooltip: 'Any rejected or unsubmitted AfC submission, or a blank draft, unedited in over 6 months,',
```
QEDK: This implies only blank drafts are eligible which is not true.

Amorymeltzer:
Tests seem good!  I've been in template/module hell trying to confirm where db-multiple seems to support R4...

Amorymeltzer:
@JJMC89 Was just editing to clarify that I meant what my import busted on testwiki, while it works just fine on enwiki.

modules/twinklespeedy.js
siddharthvp: Doesn't the label already say "Old, abandoned AfC submission" which means a draft unedited in 6 months? Anyways, please post your suggested wording here - I'll make the change.

modules/twinklespeedy.js
siddharthvp: Like above, I don't see how the wording implies that. What wording are you suggesting?

modules/twinklespeedy.js
QEDK: @siddharthvp Simply make it Old, abandoned AfC submissions and draftspace pages and for the tooltip write the same including pages with no content, unedited in more than 6 months.

Amorymeltzer: So theoretically this particular message will only show up on user pages, so it's okay not to clarify the draftspace content.  It's probably worth updating the other one though.

siddharthvp:
@Amorymeltzer @QEDK I've completely changed the labels and tooltips of G13 in both userList and generalList, to avoid any confusion. Note that in userList, it's not necessary to mention draftspace pages. And in generalList, it's not worth mentioning blank drafts in the label because Article Wizard-generated blank drafts occur only in userspace, and the userList criteria are anyway shown above generalList criteria  if the page is indeed in userspace.

Amorymeltzer:
@siddharthvp Thanks for the speedy updates.  This request got lost amid the other comments, but can you deal with the `'blankdraft': 'g6',` line in `normalizedhash`?

Issue #339, opened 2016-05-10
Should not skip placing {{merge from}} if proposal is for another article


nyuszika7h:
I nominated "Randy Cunningham: 9th Grade Ninja (season 1)" and "Randy Cunningham: 9th Grade Ninja (season 2)" for merging with "List of Randy Cunningham: 9th Grade Ninja episodes". Twinkle didn't place a `{{merge from}}` on the episode list article for season 2 because there was already a tag. Except that was for season 1.

PR #471, opened 2018-12-11
tag: avoid putting tags into {{multiple issues|section=y|  }} boxes; remove miOldStyleTest + more related fixes

### LINKED ISSUE #221: Twinkle added my maintenance tag in the wrong place
I wanted to add a maintenance tag to [Poway High School](http://en.wikipedia.org/wiki/Poway_High_School). However, since the first section contained a multiple issues maintenance tag, it added the tag to the first section's multiple issues tag instead of to the top of the article. Example revision: https://en.wikipedia.org/w/index.php?title=Poway_High_School&diff=612281820&oldid=612281653

siddharthvp:
Closes #221.

Since all multiple issues tags within sections have |section= parameter, the issue can be resolved simply by adding a negative lookahead for |section= to the regex that checks for {{multiple issues}} template.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -86,7 +86,7 @@ Twinkle.tag.callback = function friendlytagCallback() {
 							label: 'Group inside {{multiple issues}} if possible',
 							value: 'group',
 							name: 'group',
-							tooltip: 'If applying three or more templates supported by {{multiple issues}} and this box is checked, all supported templates will be grouped inside a {{multiple issues}} template.',
+							tooltip: 'If applying two or more templates supported by {{multiple issues}} and this box is checked, all supported templates will be grouped inside a {{multiple issues}} template.',
 							checked: Twinkle.getFriendlyPref('groupByDefault')
 						}
 					]
@@ -569,7 +569,7 @@ Twinkle.tag.article.tagCategories = {
 		],
 		"Timeliness": [
 			"current",
-			"update"			
+			"update"
 		],
 		"Neutrality, bias, and factual accuracy": [
 			"autobiography",
@@ -1006,7 +1006,7 @@ Twinkle.tag.callbacks = {
 						Morebits.status.warn( 'Info', 'Found {{' + tag +
 							'}} on the article already...excluding' );
 						// don't do anything else with merge tags
-						if ( ['merge', 'merge to', 'merge from'].includes(tag) ) {
+						if ( ['merge', 'merge to'].indexOf(tag) !== -1 ) {
 							params.mergeTarget = params.mergeReason = params.mergeTagOther = null;
 						}
 					}
@@ -1028,12 +1028,12 @@ Twinkle.tag.callbacks = {
 				if( tags.length > 0 ) {
 					summaryText += ', and';
 				}
-				
+
 				var miRegex = new RegExp("(\\{\\{\\s*" + miTest[1] + "\\s*(?:\\|(?:\\{\\{[^{}]*\\}\\}|[^{}])*)?)\\}\\}\\s*", "im");
 				pageText = pageText.replace(miRegex, "$1" + tagText + "}}\n");
 				tagText = "";
 
-			} else if( params.group && groupableTags.length >= 3 ) {
+			} else if( params.group && groupableTags.length >= 2 ) {
 				Morebits.status.info( 'Info', 'Grouping supported tags inside {{multiple issues}}' );
 
 				groupableTags.sort();
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1011,10 +1011,8 @@ Twinkle.tag.callbacks = {
 			}
 
 			var miTest = /\{\{(multiple ?issues|article ?issues|mi)(?!\s*\|\s*section\s*=)[^}]+\{/im.exec(pageText);
-			var miOldStyleRegex = /\{\{(multiple ?issues|article ?issues|mi)\s*\|([^{]+)\}\}/im;
-			var miOldStyleTest = miOldStyleRegex.exec(pageText);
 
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I'm not particularly familiar with {{mi}}, so is there a reason this addition isn't needed in `miOldStyleRegex` as well?

siddharthvp:
Now that I investigate, [old-style multiple issues](https://en.wikipedia.org/wiki/Template:Multiple_issues/doc/old) is really old-style and no longer supported (since at least 2014). It now [just gives an error msg](https://en.wikipedia.org/w/index.php?title=Ume%C3%A5_Arts_Campus&oldid=401252338). All existing instances have been either fixed, or in very rare cases, commented out from pages (eg. [1](https://en.wikipedia.org/w/index.php?title=Ume%C3%A5_Arts_Campus&diff=609689584&oldid=579559265&diffmode=source) [2](https://en.wikipedia.org/w/index.php?title=Silver_Bell_Mountains&action=edit)). [Category:Pages using multiple issues with unknown parameters](https://en.wikipedia.org/wiki/Category:Pages_using_multiple_issues_with_unknown_parameters) is empty. 

Hence we can remove this mioldstyleregex from TW, as anyway it doesn't make sense to add in tags to a commented out template. And trying to uncomment the thing and changing it to new-style mi is quite unnecessary, and would be confusing to the tagging editor as they cannot see the commented-out mi from the 'Read' mode.

siddharthvp:
> * Closes #339
apparently this doesn't work so #339 would need to be closed manually

modules/friendlytag.js
```diff
@@ -991,30 +991,31 @@ Twinkle.tag.callbacks = {
...
-					if (params.tags[i] === "merge" || params.tags[i] === "merge from" ||
-						params.tags[i] === "merge to") {
-						params.mergeTarget = params.mergeReason = params.mergeTagOther = false;
+					if(tag === 'merge from') {
+						tags = tags.concat( tag );
+					} else {
+						Morebits.status.warn( 'Info', 'Found {{' + tag +
+							'}} on the article already...excluding' );
+						// don't do anything else with merge tags
+						if ( ['merge', 'merge to', 'merge from'].includes(tag) ) {
```
Amorymeltzer: As in https://github.com/azatoth/twinkle/pull/487#discussion_r244915402 we should avoid the use of `includes`

modules/friendlytag.js
```diff
@@ -1028,19 +1029,11 @@ Twinkle.tag.callbacks = {
...
-					var miRegex = new RegExp("(\\{\\{\\s*" + miTest[1] + "\\s*(?:\\|(?:\\{\\{[^{}]*\\}\\}|[^{}])*)?)\\}\\}\\s*", "im");
-					pageText = pageText.replace(miRegex, "$1" + tagText + "}}\n");
-				}
+				var miRegex = new RegExp("(\\{\\{\\s*" + miTest[1] + "\\s*(?:\\|(?:\\{\\{[^{}]*\\}\\}|[^{}])*)?)\\}\\}\\s*", "im");
+				pageText = pageText.replace(miRegex, "$1" + tagText + "}}\n");
 				tagText = "";
-			} else if( params.group && groupableTags.length >= 3 ) {
+
+			} else if( params.group && groupableTags.length >= 2 ) {
 				Morebits.status.info( 'Info', 'Grouping supported tags inside {{multiple issues}}' );
```
Amorymeltzer: I think this might run into a situation with section MIs where this message would still show.  I can't quite place my finger on it anymore having only replicated it once, but perhaps to be safe it'd be better to write _Grouping **any** supported tags_ ?

modules/friendlytag.js
```diff
@@ -991,30 +991,31 @@ Twinkle.tag.callbacks = {
...
-						params.mergeTarget = params.mergeReason = params.mergeTagOther = false;
+					if(tag === 'merge from') {
+						tags = tags.concat( tag );
+					} else {
+						Morebits.status.warn( 'Info', 'Found {{' + tag +
+							'}} on the article already...excluding' );
+						// don't do anything else with merge tags
+						if ( ['merge', 'merge to', 'merge from'].includes(tag) ) {
+							params.mergeTarget = params.mergeReason = params.mergeTagOther = null;
+						}
```
Amorymeltzer: Am I wrong that I think you meant to put this before the `don't do` comment?  As it stands, `merge from` will never end up here, as this is restricted to the innermost `else` block rather than the outer one.

Amorymeltzer:
Since it's not in the commit message, noting here for the record:
- The old style of multiple issues was [changed July 2012 by MSGJ](https://en.wikipedia.org/w/index.php?title=Template:Multiple_issues&diff=501152686&oldid=500713856)
- Discussions archived [here](https://en.wikipedia.org/wiki/Template_talk:Multiple_issues/Archive_10#Suggestion_2) and [here](https://en.wikipedia.org/wiki/Template_talk:Multiple_issues/Archive_11#Invoke_standalone_templates)
- Updated here in 99e7d7a0f8bc44d10740a75dd1a0890eb777fe5a and improved with nifty conversion code the following month in 1bfc2249bc5be90bf40d13e6f8585d2a84372c76

Regarding moving from `>=3` to `>=3`, the [MI doc is mixed](https://en.wikipedia.org/wiki/Template:Multiple_issues/doc): one place says for three or greater, another suggests it's fine for two.  Mobile-friendly or not, I think two is better, so glad for this.  The suggestion for >=3 seems likely to have come from twinkle (well, FurMe) anyway, so I feel fine overruling.

modules/friendlytag.js
siddharthvp: What with section MIs? Section tags or section MI tags are completely ignored (first commit).

Amorymeltzer: Yeah, I haven't been able to replicate this in the past few days, must've misread a diff when testing

PR #475, opened 2018-12-13
Change refimprove to more citations needed

AntiCompositeNumber:
Template has been renamed, see https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Refimprove


AntiCompositeNumber:
How did I not notice that the first time...Oh well, fixed.

PR #476, opened 2018-12-13
fluff: show rollback/vandalism links for IP ranges

Amorymeltzer:
Follow-up to https://github.com/azatoth/twinkle/pull/439#issuecomment-436102516

Relies upon the id in `MediaWiki:Sp-contributions-footer-anon-range` to discover if we're viewing a range's contributions.  This is required because while using wgRelevantUserName (#439) avoids the previous iteration's console error, it isn't (reliably) refined for IP range, and doesn't exist when viewing their contributions (see https://phabricator.wikimedia.org/T206954).

Incidentally, this doesn't break for ranges outside the CIDR limit (/16 on enWiki) because mediawiki incorrectly shows `MediaWiki:Sp-contributions-footer-anon` in that case. Depending on how (or if) https://phabricator.wikimedia.org/T211910 is resolved, this may or may not need updating.

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -38,7 +38,10 @@ Twinkle.fluff = {
 			return span;
 		};
 
-		if( mw.config.get('wgCanonicalSpecialPageName') === "Contributions" && mw.config.exists('wgRelevantUserName') ) {
+		// $('sp-contributions-footer-anon-range') relies on the fmbox
+		// id in [[MediaWiki:Sp-contributions-footer-anon-range]] and
+		// is used to show rollback/vandalism links for IP ranges
+		if( mw.config.get('wgCanonicalSpecialPageName') === "Contributions" && (mw.config.exists('wgRelevantUserName') || !!$('#sp-contributions-footer-anon-range')[0])) {
 			//Get the username these contributions are for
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
If this is merged, I'll put a note in [MediaWiki:Sp-contributions-footer-anon-range](https://en.wikipedia.org/wiki/MediaWiki:Sp-contributions-footer-anon-range) to not change the `id` without dropping a line at WT:TW.

One other thing to note: this means that `username` (Line XXX) will be undefined for ranges until `wgRelevantUserName` is expanded to include ranges.  Theoretically, we could get around that with by using something like `$('input[name=target].mw-input')[0].value` as a fallback, or even just in place of `wgRelevantUserName`, but that doesn't work on `Special:Contributions/`, which is what #215 was about; more importantly, though, it doesn't matter, as this `username` is only used to compare to the twinkle user vis à vis their preferences, so it need not ever be defined for IP ranges, much less IPs.

MusikAnimal:
I like it! It may be a little hacky, but it works and is helpful. I shall merge :)

Since I introduced IP range contributions (one of my very few MediaWiki core patches!), I may take a stab at [T206954](https://phabricator.wikimedia.org/T206954) and once merged we'll update Twinkle accordingly.

PR #480, opened 2018-12-17
 tag: add custom tags for files and redirects as well

siddharthvp:
Added custom tags support for files per request at [Wikipedia_talk:Twinkle#Custom_tags_for_non-article_namespace](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Custom_tags_for_non-article_namespace). Then realised that now only redirects are being left out, so added it to that too.

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -555,6 +555,22 @@ Twinkle.config.sections = [
 			type: "customList",
 			customListValueTitle: "Template name (no curly brackets)",
 			customListLabelTitle: "Text to show in Tag dialog"
+		},
+		{
+			name: "customFileTagList",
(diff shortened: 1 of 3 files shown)


modules/twinkleconfig.js
```diff
@@ -549,12 +549,28 @@ Twinkle.config.sections = [
 			enumValues: { "cat": "By categories", "alpha": "In alphabetical order" }
 		},
 		{
-			name: "customTagList",
```
Amorymeltzer: I won't be able to check this out until the new year, but won't this cause an issue with people's existing configs?

siddharthvp: ooohh! You're right, of course. Fixed.

Amorymeltzer:
Nice!

PR #482, opened 2018-12-20
Add new G14 option, previously G6 disambig

Amorymeltzer:
https://en.wikipedia.org/w/index.php?oldid=874654073#Proposal:_Move_deletion_of_disambiguation_pages_from_G6_%E2%86%92_G14


modules/twinklespeedy.js
```diff
@@ -1993,7 +1992,7 @@ Twinkle.speedy.callback.evaluateUser = function twinklespeedyCallbackEvaluateUse
 	if (form.notify.checked) {
 		$.each(normalizeds, function(index, norm) {
 			if (Twinkle.getPref('notifyUserOnSpeedyDeletionNomination').indexOf(norm) !== -1) {
-				if (norm === 'g6' && ['disambig', 'copypaste'].indexOf(values[index]) === -1) {
+				if (norm === 'g6' && 'copypaste'.indexOf(values[index]) === -1) {
```
MusikAnimal: I think `indexOf` in this case is meant to be used on an array. Maybe you meant `'copypaste' === values[index]`? Though I think your code might still work, coincidentally.

Amorymeltzer: `'copypaste' !== values[index]` but yeah.  The blocks around it do the same thing, weirdly.

Amorymeltzer: @MusikAnimal Done

siddharthvp: > I think `indexOf` in this case is meant to be used on an array. 

It can also be used on an string, in which case it is treated as an array of its characters. Though obviously that isn't intended here.

PR #487, opened 2018-12-28
tag: further improvements

siddharthvp:
- simplified `Twinkle.tag.callback.evaluate`, reducing the lines of code as well as making to easier to maintain. Now if you add subgroups to some tags, you no longer have to update anything in this function.
- file tags: convert tag-specific prompt dialogs to subgroups for better user experience

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -28,7 +28,7 @@ Twinkle.tag = function friendlytag() {
 		Twinkle.addPortletLink( Twinkle.tag.callback, "Tag", "friendly-tag", "Add maintenance tags to file" );
 	}
 	// article/draft article tagging
-	else if( [0,118].includes(mw.config.get('wgNamespaceNumber')) && mw.config.get('wgCurRevisionId') ) {
+	else if( [0,118].indexOf(mw.config.get('wgNamespaceNumber')) !== -1 && mw.config.get('wgCurRevisionId') ) {
 		Twinkle.tag.mode = 'article';
 		Twinkle.addPortletLink( Twinkle.tag.callback, "Tag", "friendly-tag", "Add maintenance tags to article" );
 	}
@@ -569,7 +569,7 @@ Twinkle.tag.article.tagCategories = {
 		],
 		"Timeliness": [
 			"current",
-			"update"			
+			"update"
 		],
 		"Neutrality, bias, and factual accuracy": [
 			"autobiography",
@@ -781,9 +781,33 @@ Twinkle.tag.file.licenseList = [
 Twinkle.tag.file.commonsList = [
 	{ label: '{{Copy to Commons}}: free media that should be copied to Commons', value: 'Copy to Commons' },
 	{ label: '{{Do not move to Commons}} (PD issue): file is PD in the US but not in country of origin', value: 'Do not move to Commons' },
-	{ label: '{{Do not move to Commons}} (other reason)', value: 'Do not move to Commons_reason' },
-	{ label: '{{Keep local}}: request to keep local copy of a Commons file', value: 'Keep local' },
-	{ label: '{{Now Commons}}: file has been copied to Commons', value: 'subst:ncd' }
+	{ 	label: '{{Do not move to Commons}} (other reason)',
+		value: 'Do not move to Commons_reason',
+		subgroup: {
+			type: 'input',
+			name: 'DoNotMoveToCommons',
+			label: 'Reason: ',
+			tooltip: 'Enter the reason why this image should not be moved to Commons (required)'
+		}
+	},
+	{ 	label: '{{Keep local}}: request to keep local copy of a Commons file',
+		value: 'Keep local',
+		subgroup: {
+			type: 'input',
+			name: 'keeplocalName',
+			label: 'Commons image name if different: ',
+			tooltip: 'Name of the image on Commons (if different from local name), excluding the File: prefix:'
+		}
+	},
+	{ 	label: '{{Now Commons}}: file has been copied to Commons',
+		value: 'subst:ncd',
+		subgroup: {
+			type: 'input',
+			name: 'ncdName',
+			label: 'Commons image name if different: ',
+			tooltip: 'Name of the image on Commons (if different from local name), excluding the File: prefix:'
+		}
+	}
 ];
 
 Twinkle.tag.file.cleanupList = [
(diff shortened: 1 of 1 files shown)


modules/friendlytag.js
```diff
@@ -28,7 +28,7 @@ Twinkle.tag = function friendlytag() {
 		Twinkle.addPortletLink( Twinkle.tag.callback, "Tag", "friendly-tag", "Add maintenance tags to file" );
 	}
 	// article/draft article tagging
-	else if( ( mw.config.get('wgNamespaceNumber') === 0 || mw.config.get('wgNamespaceNumber') === 118 || /^Wikipedia( talk)?:Articles for creation\//.exec(Morebits.pageNameNorm) ) && mw.config.get('wgCurRevisionId') ) {
+	else if( [0,118].includes(mw.config.get('wgNamespaceNumber')) && mw.config.get('wgCurRevisionId') ) {
```
MusikAnimal: As much as I'd like to ignore IE, we should probably support IE 11 as MediaWiki does. Hence `includes` isn't an option... yet!

siddharthvp: Jeez! Microsoft would be doing every web developer on the planet a great favour if they just kill IE, and kill it with fire ... 
@MusikAnimal How about creating a shim in morebits.js to support this function? 
`if (!Array.prototype.includes) { Array.prototype.includes = function(el) { return this.indexOf(el) !== -1; } }`
I find `indexOf( ) !== -1` to be quite confusing, unintuitive and unreadable. Even after some months of JS, I still confuse `!== -1` with `=== -1` and vice versa.

siddharthvp: Fixed for now.

Amorymeltzer:
@siddharthvp 	This looks mostly fine from reading through the code, although I'll hold off on testing and doing a thorough review until the merge conflicts are dealt with.  I'll probably be squashing them given things like the leftover `console.log` and the `includes`/`indexOf`.

siddharthvp:
> until the merge conflicts are dealt with.

Done.

modules/friendlytag.js
```diff
@@ -907,6 +975,14 @@ Twinkle.tag.file.replacementList = [
 	{ label: '{{PNG version available}}', value: 'PNG version available' },
 	{ label: '{{Vector version available}}', value: 'Vector version available' }
 ];
+Twinkle.tag.file.replacementList.forEach(function(idx,el) {
```
Amorymeltzer: It can get maddening, but jQuery's `.each` uses `index, value` while javascript's `forEach` uses `value, index`.

```suggestion
Twinkle.tag.file.replacementList.forEach(function(el,idx) {
```

modules/friendlytag.js
```diff
@@ -959,13 +1035,13 @@ Twinkle.tag.callbacks = {
 						currentTag += '|source=' + params.tagParameters.closeParaphrasing;
```
Amorymeltzer: There are a few more of these (seven total?) that still need removing.

```suggestion
						currentTag += '|source=' + params.closeParaphrasing;
```

modules/friendlytag.js
```diff
@@ -1438,76 +1467,97 @@ Twinkle.tag.callback.evaluate = function friendlytagCallbackEvaluate(e) {
...
-			params.translationComments = form["articleTags.translationComments"] ? form["articleTags.translationComments"].value : null;
+
+			// Save values of input text and checkbox parameters, this works as
+			// Quickform input fields within subgroups of elements with name 'articleTags'
+			// have their name attribute as 'articleTags.' + name of the subgroup element
+			$(form).find('input[type=text]').each(function(idx,el) {
+				// el are the HTMLInputElements, el.name gives their name attribute
+				params[el.name.slice(12)] = form[el.name].value;
+			});
+			$(form).find("input[type=checkbox][name^='articleTags.']").each(function(idx,el) {
```
Amorymeltzer: Nifty though this is, `select` dropdown menus are left out, breaking `notability`, `globalize`, and `Should be PNG`.

Amorymeltzer:
@siddharthvp In addition to the above changes, the changes to file tagging simply don't work: they all end up as `undefined` and a number of the parameters are repeated.  Compare the changes made with [the current tag module](https://test.wikipedia.org/w/index.php?title=File:Amory_St.jpg&diff=prev&oldid=385894) to those made with [this version](https://test.wikipedia.org/w/index.php?title=File:Amory_St.jpg&diff=385896&oldid=385895)

modules/friendlytag.js
```diff
@@ -1301,86 +1377,42 @@ Twinkle.tag.callbacks = {
 					"Now Commons"].indexOf(tag) !== -1) {
 					text = text.replace(/\{\{(mtc|(copy |move )?to ?commons|move to wikimedia commons|copy to wikimedia commons)[^}]*\}\}/gi, "");
 				}
-				if (tag === "Vector version available") {
-					text = text.replace(/\{\{((convert to |convertto|should be |shouldbe|to)?svg|badpng|vectorize)[^}]*\}\}/gi, "");
-				}
 
-				currentTag = "{{" + (tag === "Do not move to Commons_reason" ? "Do not move to Commons" : tag);
```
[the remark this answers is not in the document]
siddharthvp: Ah, accidentally removed. This was why file tagging wasn't working.

modules/friendlytag.js
siddharthvp: oh, they crept in from the merge commit.

modules/friendlytag.js
siddharthvp: Dealt with. Textareas covered as well.

modules/friendlytag.js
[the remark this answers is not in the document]
Amorymeltzer: Okay, that makes sense!  Weird way of handling that

modules/friendlytag.js
```diff
@@ -1434,80 +1465,96 @@ Twinkle.tag.callback.evaluate = function friendlytagCallbackEvaluate(e) {
...
 	}
 
+	params.tags = form.getChecked( Twinkle.tag.mode + 'Tags' );
+
+	// Save values of input fields into params object. This works as quickform input
+	// fields within subgroups of elements with name 'articleTags' (say) have their
+	// name attribute as 'articleTags.' + name of the subgroup element
+
+	var name_prefix = Twinkle.tag.mode + 'Tags.';
+	$(form).find("[name^='" + name_prefix + "']").each(function(idx,el) {
```
Amorymeltzer: Wouldn't it be somewhat more parsimonious to add a negative selector for the checkboxes, since they're covered below?

```suggestion
	$(form).find("[name^='" + name_prefix + "']:not(input[type=checkbox])").each(function(idx,el) {
```

Amorymeltzer:
It's certainly different, but I really like the changes here!  Smart and nifty, cleans things up real well, and should make it **much** easier to maintain 🏆

modules/friendlytag.js
siddharthvp: To (slightly) improve selector efficiency, I think a better way is to use a single `$(form).find()`, and use `el.type` to see if el is a checkbox, and hence treat it differently.

Amorymeltzer: Super clean! :100:

PR #492, opened 2019-01-11
fluff: Make summaries uniform, normalize notices, remove period

### LINKED ISSUE #490: Feature request: Ability to edit automatic summaries (e.g. remove auto-capitalization and period in reverting message)
I wonder if it would be possible for users to be able to change parts of the automatic summaries when rollbacking with an explanation message. In particular, I feel that the addition of a period and the auto-capitalization of the first character makes the summary seem terse, considering that the default rollback and undo message don't include periods. I also personally think a separator of " - " between the "Reverted edits by X (talk)" and the revert reasoning looks slightly clearer than the current separator of ": ".

E.g. instead of _Reverted edits by X (talk): Test edit. (TW)_, I'm suggesting an option to edit the summary to show up as _Reverted edits by X (talk) - test edit (TW)_

QEDK:
Closes #490

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -312,7 +312,7 @@ Twinkle.fluff.callbacks = {
 		var revs = $(xmlDoc).find('rev');
 
 		if( revs.length < 1 ) {
-			self.statelem.error( 'We have less than one additional revision, thus impossible to revert' );
+			self.statelem.error( 'We have less than one additional revision, thus impossible to revert.' );
 			return;
 		}
 		var top = revs[0];
@@ -326,7 +326,7 @@ Twinkle.fluff.callbacks = {
 			if( lastuser === self.params.user ) {
 				switch( self.params.type ) {
 				case 'vand':
-					Morebits.status.info( 'Info', [ 'Latest revision was made by ', Morebits.htmlNode( 'strong', self.params.user ) , '. As we assume vandalism, we continue to revert' ]);
+					Morebits.status.info( 'Info', [ 'Latest revision was made by ', Morebits.htmlNode( 'strong', self.params.user ) , '. As we assume vandalism, we continue to revert.' ]);
 					break;
 				case 'agf':
 					Morebits.status.warn( 'Warning', [ 'Latest revision was made by ', Morebits.htmlNode( 'strong', self.params.user ) , '. As we assume good faith, we stop reverting, as the problem might have been fixed.' ]);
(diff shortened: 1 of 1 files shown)


modules/twinklefluff.js
```diff
@@ -356,7 +356,7 @@ Twinkle.fluff.callbacks = {
 				self.params.user = revs[1].getAttribute( 'user' );
 				break;
 			case 'agf':
-				Morebits.status.warn( 'Notice', [ 'Good faith revert was chosen on ', Morebits.htmlNode( 'strong', self.params.user ), '. This is a whitelisted bot, and since bots have no faith, AGF rollback will not proceed.' ] );
```
Amorymeltzer: This is hilarious

Amorymeltzer:
"Stopping revert" is added twice; does that sound like twinkle stopped mid-way and something might have gotten half-done?  Is "stopping reversion" better?

These messages have a weird mix of tenses/person/activity in them, I wonder if it's possible to get them in more agreement?  I might mull on that.

QEDK:
> "Stopping revert" is added twice; does that sound like twinkle stopped mid-way and something might have gotten half-done? Is "stopping reversion" better?
> 
> These messages have a weird mix of tenses/person/activity in them, I wonder if it's possible to get them in more agreement? I might mull on that.

I agree, it's very mixed.

modules/twinklefluff.js
```diff
@@ -326,13 +326,13 @@ Twinkle.fluff.callbacks = {
 			if( lastuser === self.params.user ) {
 				switch( self.params.type ) {
 				case 'vand':
-					Morebits.status.info( 'Info', [ 'Latest revision was made by ', Morebits.htmlNode( 'strong', self.params.user ) , '. As we assume vandalism, we continue to revert' ]);
+					Morebits.status.info( 'Info', [ 'Latest revision was made by ', Morebits.htmlNode( 'strong', self.params.user ) , '. As we assume vandalism, we proceed to revert.' ]);
```
Amorymeltzer: Match tense and structure with other two.

```suggestion
					Morebits.status.info( 'Info', [ 'Latest revision was made by ', Morebits.htmlNode( 'strong', self.params.user ) , '. As we assume vandalism, we will proceed with the revert.' ]);
```

modules/twinklefluff.js
Amorymeltzer: Thinking on it further, hilarity aside this might actually be clearer.  If we want to remove it, maybe something like `This is a whitelisted bot so AGF rollback will not proceed.`

modules/twinklefluff.js
```diff
@@ -566,15 +566,11 @@ Twinkle.fluff.callbacks = {
...
+	// append user's custom reason
 	if (userString) {
 		result += ': ' + Morebits.string.toUpperCaseFirstChar(userString);
-		if (userString.search(/[.?!;]$/) === -1) {
-			result += '.';
-		}
-	} else {
-		result += '.';
 	}
+
```
Amorymeltzer: Extra tab.

```suggestion

```

modules/twinklefluff.js
QEDK: Fixed.

modules/twinklefluff.js
QEDK: Fixed. I left it for adequate resolving space probably.

modules/twinklefluff.js
QEDK: Fixed.

PR #497, opened 2019-01-14
speedy: Enable U2 on redirects

Amorymeltzer:
Requested at [WT:TW](https://en.wikipedia.org/w/index.php?oldid=878315943#Function_request).  Given the data I posted there, this would get used much at all, but as far as U2 noms go it'd be about 1/3 of them, so that seems worth allowing.

U1 was enabled in 61c2f4a33

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -771,15 +771,15 @@ Twinkle.speedy.userAllList = [
 			size: 60
 		} : null),
 		hideSubgroupWhenMultiple: true
-	}
-];
-
-Twinkle.speedy.userNonRedirectList = [
+	},
 	{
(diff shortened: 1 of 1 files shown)


MusikAnimal:
Looks good! At some point we might consider checking wgRelevantUsername to determine when U2 should be available. Not important, just an idea.

Amorymeltzer:
Just jotting down some notes on that front:

- If `wgRelevantUsername` doesn't exist and namespace is 2 or 3, should be enabled
- If it _does_ exist but mw.util.isIpAddress is true and (namespace is 2 or (namespace is 3 and it's a subpage)), should be enabled
- To hell with IP ranges

PR #506, opened 2019-01-22
morebits: Partially revert #481

Amorymeltzer:
#481 removed a ton of functions

cc @MusikAnimal and @siddharthvp


siddharthvp:
So sorry, man! I'm so glad you found it before it went on-wiki. [This commit](https://github.com/azatoth/twinkle/pull/481/commits/b3a54386aad79ce91c588da85667237001a18da5
) must have been the problem of course. It could never ocurred to me to look at the diff counts for a final sanity check, I had just done some spot-checks to ensure nothing got removed.

_Right now_, don't have the time don't have the time to look into this, but I certainly see that some (many?) functions like getPageName weren't removed. You've duplicated them.

Amorymeltzer:
Indeed — I did a quick copy paste just to get something up as a starter.  Closing in favor of #507

PR #508, opened 2019-01-23
warn: Remove outdated hack from f53706a

Amorymeltzer:
Basically, in 26 May 2016, the devs turned on in-place rollback (i.e. no post-rollback diff) and confirmation (if opened in a new window) with [phab:T88044](https://phabricator.wikimedia.org/T88044).  This broke a bunch of important stuff (see [phab:T136375](https://phabricator.wikimedia.org/T136375)) and [made a lot of people very angry](https://en.wikipedia.org/wiki/Wikipedia:Village_pump_(technical)/Archive_147#Rollback_function_has_been_changed).  @atlight hacked together a quick fix for Twinkle in f53706a while the devs worked it out.  Widely regarded as a bad move, the change to rollback was [quickly rolled back](https://gerrit.wikimedia.org/r/c/mediawiki/core/+/291249).  The change in Twinkle, however, has remained ever since, but hasn't been necessary.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -18,74 +18,22 @@ Twinkle.warn = function twinklewarn() {
 			Twinkle.addPortletLink( Twinkle.warn.callback, "Warn", "tw-warn", "Warn/notify user" );
 	}
 
-	// Modify URL of talk page on rollback success pages. This is only used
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Change is currently live on test.wiki fwiw

PR #509, opened 2019-01-24
warn: Add option to allow automatic opening of menu after revert

### LINKED ISSUE #162: [Rollback vandal] to open user's talk page with warn module instead of in edit mode.
A majority of the time it is appropriate to post a warning on the talkpage of an edit that is being rolled back as vandalism.  It would be easier to do this if Twinkle loaded the warn module on those user talk pages than opening the page in edit mode.  It may also as a side effect reduce the number of people rolling things back as vandalism incorrectly (and I wouldn't be opposed to the idea of it requiring a warning to be sent in order to roll it back as a vandal edit).  If the rollbacker doesn't want to leave a warning, they could always still use regular rollback or AGF rollback.

Amorymeltzer:
Closes #162

Still loads the page in edit mode, but optionally opens the warning menu.  Defaults to previous behavior (although perhaps there's an argument that it shouldn't).  I think this is likely to be popular!

One thing to consider is whether the config option should be in the warn or rollback on WP:TW/PREF.

The first commit is actually to `twinkleconfig` and moves an xfd-relevant item (`markXfdPagesAsPatrolled`) to the proper section (introduced in 6651b059).

I should note that I've put this live on test.wiki

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -16,6 +16,9 @@
 Twinkle.warn = function twinklewarn() {
 	if( mw.config.get( 'wgRelevantUserName' ) ) {
 			Twinkle.addPortletLink( Twinkle.warn.callback, "Warn", "tw-warn", "Warn/notify user" );
+			if (Twinkle.getPref('autoMenuAfterRollback') && mw.config.get('wgNamespaceNumber') === 3 && mw.util.getParamValue('vanarticle')) {
+				Twinkle.warn.callback();
+			}
 	}
 
 	// Modify URL of talk page on rollback success pages, makes use of a
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -669,12 +669,6 @@ Twinkle.config.sections = [
 			type: "customList",
 			customListValueTitle: "Template name (no curly brackets)",
 			customListLabelTitle: "Text to show in warning list (also used as edit summary)"
-		},
-
-		{
-			name: "markXfdPagesAsPatrolled",
-			label: "Mark page as patrolled when nominating for AFD (if possible)",
-			type: "boolean"
 		}
 	]
 },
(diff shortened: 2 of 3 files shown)


modules/twinklewarn.js
```diff
@@ -16,6 +16,9 @@
 Twinkle.warn = function twinklewarn() {
 	if( mw.config.get( 'wgRelevantUserName' ) ) {
 			Twinkle.addPortletLink( Twinkle.warn.callback, "Warn", "tw-warn", "Warn/notify user" );
+			if (Twinkle.getPref('autoMenuAfterRollback') && mw.config.get('wgNamespaceNumber') === 3 && mw.util.getParamValue('vanarticle')) {
```
Amorymeltzer: I feel like there's a better/more twinkley way of doing this...

siddharthvp:
Very nice, though I'd suggest giving a more specific name (`tw-vanarticle` ?) to the param value to rule out any possibility of conflict with other scripts.

Amorymeltzer:
`vanarticle` is used in quite a few twinkle modules (this one is from the rollback links), and has been for a while; at this point, I think there are more likely to be other scripts that _depend_ on it, particularly the rollback links.

MusikAnimal:
> I think this is likely to be popular!

I think so too!

> One thing to consider is whether the config option should be in the warn or rollback on WP:TW/PREF.

The new option only applies if "_Open user talk page when invoking rollback from user contributions_" (under Rollback) is enabled, so I think it should be right next to it. Perhaps if it's not too complicated, show/hide the new option accordingly.

modules/twinklewarn.js
MusikAnimal: This approach seems fine to me.

Amorymeltzer:
@MusikAnimal Good point on the location, although `openTalkPageOnAutoRevert` only applies to the contribs page; regardless of that state, the boxes from `openTalkPage` alone will dictate behavior on diffs.  I've moved it above them (to keep the references from this and `openTalkPageOnAutoRevert` clear) with a little note, how's that scan?  I don't think the config has any show/hide options, although that'd be nice in the future.

MusikAnimal:
Looks good! Thank you. I will make heavy use of this feature!

PR #510, opened 2019-01-25
morebits: enable multiline unbinder

siddharthvp:
Regex s flag appears to be resulting in an exception in IE 11 and legacy browsers. I have used [\s\S]*? instead of it so that newlines get matched. Eslint doesn't like this syntax for whatever reason, though this usage seems to be fine and is suggested online. 

While at some point (after making the commit) it came to my attention that the regex s flag is not supported in IE 11 and the like, I did not realise that it would cause an exception - I thought that these browsers would just ignore the flag. I was working in Ubuntu and so did not have IE handily available to test on.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1241,7 +1241,7 @@ Morebits.unbinder.prototype = {
 	 * @param {string} postfix
 	 */
 	unbind: function UnbinderUnbind( prefix, postfix ) {
-		var re = new RegExp( prefix + '([\s\S]*?)' + postfix, 'g' ); //eslint-disable-line no-useless-escape
+		var re = new RegExp( prefix + '([\\s\\S]*?)' + postfix, 'g' );
 		this.content = this.content.replace( re, Morebits.unbinder.getCallback( this ) );
 	},
 
--- a/morebits.js
+++ b/morebits.js
@@ -1241,7 +1241,7 @@ Morebits.unbinder.prototype = {
 	 * @param {string} postfix
 	 */
 	unbind: function UnbinderUnbind( prefix, postfix ) {
-		var re = new RegExp( prefix + '(.*?)' + postfix, 'gs' );
+		var re = new RegExp( prefix + '([\s\S]*?)' + postfix, 'g' ); //eslint-disable-line no-useless-escape
 		this.content = this.content.replace( re, Morebits.unbinder.getCallback( this ) );
 	},
 


siddharthvp:
Ah, I was going to make an editreq at Mediawiki talk:Gadget-morebits.js before seeing that Amorymeltzer has fixed it already. So this is no longer urgent ...

siddharthvp:
@Amorymeltzer While processing wikitext, it is useful to disregard everything between nowiki tags and HTML comments - this is what Unbinder elegantly helps to do. While it's currently used within a couple of morebits functions only, the Unbinder is quite cool and could be used in external scripts. In context of potential wider usage, this is as good as a bug - since the opening and closing of nowiki tags could appear in different paragraphs (though it may not be common).

/s works fine on both Chrome and Firefox on Ubuntu, and if I remember it right even on Chrome on Windows.

morebits.js
```diff
@@ -1241,7 +1241,7 @@ Morebits.unbinder.prototype = {
 	 * @param {string} postfix
 	 */
 	unbind: function UnbinderUnbind( prefix, postfix ) {
-		var re = new RegExp( prefix + '(.*?)' + postfix, 'gs' );
+		var re = new RegExp( prefix + '([\s\S]*?)' + postfix, 'g' ); //eslint-disable-line no-useless-escape
```
Amorymeltzer: lint is whining appropriately here, because this won't do what you think it will.  When using the contructor format instead of the regex literal, you have to escape the escape.
```suggestion
		var re = new RegExp( prefix + '([\\s\\S]*?)' + postfix, 'g' );
```

siddharthvp: Right! I hastily tested w/o using regex constructor form and decided lint was wrong.

Amorymeltzer:
@siddharthvp 	Do you have a good way to show the difference this will make?  I couldn't find a difference with AfD or with PROD when doing <nowiki> on a different line than </nowiki>

Amorymeltzer:
@siddharthvp Reping

siddharthvp:
@Amorymeltzer  Sorry for not replying earlier, it was a mixture of not seeing your msg and not knowing what to reply when I finally saw it. 

There are no advantages of doing this within TW at present. 

In [User:Enterprisey/reply-link](https://en.wikipedia.org/wiki/User:Enterprisey/reply-link), for instance, the the 54-line function `hasSig` can be implemented in just 3 lines if we make this tiny change to unbinder, as : `function hasSig(text) { var u = new Morebits.unbinder(text); u.unbind('<nowiki'>, '</no' + 'wiki>'); return u.content.includes('~~~~'); }`  Similarly, the function `getSectionWikitext` in the same script can also implemented way more neatly with unbinder if it supported multi-line.

Amorymeltzer:
Good enough for me

Issue #500, opened 2019-01-17
Expand and enforce linting guidelines


Amorymeltzer:
Let's talk about linting.  As noted in the docs, there isn't really a consistent style applied to Twinkle.  That's okay!  But with around 17,000 lines of code, maybe it's also okay to have some more options in place.  As of #400 we have the basics via `eslint:recommended`, and `no-caller` was added in cd45282 (`max-len` is listed but unused).

I went through the [eslint rules](https://eslint.org/docs/rules/) and picked out a bunch of possibilities; the list is a combination of options that we already follow, options we don't, options I just personally prefer, and options that I thought might seem reasonable to at least someone out there.  Some are listed with different configurations.  This is a _long_ list; still, it is by no means meant to be exhaustive or limiting, so feel free to proffer others.  Many are probably unnecessary or overkill.

I generated a table with the current results from each, arranged alphabetically by category, and intend to update the ✔️ or ❌ marks as this conversation goes forward.  The links will take you to the online docs page for each option, which should explain each in full.  I've made some (heavy-handed?) suggestions for a lot of them, in particular those with multiple configurations listed, to get us started, but feel free to disagree with those choices.

This isn't particularly urgent (see also #477), and there are a number of PRs we should review/merge beforehand to avoid conflicts, but once we've got a firm list I'll go through and make the changes.

cc @MusikAnimal @atlight @kevinji @azatoth

| Use?      | Option      | Config      | Curr errors | Fix?    | Category    | Notes?      |
| ----------- | ----------- | ----------- | ----------- | ----------- | ----------- | ----------- |
| ❌  | [no-extra-parens](https://eslint.org/docs/rules/no-extra-parens) | "error" | 146 | Yes | Possible errors | Alt|
| ❌  |   [no-extra-parens](https://eslint.org/docs/rules/no-extra-parens) | ["error", "functions"] | 0 | Yes | Possible errors | Alt; might be allow additional clarity in ternaries|
| ✔️  |   [no-extra-parens](https://eslint.org/docs/rules/no-extra-parens) | ["error", "all", { "nestedBinaryExpressions": false }] | 113 | Yes | Possible errors | Alt; required for `no-mixed-operators` |
| ✔️  |   [no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins) | "error" | 1 | No |  Possible errors| |
[...]

PR #511, opened 2019-02-02
fluff: Add [restore this revision] link to old revisions

### LINKED ISSUE #131: Add [restore this version] link when viewing old revisions
For wikipedia, twinkle only shows rollback options on select diff pages and contributions pages. Is there a way to have the 3 rollback buttons in other locations, such as the 'revision history/view history' section of an article? That would be very handy. Not sure if the coding makes this possible or if this all has been done before as I remember something like this in the past, though I'm not sure. Thanks anyway.

Amorymeltzer:
1st commit: Tidies up and removes duplicate/unnecessary code
- The same four lines, three of which used `spanTag`, were repeated frequently
- Newer revisions either are the latest or not
- No idea what `.firstrevisionheader` did but I can't find it anywhere on-wiki, so removed

2nd commit: Add [restore this revision] link to oldid pages

fluff: Add [restore this revision] link to oldid pages, closes #131

Checking for `wgDiffOldId` and `wgDiffNewId` excludes Special:Undelete, and so should be sufficient to avoid problematic diffs.  Some of the logic/checks may be a tad confusing since `wgDiffNewId` and `wgDiffOldId` have different behaviors (see [[[phab:T214985]]](https://phabricator.wikimedia.org/T214985)).  One advantage of the restructuring is that [restore this revision] will now appear when viewing the first revision as a diff with `diff=prev`, as may sometimes happen.

One thing to note is that when viewing old revisions, checking for `div#mw-revision-info` sufficiently prevents the link from appearing on the most recent revision, since `div#mw-revision-info-current` is used instead.  However, that is only the case if there is anything present at [[MediaWiki:Revision-info-current]], elsewise the fallback is to use `div#mw-revision-info`; this may have issues for other wikis if that page doesn't exist.  Moreover, unlike revisionasof, [$3 in revision-info](translatewiki.net/wiki/MediaWiki:Revision-info/qqq) doesn't expose a link we can capture, hence the imperfect `Morebits.query('oldid')` appears the best we can do.

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -37,6 +37,13 @@ Twinkle.fluff = {
 			span.appendChild( document.createTextNode( content ) );
 			return span;
 		};
+		var buildLink = function(color, text) {
+			var link = document.createElement('a');
+			link.appendChild(spanTag('Black', '['));
+			link.appendChild(spanTag(color, text));
+			link.appendChild(spanTag('Black', ']'));
+			return link;
+		};
 
 		// $('sp-contributions-footer-anon-range') relies on the fmbox
 		// id in [[MediaWiki:Sp-contributions-footer-anon-range]] and
@@ -50,17 +57,11 @@ Twinkle.fluff = {
 				var list = $("#mw-content-text").find("ul li:has(span.mw-uctop)");
 
 				var revNode = document.createElement('strong');
-				var revLink = document.createElement('a');
-				revLink.appendChild( spanTag( 'Black', '[' ) );
-				revLink.appendChild( spanTag( 'SteelBlue', 'rollback' ) );
-				revLink.appendChild( spanTag( 'Black', ']' ) );
+				var revLink = buildLink('SteelBlue', 'rollback');
 				revNode.appendChild(revLink);
 
 				var revVandNode = document.createElement('strong');
-				var revVandLink = document.createElement('a');
-				revVandLink.appendChild( spanTag( 'Black', '[' ) );
-				revVandLink.appendChild( spanTag( 'Red', 'vandalism' ) );
-				revVandLink.appendChild( spanTag( 'Black', ']' ) );
+				var revVandLink = buildLink('Red', 'vandalism');
 				revVandNode.appendChild(revVandLink);
 
 				list.each(function(key, current) {
@@ -82,12 +83,6 @@ Twinkle.fluff = {
 				return;
 			}
 
-			var firstRev = $("div.firstrevisionheader").length;
-			if( firstRev ) {
-				// we have first revision here, nothing to do.
-				return;
-			}
-
 			var otitle, ntitle;
 			try {
 				var otitle1 = document.getElementById('mw-diff-otitle1');
@@ -104,7 +99,7 @@ Twinkle.fluff = {
 
 			var old_rev_url = $("#mw-diff-otitle1").find("strong a").attr("href");
 
-			// Lets first add a [edit this revision] link
+			// Lets first add a [restore this revision] link
 			var query = new Morebits.queryString( old_rev_url.split( '?', 2 )[1] );
 
 			var oldrev = query.get('oldid');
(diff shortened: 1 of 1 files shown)


modules/twinklefluff.js
```diff
@@ -76,124 +77,91 @@ Twinkle.fluff = {
...
 				});
-				revertToRevisionLink.appendChild( spanTag( 'Black', '[' ) );
-				revertToRevisionLink.appendChild( spanTag( 'SaddleBrown', 'restore this version' ) );
-				revertToRevisionLink.appendChild( spanTag( 'Black', ']' ) );
-				ntitle.insertBefore( revertToRevision, ntitle.firstChild );
+				revertToRevision.appendChild(revertToRevisionLink);
 
-				return;
+				var otitle = document.getElementById(onode).parentNode;
+				otitle.insertBefore( revertToRevision, otitle.firstChild );
```
MusikAnimal: Nitpick -- I don't really care about `( code )` vs. `(code)` but you probably didn't mean to put `(code )` (with the trailing space). One day we should add some linter rules and unify all this. I think there are utilities to automatically update the code, too.

modules/twinklefluff.js
```diff
@@ -50,17 +57,11 @@ Twinkle.fluff = {
 				var list = $("#mw-content-text").find("ul li:has(span.mw-uctop)");
 
 				var revNode = document.createElement('strong');
-				var revLink = document.createElement('a');
-				revLink.appendChild( spanTag( 'Black', '[' ) );
-				revLink.appendChild( spanTag( 'SteelBlue', 'rollback' ) );
-				revLink.appendChild( spanTag( 'Black', ']' ) );
+				var revLink = buildLink('SteelBlue', 'rollback' );
```
MusikAnimal: Nitpick -- I don't really care about `( code )` vs. `(code)` but you probably didn't mean to put `(code )` (with the trailing space). One day we should add some linter rules and unify all this. I think there are utilities to automatically update the code, too.

MusikAnimal:
At cursory glance this seems to work well. I am a little concerned about checking the DOM for `.mw-revision-info`. What about comparing `wgRevisionId` and `wgCurRevisionId`? Could that work?

Visually, my only concern is proximity of the [restore this version] link to the breadcrumbs (say if you're viewing a subpage). I don't have any good ideas on how to address this... but I do see misclicks happening. Minor issue.

The link is pretty tiny too, but I suppose that's just a consequence of how that part of the page is styled by MediaWiki (or Common.css, etc).

modules/twinklefluff.js
Amorymeltzer: Nice catch!  I try and match the nearby style, but (mostly because of perl stuff) I [do care](#500), so thanks.

Amorymeltzer:
@MusikAnimal:
>What about comparing `wgRevisionId` and `wgCurRevisionId`

That was my initial plan, but ran into a few issues.  The more minor one is that on deleted pages, the two are the same (0), and I was trying to simplify things by removing that.  The larger issue is that `wgRevisionId` and `wgCurRevisionId` don't help distinguish diffs from old revisions, but `#mw-revision-info` does.  It's avoidable if you test (and fail to find) `#mw-diff-otitle1` and `#mw-diff-otitle1`, but that's just relying on a different id, so isn't much help.  Using `#mw-revision-info` simplifies the logic quite a bit.

>Visually, my only concern is proximity of the [restore this version] link to the breadcrumbs (say if you're viewing a subpage).

Ditto.  I first tried putting it above the diff nav links, but I think that's even more likely to cause misclicks.  I think the *real* issue is that, unlike diffs, this stuff isn't centered, so it all seems off-kilter and tiny next to the comparatively-giant message from `#mw-revision-info`.  Discoverability is definitely an issue, although I hope folks are familiar enough with the bold brown text to spot it.

Amorymeltzer:
Ignored my own advice in https://github.com/azatoth/twinkle/pull/511/commits/855097570d9a41cdc6882c0c479ef5094e69eca0, so this took a while to get back to.  After going through all the conflicts, I ended up refactoring this too much, so I'm closing this in favor of #561 (cosmetics) and #562 (the meat).

PR #512, opened 2019-02-10
Fix and improve sync.pl

Amorymeltzer:
1st commit: Fix sync.pl
- Roughly match previous behavior, although some dependencies have changed.
- Doesn't respect previous .mwbotrc preferences, but then again they haven't worked in years and won't work here

2nd commit: Autogenerate edit summaries, add some safety/error handling, other improvements
    
- Grab a commit hash from the edit summary of the most recent revsion of the on-wiki page, - and do our best to concatenate the log from there
- Require a clean working environment for all actions, not just --pull (matches implied intent in docs)
- Require confirmation of edit parameters before initiating
- Ensure files and pages being pulled/pushed/deployed exist
- Skip files without any changes
- Use File::Slurper over File::Slurp (See <http://blogs.perl.org/users/leon_timmermans/2015/08/fileslurp-is-broken-and-wrong.html>)

cc @MusikAnimal as the other active collaborator as well as @azatoth as the original author.  I've tested this at testwiki, and would be happy to walk though any of the changes or improvements.  The new version is rather more verbose, but should be a lot less likely to muck things up, which seems the right balance to me.

I didn't feel too bad about breaking past configurations since they haven't worked for ages, but also because this method should be safer.  The `pull` and `push` functions have been left in, as I think they may be useful for developing on non-enWiki projects; likewise the makefile, although I think that's likely to be more confusing than not at this point.  I've got a few other ideas for improvements, but this is enough for the time being.

On a personal note, it's nice to be able to work in perl again!

### DIFF
--- a/sync.pl
+++ b/sync.pl
@@ -5,28 +5,28 @@ use strict;
 use warnings;
 
 use English qw(-no_match_vars);
-use Encode;
 use utf8;
 
 use Config::General qw(ParseConfig);
 use Getopt::Long::Descriptive;
 use Git::Repository;
 use MediaWiki::API;
-use File::Slurp;
+use File::Slurper qw(read_text write_text);
 
 # Config file should be a simple file consisting of keys and values:
 # username = Jimbo Wales
 # lang = en
 # etc.
+my %conf;
 my $config_file = "$ENV{HOME}/.twinklerc";
-my %conf = ParseConfig($config_file);
+%conf = ParseConfig($config_file) if -e -f -r $config_file;
 
 my ($opt, $usage) = describe_options(
     "$PROGRAM_NAME %o <files...>",
     ['username|u=s', 'username for account on wikipedia', {default => $conf{username} // q{}}],
     ['password|p=s', 'password for account on wikipedia (do not use)', {default => $conf{password} // q{}}],
-    ['lang=s', 'Target language', {default => $conf{lang} // 'en'}],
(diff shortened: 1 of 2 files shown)


sync.pl
```diff
@@ -1,136 +1,301 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
+# sync.pl by azatoth (2011), update by amorymeltzer (2019)
 
-use v5.10;
```
Amorymeltzer: `use v5.10;` is required to ensure the perl6-like `say` is available, but that's been removed

sync.pl
```diff
@@ -1,136 +1,301 @@
...
+use strict;
+use warnings;
 
-use Git::Repository;
-use MediaWiki::Bot;
```
Amorymeltzer: Dead

sync.pl
```diff
@@ -1,136 +1,301 @@
...
-use File::Slurp;
-use Getopt::Long::Descriptive;
-use Encode;
+use English qw(-no_match_vars);
 use utf8;
 
-my $config_file = "$ENV{HOME}/.mwbotrc";
-my $c = {do "$config_file"} if -f $config_file;
+use Config::General qw(ParseConfig);
```
Amorymeltzer: Excessive?  Maybe.  But the whole point of cpan is not having to reinvent the wheel, especially when it comes to loading files.

PR #514, opened 2019-02-14
speedy: Exclude G1 and G2 from userspace

Amorymeltzer:
Match policy at WP:CSD, requested at [WT:TW](https://en.wikipedia.org/w/index.php?oldid=883223538#G1_in_userspace)


Amorymeltzer:
A [quarry search](https://quarry.wmflabs.org/query/33395) suggests this is not uncommon.

PR #515, opened 2019-02-17
xfd/protect/speedy: Improve handling in Module space

### LINKED ISSUE #478: Modules can't be CSD'd
CSD on modules fail due to the edit filter because of the  Lua error introduced. 

A similar issue is #219.

Amorymeltzer:
See individual commit messages for the specifics, but basically this just hammers `wgPageContentModel` to avoid the lua error message and be helpful.  Closes #478 

**Protect**:
- Disable tagging of modules

**CSD**:
- Places the CSD tag on the /doc subpage (i.e., what is done at WP:TFD)

**XFD**:
- Place the XFD tag on the /doc subpage
- General fixes and tweaks to handle modules, in particular language changes

Regarding the changes to `twinklexfd`:
- As written, the display style will show only `module` for modules.  Another option is to always list all the options (module, standard, tiny inline, etc.) but disable module for templates and disable the others for modules.  The first seems more reasonable to me? 
 Another option is to just always list 'em all, default to the correct one, and hope people don't muck it up.
- The `noinclude` option is still available for Modules, thought I don't imagine it should ever be used.  Might be good to turn that off?
- I avoided ternaries given how involved some of the items could be, hoping that if/else would be more readable

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -219,6 +219,7 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 				type: 'div',
 				label: 'Stub types and userboxes are not eligible for TfD. Stub types go to CfD, and userboxes go to MfD.'
 			} );
+		var templateOrModule = (mw.config.get('wgPageContentModel') === 'Scribunto') ? 'module' : 'template';
 		var tfd_category = work_area.append( {
 				type: 'select',
 				label: 'Choose type of action wanted: ',
@@ -230,7 +231,7 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 						var xfdtarget = new Morebits.quickForm.element( {
 							name: 'xfdtarget',
 							type: 'input',
-							label: 'Other template to be merged: '
+							label: 'Other ' + templateOrModule + ' to be merged: '
 						} );
 						target.parentNode.appendChild(xfdtarget.render());
 					} else {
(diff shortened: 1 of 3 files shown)


Amorymeltzer:
@JJMC89 I wouldn't mind you taking a look over the xfd stuff in particular

modules/twinklexfd.js
```diff
@@ -716,8 +723,8 @@ Twinkle.xfd.callbacks = {
 			var text = pageobj.getPageText();
 			var params = pageobj.getCallbackParameters();
 
-			pageobj.setPageText((params.noinclude ? "<noinclude>" : "") + "{{subst:template for discussion|help=off|" +
-				(params.tfdtype !== "standard" ? "type=" + params.tfdtype + "|" : "") + mw.config.get('wgTitle') + (params.noinclude ? "}}</noinclude>" : "}}") + text);
```
Amorymeltzer: Is `mw.config.get('wgTitle')` here ever necessary?  I don't see that it is.

JJMC89: It could be optionally passed to `|page=`. I think it may have been `|1=` at some point.

JJMC89:
_Grumbles about how poorly modules are handled in the templates._

* I don't think type needs to be shown for modules. The current way is better than letting it be incorrectly changed.
* `noinclude` isn't needed for modules since the notice is on /doc, which isn't transcluded when the module is invoked.

The changes look good.

modules/twinklexfd.js
Amorymeltzer: From looking at [[Template:Template for discussion/dated]], it seemed like it was used for optionally changing the section linked in the template.  That really only makes sense when bulk nominating, which twinkle doesn't handle yet, so I think it's fine to remove.

Amorymeltzer:
>* `noinclude` isn't needed for modules since the notice is on /doc, which isn't transcluded when the module is invoked.

Worth removing from the prompt then, or keeping as is?  I haven't been able to think of a scenario where it would be used.

modules/twinklexfd.js
JJMC89: > I think it's fine to remove.

Agreed

JJMC89:
> > * `noinclude` isn't needed for modules since the notice is on /doc, which isn't transcluded when the module is invoked.
> 
> Worth removing from the prompt then, or keeping as is? I haven't been able to think of a scenario where it would be used.

I'd remove it. It would just hide the transclusion from the /doc to the module, which is the whole point of the notice.

----
For templates, maybe we could use `|type=disabled` or `|type=hide` instead of `noinclude`?

modules/twinklexfd.js
Amorymeltzer: Worth noting too that `taggingTemplateForMerge` likewise doesn't have it.

Amorymeltzer:
>For templates, maybe we could use `|type=disabled` instead of `noinclude`?

Isn't there still some value to be derived, though, from being able `noinclude` the tfd on a substituted template?  Even if it's rare, WP:TfD is so wonky I'd sooner have more options rather than fewer.

JJMC89:
> > For templates, maybe we could use `|type=disabled` instead of `noinclude`?
> 
> Isn't there still some value to be derived, though, from being able `noinclude` the tfd on a substituted template? Even if it's rare, WP:TfD is so wonky I'd sooner have more options rather than fewer.

Forgot about subst - `noinclude` is needed for that since tf[dm]/dated won't subst.

PR #517, opened 2019-02-19
warn: Remove curly braces from display, allow for easy menu access via keyboard

### LINKED ISSUE #491: Feature request: remove {{uw- prefix from warning menu
Often, I know what warning I want, but spend several seconds trying to find it in the warning menu. If the entry in the menu simply started "test1" instead of "{{uw-test1}}" (for example), I could just click the menu and start typing "test" to find the entry. OTOH, it makes the mapping to actual templates less clear, so I'm not sure it if's worth that confusion.

Amorymeltzer:
Closes #491.  Leaves the curlies on for custom messages to make it clear they are, in fact, templates.

I also changed the dropdown menus so that the warning level numbers come first.  That is, `General note (1)` is now `1: General note`, and so on.  This makes the entire menu easy to access via keyboard, negating the need of the cursor for most actions.  As noted in https://github.com/azatoth/twinkle/issues/491#issuecomment-465255614 I'm not wild about the look, although maybe that's just 'cause they're different.

These should be squashed if both chosen.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1178,9 +1178,10 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 				selected = true;
 			}
 
+			// Slice out leading uw- from the menu display
 			var elem = new Morebits.quickForm.element( {
 				type: 'option',
-				label: "{{" + key + "}}: " + itemProperties.label,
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Update: I got over it 😀

PR #430, opened 2018-08-08
twinklexfd - Add section to RfD nomination

Amorymeltzer:
If a redirect being nominated for RfD redirects to a section, add on that section to the rfd2 nomination template.

I'm no whiz with XML, so this just copies how the target itself is found.  Closes #409

PR #311, opened 2016-01-19
Finish adding preview to XfD

Scimonster:
The continuation of #307, to close #168.

PR #307, opened 2016-01-03
Add preview to XfD module for afd, mfd, tfd/tfm, ffd

ndrewh:
It might work for other venues too but haven't tested.
Issue #168

This is a GCI task.

PR #521, opened 2019-02-22
xfd: Finish adding preview (support RfD, CfD, and CfD/s)

### LINKED ISSUE #168: XfD Preview
I would like to propose a preview link for previewing XfD nominations, similar to the preview link available when posting a user talk page warning/notification with Twinkle. 

It's frequently necessary to include wikilinks to articles, previous discussions, and policy and guideline pages when submitting an XfD. Since it is so easy to make a typo, or to choose the wrong policy shortcut, it would be very handy to have the ability to preview a nomination before creating it.

Amorymeltzer:
Closes #168 and #311.  Pretty much all of this was done by @Scimonster in #311 as part of GCI2015 (see also #307); I just handled the intervening conflicts (in particular #430)

I ditched the "waiting" messages in favor of the "loading" from the preview box and incorporated getDiscussionWikitext into RfD's todayslist.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -61,7 +61,7 @@ Twinkle.xfd.callback = function twinklexfdCallback() {
 			event: Twinkle.xfd.callback.change_category
 		} );
 	var namespace = mw.config.get('wgNamespaceNumber');
-	
+
 	categories.append( {
 			type: 'option',
 			label: 'AfD (Articles for deletion)',
@@ -164,7 +164,6 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 			value: oldreason,
 			tooltip: 'You can use wikimarkup in your reason. Twinkle will automatically sign your post.'
 		} );
-		// TODO possible future "preview" link here
 	};
 
 	form.previewer.closePreview();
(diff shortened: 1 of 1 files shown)


modules/twinklexfd.js
```diff
@@ -1313,18 +1300,38 @@ Twinkle.xfd.callbacks = {
...
-			apiobj.params.section = section;
-			Twinkle.xfd.callbacks.rfd.main(apiobj.params);
+		},
+		// This is a closure for the callback from the above API request, which gets the target of the redirect
+		findTargetCallback: function(callback) {
+			return function(apiobj) {
+				var xmlDoc = apiobj.responseXML;
+				var target = $(xmlDoc).find('redirects r').first().attr('to');
+				if( !target ) {
+					apiobj.statelem.error( "This page is currently not a redirect, aborting" );
```
Amorymeltzer: The preview will fail silently (except for the console message) here if the page isn't a redirect.  Certainly an improvement over the current situation, but I'm not entirely sure how to treat the preview (don't quit) differently than the submission (quit with error overlay), or if it's even worth it.

Amorymeltzer: This also happens for the ffd check above, so it's not anything new.  We should handle these better in general, but I think this can go forward.

PR #525, opened 2019-02-26
improve alert text for {{now commons}}

DannyS712:
Tweak line 1316: Explain that you do need to add the file extension, since its unclear (see my mistake at https://en.wikipedia.org/w/index.php?title=File:6150K.3347642.png&diff=885151052&oldid=584194465&diffmode=source)

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1313,7 +1313,7 @@ Twinkle.tag.callbacks = {
 						/* falls through */
 					case "Keep local":
 						input = prompt( "{{" + (tag === "subst:ncd" ? "Now Commons" : tag) +
-							"}} - Enter the name of the image on Commons (if different from local name), excluding the File: prefix:", "" );
+							"}} - Enter the name of the image on Commons (if different from local name), excluding the \"File:\" prefix but including the \".format\" suffix:", "" );
 						if (input === null) {
 							return true;  // continue
 						} else if (input !== "") {


modules/friendlytag.js
```diff
@@ -1313,7 +1313,7 @@ Twinkle.tag.callbacks = {
 						/* falls through */
 					case "Keep local":
 						input = prompt( "{{" + (tag === "subst:ncd" ? "Now Commons" : tag) +
-							"}} - Enter the name of the image on Commons (if different from local name), excluding the File: prefix:", "" );
+							"}} - Enter the name of the image on Commons (if different from local name), excluding the \"File:\" prefix but including the \".format\" suffix:", "" );
```
Amorymeltzer: I don't think `".format" suffix` is particularly understandable

Amorymeltzer:
How, uh, common is this?  It's a pretty lengthy alert, and I'm not aware of anywhere else where we distinguish the file extension from the name; extensions are part of the name, so mightn't it muddy the waters to suggest there are scenarios where the file extension isn't considered part of the filename?

DannyS712:
Well, at least on my computer, the _name_ of the file is different from the _format_ of the file, hence the confusion.

Amorymeltzer:
Okay but this is mediawiki, and the extension is always part of the title.  I don't know that I've ever seen anyone make this mistake before, and as above I wouldn't want to suggest there are places where excluding the extension would be appropriate; that'd only increase the chance of this sort of thing happpening.

DannyS712:
Never mind then.

Issue #484, opened 2018-12-23
Adding {{subst:uw-subtle}} and {{subst:uw-fringe}}


iNeedSupport:
Hi there! I propose that the subtle vandalism and fringe template should be added to Wikipedia. The subtle vandalism would be useful if a user made edits that may look like vandalism, but at the same time they might not be vandalism at all. The fringe template would help the warned users not to give an opinion about a certain topic. The NPOV template in twinkle is too vague to show this.

Amorymeltzer:
I think after asking twice ([1](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_40#Subtle_Vandalism_Template), [2](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_40#Proposal_of_{{subst:uw-subtle}}_and_{{subst:uw-fringe}})) it seems there isn't much appetite?  You can always add custom templates, though, that's pretty easy.  Being cognizant of #412, too.

PR #526, opened 2019-02-27
Added uw-subtle series (1-4)

MattLongCT:
Listen, I get that it is not a _super_ popular user warning series, but for those of that _do_ use it we find it crucial. This is a request to added it under the "Other" section


Amorymeltzer:
Thanks for the work, but I think I'm gonna close per my note in #484; seemingly no appetite for it.

Issue #513, opened 2019-02-13
Replace "Talkback" heading by "New message"


ToBeFree:
The "talkback" template is directed especially at new users, yet Twinkle currently adds a meaningless heading with wrong implications above it.

https://en.oxforddictionaries.com/definition/talk_back

https://www.merriam-webster.com/dictionary/back%20talk

Amorymeltzer:
I wonder if defaulting to include your username or the date might be helpful?  I try and do that when I leave a message, but maybe it'd be too radical of a change?

ToBeFree:
@DannyS712 thanks! 🙂 

@Amorymeltzer hmm, it does sound useful. I would use that too. However, I have no idea about the technical complexity involved.

Amorymeltzer:
Username would be insanely easy, date only slightly less so.  I dunno, it's a lot for a default, but how often are these used daily?  It's helpful for archives, but only new users really get a lot of 'em.

ToBeFree:
Ooooh, we can substitute magic words / parser functions.

Amorymeltzer:
>I was wrongly assuming a need for new Twinkle code

Hmm?  You've been able to customize that heading in your preferences to say whatever you want e.g. "New Message from the exalted Jimbo" or something.

Anyway, my inclination is "New message from <insertusername>" would be a good default but I dunno how popular that would be.

ToBeFree:
(I meant that I thought we'd need JavaScript code to insert username/date
into the heading.)

PR #529, opened 2019-03-03
twinkle.js: change default talkback heading to "New message"

DannyS712:
Solving #513. Does not affect user set preferences, just the default.

### DIFF
--- a/twinkle.js
+++ b/twinkle.js
@@ -163,7 +163,7 @@ Twinkle.defaultConfig.friendly = {
 	// Talkback
 	markTalkbackAsMinor: true,
 	insertTalkbackSignature: true,  // always sign talkback templates
-	talkbackHeading: "Talkback",
+	talkbackHeading: "New message",
 	adminNoticeHeading: "Notice",
 	mailHeading: "You've got mail!",
 


Amorymeltzer:
@DannyS712 See discussion in #513

twinkle.js
```diff
@@ -163,7 +163,7 @@ Twinkle.defaultConfig.friendly = {
 	// Talkback
 	markTalkbackAsMinor: true,
 	insertTalkbackSignature: true,  // always sign talkback templates
-	talkbackHeading: "Talkback",
+	talkbackHeading: "New message",
```
Amorymeltzer:
```suggestion
	talkbackHeading: "New message from " + mw.config.get('wgUserName'),
```
See discussion in #513

DannyS712: I don't know how to accept change, but I agree with this

Amorymeltzer: @DannyS712 	Is there not an "commit suggestion" button or something similar here in this window?

DannyS712: Nope. There is "review changes" (which I think is for the overall pull request), in the top right, and when I clicked on the "start review" within this window it just adds a "1" to that button and still won't let me approve anything...

DannyS712: Maybe because I am not a collaborator?

Amorymeltzer: Should be the other way around, if anything — you own this branch.

Amorymeltzer: Hmm, did you do something weird with this branch?  I notice above it says "unknown repository" for the location, which might be causing issues.

DannyS712: I have no idea, sorry. Should I just open a new pull request incorporating your suggested change? Or, should this approved as is, and then another pull request with just your change can be committed? Sorry

DannyS712: Okay. So should this be closed as rejected?

Amorymeltzer:
Closed in #558

twinkle.js
Amorymeltzer: Thanks again for this

Issue #530, opened 2019-03-04
Allow g14 on redirects


DannyS712:
I just came across https://en.wikipedia.org/w/index.php?title=Dexit_(disambiguation)&redirect=no, which is a redirect to https://en.wikipedia.org/wiki/Dexit_(corporation). G14 applies, but the option isn't appearing in the CSD menu, likely because its a redirect. I haven't tagged it for deletion, so that this can be tested, but G14 should work on redirects in mainspace.

Amorymeltzer:
G14 applies to disambiguation pages, and redirects aren't disambiguation pages.

PR #531, opened 2019-03-04
unhide g14 on redirects

DannyS712:
Show g14 on redirects by removing 'hideWhenRedirect: true', solves #530


Amorymeltzer:
G14 applies to disambiguation pages, and redirects aren't disambiguation pages.

PR #539, opened 2019-03-08
warn: Don't autowarn when autowelcoming

Amorymeltzer:
Favor autowelcoming (via the `friendlywelcome` parameter) over autowarning (added in #509); both use vanarticle


Amorymeltzer:
cc @MusikAnimal, although I've already made it live

Issue #237, opened 2014-07-10
Request: put maintenance tags below deletion notices


Breza:
Would it be possible for Twinkle to put new maintenance tags below CSD/PROD/AFD tags?  It would prevent me from having to fix edits like this: https://en.wikipedia.org/w/index.php?title=Bell_Integrator&diff=616390399&oldid=616389887

Amorymeltzer:
This is the tag module's [massive regex](https://github.com/azatoth/twinkle/blob/b8d57ba15a219d1769f7d493509b6dc70730c252/modules/friendlytag.js#L1167).

Amorymeltzer:
#541 should do CSD and PROD, but as I said there, AfD is trickier.  It leads not with a template but with a comment:

`<!-- Please do not remove or change this AfD message until the discussion has been closed. -->`

That poses all sorts of annoyances, since that comment is followed by a template and then two more comments, one of which has template code inside of it.  It's fairly predictable, at least, but annoying.

Amorymeltzer:
So, for AfD, something like the below could work:

```javascript
<!-- Please do not remove or change this AfD.*\n\{\{Article for deletion\/dated.*\n<!-- Once discussion.*\n<!-- End of AfD message.*|
```

That's a fairly unique sequence and could probably be shortened to something like the below without losing any specificity, fidelity, or clarity:
```javascript
<!--.*AfD.*\n\{\{Article for deletion\/dated.*\n<!--.*\n<!--.*AfD.*|
```

That'd make the final line look something like the untested:
```javascript
pageText = pageText.replace(/^\s*(?:((?:\s*<!--.*AfD.*\n\{\{Article for deletion\/dated.*\n<!--.*\n<!--.*AfD.*|\{\{\s*(?:db|delete|db-.*?|speedy deletion-.*?|(?:proposed deletion|prod blp)\/dated|about|correct title|dablink|distinguish|for|other\s?(?:hurricaneuses|people|persons|places|uses(?:of)?)|redirect(?:-acronym)?|see\s?(?:also|wiktionary)|selfref|the)\d*\s*(\|(?:\{\{[^{}]*\}\}|[^{}])*)?\}\})+(?:\s*\n)?)\s*)?/i,
```

This is why /x was invented 😡

siddharthvp:
> This is why /x was invented 😡

If the issue is readability, how about breaking down the regex into something like:
```javascript
pageText = pageText.replace(
	new RegExp(
		// leading spaces
[...]

PR #541, opened 2019-03-09
friendlytag: Put tags under csd and prod nominations (not afd yet)

Amorymeltzer:
Uses regex from the speedy module, partial fix for #237.  AfD is NOT supported because it doesn't lead with a template, but rather with a <!-- comment.

This regex was already enormous and unwieldy, and is now even worse.  It'd be good to clean it up, but even modern js isn't fully pcre, so there's no /x to help us out.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1161,10 +1161,11 @@ Twinkle.tag.callbacks = {
 				pageText += '\n{{Redirect category shell|' + tagText + oldPageTags + '\n}}';
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
This should do for CSD and PROD templates.  AfD is tricky, since it doesn't use a template to start off, but rather a comment:

`<!-- Please do not remove or change this AfD message until the discussion has been closed. -->`

It's certainly doable — I'll try and do the more in-depth reading of the regex later — but for now this should suffice.

PR #546, opened 2019-03-10
speedy: Add additional named parameters

Amorymeltzer:
- Support additional named parameters in A5 (location) and T2 (rationale)
- Add additional entry for F9 (rationale)
- F9: Require either the rationale or the url to be provided in order to carry through
- Remove unused date parameter for F8
- Be explicit with remaining named parameters in preparation for #441, specifically https://github.com/azatoth/twinkle/pull/441#issuecomment-439909083

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -591,7 +591,7 @@ Twinkle.speedy.fileList = [
 		value: 'badfairuse',  // same as below
 		tooltip: 'This is only for files with a clearly invalid fair-use tag, such as a {{Non-free logo}} tag on a photograph of a mascot. For cases that require a waiting period (replaceable images or otherwise disputed rationales), use the options on Twinkle\'s DI tab.',
 		subgroup: {
-			name: 'badfairuse_reason',
+			name: 'badfairuse_rationale',
 			type: 'input',
 			label: 'Optional explanation: ',
 			size: 60
@@ -602,7 +602,7 @@ Twinkle.speedy.fileList = [
 		value: 'badfairuse',  // same as above
 		tooltip: 'Non-free images or media from a commercial source (e.g., Associated Press, Getty), where the file itself is not the subject of sourced commentary, are considered an invalid claim of fair use and fail the strict requirements of WP:NFCC.',
 		subgroup: {
-			name: 'badfairuse_reason',
+			name: 'badfairuse_rationale',
 			type: 'input',
 			label: 'Optional explanation: ',
 			size: 60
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Some of these parameters already existed; I've recently added the others, so this is a draft for now to ensure that those changes settle just fine.  I'll test in the meantime.

modules/twinklespeedy.js
```diff
@@ -626,12 +626,20 @@ Twinkle.speedy.fileList = [
...
-		}
+		subgroup: [
+			{
+				name: 'imgcopyvio_url',
+				type: 'input',
+				label: 'URL of the copyvio, including the "http://".  If the copyvio is of a non-internet source and you cannot provide a URL, you must use the deletion rationale box. ',
+				size: 60
+			},
+			{
+				name: 'imgcopyvio_reason',
```
Amorymeltzer: One thing I'd like some feedback on is whether this is a good idea, or whether it'd be better served as a popup.  Having two boxes takes up a lot of visual space, and the url message is necessarily already very long and awkward looking.  In particular I worry about the "tag multiple" option, where this will especially stand out.

Three options could be:
1. Previous status quo: Don't include the `reason` parameter for F9, keep `url` optional but heavily encouraged
2. This commit: Have both `url` and `reason` inputs, require at least one of them
3. Alt: Have `url` input box, provide a `prompt` for the `reason` parameter if they leave `url` blank.

Amorymeltzer: I've written this as option 2, but I'm leaning toward doing option 3, for overall clarity and to encourage people not to use the reason/rationale input willy-nilly

JJMC89: 2 or 3 should be OK. I'm leaning toward 3. It should still require `rationale` in order to tag if `url` wasn't specified.

AntiCompositeNumber: 2>3>1. In cases where I'm contradicting a statement by the uploader other than {{own work}}, I'd like to leave a reason along with the URL. I'm not so concerned about the two-boxes-being-bigger-than-one-box issue as they're only shown when someone has already clicked the radio button.

Amorymeltzer: @AntiCompositeNumber 	It's more that the text for url wraps likely around, so there's two lines plus a big box for that, then another line for rationale+text.  It's awkward but certainly manageable.  I think the chief advantage of option 3 is further encouragement *not* to use `rationale`, but then again we have that long-ass sentence so it should be pretty clear.  Besides, the awkward placement would surely draw attention!

Maybe the thing to do is go with 3 and if people complain can do 2.

Amorymeltzer: Fixed my above comment fwiw, I meant to suggest sticking with option 2 (this current PR) and going to 3 if folks complain.  I do suppose being able input *both* `url` and `rationale` could be productive.

PR #551, opened 2019-03-16
protect: Don't fail if flaggedrevs/pending changes aren't enabled

Amorymeltzer:
Mainly to help make testing at test.wiki a little smoother, but theoretically this should also help (down the road) support for other wikis


Amorymeltzer:
@MusikAnimal 	I got the idea for this after looking at #297.  I've opened this as a draft so it can be looked at and tested, but mainly because I wanted to avoid complicating #297; after that is merged, I can incorporate those changes and add to this.

Issue #554, opened 2019-03-19
CSD: Disable inapplicable redirect criteria


DannyS712:
Currently, R2 and R4 are options on (as far as I can tell) all redirect pages, despite only applying to redirects in mainspace and in the file namespace respectively

Amorymeltzer:
No need to open an issue if you're gonna open a PR right away.

PR #555, opened 2019-03-19
csd: Only show R2 and R4 criterion when applicable

DannyS712:
Solves #554 by separating speedy deletion criterion R2 and R4 into separate fileRedirect and mainspaceRedirect, and then adding those only when the page is a redirect in the file namespace or in mainspace

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -397,16 +397,18 @@ Twinkle.speedy.callback.modeChanged = function twinklespeedyCallbackModeChanged(
 			work_area.append( { type: 'header', label: 'User pages' } );
 			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.userAllList, mode) } );
 		}
-		else if (namespace == 0) {
+		work_area.append( { type: 'header', label: 'Redirects' } );
+		if (namespace == 0) {
 			work_area.append( { type: 'header', label: 'Mainspace redirects' } );
-			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.mainspaceRedirect, mode) } );
+			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.articleRedirect.concat(Twinkle.speedy.redirectList), mode) } );
 		}
 		else if (namespace == 6) {
 			work_area.append( { type: 'header', label: 'Mainspace redirects' } );
-			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.fileRedirect, mode) } );
+			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.redirectList.concat(Twinkle.speedy.fileRedirect), mode) } );
 		}
-		work_area.append( { type: 'header', label: 'Redirects' } );
-		work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.redirectList, mode) } );
+		else {
+			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.redirectList, mode) } );
+		}
 	}
 
 	var generalCriteria = Twinkle.speedy.generalList;
@@ -1077,7 +1079,7 @@ Twinkle.speedy.redirectList = [
 	}
 ];
 	
-Twinkle.speedy.mainspaceRedirect = [
+Twinkle.speedy.articleRedirect = [
 	{
 		label: 'R2: Redirects from mainspace to any other namespace except the Category:, Template:, Wikipedia:, Help: and Portal: namespaces',
 		value: 'rediruser',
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Good call!  I think it'd be better to put these in the same "Redirects" section rather than their own section; there's enough clutter that multiple sections with very few items probably isn't helpful.

modules/twinklespeedy.js
```diff
@@ -397,6 +397,14 @@ Twinkle.speedy.callback.modeChanged = function twinklespeedyCallbackModeChanged(
 			work_area.append( { type: 'header', label: 'User pages' } );
 			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.userAllList, mode) } );
 		}
+		else if (namespace == 0) {
+			work_area.append( { type: 'header', label: 'Mainspace redirects' } );
+			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.mainspaceRedirect, mode) } );
+		}
+		else if (namespace == 6) {
+			work_area.append( { type: 'header', label: 'Mainspace redirects' } );
```
Amorymeltzer: Mooted by my comment on the PR, but this is mislabeled

DannyS712: > Mooted by my comment on the PR, but this is mislabeled

Oh, it should be file redirects. How should I ensure that there isn't a heading? Remove lines 401 and 405?

Amorymeltzer: You'd need to refactor it to use `concat`, like we do with `userNonRedirectList` (L362) or `generalNonUser` (L415).  Basically, take `redirectList`, then:
- if it's mainspace throw `articleRedirectList` in front
- if it's filespace throw `fileRedirectList` in back
- Append that to the `Redirects` section

DannyS712: Done in 2 more commits, now uses concat.

Amorymeltzer:
@DannyS712 	This all works fine, but the filespace redirect menu ain't ideal: it shows `R3, G6, G8, R4`.  One option is to go back to the old style, but disable the appropriate option in the appropriate namespace; we don't do that elsewhere and it'll likely just look bad/broken.  The alternative is to split this *yet again*; that'd mean a group with just R2, a group with just R3, a group with just R4, and a group with the G6/G8 redirect options.  Hideous.  I'll brainstorm better ways, but that might be the path forward.

DannyS712:
> @DannyS712 This all works fine, but the filespace redirect menu ain't ideal: it shows `R3, G6, G8, R4`. One option is to go back to the old style, but disable the appropriate option in the appropriate namespace; we don't do that elsewhere and it'll likely just look bad/broken. The alternative is to split this _yet again_; that'd mean a group with just R2, a group with just R3, a group with just R4, and a group with the G6/G8 redirect options. Hideous. I'll brainstorm better ways, but that might be the path forward.

@Amorymeltzer solved in a less hideous way I think. See my latest commit.

Amorymeltzer:
@DannyS712 	For future maintainability, I think awkward `concat`/`if`-`else` chains are better than duplication of hardcoded data.  It might look complicated, but we should avoid duplicating information that may need to be updated in the future, especially since we don't elsewhere here.

DannyS712:
@Amorymeltzer sorry, I didn't see that you had assigned this to yourself. I split it up per your last comment, with a fuller explanation in the commit message.

Amorymeltzer:
@DannyS712 	No apologies needed!  Since this is a PR, I meant it as a "I'll review and be responsible for merging this."  Mainly, though, I just did it to be on my immediate to-do list for the day.

modules/twinklespeedy.js
```diff
@@ -1078,6 +1071,30 @@ Twinkle.speedy.redirectList = [
 		hideWhenMultiple: true
 	}
 ];
+Twinkle.speedy.redirectCriterion = [
```
Amorymeltzer:
```suggestion

Twinkle.speedy.redirectCriterion = [
```

modules/twinklespeedy.js
```diff
@@ -1078,6 +1071,30 @@ Twinkle.speedy.redirectList = [
...
+	{
+		label: 'R3: Redirects as a result of an implausible typo or misnomers that were recently created',
+		value: 'redirtypo',
+		tooltip: 'However, redirects from common misspellings or misnomers are generally useful, as are redirects in other languages'
+	}
+];
+	
+
```
Amorymeltzer:
```suggestion
```

modules/twinklespeedy.js
```diff
@@ -1078,6 +1071,30 @@ Twinkle.speedy.redirectList = [
...
+	
+	
+Twinkle.speedy.articleRedirect = [
+	{
+		label: 'R2: Redirects from mainspace to any other namespace except the Category:, Template:, Wikipedia:, Help: and Portal: namespaces',
+		value: 'rediruser',
+		tooltip: '(this does not include the Wikipedia shortcut pseudo-namespaces). If this was the result of a page move, consider waiting a day or two before deleting the redirect'
+	}
+];
+
```
Amorymeltzer:
```suggestion

```

modules/twinklespeedy.js
```diff
@@ -1078,6 +1071,30 @@ Twinkle.speedy.redirectList = [
...
 	}
 ];
+Twinkle.speedy.redirectCriterion = [
+	{
+		label: 'R3: Redirects as a result of an implausible typo or misnomers that were recently created',
+		value: 'redirtypo',
+		tooltip: 'However, redirects from common misspellings or misnomers are generally useful, as are redirects in other languages'
+	}
+];
+
```
Amorymeltzer:
```suggestion

```

DannyS712:
> Whitespace cleanup

@Amorymeltzer I've added all of the suggested edits

Amorymeltzer:
Thanks for all the back and forth on this!

PR #561, opened 2019-03-22
fluff: Remove duplicated code and avoid using Morebits.queryString

Amorymeltzer:
1st commit: Remove some duplicate code
- The same four lines, three of which used spanTag, were repeated frequently
- Newer revisions either are the latest or not
- No idea what .firstrevisionheader did but I can't find it anywhere on-wiki, so removed

2nd commit: Avoid using Morebits.queryString to parse the url for 'oldid'
The url on diffs and old revisions is fickle, and may not necessarily display exactly what you think or expect based solely on the value of oldid.

- wgRevisionId added in MW 1.22 circa (2013)
- wgDiffOldId and wgDiffNewId added in MW 1.30 (2017)

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -33,7 +33,7 @@ var buildLink = function(color, text) {
 
 Twinkle.fluff = {
 	auto: function() {
-		if( parseInt( Morebits.queryString.get('oldid'), 10) !== mw.config.get('wgCurRevisionId') ) {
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
The first commit here supersedes the first commit in #511, compare with https://github.com/azatoth/twinkle/pull/511/commits/33ae439f9be7c5e4c952d2509d2a22b048ad5715.

PR #562, opened 2019-03-22
fluff: Add [restore this revision] link to oldid pages

### LINKED ISSUE #131: Add [restore this version] link when viewing old revisions

Amorymeltzer:
Closes #131

Checking for wgDiffOldId and wgDiffNewId excludes Special:Undelete (where they
are each 0), and so should be sufficient to avoid problematic diffs.

One advantage of the restructuring is that [restore this revision] will now
appear when viewing the first revision as a diff with diff=prev and will not
appear on the left-hand revision when viewing the latest revision with
diff=cur, as may sometimes happen.  Similarly, revert or restore links should
no longer appear on single-revision pages.  This makes some of the
logic/checks a tad involved, as well as the fact that wgDiffNewId and
wgDiffOldId have unbalanced behavior (see [[phab:T214985]]).

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -80,101 +80,97 @@ Twinkle.fluff = {
 	},
 
 	diff: function() {
-		if( mw.config.get('wgCanonicalSpecialPageName') === "Undelete" ) {
-			//You can't rollback deleted pages!
-			return;
-		}
+		if (mw.config.get('wgDiffNewId') === mw.config.get('wgRevisionId')) {
+			// Add a [restore this revision] link to the older revision
+			// Don't show if there's a single revision or weird diff (cur on latest)
+			if (mw.config.get('wgDiffOldId') && (mw.config.get('wgDiffOldId') !== mw.config.get('wgDiffNewId'))) {
+				var revertToRevision = document.createElement('div');
+				revertToRevision.setAttribute( 'id', 'tw-revert-to-orevision' );
+				revertToRevision.style.fontWeight = 'bold';
 
-		var otitle, ntitle;
-		try {
-			var otitle1 = document.getElementById('mw-diff-otitle1');
-			var ntitle1 = document.getElementById('mw-diff-ntitle1');
-			if (!otitle1 || !ntitle1) {
-				return;
+				var revertToRevisionLink = buildLink('SaddleBrown', 'restore this version');
+				revertToRevisionLink.href = "#";
+				$(revertToRevisionLink).click(function(){
+					Twinkle.fluff.revertToRevision(mw.config.get('wgDiffOldId').toString());
+				});
+				revertToRevision.appendChild(revertToRevisionLink);
+
+				var otitle = document.getElementById('mw-diff-otitle1').parentNode;
+				otitle.insertBefore( revertToRevision, otitle.firstChild );
 			}
-			otitle = otitle1.parentNode;
-			ntitle = ntitle1.parentNode;
-		} catch( e ) {
-			// no old, nor new title, nothing to do really, return;
-			return;
-		}
 
-		// Lets first add a [restore this revision] link
+			// Add either restore or rollback links to the newer revision
+			// Don't show if there's a single revision or weird diff (prev on first)
+			var ntitle = document.getElementById('mw-diff-ntitle1').parentNode;
+			if( document.getElementById('differences-nextlink') ) {
+				// Not latest revision
+				var revertToRevisionN = document.createElement('div');
+				revertToRevisionN.setAttribute( 'id', 'tw-revert-to-nrevision' );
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Supersedes the second commit in #511, namely https://github.com/azatoth/twinkle/pull/511/commits/855097570d9a41cdc6882c0c479ef5094e69eca0

modules/twinklefluff.js
```diff
@@ -80,101 +80,97 @@ Twinkle.fluff = {
...
 		var revertToRevision = document.createElement('div');
 		revertToRevision.setAttribute( 'id', 'tw-revert-to-orevision' );
 		revertToRevision.style.fontWeight = 'bold';
 
 		var revertToRevisionLink = buildLink('SaddleBrown', 'restore this version');
-
 		revertToRevisionLink.href = "#";
 		$(revertToRevisionLink).click(function(){
-			Twinkle.fluff.revertToRevision(mw.config.get('wgDiffOldId').toString());
+			Twinkle.fluff.revertToRevision(mw.config.get('wgRevisionId').toString());
```
Amorymeltzer: Fair warning, because this tripped me up for a while, but the regular split diff view makes it look like `wgDiffOldId` is being changed to `wgRevisionId`, but it's not!  `git diff` does poorly with duplicated sections, so the stuff on the left (diff) is actually above this section (oldid).  Viewing without whitespace should make it clearer.

modules/twinklefluff.js
```diff
@@ -586,8 +582,14 @@ Twinkle.fluff.init = function twinklefluffinit() {
 
 		if ( Morebits.queryString.exists( 'twinklerevert' ) ) {
 			Twinkle.fluff.auto();
-		} else {
-			Twinkle.fluff.normal();
+		} else if( mw.config.get('wgNamespaceNumber') === -1 && mw.config.get('wgCanonicalSpecialPageName') === "Contributions" ) {
+			Twinkle.fluff.contributions();
+		} else if( mw.config.get('wgDiffOldId') || mw.config.get('wgDiffNewId') ) { // wgDiffOldId included for clarity in if else loop [[phab:T214985]]
```
Amorymeltzer: That is, `wgDiffNewId` alone is sufficient to determine if we're on a diff page, but I've included `wgDiffOldId` here to the following `else if` a little simpler.  Plus, this should be a little clearer to maintain, especially if that ever changes.

modules/twinklefluff.js
```diff
@@ -80,101 +80,97 @@ Twinkle.fluff = {
...
+
+				var revertToRevisionNLink = buildLink('SaddleBrown', 'restore this version');
+				revertToRevisionNLink.href = "#";
+				$(revertToRevisionNLink).click(function(){
+					Twinkle.fluff.revertToRevision(mw.config.get('wgDiffNewId').toString());
+				});
+				revertToRevisionN.appendChild(revertToRevisionNLink);
+
+				ntitle.insertBefore( revertToRevisionN, ntitle.firstChild );
+			} else if( Twinkle.getPref('showRollbackLinks').indexOf('diff') !== -1 && mw.config.get('wgDiffOldId') && (mw.config.get('wgDiffOldId') !== mw.config.get('wgDiffNewId') || document.getElementById('differences-prevlink'))) {
```
Amorymeltzer: This `else if` is somewhat unwieldy.  I can make it shorter, but I think the code is likely easier to maintain if we handle the left-hand restore first then tackle the right-hand separately, adding either restore or rollback links.

I'd like to try and deduplicate even more of this down the road.

Amorymeltzer:
@MusikAnimal 	Refactored and redesigned, but most of the code and the overall major thrust is the same.  Since #407 required doing away with some of the real estate savings, I could take care of your major issue from https://github.com/azatoth/twinkle/pull/511/commits/855097570d9a41cdc6882c0c479ef5094e69eca0.  The redesign also let me take care of a few edge cases where diffs behave weirdly.

Amorymeltzer:
Any lingering concerns @MusikAnimal?

Amorymeltzer:
@MusikAnimal I've gone and merged this.  Your major concern from https://github.com/azatoth/twinkle/pull/511#issuecomment-470686797 was dealt with, but the issue of its size remains.  It would be trivial to add a `fontSize = '120%'`, to match what we show on diffs (120% of 84% is roughly 100%) rather than the surrounding content, or perhaps the more modest `110`.  Worth it?  After some testing I think they're all fine, so maybe the 110?

Issue #505, opened 2019-01-22
Uncaught TypeError: originalEvent is not a function


JJMC89:
I get the below error when (un)checking CSD criteria when multiple is cheched.

```
Uncaught TypeError: originalEvent is not a function at HTMLInputElement.criterion.event (load.php?debug=false&lang=en&modules=ext.gadget.Twinkle&skin=vector&version=0f2ee8y:formatted:5582)
criterion.event @ load.php?debug=false&lang=en&modules=ext.gadget.Twinkle&skin=vector&version=0f2ee8y:formatted:5582
```
It refers to
https://github.com/azatoth/twinkle/blob/b91e9dfdb72e15ed93358e6c2b03ad565e8d66ae/modules/twinklespeedy.js#L501

Amorymeltzer:
I've got an implementation for #520 that takes care of this.

Amorymeltzer:
#563

Amorymeltzer:
Done in #563

PR #563, opened 2019-03-24
speedy: New option for sysops: notify creator upon deletion

### LINKED ISSUE #520: Speedy: Allow sysops to issue warning upon csd
Requested on IRC.

It's not really standard practice, mainly since most pages are already tagged with users warned, and we offer the option to open the user's page after deletion.  Still, I suppose it could be helpful?  There are already a lot of options there.

Amorymeltzer:
Closes #520 and #505.  Removes the old "open talk page" sysop option and folds the non-sysop tagging system into a unified `noteToCreator` notification system for both tagfing and deleting.  Uses the existing system for tagging to determine whether or not to notify, rather than the bespoke option for `openUserTalk` from #300; that is, it will notify only if the corresponding box is checked AND (at least one of) the CSD in question is checked in the user preferences.

The box is checked by default, unless it finds `$('#delete-reason')` on the page, indicating some sort of XfD or CSD or the like; in the latter scenario, if the box is subsequently checked, there is an additional prompt (this might be a good candidate for removal down the line, but I think it's important to include now, especially since it's a new feature).  It uses the existing preference about whether to welcome the user alongside the notification rather than create a separate one just for post-deletion notifications.

New templates all listed at [[Template:Speedy deletion deleted]].

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -397,10 +397,10 @@ Twinkle.config.sections = [
 		},
 
 		// TwinkleConfig.watchSpeedyPages (array)
-		// Whether to add speedy tagged pages to watchlist
+		// Whether to add speedy tagged or deleted pages to watchlist
 		{
 			name: "watchSpeedyPages",
-			label: "Add page to watchlist when tagging with these criteria",
+			label: "Add page to watchlist when using these criteria",
 			type: "set",
 			setValues: Twinkle.config.commonSets.csdCriteria,
 			setDisplayOrder: Twinkle.config.commonSets.csdCriteriaDisplayOrder
@@ -414,24 +414,36 @@ Twinkle.config.sections = [
 			type: "boolean"
 		},
 
+		// TwinkleConfig.welcomeUserOnSpeedyDeletionNotification (array of strings)
+		// On what types of speedy deletion notifications shall the user be welcomed
+		// with a "firstarticle" notice if their talk page has not yet been created.
+		{
+			name: "welcomeUserOnSpeedyDeletionNotification",
+			label: "Welcome page creator when notifying with these criteria",
+			helptip: "The welcome is issued only if the user is notified about the deletion, and only if their talk page does not already exist. The template used is {{firstarticle}}.",
+			type: "set",
+			setValues: Twinkle.config.commonSets.csdCriteriaNotification,
+			setDisplayOrder: Twinkle.config.commonSets.csdCriteriaNotificationDisplayOrder
+		},
+
 		// TwinkleConfig.notifyUserOnSpeedyDeletionNomination (array)
-		// What types of actions should result that the author of the page being notified of nomination
+		// What types of actions should result in the author of the page being notified of nomination
 		{
 			name: "notifyUserOnSpeedyDeletionNomination",
-			label: "Notify page creator only when tagging with these criteria",
+			label: "Notify page creator when tagging with these criteria",
 			helptip: "Even if you choose to notify from the CSD screen, the notification will only take place for those criteria selected here.",
 			type: "set",
 			setValues: Twinkle.config.commonSets.csdCriteriaNotification,
 			setDisplayOrder: Twinkle.config.commonSets.csdCriteriaNotificationDisplayOrder
 		},
 
-		// TwinkleConfig.welcomeUserOnSpeedyDeletionNotification (array of strings)
-		// On what types of speedy deletion notifications shall the user be welcomed
-		// with a "firstarticle" notice if his talk page has not yet been created.
+		// TwinkleConfig.warnUserOnSpeedyDelete (array)
+		// What types of actions should result in the author of the page being notified of speedy deletion (admin only)
 		{
-			name: "welcomeUserOnSpeedyDeletionNotification",
-			label: "Welcome page creator alongside notification when tagging with these criteria",
-			helptip: "The welcome is issued only if the user is notified about the deletion, and only if their talk page does not already exist. The template used is {{firstarticle}}.",
+			name: "warnUserOnSpeedyDelete",
+			label: "Notify page creator when deleting under these criteria",
+			helptip: "Even if you choose to notify from the CSD screen, the notification will only take place for those criteria selected here.",
+			adminOnly: true,
 			type: "set",
 			setValues: Twinkle.config.commonSets.csdCriteriaNotification,
 			setDisplayOrder: Twinkle.config.commonSets.csdCriteriaNotificationDisplayOrder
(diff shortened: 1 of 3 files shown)


Amorymeltzer:
@MusikAnimal 	There's some more tidying I can do — one or two of the variables are unnecessary, but helpful for clarity when reviewing — but here's the draft code for #520.  I think I explain the changes in the commit message well enough to reply to the points there, but to be explicit:

- It takes the system for notifying users of a tag and uses that for both notifying (tag) or warning (delete).  The box must be checked and the corresponding csd must be checked in your preferences.
- Does away with the `openUserTalk` option after deleting a page, and likewise the routine you wrote to process the checkbox.
- Uses warnings of the same naming style as the notifications (e.g. `db-reason-warn` to match `db-reason-notify`)

I think using the same structure of templates as we do for notifications but keeping them separate should make things simpler and easier for others to use, but that can be changed I suppose.  Regardless, whatever we decide, I'll go and make all the templates.  A few exist under a different naming scheme (see [Category:CSD_warning_templates](https://en.wikipedia.org/wiki/Category:CSD_warning_templates)) but I think it'd be fine to move those to a more unified name.

More to the point, the larger issue is default behavior.  As of now, the checkbox defaults to being checked if there isn't `$('#delete-reason')` on the page, and I sort of half-assed looked over the default items to be checked in your preferences.  I think what should be done, since this will be a *big* change in behavior, is to ask two questions at AN:
- Should the box be checked or unchecked by default?
- What should the default set of criteria to notify include?
We can discuss here and crib off the other defaults, but I think that will need broader input.  Can also ask about moving the above templates if need be.

Amorymeltzer:
@MusikAnimal I'll open a discussion on WP:AN in the next day or two

Amorymeltzer:
Minimal input: https://en.wikipedia.org/wiki/Wikipedia:Administrators%27_noticeboard/Archive309#Should_Twinkle_warn_after_CSD_deletion?_And_default_behavior

Amorymeltzer:
@MusikAnimal I'm gonna start creating the templates in the next few days, I've decided to go with `{{db-g12-deleted}}` (instead of `{{db-g12-warn}}`) to match `{{db-g12-notice}}`, how does that sound?

MusikAnimal:
> @MusikAnimal I'm gonna start creating the templates in the next few days, I've decided to go with `{{db-g12-deleted}}` (instead of `{{db-g12-warn}}`) to match `{{db-g12-notice}}`, how does that sound?

That sounds fine to me! If people don't like it they can redirect the templates and we'll change Twinkle at leisure. Thanks for your work on this!

Amorymeltzer:
Workflow at [User:Amorymeltzer/sandbox/tw](https://en.wikipedia.org/wiki/User:Amorymeltzer/sandbox/tw) fwiw

Amorymeltzer:
@MusikAnimal FYI I pushed these changes live to testwiki in case you want to take a spin.  I haven't looked at or touched the code much in weeks, and will hopefully do some more thorough tests in any spare time the next few days, but it seems to work just dandy.

MusikAnimal:
> @MusikAnimal FYI I pushed these changes live to testwiki in case you want to take a spin. I haven't looked at or touched the code much in weeks, and will hopefully do some more thorough tests in any spare time the next few days, but it seems to work just dandy.

Looking good!

A few possible issues:
* When I chose "Tag page only" with "Notify page creator" checked, it did not notify the user. I used G11, and I have my Twinkle preferences set to notify for this criteria.
* With [this notification](https://test.wikipedia.org/w/index.php?title=User_talk:MusikPuppet&diff=395297&oldid=365380&diffmode=source) it did not create a new heading for May 2019.
* Minor nitpick... wouldn't the word "notify" be more fitting in this context than "warn"? E.g. "Notify page creator of page deletion" instead of "Warn page creator of page deletion". The templates (at least the ones I checked) do actually warn the user about recreating the page, but the main idea here I think is to simply let the user know you deleted the page and why.

Amorymeltzer:
>When I chose "Tag page only" with "Notify page creator" checked, it did not notify the user.

@MusikAnimal Heh, I think you ran into what I did for a while with some xfd stuff: User:MusikAnimal created the page User:MusikPuppet, so it skipped notifying yourself.  Trying it on [User:MusikPuppet/test](https://test.wikipedia.org/wiki/User:MusikPuppet/test) should work, as it [just](https://test.wikipedia.org/w/index.php?title=User:MusikPuppet/test&diff=395310&oldid=362532) did for [me](https://test.wikipedia.org/w/index.php?title=User_talk:MusikPuppet&diff=395311&oldid=395297) (see also below).

>With this notification it did not create a new heading for May 2019.

That's because the new templates don't exist on testwiki.  If you copy-paste the wikitext it added to the talk page (`{{subst:db-csd-deleted-custom|1=User:MusikPuppet/test|2=notwebhost}}`) to something on en.wiki, it should be fine. 
 I didn't go through the process of importing them some or all of them, but I suppose I should.

>Minor nitpick...

100%, thanks.  I've got a few other language tweaks to do, this is one of them.  I made a distinction in the code between being "notified" and "warned" so it was clear which was which, and as the seven existing post-deletion templates were all named `criterion-warn-deletion`, e.g. [Template:Nn-warn-deletion](https://en.wikipedia.org/wiki/Template:Nn-warn-deletion).  I dropped that language when making the templates for exactly that reason, favoring `{{db-spam-deleted}}` over `{{db-spam-warn}}` or `{{db-spam-warning}}`, but need to go back here and adjust the front-facing language.

Amorymeltzer:
@MusikAnimal 	Tweaked language (including commit message), look good? https://github.com/azatoth/twinkle/compare/f0d2f6986bd972d4fd29ffac93cbfd98039704c0..e82e7346cb7f0ff1b00b2b36e1912cd9d10e6d86

Also, I changed the default list to just match what we do for tagging notifications (https://github.com/azatoth/twinkle/compare/e82e7346cb7f0ff1b00b2b36e1912cd9d10e6d86..13ebf0e1784f235544c044d61b66efe8c051f115); that's adding A2, A5, F1, F2, F10, T3, and P2 from the `openUserTalk` system.  Seemed simpler, and more likely to be expected behavior.

MusikAnimal:
I figured my testing was flawed. Those bugs I found were too obvious for you to have missed :)

:+1: and a :beer: from me. I admit I have not done a thorough code review so I'll leave merging to you or whoever else. I'm sure admins are going to really enjoy this feature!

Amorymeltzer:
I think what I'll do is push the `twinkle.js` and `twinkleconfig.js` changes in the next day or two and leave a note in the Admin Newsletter that the feature is coming soon (I'm shooting for the 12th for next on-wiki update) but they can update preferences now.

PR #566, opened 2019-03-27
prod: Check for old PROD templates on talkpage or an existing AfD, abort if found

### LINKED ISSUE #266: Warn before allowing PROD if the talk page is a member of Category:Past proposed deletion candidates
With the exception of BLP PROD, if an article has been nominated for proposed deletion before, and the PROD has been contested by someone else (either by removing the PROD tag or by requesting undeletion of a deleted PRODed article at [[Wikipedia:Requests for undeletion]]), the article cannot be proposed for deletion again and must instead be taken to Articles for Deletion. In addition, anyone who contests a PROD is supposed to add the template [[Template:Old prod full]] to the article's talk page, which will make it a member of [[:Category:Past proposed deletion candidates]].

I propose that Twinkle should check to see if the article's talk page is a member of this category when a user chooses to propose an article for deletion via PROD, and if it is, it should warn the user with a message that says that the article has already been nominated for PROD in the past and suggest that the user use AfD instead.

Note that the above only applies to regular PROD, not to sticky BLP PROD (which can be done at any time if a BLP becomes unsourced, and must remain as long as it is unsourced).

Amorymeltzer:
Closes #266.  Can't check for prior FfDs, since they don't leave a subpage or a log/category somewhere

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -172,6 +172,58 @@ Twinkle.prod.callback.prodtypechanged = function(event) {
 };
 
 Twinkle.prod.callbacks = {
+	checkpriors: function(apiobj) {
+		var xmlDoc = apiobj.responseXML;
+		var statelem = apiobj.statelem;
+		var params = apiobj.params;
+
+		// Check existence of AfD, would be easier if mw.Title.exists() worked [[phab:T184953]]
+		var exists;
+		if (namespace === 'article') {
+			var query = {
+				'action': 'query',
+				'prop': 'revisions',
+				'titles': 'Wikipedia:Articles for deletion/' + Morebits.pageNameNorm
+			};
+			var wikipedia_api = new Morebits.wiki.api('Checking for prior AfDs', query, function(apiobj) {
+				var xmlDoc = apiobj.responseXML;
+				exists = $(xmlDoc).find('rev').attr('revid');
+			});
+			// Wait for API call to finish
+			wikipedia_api.post({
+				async: false
+			});
+		}
+		if (exists) {
+			statelem.warn( 'Previous AfD for this page title found, aborting procedure' );
+			return;
+		}
+
+		// Check talk page for category indicating prior PROD
+		var cats = $(xmlDoc).find('categories');
+		if(cats.length) {
+			if(params.blp) {
+				if( !confirm( 'Previous PROD nomination found on talk page. Do you still want to continue applying BLPPROD? ' ) ) {
+					statelem.warn( 'Previous PROD found on talk page, aborted by user' );
+					return;
+				} else {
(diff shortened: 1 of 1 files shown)


modules/twinkleprod.js
```diff
@@ -172,6 +172,58 @@ Twinkle.prod.callback.prodtypechanged = function(event) {
...
+				'action': 'query',
+				'prop': 'revisions',
+				'titles': 'Wikipedia:Articles for deletion/' + Morebits.pageNameNorm
+			};
+			var wikipedia_api = new Morebits.wiki.api('Checking for prior AfDs', query, function(apiobj) {
+				var xmlDoc = apiobj.responseXML;
+				exists = $(xmlDoc).find('rev').attr('revid');
+			});
+			// Wait for API call to finish
+			wikipedia_api.post({
```
Amorymeltzer: `async` used so we can avoid creating separate functions for mainspace and as well as not showing "Checking for AfDs" for filespace PRODs

modules/twinkleprod.js
```diff
@@ -372,13 +424,17 @@ Twinkle.prod.callback.evaluate = function twinkleprodCallbackEvaluate(e) {
 	Morebits.simpleWindow.setButtonsEnabled( false );
 	Morebits.status.init( form );
 
-	Morebits.wiki.actionCompleted.redirect = mw.config.get('wgPageName');
-	Morebits.wiki.actionCompleted.notice = "Tagging complete";
+	var talk_title = new mw.Title(mw.config.get('wgPageName')).getTalkPage().getPrefixedText();
```
Amorymeltzer: `mw.Title` isn't particularly Twinkle-y (although it is used elsewhere), but imo it's a lot cleaner for specifically this sort of thing: getting the talk page name.  `Morebits.wikipedia.namespaces` isn't too unwieldy — nor is `wgNamespaceIds` — but this is just so much simpler looking.

modules/twinkleprod.js
```diff
@@ -172,6 +172,58 @@ Twinkle.prod.callback.prodtypechanged = function(event) {
...
+		if (exists) {
+			statelem.warn( 'Previous AfD for this page title found, aborting procedure' );
+			return;
+		}
+
+		// Check talk page for category indicating prior PROD
+		var cats = $(xmlDoc).find('categories');
+		if(cats.length) {
+			if(params.blp) {
+				if( !confirm( 'Do you still want to continue applying BLPPROD? ' ) ) {
```
siddharthvp: Should be `!confirm( 'Previous PROD nomination found on talk page. Do you still want to continue applying BLPPROD? ' )`. Right?

Amorymeltzer: Definite improvement!

PR #568, opened 2019-03-29
unlink: remove unnecessary api call, and hardcoded api limits

siddharthvp:
- remove unnecessary API call to check for extendconfirmed right
- avoid hardcoding maximum api limits

### DIFF
--- a/modules/twinkleunlink.js
+++ b/modules/twinkleunlink.js
@@ -77,8 +77,8 @@ Twinkle.unlink.callback = function(presetReason) {
 			'list': [ 'backlinks', 'imageusage' ],
 			'bltitle': mw.config.get('wgPageName'),
 			'iutitle': mw.config.get('wgPageName'),
-			'bllimit': Morebits.userIsInGroup( 'sysop' ) ? 5000 : 500, // 500 is max for normal users, 5000 for bots and sysops
-			'iulimit': Morebits.userIsInGroup( 'sysop' ) ? 5000 : 500, // 500 is max for normal users, 5000 for bots and sysops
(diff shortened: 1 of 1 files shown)


modules/twinkleunlink.js
```diff
@@ -14,15 +14,11 @@
...
+	if( mw.config.get('wgNamespaceNumber') < 0 || mw.config.get('wgPageName') === 'Wikipedia:Sandbox' ||
+		!Morebits.userIsInGroup('extendedconfirmed')) {
 		return;
 	}
-	// Restrict to extended confirmed users. This is done here instead of above to avoid an unnecessary API call.
-	mw.user.getRights().then(function (rights) {
-		if( rights.indexOf('extendedconfirmed') !== -1 ) {
-			Twinkle.addPortletLink( Twinkle.unlink.callback, "Unlink", "tw-unlink", "Unlink backlinks" );
-		}
-	});
```
JJMC89: Rights need to be checked, not groups. `sysop` has the `extendedconfirmed` right, but `sysop`s are not in the `extendedconfirmed` group.

siddharthvp: Ah, still we can manage without an API call.

Amorymeltzer: Just to note that this will disallow `bot`s, unlike the previous check.  I have a hard time imagining someone using a bot to unlink, much less doing so with Twinkle's unlink module, but that's the other group with the `extendedconfirmed` right.  I don't think it's necessary.

Amorymeltzer:
Just to link the threads on gh, this was added by @MusikAnimal in #428

PR #569, opened 2019-04-01
General minor clean up

DannyS712:
Move "// notify user" comment to the line directly above the relevant code, separate "Morebits.wiki.removeCheckpoint();" from the chunk of code used for notifications (only whitespace was changed)

### DIFF
--- a/modules/friendlywelcome.js
+++ b/modules/friendlywelcome.js
@@ -27,7 +27,7 @@ Twinkle.welcome = function friendlywelcome() {
 
 Twinkle.welcome.auto = function() {
 	if( Morebits.queryString.get( 'action' ) !== 'edit' ) {
-		// userpage not empty, aborting auto-welcome
+		// Userpage not empty, aborting auto-welcome
 		return;
 	}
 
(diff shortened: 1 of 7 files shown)


Amorymeltzer:
A lot of this would be better dealt with by some input at #500

DannyS712:
> A lot of this would be better dealt with by some input at #500

I read through that thread - I'm not familiar with ESlint at all, and it seems out of my depth

Amorymeltzer:
It would (among other things) enforce a lot of the stylistic changes here that we have up to now not chosen to prefer, such as spaces after `if` and the like.

DannyS712:
@Amorymeltzer I think I'm done with this round of clean up, if you are willing to merge it

siddharthvp:
I don't think this PR should be accepted as it would result in completely unnecessary merge conflicts with other PRs. Also, I don't agree with many of the changes:
- Capitalisation of others' comments is plain silly. I wouldn't make such changes unless I'm editing the surrounding code or making some overall structural changes.
- Removal of newlines in some places is unnecessary and makes stuff harder to read.
- Changes to styling of if conditions should be handled along with #500, as Amorymeltzer said.

DannyS712:
Okay, then I'll withdrawn this. Sorry for the trouble

PR #570, opened 2019-04-01
Warn: whitespace cleanup

DannyS712:
Remove unneeded new line to put the inline comment right next to the code it refers to


Amorymeltzer:
I'm not inherently opposed to purely cosmetic changes, but if they're going to be as incredibly minor as this and #569, please put them in one commit.

PR #574, opened 2019-04-06
Speedy: Don't show G13 when not applicable

### LINKED ISSUE #573: Speedy: Don't show G13 when not applicable
G13 only applies to userspace and draftspace pages, so it should only be shown on them.

I might have a chance to code a patch this weekend, but I'm posting this as an issue in case I don't have time/forget.

DannyS712:
Closes #573 (it turns out I did have time) by:
1. Separating the generalList criteria into generalNonDraft, draftCritereon, and generalNonDraft2
2. Starting with the generalNonDraft, if in user or draft namespace append the draftCritereon
3. Regardless of 2, append generalNonDraft2 to the end

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -396,11 +396,16 @@ Twinkle.speedy.callback.modeChanged = function twinklespeedyCallbackModeChanged(
 		}
 	}
 
-	var generalCriteria = Twinkle.speedy.generalList;
+	var generalCriteria = Twinkle.speedy.generalNonDraft;
 	// G1 and G2 aren't supposed to be used on userpages
 	if (namespace !== 2) {
 		generalCriteria = Twinkle.speedy.generalNonUser.concat(generalCriteria);
 	}
+	if (namespace == 2 || namespace == 118) {
+		generalCriteria = generalCriteria.concat(Twinkle.speedy.draftCriterion);
+	}
+	generalCriteria = generalCriteria.concat(Twinkle.speedy.generalNonDraft2);
+
 	// custom rationale lives under general criteria when tagging
 	if(!Twinkle.speedy.mode.isSysop(mode)) {
 		generalCriteria = Twinkle.speedy.customRationale.concat(generalCriteria);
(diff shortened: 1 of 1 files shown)


siddharthvp:
Can't say I'm a fan of all these changes optimizing the namespaces on which certain criteria are shown. The present commit introduces such lists like `generalNonDraft`, 'generalNonUser`, `generalNonDraft2`, making it all very confusing, and more importantly making the codebase averse to internationalization. Think of all the headache other wikis will have to go through to port this code. This, I think, is the wrong way to go.

I think we should come up with a better infrastructure of handling namespaces suitability of criteria. We could probably use properties like `hideInNamespaces` (an array of namespace numbers) and `hideOnRedirect` (boolean) for each criteria, and get them to work by revamping the `Twinkle.speedy.callback.modeChanged` function. Will look into this when I have time.

DannyS712:
@siddharthvp Yes, I agree that that would be a lot easier. I'll note that this only introduces 2 new lists, generalNonUser was introduced in #514 by @Amorymeltzer. I can help look into developing the namespace framework you describe, but until then are you okay with this commit?

Amorymeltzer:
>come up with a better infrastructure of handling namespaces suitability of criteria

Agree 100% @siddharthvp.  I thought about this with with #555, as that really was quite a pain, but in the short term these aren't hard to do nor hard to undo.  A `hideInNamespaces` would be excellent, I don't think too difficult?  I'm not certain if `hideOnRedirect` has the same value — we do explicitly want a separate category for redirects from the standard G set, much like with files or categories.

>making the codebase averse to internationalization

While `hideInNamespaces` would certainly help with portability, these changes aren't a step back since the criteria themselves will almost completely and the whole thing will have to be rewritten.  `hideInNamespaces` would certainly help, though.

siddharthvp:
> , so the infrastructure for hideInNamespaces is already there (starting at L454 of speedy.js).

Thanks, I see. That means implementing this just takes 2 lines of extra code! But I note that on further review, a `showInNamespaces` is helpful rather than `hideInNamespaces`.  Done in #575. I've also incorporated changes made here and in #555 in that. 


> I'm not certain if `hideOnRedirect` has the same value

Yeah, I see. Though we already have it as `hideWhenRedirect` which is used only on the G set.

DannyS712:
It may make sense to close this in favor of #575, which is more elegant and wider-ranging.

PR #575, opened 2019-04-06
speedy: add `showInNamespaces` and `hideInNamespaces` properties to csd criteria lists

siddharthvp:
Per the discussion in #574, added a `showInNamespaces` property to CSD criteria list objects to simplify the code that causes some criteria to be shown only on applicable namespaces.

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -385,15 +385,7 @@ Twinkle.speedy.callback.modeChanged = function twinklespeedyCallbackModeChanged(
 			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.userAllList, mode) } );
 		}
 		work_area.append( { type: 'header', label: 'Redirects' } );
-		if (namespace == 0) {
-			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.articleRedirect.concat(Twinkle.speedy.redirectCriterion, Twinkle.speedy.generalRedirectList), mode) } );
-		}
-		else if (namespace == 6) {
-			work_area.append( { type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.redirectCriterion.concat(Twinkle.speedy.fileRedirect, Twinkle.speedy.generalRedirectList), mode) } );
-		}
-		else {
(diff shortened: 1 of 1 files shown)


DannyS712:
@siddharthvp Have you considered also having a hideInNamespaces? Eg for use with G1 and G2? Something along the lines of returning null when:
`criterion.hideInNamespaces && criterion.hideInNamespaces.indexOf(mw.config.get('wgNamespaceNumber')) > -1` (when hideInNamespaces is defined and the current namespaces is in the array). I would commit it myself, but I don't know how to do that for someone else's pull.

siddharthvp:
> Have you considered also having a hideInNamespaces? Eg for use with G1 and G2?

Neat, done! 

Meanwhile, I'm still wondering how to unify and `userAllList` and `userNonRedirectList`. Maybe the fix is to add `$.extend({showInNamespaces: [ 2 ]}, Twinkle.speedy.userList[0])` and `$.extend({showInNamespaces: [ 2 ]}, Twinkle.speedy.userList[1])` to the `redirectList`? (so that U1 and U2 show up on userspace redirects under the 'Redirects' section?)

DannyS712:
@siddharthvp just add 'hideWhenRedirect: true' to all of them, and combine the 'userAllList' with 'userNonRedirectList'

Amorymeltzer:
>Just remove the G11 and G13 instances in userNonRedirectList

I wouldn't do that — the items there are specific instances of those, as they are oft-used varieties language

DannyS712:
@siddharthvp can you also set 'G8: Subpages with no parent page' hidden for mainspace (ns=0) since subpages are disabled there?

DannyS712:
@siddharthvp Also, db-subpage should be disabled in namespaces 6 and 8 (file and mediawiki) per [gerrit](https://gerrit.wikimedia.org/r/plugins/gitiles/operations/mediawiki-config/+/refs/heads/master/wmf-config/InitialiseSettings.php#4939)

siddharthvp:
I haven't done the unification of `userAllList` and `userNonRedirectList` (into a single `userList`) because neither of the two proposed methods are neat.  

> add 'hideWhenRedirect: true' to U3 and U5, and merge those into 'userAllList'. 

This means we'd be showing all userList criteria on userspace redirects. Does not make semantic sense.

DannyS712:
@siddharthvp I've been waiting for this to be merged to go and clean up the entire list, now that we have a better system in place. I'll try to work on that too

Amorymeltzer:
Given this code, I think it's now less about unifying the lists but whether having a separate header is helpful.  As mentioned in https://github.com/azatoth/twinkle/pull/575#issuecomment-480610055, I believe it so, although it wouldn't hurt to look.

PR #576, opened 2019-04-07
block: Remove old fix for indef expirys that is no longer necessary, refactor some regex

Amorymeltzer:
Various synonyms for infinite weren't always considered in the blocking api, see https://phabricator.wikimedia.org/T68646.  This was fixed in https://gerrit.wikimedia.org/r/c/mediawiki/core/+/140276 so that infinite, indeinite, infinity, and never are treated the same.  Also allowed for infinity and infinite in the regex matching.


Amorymeltzer:
cc @MusikAnimal as this was part of your original creation of this module, although amusingly the underlying was merged about a week later.

PR #578, opened 2019-04-11
speedy/prod: don't log notification if that failed

### LINKED ISSUE #577: Speedy: Bug - CSD log shouldn't say "notified ___" if they weren't notified
https://en.wikipedia.org/wiki/User_talk:Cydebot is protected, meaning that I was unable to notify Cydebot when tagging https://en.wikipedia.org/wiki/Category:Savoy_portal for deletion per C1. However, my CSD log still reports that they were notified.

Expected behavior: Twinkle checks if the notification was actually delivered before adding an entry to the CSD log (or at least checks if the page is editable)

siddharthvp:
Closes #577. If notification to page creator couldn't be sent (eg. when the
user talk page is protected), then don't say "notified ___" in log entry

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -97,7 +97,7 @@ Twinkle.prod.callback = function twinkleprodCallback() {
 	if(namespace !== 'article') {
 		$(result).find('#prodtype_fieldset').hide();
 	}
-	
+
 	// Fake a change event on the first prod type radio, to initialize the type-dependent controls
 	var evt = document.createEvent( "Event" );
 	evt.initEvent( 'change', true, true );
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
I'll test/review this when I have a moment later, but smart and morebits-y to use `onSuccess`/`Failure`

DannyS712:
@Amorymeltzer this is still happening - any update on when the change will be deployed? See [this edit](https://en.wikipedia.org/w/index.php?title=User:DannyS712/CSD_log&diff=893857791&oldid=893857783&diffmode=source) for my latest run in with the bug

Amorymeltzer:
Next week, maybe the week after.  I've been swamped so haven't had much time, but there are a few others I want to get in on the next on-wiki update, in particular #485.

Also, don't notify bots.  We could maybe set up a blacklist for notifications, I think it'd just really by Cydebot, yeah?

Amorymeltzer:
@DannyS712 #593

PR #580, opened 2019-04-11
xfd: add delsort to afd

### LINKED ISSUE #143: Add deletion sorting
https://en.wikipedia.org/wiki/User:John_Vandenberg/Deletion_sorting_tool required TW to work but has since started not working and, according to John Vandenberg, needs to be completely rewritten. Instead of having a standalone deletion sorting tool that requires TW it would be nice if deletion sorting was incorporated into TW.

siddharthvp:
Allow listing of the discussion on deletion sorting pages while creating an AFD. Closes #143. While there may be no need for Twinkle to provide a standalone delsort tool, it would be probably helpful to have one within xfd module, allowing afd nominators to sort their afds while creating it. 

This makes use of the `jquery.chosen` module/library, which I have loaded using `mw.loader.using`. Do we have to add it to the list of dependencies or do something funny like that? I'd think not. @MusikAnimal ? 

This is inspired from Enterprisey's [delsort tool](https://en.wikipedia.org/wiki/User:Enterprisey/delsort), from where the the list of delsort pages has been copied from.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -203,9 +203,64 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 		afd_category.append( { type:'option', label:'Indiscernible or unclassifiable topic', value:'I' } );
 		afd_category.append( { type:'option', label:'Debate not yet sorted', value:'U' } );
 
+		// delsort categories list copied off [[User:Enterprisey/delsort.js]], originally taken from [[WP:DS/C]]
+		var delsortCategories = {
+			"People": ["People", "Academics and educators", "Actors and filmmakers", "Artists", "Authors", "Bands and musicians", "Businesspeople", "Politicians", "Sportspeople", "Women", "Lists of people"],
+			"Arts": ["Arts", "Fictional elements", "Science fiction"],
+			"Arts/Culinary": ["Food and drink", "Wine"],
+			"Arts/Language": ["Language", "Academic journals", "Bibliographies", "Journalism", "Literature", "Logic", "News media", "Philosophy", "Poetry"],
+			"Arts/Performing": ["Albums and songs", "Dance", "Film", "Magic", "Music", "Radio", "Television", "Theatre", "Video games"],
+			"Arts/Visual arts": ["Visual arts", "Architecture", "Fashion", "Photography"],
+			"Arts/Comics and animation": ["Comics and animation", "Anime and manga", "Webcomics"],
+			"Places of interest": ["Museums and libraries", "Shopping malls"],
+			"Topical": ["Animal", "Bilateral relations", "Business", "Conservatism", "Conspiracy theories", "Crime", "Disability", "Discrimination", "Ethnic groups", "Events", "Games", "Health and fitness", "History", "Law", "Military", "Organizations", "Paranormal", "Piracy", "Politics", "Terrorism"],
+			"Topical/Business": ["Business", "Advertising", "Companies", "Management", "Finance"],
+			"Topical/Culture": ["Beauty pageants", "Fashion", "Mythology", "Popular culture", "Sexuality and gender"],
+			"Topical/Education": ["Education", "Fraternities and sororities", "Schools"],
+			"Topical/Religion": ["Religion", "Atheism", "Bible", "Buddhism", "Christianity", "Islam", "Judaism", "Hinduism", "Paganism", "Sikhism", "Spirituality"],
+			"Topical/Science": ["Science", "Archaeology", "Astronomy", "Behavioural science", "Economics", "Environment", "Geography", "Mathematics", "Medicine", "Organisms", "Social science", "Transportation"],
+			"Topical/Sports": ["Sports", "American football", "Baseball", "Basketball", "Bodybuilding", "Boxing", "Cricket", "Cycling", "Football", "Golf", "Horse racing", "Ice hockey", "Rugby union", "Softball", "Martial arts", "Wrestling"],
+			"Topical/Technology": ["Technology", "Aviation", "Computing", "Firearms", "Internet", "Software", "Websites"],
+			"Wikipedia page type": ["Disambiguations", "Lists"],
+			"Geographic/Africa": ["Africa", "Egypt", "Ethiopia", "Ghana", "Kenya", "Laos", "Mauritius", "Morocco", "Nigeria", "Somalia", "South Africa", "Zimbabwe"],
+			"Geographic/Asia": ["Asia", "Afghanistan", "Bangladesh", "Bahrain", "Brunei", "Cambodia", "China", "Hong Kong", "India", "Indonesia", "Japan", "Korea", "Malaysia", "Maldives", "Mongolia", "Myanmar", "Nepal", "Pakistan", "Philippines", "Singapore", "South Korea", "Sri Lanka", "Taiwan", "Thailand", "Vietnam"],
+			"Geographic/Asia/Central Asia": ["Central Asia", "Kazakhstan", "Kyrgyzstan", "Tajikistan", "Turkmenistan", "Uzbekistan"],
+			"Geographic/Asia/Middle East": ["Middle East", "Iran", "Iraq", "Israel", "Jordan", "Kuwait", "Lebanon", "Libya", "Palestine", "Saudi Arabia", "Syria", "United Arab Emirates", "Yemen", "Qatar"],
+			"Geographic/Europe": ["Europe", "Albania", "Armenia", "Austria", "Azerbaijan", "Belarus", "Belgium", "Bosnia and Herzegovina", "Bulgaria", "Croatia", "Cyprus", "Czech Republic", "Denmark", "Estonia", "Finland", "France", "Georgia (country)", "Germany", "Greece", "Hungary", "Iceland", "Ireland", "Italy", "Jersey", "Kosovo", "Latvia", "Lithuania", "Luxembourg", "Macedonia", "Malta", "Moldova", "Montenegro", "Netherlands", "Norway", "Poland", "Portugal", "Romania", "Russia", "Serbia", "Slovakia", "Slovenia", "Spain", "Sweden", "Switzerland", "Turkey", "Ukraine", "Yugoslavia"],
+			"Geographic/Europe/United Kingdom": ["United Kingdom", "England", "Northern Ireland", "Scotland", "Wales"],
+			"Geographic/Oceania": ["Oceania", "Antarctica", "Australia", "New Zealand"],
+			"Geographic/Americas/Canada": ["Canada", "British Columbia", "Manitoba", "Nova Scotia", "Ontario", "Quebec", "Alberta"],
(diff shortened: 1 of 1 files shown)


modules/twinklexfd.js
```diff
@@ -740,6 +809,13 @@ Twinkle.xfd.callbacks = {
 			}
 			usertalkpage.setFollowRedirect(true);
 			usertalkpage.append();
+		},
+		delsortListing: function(pageobj) {
+			var discussionPage = pageobj.getCallbackParameters().discussionPage;
+			var text = pageobj.getPageText().replace("directly below this line -->", "directly below this line -->\n{{" + discussionPage + "}}");
+			pageobj.setPageText(text);
+			pageobj.setEditSummary('Listing [[:' + discussionPage +']].' + Twinkle.getPref('summaryAd'));
+			pageobj.save();
```
```suggestion
			pageobj.setCreateOption('nocreate');
			pageobj.save();
```

Just to be safe, eh?

modules/twinklexfd.js
```diff
@@ -203,9 +203,61 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
+			var grp = delsort.append( { type: 'optgroup', label: grpname } );
+			list.forEach(function(item) {
+				grp.append( { type: 'option', label: item, value: item } );
+			});
+		});
+
 		appendReasonBox();
 		work_area = work_area.render();
 		old_area.parentNode.replaceChild( work_area, old_area );
+		mw.loader.using('jquery.chosen').then(function chosenInit() {
```
Amorymeltzer: Fairly certain this shouldn't be here, but rather it is dealt with at [MediaWiki:Gadgets-definition](https://en.wikipedia.org/wiki/MediaWiki:Gadgets-definition). 
 That is, this should be changed in the [readme](https://github.com/azatoth/twinkle/blob/master/README.md#updating-scripts-on-wikipedia), although this clearly isn't used by anything else at the moment.

Amorymeltzer:
Left a couple comments, not a full review — would like to hear what @MusikAnimal thinks first.  Also cc @enterprisey.

@enterprisey How did you come up with that list, I'm sure that's not the full extent of possibilities?  If this does go forward, I think it'd be good for us to come up with a set criterion for inclusion/exclusion here.

JJMC89:
> @enterprisey How did you come up with that list, I'm sure that's not the full extent of possibilities? If this does go forward, I think it'd be good for us to come up with a set criterion for inclusion/exclusion here.

It is based on [Wikipedia:WikiProject Deletion sorting/Compact](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Deletion_sorting/Compact). WP:DS/C is likely missing some pages, and the list in the script it is likely out of sync with WP:DS/C since I haven't been keeping up with it.

modules/twinklexfd.js
siddharthvp: We don't need to load jquery.chosen unless the TW xfd module is opened and mode is set to AFD. I don't see the benefit in having it load on every page.

siddharthvp: BTW, jquery.chosen will look good on warn module too, for warning selection dropdown, as I have said elsewhere earlier.

MusikAnimal:
Deletion sorting is one thing I've virtually no experience in... so I may not have much to offer here. That said, I don't see a lot of new code, so this seems like an acceptable addition that isn't too far out of scope. I would be weary of letting it grow into a large, full-on delsort module (should that be the concern), since it is a very niche area of the project.

MusikAnimal:
Regarding `jquery.chosen`, it's certainly cleaner to specify this in the gadget definition, though I agree XfD is among the less popular modules. It will load very fast, and the client will cache it (so negligible difference in a single session), but it's still 10 KB of JavaScript for something that won't be used often.

Have we considered the native [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist)? It may not look as pretty but it may work just as well here, and is safe for us to use in terms of browser compatibility.

siddharthvp:
`datalist` allows user input that's not a part of the options in the dropdown. I don't think we want that here. Also it doesn't seem to support multiple inputs, which is necessary here. 

jquery.chosen is good in terms of browser compatibility: I've tested it in IE too. On mobile browsers, the chosen features don't work - the field fallbacks to the native multi-select, which still is usable. 

I don't have any strong preference regarding how we load it. Its used on several of Enterprisey's scripts and we could use it in the Warn module too, so maybe the 10kb JS won't be that unused if we put it in the gadget definition.

Amorymeltzer:
I'd much prefer it in the definition, if only for cleanliness and consistency, but js does already take a noticeable time to load for some (although I think popups is far and away the worst offender?), so I'm sympathetic to the inline style.

That being said, we *could* use it for nearly everything with a longish dropdown, no?  I dunno if it's a good idea — at a minimum it clashes a bit with our non-ooui style — but it would avoid things like #491, where you first mentioned it, and might be appreciated in @MusikAnimal's `block` module. 🤔

MusikAnimal:
Sounds fine to me!

modules/twinklexfd.js
Amorymeltzer: Per comments below

modules/twinklexfd.js
```diff
@@ -203,9 +203,61 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
+			"Geographic/Americas/USA": ["United States of America", "Alabama", "Alaska", "Arizona", "Arkansas", "California", "Colorado", "Connecticut", "Delaware", "Florida", "Georgia (U.S. state)", "Hawaii", "Idaho", "Illinois", "Indiana", "Iowa", "Kansas", "Kentucky", "Louisiana", "Maine", "Maryland", "Massachusetts", "Michigan", "Minnesota", "Mississippi", "Missouri", "Montana", "Nebraska", "Nevada", "New Hampshire", "New Jersey", "New Mexico", "New York", "North Carolina", "North Dakota", "Ohio", "Oklahoma", "Oregon", "Pennsylvania", "Rhode Island", "South Carolina", "South Dakota", "Tennessee", "Texas", "Utah", "Vermont", "Virginia", "Washington", "Washington, D.C.", "West Virginia", "Wisconsin", "Wyoming"],
+			"Geographic/Unsorted": ["Islands"]
+		};
+
+		var delsort = work_area.append( {
+			type: 'select',
+			multiple: true,
+			name: 'delsort',
+			label: 'Choose deletion sorting pages to list in: '
```
Amorymeltzer: Awkward phrasing.  Maybe something like `Deletion sorting categories to list article in` or `Deletion categories to sort article into`?  I'd prefer `in which to sort article` but I think I'm likely alone there!  I also think it'd be a good idea to have a tooltip with some additional guidance, in particular around not going too crazy.  There's been some abuse lately, and a discussion is currently ongoing about this, so better to nip it in the bud.  Not sure if the below is best.

```suggestion
			label: 'Deletion categories to sort article into: ',
			tooltip: 'Only select a few specific categories that are meaningful and significant to the subject of the article.'
```

modules/twinklexfd.js
```suggestion
		// TODO: Add to the gadget definition if used elsewhere
		mw.loader.using('jquery.chosen').then(function chosenInit() {
```

modules/twinklexfd.js
```diff
@@ -520,6 +579,7 @@ Twinkle.xfd.callbacks = {
 		if (form.xfdtarget2) {
 			params.target2 = form.xfdtarget2.value;
 		}
+		params.delsort_cats = $(form.delsort).val();
```
Amorymeltzer: I know this will likely be dealt with in another of your pulls, but why the jquery here compared to the others?

modules/twinklexfd.js
```diff
@@ -203,9 +203,61 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
+			});
+		});
+
 		appendReasonBox();
 		work_area = work_area.render();
 		old_area.parentNode.replaceChild( work_area, old_area );
+		mw.loader.using('jquery.chosen').then(function chosenInit() {
+			$(work_area).find('[name=delsort]')
+				.attr('data-placeholder', 'Select delsort pages')
+				.chosen({width: "100%"});
```
Amorymeltzer: CSS isn't my strong suit but I don't suppose there's any way to get this to start on the same line as the label, but then be at 100% when multiline?  I think it's either/or.

Amorymeltzer:
One other thing I think would be good to commit here is what to include/exclude in the future.  I'm fine cribbing off @enterprisey and/or WP:DS/C, but we should definitely say so.

enterprisey:
Looks good! I originally got it from DS/C, and then people would periodically tell me about new lists to add on my user talk page (or submit pull requests to my github repo).

modules/twinklexfd.js
Amorymeltzer: I had thought this would be nullified by #597 but seems not.  Still think this should match the others.

siddharthvp: Can't do that, `value` property for a multi-select input gives only the first selected option rather than the list of all selected options, which we need.

Amorymeltzer: Indeed!

Amorymeltzer:
@siddharthvp I know you're busy, but my changes are just comment-tweaks, so if you don't mind, I think I can handle that in the terminal and get this merged and pushed live today.

modules/twinklexfd.js
siddharthvp: Don't really follow. Are u saying that the input box should be on the same line as the label? I don't think so since the label is kinda long, also if there are multiple delsorts, they'd take up space.

modules/twinklexfd.js
siddharthvp: Made the label "Choose deletion sorting categories: ", as I think it'd be good to preserve the original phrasing "deletion sorting" somewhere, so that ppl know what exactly this is. 

Dont really agree with the tooltip. My reading of the consensus at [the discussion](https://en.wikipedia.org/wiki/Wikipedia_talk:WikiProject_Deletion_sorting#Is_there_such_a_thing_as_too_much_deletion_sorting?) is that generally AfDs may be put into as many delsorts as possible.

modules/twinklexfd.js
Amorymeltzer: I don't think it's possible, but what I *really* want is for it to be on the same line, but then when it would overflow, the input would be on it's own line.  Not something I intended for you to fix, just wishin'

modules/twinklexfd.js
Amorymeltzer: 👍 on the first point.  As for that discussion, I don't think there was much of a consensus reached either way, but as this will be new behavior for editors and most do not DELSORT after nomination, some level of how to chose might be appreciated.

Maybe if I tone down my initial suggestion to something like
`tooltip: 'Select a few categories that are relevant to the subject of the article.'`

If you object to the "a few" that can be removed.

siddharthvp: Done.

modules/twinklexfd.js
siddharthvp: Not possible it seems. But even if it was, I'd say such "jumping" elements are poor user interfaces.

PR #584, opened 2019-04-18
speedy: Further improve extra parameters in CSD log

Amorymeltzer:
1st commit: Ignore blank f9 rationale or url in log (bugfix)
- Fix logic from #546 that would allow url or rationale as an empty string when logging in #441

2nd commit: link additional parameters in the CSD log if possible (enhancement)
- wikilink: G4 `xfd`; G6 `page`, `fullvotepage`, `sourcepage`; A2 `source`; A10 `article`; F5 `replacement`
- user: G5 `user`
- file: F1 `filename`
- template: T3 `template`
- external: G12 `url`, `url2`, `url3` (only if beginning with http)
- commons: F8 `filename`
- WP:CSD section: P1 `criterion`

Style the link using curly braces `{}` instead of square brackets `[]`.  Also adds a tooltip to the G6 xfd input.

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1650,7 +1650,7 @@ Twinkle.speedy.callbacks = {
 						params.templateParams.forEach(function(item, index) {
 							var keys = Object.keys(item);
 							if (keys[0] !== undefined && keys[0].length > 0) {
-								//Second loop required since G12 can have multiple urls
+								//Second loop required since some items (G12, F9) may have multiple keys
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Two minor points:
- The G12 links could have the `plainlinks` class, but I think that might be surprising to folks who could expect/want the external link icons
- I tweaked the language around the P1 criterion; there's no sanitizing of the entry, but a single item is required for the db-p1 template anyway

PR #585, opened 2019-04-18
prod: Don't insert a stray comma when endorsing prod

Amorymeltzer:
Caused by a capturing group introduced in b7f79ed


Amorymeltzer:
@siddharthvp

PR #589, opened 2019-04-22
fluff: Add vanarticle parameter to user talk links on diffs

Amorymeltzer:
Allows user to click to an editor involved in a diff and provide a warning with the article title filled-in, without having to revert the diff themself (e.g., already-reverted vandalism that went unwarned).  This has been requested a few times, and is fairly trivial.  The link is bolded, just like the user talk link on the mw rollback done page.  Clicking the talkpage link will NOT trigger the autowarn feature (#509) even if enabled.

Also sorted some functions from #562 into the `Twinkle.fluff` object.

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -17,21 +17,38 @@
  Twinklefluff revert and antivandalism utility
  */
 
-var spanTag = function( color, content ) {
-	var span = document.createElement( 'span' );
-	span.style.color = color;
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
Thinking about #605, perhaps this change would be better suited in the `warn` module, much like is currently done for `welcome`

Amorymeltzer:
Closed in favor of #626

Issue #533, opened 2019-03-04
Discussion: What criteria should determine inclusion into the warn menu?


Amorymeltzer:
Additions to the `twinklewarn` menu are commonly suggested, and right now our inclusion criteria are a bit _ad hoc_, so it'd be good to nail those done, especially for #477.  As @atlight [nicely summarized](https://github.com/azatoth/twinkle/pull/412#issuecomment-397842925), there are two major points we should consider as guiding concerns:

1. Length of the list
2. Not disrupting current practice

A swollen list makes it difficult for anyone, in particular newish users, to find what they want (#517 may help).  We're kind of begging the question here, given that most warnings are likely given by Twinkle and Huggle, but regardless we do not want to overwhelm or aggravate antivandalism efforts.

#412 is probably where the most consolidated discussion has taken place, but as a PR has gotten unwieldy and isn't a good starting place, so I'm opening this to supersede and centralize discussion of the merits.  The last major pruning was done in #302 by @MusikAnimal.

NovemLinguae:
I propose that we make the guideline >25 substs. We can determine this via insource search, because subst'd templates leave a comment that looks like `<!-- Template:Uw-test -->`

PR #592, opened 2019-04-24
warn: Reorganize messages object, should make it more clear to maintain

Amorymeltzer:
Setting it up this way means there is even more of a difference between the scaled levels (level2, level4im, etc.) and singlenotice/singlewarning/custom,
but doing so should make it very straightforward to remove or add templates (pursuant to any discussion in #533).

The only change to the enduser is that this will now generate a submenu category for every grouping, even if there are no templates in that group at that level.  At the moment, this is only the `4im` level for the "Removal of deletion tags" category.  I'm not sure this is a bad thing.

One header was mistmatched ('other editors' versus 'other users'), resolves that.

----
Draft for reviewing and feedback as I think this will be helpful, but I'm not sure!
- I whipped up some perl ([`procjson.pl`](https://github.com/Amorymeltzer/wiki/blob/54221374cc541bf2abb36b5a235db7e4bc6d2f0f/procjson.pl)) to handle most of the processing, in case anyone wants to check it out.
- Some processing of the level `value` is needed to build the templates, this seems simpler than providing an optional parameter to `createEntries`
- As noted above, the current structure creates an empty group in the `4im` menu for "Removal of deletion tags" whereas the old system didn't.  I don't think this is too bad, if bad at all, but it should be the only visible change to Twinkle users.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -119,655 +119,663 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 //   summary (required): The edit summary used. If an article name is entered, the summary is postfixed with "on [[article]]", and it is always postfixed with ". $summaryAd"
 //   suppressArticleInSummary (optional): Set to true to suppress showing the article name in the edit summary. Useful if the warning relates to attack pages, or some such.
 Twinkle.warn.messages = {
-	level1: {
+	levels: {
 		"Common warnings": {
-			"uw-vandalism1": {
-				label: "Vandalism",
-				summary: "General note: Unconstructive editing"
+			"uw-vandalism": {
+				level1: {
+					label: "Vandalism",
+					summary: "General note: Unconstructive editing"
+				},
+				level2: {
+					label: "Vandalism",
+					summary: "Caution: Unconstructive editing"
+				},
+				level3: {
+					label: "Vandalism",
+					summary: "Warning: Vandalism"
+				},
+				level4: {
+					label: "Vandalism",
+					summary: "Final warning: Vandalism"
+				},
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Placeholder comment to remind myself to remove the perl script

modules/twinklewarn.js
```diff
@@ -119,655 +119,663 @@ Twinkle.warn.callback = function twinklewarnCallback() {
...
+					label: "Removing {{blp prod}} templates",
+					summary: "General note: Removing {{blp prod}} templates"
+				},
+				level2: {
+					label: "Removing {{blp prod}} templates",
+					summary: "Caution: Removing {{blp prod}} templates"
+				},
+				level3: {
+					label: "Removing {{blpprod}} templates",
+					summary: "Warning: Removing {{blpprod}} templates"
```
kenrick95: Hi @Amorymeltzer 

Is there any reason the 3rd level template is `{{blpprod}}` but other levels are `{{blp prod}}`?

Amorymeltzer: @kenrick95 No good reason at all!  The space was missing for both level 3 and level 4 when introduced in 2010, and the space was added to level 4 in 96e6ac3 but not level 3.  This is just the label so it doesn't make a difference beyond sloppiness.

Amorymeltzer: Although it should probably be neither of those, but rather `{{prod blp}}`, or even `WP:BLPPROD`

kenrick95: Alright, thanks for the insight! 👍

PR #595, opened 2019-04-24
warn: remove old fix for out-of-range defaultGroup

Amorymeltzer:
This dates back to 2008 (<https://en.wikipedia.org/w/index.php?diff=247182138&oldid=245908598&title=User:AzaToth/twinklewarn.js>).  There used to be a group 8 with block templates for sysops, hence the `Morebits.userIsInGroup` switch, but after #260 it was removed and this persisted, meaning that we don't need any special handling for sysops or nonsysops here.  That is, `defaultGroup > 7` would handle any remaining unmodified options (there are some).  <s>That being said, if even *that* isn't present, I don't get any errors in Chrome, FireFox, or Safari, so it doesn't seem necessary.  Perhaps it throws/threw an error in IE?</s>  Similarly, `undefined` or -1 groups were possible due to how the config was structured back then: in your skin.js rather than a separate file.

Perhaps old browsers threw an error, but for now I can't seem to trigger any errors and regardless, the default behavior if something isn't true is the first option, so none of this is needed.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -66,7 +66,7 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 		} );
 
 	var defaultGroup = parseInt(Twinkle.getPref('defaultWarningGroup'), 10);
-	main_group.append( { type: 'option', label: '1: General note', value: 'level1', selected: ( defaultGroup === 1 || defaultGroup < 1 || ( Morebits.userIsInGroup( 'sysop' ) ? defaultGroup > 8 : defaultGroup > 7 ) ) } );
+	main_group.append( { type: 'option', label: '1: General note', value: 'level1', selected: ( defaultGroup === 1 ) } );
 	main_group.append( { type: 'option', label: '2: Caution', value: 'level2', selected: ( defaultGroup === 2 ) } );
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
@siddharthvp You mentioned elsewhere you checked IE, what version?  Would you mind testing this real quick (edit `defaultWarningGroup` in `twinkleoption.js` to be 8 or 10 or something then load the code) 'cause I can't get an error out of this.

siddharthvp:
Don't get any errors on IE. I think the first item in `<select>` is selected by default if the selection condition of none of the options are true. So, there's no need for that `defaultGroup < 1`, too.

Amorymeltzer:
@siddharthvp Thanks — hard for me to test IE stuff on my mac!  Not sure why `-1` was there, I presumed it was a potential issue with Twinkle's getting of preferences, but yeah, can't find it.  There used to be an `undefined` check, though I think that's more likely to be a Twinkle-based issue than a browser one.

Amorymeltzer:
Forgot that back in the old days, we put our TwinkleConfig in our skin.js, so there was no guaranteed default (as we now have in Twinkle.js) thus things like -1 or undefined were possible..

PR #597, opened 2019-04-24
xfd: automate collection of filled form data

siddharthvp:
In Twinkle.callbacks.evaluate, use jQuery to save all entered form values into an object, rather than saving each one explicitly into a variable. Something similar was done for tag module in e54abb32.


siddharthvp:
I am rethinking the utility of this change. I agree that this raises a lot of unnecessary confusion. Using an object to store the form entries was perhaps warranted in tag module due to the huge number of parameters and fields. I now think it is not worth it here. To say nothing of the conflict with #580, which uses jQuery syntax. 

What is needed actually is to simplify the conditions and remove the pointless condition checks, which surprisingly results in much neater code. See #634. Suggesting closure of this one in favour of that.

PR #598, opened 2019-04-24
morebits: Remove Morebits.bytes

siddharthvp:
Morebits.bytes is not used anywhere in Twinkle or any other Morebits-based gadget/script on enwiki. See [search result](https://en.wikipedia.org/w/index.php?search=insource%3A%2FMorebits%5C.bytes%2F&title=Special%3ASearch&profile=advanced&fulltext=1&advancedSearch-current=%7B%7D&ns0=1&ns1=1&ns2=1&ns3=1&ns4=1&ns5=1&ns6=1&ns7=1&ns8=1&ns9=1&ns10=1&ns11=1&ns12=1&ns13=1&ns14=1&ns15=1&ns100=1&ns101=1&ns108=1&ns109=1&ns118=1&ns119=1&ns446=1&ns447=1&ns710=1&ns711=1&ns828=1&ns829=1&ns2300=1&ns2301=1&ns2302=1&ns2303=1). 

If this code is of any use, we should find another home for it. No point in inflating the size of this already 4000-line library with unused stuff.


Amorymeltzer:
I did some sleuthing, this was used by `traverseimages` (removed in 9656f8cd), which was a sysop-only module that seems to have been used for going through image categories and selecting an action for each, one-by-one.

I see no need for this, but I think it ''does'' open up questions about how Twinkle-focused `Morebits` should be.  Pinging @MusikAnimal @kevinji @atlight @azatoth on the offchance any of them have an opinion.  I've got some changes in mind (as I imagine you do!) depending on that viewpoint.

PR #599, opened 2019-04-24
morebits: Remove string trimming shims

siddharthvp:
- String.prototype.trim is now supported on all target browsers
- String.prototype.trimLeft is not used in Twinkle or other Morebits-based scripts
- String.prototype.trimRight was used only once which I removed as we can just use trim there

trimLeft and trimRight are unsupported only in IE 11.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1020,24 +1020,6 @@ Morebits.bytes.prototype.toString = function( magnitude ) {
  * **************** String; Morebits.string ****************
  */
 
-if (!String.prototype.trimLeft) {
-	String.prototype.trimLeft = function stringPrototypeLtrim( ) {
-		return this.replace( /^[\s]+/g, "" );
-	};
-}
-
-if (!String.prototype.trimRight) {
-	String.prototype.trimRight = function stringPrototypeRtrim( ) {
-		return this.replace( /[\s]+$/g, "" );
-	};
-}
-
-if (!String.prototype.trim) {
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Aren't we explicitly trying to cover IE11, though?  MediaWiki in general still is, and though I don't know about enWiki specifically, the [analytics dashboard](https://analytics.wikimedia.org/dashboards/browsers/#desktop-site-by-browser/browser-family-and-major-hierarchical-view) says that IE11 has been (coincidentally) 11% of desktop browsers to WMF sites the last four months.

siddharthvp:
I should have been clearer: `trimLeft` is totally unused. `trimRight` had one usage in TW which I removed. It is still used in TTO's [masstag script](https://en.wikipedia.org/wiki/User:This,_that_and_the_other/masstag.js). Maybe we could retain the shim for trimRight just to cover that usage, but I don't think it's worth it since `trim` (which is supported on IE) may well be used all the time instead of `trimRight`. 

I don't think we should worry about breaking that script since it's for TTO's personal use and let's hope he doesn't use IE.

Amorymeltzer:
Ah yeah I get ya now.  Agreed that's not a huge issue, as TTObot is inactive, but ping @atlight anyway.

atlight:
I don't use IE :) Feel free to remove it.

Amorymeltzer:
@atlight I knew you'd have good taste!

PR #600, opened 2019-04-24
morebits: Remove Morebits.wikipedia

siddharthvp:
- On enwiki, `Morebits.wikipedia.namespaces` is just a copy of `mw.config.get('wgFormattedNamespaces')`. For other wikis, it is useless.
- `mw.config.get('wgFormattedNamespaces')` is better and works across wikis.
- `Morebits.wikipedia.namespacesFriendly` (also enwiki-only) iss used just once in the codebase. This usage has been changed to use wgFormattedNamespaces.

 Also removed the comment about removal of Twinkle blacklist 8 years ago.


modules/twinkleunlink.js
```diff
@@ -164,7 +164,8 @@ Twinkle.unlink.callbacks = {
 					apiobj.params.form.append( { type:'header', label: 'File usage' } );
 					namespaces = [];
 					$.each(Twinkle.getPref('unlinkNamespaces'), function(k, v) {
-						namespaces.push(Morebits.wikipedia.namespacesFriendly[v]);
+						var nsName = mw.config.get('wgFormattedNamespaces')[v];
```
Amorymeltzer: This might just be me, and it's certainly nitpicking, but I see:
````
var = value;
array.push(ternary)
````
As more complicated than both
````
var = ternary
array.push(var)
````
and
````
array.push(ternary)
````

File: modules/twinkleunlink.js [hunk 2]
```diff
@@ -210,7 +211,8 @@ Twinkle.unlink.callbacks = {
 				apiobj.params.form.append( { type:'header', label: 'Backlinks' } );
 				namespaces = [];
 				$.each(Twinkle.getPref('unlinkNamespaces'), function(k, v) {
-					namespaces.push(Morebits.wikipedia.namespacesFriendly[v]);
+					var nsName = mw.config.get('wgFormattedNamespaces')[v];
```
Amorymeltzer: Should match above, if that is changed.

Amorymeltzer:
I thought I already opened a PR for this and nobody joined in, oops!  Thanks for picking up the ball.  I might leave this open for a bit for the sake of it but 👍

modules/twinkleunlink.js
siddharthvp: just wanted to avoid writing `mw.config.get('wgFormattedNamespaces')` twice!

Amorymeltzer: Why twice?  `v` holds the number, we only care if `v === 0`, right?  So wouldn't something like either
```suggestion
						namespace.push(v === 0 ? '(Article)' : mw.config.get('wgFormattedNamespaces')[v];
```
or
```suggestion
						var nsName = (v === 0 ? '(Article)' : mw.config.get('wgFormattedNamespaces')[v]);
						namespaces.push(nsName);
```
do the trick?

siddharthvp: Oh yeah, that's neat. Didn't think of testing on `v`. It's been a long night...

modules/twinkleunlink.js
```diff
@@ -164,7 +164,7 @@ Twinkle.unlink.callbacks = {
 					apiobj.params.form.append( { type:'header', label: 'File usage' } );
 					namespaces = [];
 					$.each(Twinkle.getPref('unlinkNamespaces'), function(k, v) {
-						namespaces.push(Morebits.wikipedia.namespacesFriendly[v]);
+						namespaces.push(v === 0 ? '(Article)' : mw.config.get('wgFormattedNamespaces')[v]);
```
Amorymeltzer:
```suggestion
						namespaces.push(v === '0' ? '(Article)' : mw.config.get('wgFormattedNamespaces')[v]);
```
Sorry, led you slightly astray; `$.each` casts v as a string, while the raw `0` gets treated as `number`

modules/twinkleunlink.js
```diff
@@ -210,7 +210,7 @@ Twinkle.unlink.callbacks = {
 				apiobj.params.form.append( { type:'header', label: 'Backlinks' } );
 				namespaces = [];
 				$.each(Twinkle.getPref('unlinkNamespaces'), function(k, v) {
-					namespaces.push(Morebits.wikipedia.namespacesFriendly[v]);
+					namespaces.push(v === 0 ? '(Article)' : mw.config.get('wgFormattedNamespaces')[v]);
```
Amorymeltzer:
```suggestion
					namespaces.push(v === '0' ? '(Article)' : mw.config.get('wgFormattedNamespaces')[v]);
```

PR #601, opened 2019-04-24
batchundelete: restore talk pages too

siddharthvp:
Add option to restore talk pages of undeleted pages, if they existed.

### DIFF
--- a/modules/twinklebatchundelete.js
+++ b/modules/twinklebatchundelete.js
@@ -33,6 +33,17 @@ Twinkle.batchundelete.callback = function twinklebatchundeleteCallback() {
 
 	var form = new Morebits.quickForm( Twinkle.batchundelete.callback.evaluate );
 	form.append( {
+			type: 'checkbox',
+			list: [
+				{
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
A quick test shows the messages don't play quite right here.  Maybe it'd be better with more pages?  For a single page+talk restoration, the "success" shows up on top of the "..." message, with the latter remaining even after the talkpage is done restoring.  That is,

````
Undeleting pages: Done (1/1 actions completed successfully)
Undeleting page User:Amory: undeleting...
Undeleting talk page: done
Status: batch undeletion is now complete
````

That seems unexpected?

Amorymeltzer:
@siddharthvp I've got something for #613 that would restructure things around a bit, meaning this would need to be refactored.  I'd like to get that in first to try and deal with the issue, mainly to keep things as simple as can be.  If you don't mind, I can do that work and open a separate PR that redoes this, credited to you.  Sound okay?

Amorymeltzer:
On second thought, never mind, I'll deal with the rebasin' myself

PR #607, opened 2019-04-28
Remove extra blank line for UAA reports

### LINKED ISSUE #606: Bad list markup when reporting to UAA
When reporting a user to UAA, Twinkle adds a blank line before the report. This is not desirable because it causes the HTML markup to show as lots of separate lists when it is in reality only one list.

pppery:
Fixes #606.


Amorymeltzer:
Added in c029ae93 after a request at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_30#UAA_reporting_requests.  I like it, but it doesn't seem necessary nor does it match AIV.

PR #609, opened 2019-04-28
Talkback: Go to page footer on reloading

DannyS712:
Since talkback appends new content, when the page is reloaded the focus should be at the bottom, which can be accessed as "#footer"

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -352,7 +352,7 @@ var callback_evaluate = function( e ) {
 	Morebits.simpleWindow.setButtonsEnabled( false );
 	Morebits.status.init( e.target );
 
-	Morebits.wiki.actionCompleted.redirect = fullUserTalkPageName;
+	Morebits.wiki.actionCompleted.redirect = fullUserTalkPageName + '#footer';
 	Morebits.wiki.actionCompleted.notice = "Talkback complete; reloading talk page in a few seconds";
 
 	var talkpage = new Morebits.wiki.page(fullUserTalkPageName, "Adding talkback");


Amorymeltzer:
Shouldn't this go to the actual new section rather than the footer?

DannyS712:
@Amorymeltzer that would be better, but it would be a _lot_ harder to implement from what I can see. The section headings aren't defined yet when the redirect is set. Furthermore, some of the talkback notice templates include their own headings, while others are explicitly defined, either in the code here or based on user preferences, including the `adminNoticeHeading`, `mailHeading`, and `talkbackHeading`. In short, while going to the actual new section would be ideal, it would also being a lot harder to implement, especially because the headers included in some templates can change.

siddharthvp:
Even if could do it, it results in a problem if there is already a section
with the same title on the page, as we'd end up at that section instead.
Going straight to the #footer is pretty neat idea, I think.

Amorymeltzer:
Ah yeah, that makes sense, way too much going on here.

Amorymeltzer:
Apparently `Morebits.wiki.actionCompleted.redirect` just uses `window.location` so this, being called from the very same user talk page, will not work.  I guess we'd have to modify the `actionCompleted` process?

DannyS712:
> Apparently `Morebits.wiki.actionCompleted.redirect` just uses `window.location` so this, being called from the very same user talk page, will not work. I guess we'd have to modify the `actionCompleted` process?

In that case, why do we currently specify `Morebits.wiki.actionCompleted.redirect = fullUserTalkPageName`?

Amorymeltzer:
talkback can be used from other pages besides just the user talk page.  The `#foooter` trick works for those, but breaks reloading entirely if used on the root user talk.

DannyS712:
Why would it break reloading?

Amorymeltzer:
@DannyS712 I'm guessing you didn't test it?  Setting `window.location` will load that page, but if `hash` is different, not the page, it doesn't reload.

DannyS712:
> @DannyS712 I'm guessing you didn't test it? Setting `window.location` will load that page, but if `hash` is different, not the page, it doesn't reload.

How exactly would I test it? And, if this is not an option, how can we have it reload and scroll to the bottom using a different method?

Amorymeltzer:
@DannyS712 Sorry — I've been (slowly) working on contributing guidelines, but have largely been waiting for #500 before making the change.  Basically, you copy the text of the script, dump it into your browser's console, and then run it.  And I'm not sure; I haven't looked into it too deeply, but `window.location` is weird; I don't think `.replace` or `reload` quite do what we want.  Setting the `href` *and* `reload`ing has weird results, none of which are great.

Amorymeltzer:
This would seem to also be the case for `friendlywelcome` and `twinklewarn`, although I think in the case of the latter we likely know the section ahead of time.

NovemLinguae:

PR #610, opened 2019-04-29
Shift morebits lookupCreator to lookupCreation, also include timestamp; use in prod to limit BLP category alert

Amorymeltzer:
1st commit: morebits: Shift lookupCreator to lookupCreation, also include timestamp

It's a free lookup and the time of creation can be useful.  Includes a notice about deprecation of `lookupCreator`, but still let's it be used.  Should probably notify any users with scripts depending on this.

2nd commit: prod: Wait three days before issuing BLPPROD alert if not in BLP cat

There was some annoyance with the alert given that BLPPORD taggings may frequently happen as part of NPP, where the category wouldn't be added.  See https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_40#BLPPROD_Issue_-_Pseudo-Improvement

Since we've added a callback to get the time of creation before tagging, we can restructure the user notification to remove that callback, essentially staying net-neutral

----
There are a few different ways to do this, but figured including the timestamp in the lookup could be helpful down the road for other options, and wouldn't cost anything.  @siddharthvp as you wrote the initial alert in #458 and are up-to-date on `morebits` so I'd appreciate you taking a look over this.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1328,7 +1328,7 @@ Twinkle.tag.callbacks = {
 					});
 				}
 				if (params.translationNotify) {
-					pageobj.lookupCreator(function(innerPageobj) {
+					pageobj.lookupCreation(function(innerPageobj) {
 						var initialContrib = innerPageobj.getCreator();
 
 						// Disallow warning yourself
--- a/modules/twinkleimage.js
+++ b/modules/twinkleimage.js
@@ -255,7 +255,7 @@ Twinkle.image.callback.evaluate = function twinkleimageCallbackEvaluate(event) {
 
 	// Notifying uploader
 	if( notify ) {
-		wikipedia_page.lookupCreator(Twinkle.image.callbacks.userNotification);
+		wikipedia_page.lookupCreation(Twinkle.image.callbacks.userNotification);
 	} else {
 		// add to CSD log if desired
 		if (lognomination) {
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -201,6 +201,17 @@ Twinkle.prod.callbacks = {
 			}
 		}
 
+		var ts = new Morebits.wiki.page(mw.config.get('wgPageName'));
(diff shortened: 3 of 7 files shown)


Amorymeltzer:
Will need adjusting around #563

siddharthvp:
Code looks fine to me, though I haven't tested...

Amorymeltzer:
@MusikAnimal Curious what you think of the deprecation message/warning.  I don't think it's something we've done much in the past, but while I think the rename is worth doing if this goes forward, I'm wary of changing something so long-standing.

MusikAnimal:
> @MusikAnimal Curious what you think of the deprecation message/warning. I don't think it's something we've done much in the past, but while I think the rename is worth doing if this goes forward, I'm wary of changing something so long-standing.

Only so many uses on enwiki https://en.wikipedia.org/w/index.php?title=Special:Search&limit=75&offset=0&ns2=1&ns8=1&search=%22lookupCreator%22&advancedSearch-current={} and many of those are probably unused scripts. There are other wikis too, of course.

Could we make it backwards-compatible? So `lookupCreator` will use `lookupCreation` but only return the creator, in whatever format it expected. An API warning seems appropriate, too. Morebits should really be using JSDoc (with [`@deprecated`](https://jsdoc.app/tags-deprecated.html) in this case), but I'd put a `NOTE:` or other tag to make it stand out.

Amorymeltzer:
(Was waiting to merge all others to fix this up, wanted to avoid conflicts)

>Could we make it backwards-compatible?

@MusikAnimal My deprecation warning was leftover from a previous rewrite, sorry for that; this was basically already done.  The way it's structured means that the only change is `lookupCreator` to `lookupCreation`; `getCreator` is still the appropriate thing to call later on, so there's no change in return format or anything.

>An API warning seems appropriate, too. Morebits should really be using JSDoc (with @deprecated in this case), but I'd put a `NOTE:` or other tag to make it stand out.

Done, done, and done!

PR #612, opened 2019-04-30
Config: Add <nowiki> tags to user config pages

DannyS712:
Add <nowiki> at the start of users' twinkleoptions.js, and </nowiki> at the end, to ensure that templates listed in them aren't parsed. Currently, [this tracking category](https://en.wikipedia.org/wiki/Category:Pages_with_templates_in_the_wrong_namespace) lists https://en.wikipedia.org/wiki/User:BilCat/twinkleoptions.js and https://en.wikipedia.org/wiki/User:Senator2029/twinkleoptions.js as in the wrong namespace, likely because they have templates in the source. The nowiki tags are written in 2 parts to avoid them being parsed at  [the gadget page itself](https://en.wikipedia.org/wiki/MediaWiki:Gadget-twinkleconfig.js)

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -1689,7 +1689,7 @@ Twinkle.config.writePrefs = function twinkleconfigWritePrefs(pageobj) {
 		"// changing the configuration parameters in a valid-JavaScript way) will be\n" +
 		"// overwritten the next time you click \"save\" in the Twinkle preferences\n" +
 		"// panel.  If modifying this file, make sure to use correct JavaScript.\n" +
-	    	"// <no" + "wiki>\n"+
+		"// <no" + "wiki>\n" +
 		"\n" +
 		"window.Twinkle.prefs = ";
 	text += JSON.stringify(newConfig, null, 2);
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -1689,12 +1689,14 @@ Twinkle.config.writePrefs = function twinkleconfigWritePrefs(pageobj) {
 		"// changing the configuration parameters in a valid-JavaScript way) will be\n" +
 		"// overwritten the next time you click \"save\" in the Twinkle preferences\n" +
 		"// panel.  If modifying this file, make sure to use correct JavaScript.\n" +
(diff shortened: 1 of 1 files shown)


modules/twinkleconfig.js
```diff
@@ -1689,12 +1689,14 @@ Twinkle.config.writePrefs = function twinkleconfigWritePrefs(pageobj) {
 		"// changing the configuration parameters in a valid-JavaScript way) will be\n" +
 		"// overwritten the next time you click \"save\" in the Twinkle preferences\n" +
 		"// panel.  If modifying this file, make sure to use correct JavaScript.\n" +
+	    	"// <no" + "wiki>\n"+
```
Amorymeltzer:
```suggestion
		"// <no" + "wiki>\n" +
```

Amorymeltzer: I always get confused by `<nowiki>` functionality, but doesn't the fact that the page is wrapped with `<nowiki>..</nowiki>` mean that the splitting is unnecessary?  Or does it break anyway?

DannyS712: It will break anyway, since, without splitting the tags, the gadget file would be parsed as <nowiki>AAA<nowiki>BBB</nowiki>CCC</nowiki>, which doesn't work as expected - a nowiki tag within another nowiki tag is ignore - they can't be nested. See https://en.wikipedia.org/wiki/User:DannyS712/sandbox#Nowiki_test for a demonstration

Amorymeltzer: Sure, in wikitext, but on a js contentModel page?  See https://en.wikipedia.org/wiki/User:Amorymeltzer/nowiki.js

DannyS712: Yes, the problem is _after_ the first nowiki tag is closed (sorry, in my previous edit the brackets didn't display, I've fixed that) the remaining content is no longer escaped - see https://en.wikipedia.org/wiki/User:DannyS712/sandbox.js.

Amorymeltzer: Right you are!  No wonder I get tripped up.  Totally lame!

Amorymeltzer:
fwiw I've dealt with the two mentioned above

PR #614, opened 2019-05-03
Config: Update inline comment

DannyS712:
Update comment about default unlink namespaces ([ 0, 100 ] -> [ 0, 10, 100, 118 ]) per actual defaults in twinkle.js


Amorymeltzer:
Changed in 2aae77d

Issue #613, opened 2019-04-30
Batch undelete mostly unsuccessful for large numbers of undeletions


MER-C:
It took me over 10 attempts to undelete the complete list of pages at https://en.wikipedia.org/w/index.php?title=User:MER-C/Sandbox&oldid=894909090 using the batch deletion tool. The failed undeletions gave the following error:

```Undeleting page Portal:Atlas/Maps/Historical/5: [XMilCwpAEMIAAJoUaqEAAABW] Caught exception of type Wikimedia\Rdbms\DBQueryError```

While this is a server-side performance issue, it can be mitigated by adding a little delay between the undeletions.

Amorymeltzer:
Also reported at [[[Special:Permalink/894839829#Batch undelete problem]]](https://en.wikipedia.org/wiki/Special:Permalink/894839829#Batch_undelete_problem)

Amorymeltzer:
I haven't had a ton of availability lately, but I'm looking into this.  I did a little rewrite of some stuff, which seems to help a bit, but not entirely.  One option could be to up `maxretries` but I think it may need work on the db end.

Amorymeltzer:
The changes in #616 are now live, and should alleviate much of this, so I'm going to close it.  I believe the underlying issue is at the database level, but until that's resolved this should hopefully suffice.

PR #616, opened 2019-05-03
batchundelete: Improvements to morebits.wiki.page, short- and long-term alleviation of #613, color like batchdelete

Amorymeltzer:
tl;dr — Add `undeletePage` to `Morebits.wiki.page`, use it in `twinklebatchundelete` (helps with #613), bump `maxRetries` (helps with #613), restructure talkpage undeletion (from #601), and add red color for fully create protected pages.

- 1st commit: morebits: Add undeletePage to morebits.wiki.page
Basically just copy the stuff for deletePage, less following redirects.  deletePage was added in 4b92869 along with move, but unlike protect (68132e5) delete and undelete are one-way streets.  Uses a dumb kludge to get around the lack of a proper token until we can rewrite the token handling stuff in `morebits` (#615).
- 2nd commit: batchundelete: convert to use Morebits.wiki.page.undeletePage
Added in prior commit (47c9cf9).  Some testing suggests this helps a bit with #613, but regardless, it's clearer this way anyway.
- 3rd commit: batchundelete: temporarily bump maxRetries up to 3 to avoid DB errors and alleviate #613
Increase from 2 to 3, makes `twinklebatchundelete` much more likely to succeed until phabricator.wikimedia.org/T222402 can be dealt with.  We've been handling `internal_api_error_DBQueryError` for deletion since #198, before even `batchOperation` was added (ecfff3b).
- 4th commit: batchundelete: Use same format in talkpage restoration as batchdelete
Follows up on #601.  Allows for a better, more clear message when listing deletions, as well as more full error handling and clearer code.  Easier to expand, should we desire.
- 5th commit: batchundelete: Add red color and confirmation for create-protected pages a la batchdelete

### DIFF
--- a/modules/twinklebatchundelete.js
+++ b/modules/twinklebatchundelete.js
@@ -138,6 +138,7 @@ Twinkle.batchundelete.callback.evaluate = function( event ) {
 		wikipedia_page.setCallbackParameters(params);
 		wikipedia_page.setEditSummary(reason + Twinkle.getPref('deletionSummaryAd'));
 		wikipedia_page.suppressProtectWarning();
+		wikipedia_page.setMaxRetries(3); // temporary increase from 2 to make batchundelete more likely to succeed [[phab:T222402]] #613
 		wikipedia_page.undeletePage(function onSuccess(apiobj) {
 			pageUndeleter.workerSuccess();
 			var talkpagename = new mw.Title(apiobj.query.title).getTalkPage().getPrefixedText();
--- a/modules/twinklebatchundelete.js
+++ b/modules/twinklebatchundelete.js
@@ -130,6 +130,7 @@ Twinkle.batchundelete.callback.evaluate = function( event ) {
 	pageUndeleter.run(function(pageName) {
 		var params = {
 			page: pageName,
+			undel_talk: undel_talk,
 			reason: reason,
 			pageUndeleter: pageUndeleter
 		};
@@ -139,15 +140,53 @@ Twinkle.batchundelete.callback.evaluate = function( event ) {
 		wikipedia_page.setEditSummary(reason + Twinkle.getPref('deletionSummaryAd'));
 		wikipedia_page.suppressProtectWarning();
 		wikipedia_page.setMaxRetries(3); // temporary increase from 2 to make batchundelete more likely to succeed [[phab:T222402]] #613
-		wikipedia_page.undeletePage(function onSuccess(apiobj) {
-			pageUndeleter.workerSuccess();
-			var talkpagename = new mw.Title(apiobj.query.title).getTalkPage().getPrefixedText();
-			if (apiobj.query.undel_talk && apiobj.query.title !== talkpagename) {
-				// Restore talk page too if it existed
-				new Morebits.wiki.api("Undeleting talk page", $.extend(apiobj.query, {title: talkpagename})).post();
-			}
-		}, pageUndeleter.workerFailure);
+		wikipedia_page.undeletePage(Twinkle.batchundelete.callbacks.doExtras, pageUndeleter.workerFailure);
 	});
+};
+
+Twinkle.batchundelete.callbacks = {
+	// this stupid parameter name is a temporary thing until I implement an overhaul
+	// of Morebits.wiki.* callback parameters
+	doExtras: function( thingWithParameters ) {
+		var params = thingWithParameters.parent ? thingWithParameters.parent.getCallbackParameters() :
+			thingWithParameters.getCallbackParameters();
+		// the initial batch operation's job is to delete the page, and that has
+		// succeeded by now
+		params.pageUndeleter.workerSuccess(thingWithParameters);
+
+		var query, wikipedia_api;
+
+		if ( params.undel_talk ) {
+			var talkpagename = new mw.Title(params.page).getTalkPage().getPrefixedText();
+			if (talkpagename !== params.page) {
+				query = {
+					'action': 'query',
+					'prop': 'deletedrevisions',
+					'drvprop': 'ids',
+					'drvlimit': 1,
+					'titles': talkpagename
+				};
+				wikipedia_api = new Morebits.wiki.api( 'Checking talk page for deleted revisions', query, Twinkle.batchundelete.callbacks.undeleteTalk );
+				wikipedia_api.params = params;
+				wikipedia_api.params.talkPage = talkpagename;
+				wikipedia_api.post();
+			}
+		}
+	},
+	undeleteTalk: function( apiobj ) {
+		var xml = apiobj.responseXML;
+		var exists = $(xml).find('page:not([missing])').length > 0;
+		var delrevs = $(xml).find('rev').attr('revid');
+
+		if( exists || !delrevs ) {
+			// page exists or has no deleted revisions; forget about it
+			return;
+		}
+
+		var page = new Morebits.wiki.page(apiobj.params.talkPage, "Undeleting talk page of " + apiobj.params.page);
+		page.setEditSummary("Undeleting [[Help:Talk page|talk page]] of \"" + apiobj.params.page + "\"" + Twinkle.getPref('deletionSummaryAd'));
+		page.undeletePage();
+	}
 };
 
 })(jQuery);
--- a/modules/twinklebatchundelete.js
+++ b/modules/twinklebatchundelete.js
@@ -122,30 +122,30 @@ Twinkle.batchundelete.callback.evaluate = function( event ) {
 		return;
 	}
 
-	var batchOperation = new Morebits.batchOperation("Undeleting pages");
-	batchOperation.setOption("chunkSize", Twinkle.getPref('batchUndeleteChunks'));
-	batchOperation.setOption("preserveIndividualStatusLines", true);
-	batchOperation.setPageList(pages);
-	batchOperation.run(function(pageName) {
-		var query = {
-			'token': mw.user.tokens.get().editToken,
-			'title': pageName,
-			'action': 'undelete',
-			'reason': reason + Twinkle.getPref('deletionSummaryAd'),
-			'undel_talk': undel_talk	// Not used in the API request, but added here for access in onSuccess()
+
+	var pageUndeleter = new Morebits.batchOperation("Undeleting pages");
+	pageUndeleter.setOption("chunkSize", Twinkle.getPref('batchUndeleteChunks'));
+	pageUndeleter.setOption("preserveIndividualStatusLines", true);
+	pageUndeleter.setPageList(pages);
+	pageUndeleter.run(function(pageName) {
+		var params = {
+			page: pageName,
+			reason: reason,
+			pageUndeleter: pageUndeleter
 		};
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
cc @MusikAnimal I mentioned this at https://github.com/azatoth/twinkle/pull/601#issuecomment-489143056 to @siddharthvp but I'd like to merge this sooner rather than later to get deal with `batchundelete`.  I did a bunch of testing and this all seems to work well, but it ended up being a fair bit of stuff.  Regardless, I'll probably just bump `maxRetries` on-wiki later tonight/tomorrow when I have a moment, which should handle things temporarily anyway.

morebits.js
```diff
@@ -2406,6 +2412,44 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+	 * Undeletes a page (for admins only)
+	 * @param {Function} onSuccess - callback function to run on success
+	 * @param {Function} [onFailure] - callback function to run on failure (optional)
+	 */
+	this.undeletePage = function(onSuccess, onFailure) {
+		ctx.onUndeleteSuccess = onSuccess;
+		ctx.onUndeleteFailure = onFailure || emptyFunction;
+
+		// if a non-admin tries to do this, don't bother
+		if (!Morebits.userIsInGroup('sysop')) {
```
siddharthvp: This would unnecessarily slow down the code in batchundelete. Better to put this in `fnProcessUndeleteError`. No decently written script would show an undelete button to non-admins anyway.

This would apply to deletePage function also.

modules/twinklebatchundelete.js
```diff
@@ -122,33 +144,71 @@ Twinkle.batchundelete.callback.evaluate = function( event ) {
...
+			thingWithParameters.getCallbackParameters();
+		// the initial batch operation's job is to delete the page, and that has
+		// succeeded by now
+		params.pageUndeleter.workerSuccess(thingWithParameters);
+
+		var query, wikipedia_api;
+
+		if ( params.undel_talk ) {
+			var talkpagename = new mw.Title(params.page).getTalkPage().getPrefixedText();
+			if (talkpagename) {
```
siddharthvp:
```suggestion
			if (talkpagename !== params.page) {
```
This is supposed to check that we're not already dealing with a talk page. `getTalkPage()` for a talk page returns the same page.

modules/twinklebatchundelete.js
```diff
@@ -122,33 +144,71 @@ Twinkle.batchundelete.callback.evaluate = function( event ) {
...
-			'undel_talk': undel_talk	// Not used in the API request, but added here for access in onSuccess()
+
+	var pageUndeleter = new Morebits.batchOperation("Undeleting pages");
+	pageUndeleter.setOption("chunkSize", Twinkle.getPref('batchUndeleteChunks'));
+	pageUndeleter.setOption("preserveIndividualStatusLines", true);
+	pageUndeleter.setPageList(pages);
+	pageUndeleter.run(function(pageName) {
+		var params = {
+			page: pageName,
+			undel_talk: undel_talk,	// Not used in the API request, but added here for access in onSuccess()
```
siddharthvp:
```suggestion
			undel_talk: undel_talk,
```
No need for that comment now, as this is now in `params`, not in `query`

morebits.js
```diff
@@ -1737,7 +1739,7 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		botEdit: false,
 		pageSection: null,
 		maxConflictRetries: 2,
-		maxRetries: 2,
+		maxRetries: 3,	// temporary increase from 2 to make batchundelete more likely to succeed [[phab:T222402]] #613
```
siddharthvp: Isn't it also okay to use `wikipedia_page.setMaxConflictRetries(3)` in batchundelete.js, rather than changing the default for every object?

morebits.js
```diff
@@ -2879,36 +2923,121 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+			if (editprot.length > 0 && editprot.attr('level') === 'sysop' && !ctx.suppressProtectWarning &&
+				!confirm('You are about to undelete the fully create protected page "' + ctx.pageName +
+				(editprot.attr('expiry') === 'infinity' ? '" (protected indefinitely)' : ('" (protection expiring ' + editprot.attr('expiry') + ')')) +
+				'.  \n\nClick OK to proceed with the undeletion, or Cancel to skip this undeletion.')) {
+				ctx.statusElement.error("Undeletion of fully create protected page was aborted.");
+				ctx.onUndeleteFailure(this);
+				return;
+			}
+
+			// Begin shitty kludge
```
```suggestion
			// KLUGE:
```
Let's use standard commenting conventions rather than resorting to obscenity!

morebits.js
```diff
@@ -2879,36 +2923,121 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+	var fnProcessUndelete = function() {
+		var pageTitle, token;
+
+		// The whole handling of tokens in Morebits is outdated (#615)
+		// but has generally worked since intoken has been deprecated
+		// but remains.  intoken does not, however, take undelete.
+		// Everything except watching and patrolling should eventually
+		// use csrf, but until then (#615) the stupid hack below
+		// should work for undeletion
+		if (fnCanUseMwUserToken('undelete')) {
```
siddharthvp: I don't understand, `token = mw.user.tokens.get('editToken');` is being used whether or not this condition is true. Why have this condition at all?

modules/twinklebatchundelete.js
```diff
@@ -70,7 +72,20 @@ Twinkle.batchundelete.callback = function twinklebatchundeleteCallback() {
 			$pages.each(function(index, page) {
 				var $page = $(page);
 				var title = $page.attr('title');
-				list.push({ label: title, value: title, checked: true });
+				var $editprot = $page.find('pr[type="create"][level="sysop"]');
+				var isProtected = $editprot.length > 0;
+
+				var metadata = [];
```
siddharthvp: The array is not needed as there's just one piece of metadata, and we're unlikely to add more in the future.

siddharthvp:
I'm not sure if querying for deleted revisions of talk page, and then undeleting if they exist is a good idea, as this is a lot of extra API calls. What about doing it like this: store the list of pages whose undeletion was successful, in an array, and then use a separate batch operation to restore the talks. For this batch-op, unset the `preserveIndividualStatusLines` property, so that the status lines for these pages don't show up at all. We could also probably display just the number of talk pages undeleted. 

How is this handled in batchdelete? Is a status line shown only for the main pages, or for talk pages and redirects too?

morebits.js
Amorymeltzer: Heh, thanks — leftover from testing, missed it in the diffs!  Sorry https://twitter.com/gitlost

morebits.js
Amorymeltzer: This is what I was trying to convey in the comment: until I can radically modernize the handling of tokens, I wanted to keep the code structure parallel.  I think it's easier to understand and fix/remove/rewrite if as much as possible follows a similar, understandable pathway rather than a bespoke system for each.  Yes, we can remove this `fnCanUseMwUserToken`, I suppose that'd be fine given the context of the comment.

morebits.js
Amorymeltzer: `setMaxRetries`, but yes, I take your point.  Doing it this way shouldn't hammer the API elsewhere (nothing *else* is failing AFAIK) and, more importantly, is something I can do on-wiki to alleviate the issue without needing code review.

morebits.js
Amorymeltzer: Isn't this backwards?  Failing here means non-sysop don't do the later protection query.  At any rate, it's quite quick as is (although admittedly slowing things down isn't bad in our current situation).

modules/twinklebatchundelete.js
Amorymeltzer: Good point!  I dunno why I thought `mw.Title` would fail...

modules/twinklebatchundelete.js
Amorymeltzer: I did it to be parallel with `batchdelete`, although I do agree we're unlikely to add more.  I don't think size or user is likely to be desired.

morebits.js
siddharthvp: To clarify, this would slow down the code _for sysops_. Yes, it would speed up things a great deal for non-sysops but what good is that since they won't be able to undelete anything after all?

Amorymeltzer:
`batchdelete` shows the status lines for each talk; I think it's helpful there (see also #581) and even moreso here, as undeleting talkpage is potentially unexpected.  More to the point, a talkpage could exist *and* have deleted revisions, which might be a significant surprise.  `batchdelete` does indeed do a query for each talkpage, to confirm that it exists and to quietly skip the page if it doesn't (rather than fail loudly).

As I'm writing this, I'm realizing that the check for `missing` in `fnProcessUndelete` is, as here, insufficient, and will need to be rewritten.

morebits.js
Amorymeltzer: Oh you know, nevermind; rewriting `batchundelete` to use `morebits.wiki.page.undeletePage` is what made `maxRetries` even relevant in the first place, so my point is dumb and you're right!

morebits.js
Amorymeltzer: I imagine the impetus was to allow for the "retrieving undelete token..." message

morebits.js
Amorymeltzer: Thought I [dealt](https://github.com/azatoth/twinkle/compare/f4b189e7d85639f50c6d9f40ef5c2b51d06b3319..4f85861eda659e8f8bb1be510b61003552189ee7) with this, but that meant morebits failed, so I've [undone](https://github.com/azatoth/twinkle/compare/2a5e111dc294abc635f4ebad5d9f6a497ce69a2b..a738d5b30e32ba1115013b86f5158652dd24c205) it for now as I don't have any more time today.

siddharthvp:
> I edited `batchundelete` to check for both `missing` and deleted revisions.

Thanks, that makes a lot of sense and also makes my point moot, as my suggested method also would have ended up restoring deleted edits at existing talk pages.

morebits.js
Amorymeltzer: Oh duh!  If `suppressProtectWarning` isn't set, the pages fail the "poor man's normalisation" in `fnCanUseMwUserToken`.  `suppressProtectWarning` is set for the batch but not for talkpages so the kludge was only reached (in this case) for talk pages.  This matches the behavior in `twinklebatchdelete`, and is required; since the batches check, display, and confirm on protection status for the pages listed (well, *now* `batchundelete` does) it's not needed for those, but the confirms(s) are needed for any talkpages found later.  Sorry for leading you on @siddharthvp, I knew this a few days ago but `Morebits` gets confusing!

As far as `fnCanUseMwUserToken` goes, `followRedirect` doesn't really make sense for `undeletePage` so the `suppressProtectWarning` is the real killer.  The real goal of that whole thing seems to be to avoid another API call to get the protection status if `suppressProtectWarning` isn't set, none of that is relevant for batch-processing like this.  I'm inclined to leave it as is for now on the off chance we decide to use it elsewhere before revamping `morebits`, as well as, as mentioned elsewhere, making what will surely be an revamping process easier.

PR #619, opened 2019-05-05
xfd: Use indexOf instead of startsWith, the latter isn't supported in IE 11

Amorymeltzer:
Added in #459.  Haven't seen any reports of issues, but if we're trying to maintain support for IE 11...

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -93,7 +93,7 @@ Twinkle.xfd.callback = function twinklexfdCallback() {
 	categories.append( {
 			type: 'option',
 			label: 'MfD (Miscellany for deletion)',
-			selected: [ 0, 6, 10, 14, 828 ].indexOf(namespace) === -1 || Morebits.pageNameNorm.startsWith("Template:User "),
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
cc @siddharthvp fwiw

siddharthvp:
Ain't it better to shim these useful functions (startsWith, endsWith, includes) rather than use outdated and poorly readable alternatives? We recently removed some useless shims from morebits, we might as well add some useful ones.

Amorymeltzer:
Maybe, but `includes` is the only one that really comes up frequently it seems. Here and #584 are AFAICS the only incidences of getting around `starts` or `ends`, so that seems unneeded.

On `includes`, we do have quite a few `indexOf`... `!== -1` though, which you mentioned dealing with in https://github.com/azatoth/twinkle/pull/487#discussion_r245616438.  The [string polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Polyfill) looks considerably simpler than the [array polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Polyfill).

Issue #94, opened 2012-04-10
Remove or modify alerts when trying to invoke modules on one's own talk page


atlight:
These humorous alert dialogs (Ioeth's brainchild) prevent users from practising with these tools, unless/until they come across the sandbox for user warnings. A section has been added to the documentation, but that is not read by most people.

The Welcome module has had this alert replaced by a confirm. The humour is lost, but this module can actually now be used on one's own talk page (for testing only, obviously).

PR #622, opened 2019-05-06
ARV: Remove completely from user's own pages

DannyS712:
Rather than showing the option, and then asking the user if they want to report themselves, and then closing (not actually letting users report themselves), the option should be hidden entirely.

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -15,7 +15,7 @@
 
 Twinkle.arv = function twinklearv() {
 	var username = mw.config.get('wgRelevantUserName');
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I would've thought we'd go the other way, since everything else has a confirmation before proceeding, but @atlight brought that very point up in #94.  The difference being, of course, that faffing about with warning or talkbacking yourself doesn't affect anyone, but an arv report would.

DannyS712:
I completely agree with atlight's point in 94, but its misleading to have the option visible but not functional, so it should be removed

PR #623, opened 2019-05-06
Image: Link "speedy deletion policy" directly to #Files

DannyS712:
Since the DI tab is only available on files, and only includes speedy deletion criteria applicable to files, the link should point to the specific section of the speedy deletion policy.


Amorymeltzer:
Seems reasonable; unchanged since ff3f96b

PR #625, opened 2019-05-08
tag: fix bug when processing items with type=div

### LINKED ISSUE #624: {{R from alternative language}} tag broken
Currently, when I try to add {{R from alternative language}} tags to redirects Twinkle silently fails. It doesn't appear to matter whether I enter language codes or fill in other parameters. I am still able to use other tags, even after attempting to add the {{R from alternative language tag}}

System specs: Mac OS 10.14.4, bug is present on  the latest versions of Safari and Firefox.

Amorymeltzer:
Closes #624.  {{R from alternative name}} has a div element, and the processing schema introduced in #487 dies on it

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1541,7 +1541,7 @@ Twinkle.tag.callback.evaluate = function friendlytagCallbackEvaluate(e) {
 	// name attribute as 'articleTags.' + name of the subgroup element
 
 	var name_prefix = Twinkle.tag.mode + 'Tags.';
-	$(form).find("[name^='" + name_prefix + "']").each(function(idx,el) {
+	$(form).find("[name^='" + name_prefix + "']:not(div)").each(function(idx,el) {
 		// el are the HTMLInputElements, el.name gives the name attribute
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
@siddharthvp I'm sure you can cook up a smarter way to deal with this but I'm tired and heading to bed in a moment so this should suffice for the time being.

modules/friendlytag.js
```diff
@@ -1541,7 +1541,7 @@ Twinkle.tag.callback.evaluate = function friendlytagCallbackEvaluate(e) {
 	// name attribute as 'articleTags.' + name of the subgroup element
 
 	var name_prefix = Twinkle.tag.mode + 'Tags.';
-	$(form).find("[name^='" + name_prefix + "']").each(function(idx,el) {
+	$(form).find("[name^='" + name_prefix + "']:not(div)").each(function(idx,el) {
```
siddharthvp:
```suggestion
	$(form).find("input[name^='" + name_prefix + "']").each(function(idx,el) {
```
Slightly cleaner.

PR #626, opened 2019-05-09
warn: Add vanarticle parameter to user talk links on diffs

Amorymeltzer:
Allows user to click to an editor involved in a diff and provide a warning with the article title filled-in, without having to revert the diff themself (e.g., already-reverted vandalism that went unwarned).  This has been requested a few times, and is fairly trivial.  The link is bolded, just like the user talk link on the mw rollback done page.  Clicking this talkpage link will NOT trigger the autowarn feature (#509) even if enabled (the list of && statements in the main `if` is getting quite long).

As noted in #562, `wgDiffNewId` appears to be sufficient for determining if we're on a diff, but having `wgDiffOldId` makes things clearer.

Supersedes #589

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -9,14 +9,16 @@
  *** twinklewarn.js: Warn module
  ****************************************
  * Mode of invocation:     Tab ("Warn")
- * Active on:              User (talk) pages and (deleted) Contributions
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Just to close the loop, doing this here rather than #589 makes #632 simpler.

Issue #605, opened 2019-04-27
Twinkle shouldn't show rollback links for pages you can't edit


pppery:

Amorymeltzer:
`diff` and `oldid` should be straightforward — being cognizant of #589 — but getting the protection status for every page on `contributions` and then checking it against against the user's perms would be expensive.

Amorymeltzer:
A bit of a headache, but I've got a better way.  Should have a pr soon.

PR #630, opened 2019-05-11
fluff: don't create rollback links for single-revision pages on contribs

Amorymeltzer:
Possibly a result from the fix in #467, but either way, this would create rollback/vandal links for every page marked with `mw-uctop`, even if it had only a single revision and thus couldn't be reverted.  The rollback links would fail (using & instead of ?) but wouldn't have been properly constructed anyway.  Check for the existence of a `href` here should remove those links.


Amorymeltzer:
I'm going to merge this since it's a potentially significant and obvious user-facing bug, but there are probably better ways of doing this, especially as concerns #605.  Indeed, though not protection-related, this does solve one part of #605.

PR #632, opened 2019-05-12
fluff: Only add links if the user has necessary perms to edit the page

### LINKED ISSUE #605: Twinkle shouldn't show rollback links for pages you can't edit

Amorymeltzer:
Closes #605.  Basically, we need a way to check for ability to edit the covers both explicit (protection) and implicit (namespace and content model) restrictions.

Since we're on the page in question, `diff` and `oldid` (#562) can be handled together with mostly-accurate results by simply checking `wgIsProbablyEditable`.  It won't handle cascading protection or TitleBlacklist restrictions, but neither does MediaWiki when choosing to display either "edit this page" or "view source"; being more accurate than the software would require an API check a la contributions (see following section as well as https://gerrit.wikimedia.org/r/c/mediawiki/core/+/65009).

`contributions` is the more <s>complex</s> interesting implementation, since we've got to check each and every page.  Querying for `edit` in `intestactions` is excellent, because it covers all bases (i.e., it includes the "expensive" checks that `wgIsProbablyEditable` skips, hence `Probably`).  I also restructured/split the contributions routine to avoid doing `async` and to look more Twinkley.

`intestactions` was added in MW 1.25 (https://www.mediawiki.org/wiki/MediaWiki_1.25) circa 2014-2015, and should probably be used more frequently by Twinkle, particularly in `Morebits`, as it's a more correct determiner of whether the user can edit (or move or protect or...) the page in question.

The first commit (remove redundant checks; better way of doing #630 aka e0f3d9c) is worth doing regardless.

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -49,13 +49,13 @@ Twinkle.fluff = {
 		// $('sp-contributions-footer-anon-range') relies on the fmbox
 		// id in [[MediaWiki:Sp-contributions-footer-anon-range]] and
 		// is used to show rollback/vandalism links for IP ranges
-		if( mw.config.get('wgCanonicalSpecialPageName') === "Contributions" && (mw.config.exists('wgRelevantUserName') || !!$('#sp-contributions-footer-anon-range')[0])) {
+		if( mw.config.exists('wgRelevantUserName') || !!$('#sp-contributions-footer-anon-range')[0]) {
 			//Get the username these contributions are for
 			var username = mw.config.get('wgRelevantUserName');
 			if( Twinkle.getPref('showRollbackLinks').indexOf('contribs') !== -1 ||
 				( mw.config.get('wgUserName') !== username && Twinkle.getPref('showRollbackLinks').indexOf('others') !== -1 ) ||
 				( mw.config.get('wgUserName') === username && Twinkle.getPref('showRollbackLinks').indexOf('mine') !== -1 ) ) {
-				var list = $("#mw-content-text").find("ul li:has(span.mw-uctop)");
+				var list = $("#mw-content-text").find("ul li:has(span.mw-uctop):has(.mw-changeslist-diff)");
 
 				var revNode = document.createElement('strong');
 				var revLink = Twinkle.fluff.buildLink('SteelBlue', 'rollback');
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
@MusikAnimal, Pinging you since you had thoughts on #562 aka #511.  Despite delaying loading of links on Special:Contributions, I think it'd be a good addition.  If you think it'd be worth it, we could add a query check before loading on diffs/oldid to be more accurate than the software, but diffs and old revisions can take long enough to load and the edge cases (cascading, TitleBlacklist) are comparatively infrequent.

cc @pppery

modules/twinklefluff.js
```diff
@@ -45,40 +45,61 @@ Twinkle.fluff = {
...
+					var titles = $(list).find(".mw-changeslist-date").map(function(){return $(this).attr("title");}).get();
+
+					var query = {
+						'action': 'query',
+						'prop': 'info',
+						'intestactions': 'edit',
+						'titles': titles.join('|').replace(/ /g,'_')
+					};
+
+					var wikipedia_api = new Morebits.wiki.api('Checking for prior AfDs', query, Twinkle.fluff.contributions.main);
```
pppery: That is not at all what this API request is doing

Amorymeltzer: Ahahaha good catch.  Will fix when I get home.

siddharthvp:
@Amorymeltzer I should probably have said earlier but I think this change was quite unnecessary. Can we revert this please? It slows down what I think is an already slow process of loading of rollback links on contribs pages. It would somewhat affect the work of RCP folks who are trying to revert vandalism _fast_. 

Also there is this obvious bug that no rollback links show up on pages with >50 current revs, as `intestactions` supports only upto 50 titles. I imagine it would be possible to skip the tests in such a case, but it'd be quite tricky to code. 

Easier option is to revert, as I don't see any major advantage out of this, that accounts for the additional delay in loading.

Or if this is that important, a better way to do this would be to show all rollback links at first, then silently do the API call and remove the rollback links on non-editable pages.

Amorymeltzer:
*sigh* All fair points.

>It slows down what I think is an already slow process of loading of rollback links on contribs pages. It would somewhat affect the work of RCP folks who are trying to revert vandalism fast.

This is what I was most worried about, but I don't think it's that much of a slowdown.  My (admittedly inadequate) tests suggested it was noticeable but not significant, and split about halfway between the query and the `each`.  I've got an old machine with a ton of user-loaded js so I generally find if it's not slow for me then it won't be for most users.  Do you mind timing it out?  I'm curious what you're seeing.  Obviously this matters more for folks with larger contribs defaults (see below) but I'm not sure how frequent that is (although I'm one of them).  Regardless, for faster machines/less weighed-down users, I guess the `.each` is quicker so the query would be the limiting step?

>no rollback links show up on pages with >50 current revs, as intestactions supports only upto 50 titles

Yeah, no idea how I missed that.  I tested on long contribs with my non-sysop alt, but I'm guessing my test users just didn't have enough unique pages edited at the time.  This for me is a total dealbreaker, so I've partially reverted this (that is, just the contribs bit) on-wiki, where it was [already raised on-wiki](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Checking_ability_to_edit_each_page%3A_Too_many_values_supplied_for_parameter_%22titles%22._The_limit_is_50) while we discuss.  It *should* be possible to parse and break down the titles into discrete batches, but that's only going to make the above issue much worse.

>show all rollback links at first, then silently do the API call and remove the rollback links on non-editable pages.

I really don't like lying to folks by showing these links, but I think that's worse.  If it's noticeable enough to be an issue on their loading, users will certainly notice them leaving.  Best-case scenario it's confusing, worst-case is zOMG I iz haxed.


Anyway, I think the best thing to do is basically what I did on-wiki: revert the contributions changes, leaving in the check on diffs.  I think we can be smarter with `auto()`, covering it with the same `wgIsProbablyEditable` check, but that might be it.

siddharthvp:
> I really don't like lying to folks by showing these links, but I think that's worse. If it's noticeable enough to be an issue on their loading, users will certainly notice them leaving. Best-case scenario it's confusing, worst-case is zOMG I iz haxed.

Disappearing of links will be noticeable, but I think that's ok. We can split the pagelist to batches of 50 (the limit is 50 for sysops too from what I read) and do the api calls after the links have been loaded. I really think this is the way to go.

I don't think this "lying" (for a few seconds, after we do the above) is an issue. Twinkle has been an extensively popular tool for reverting vandalism for 12 years, but how many times has anyone complained about this? Never, right?

Amorymeltzer:
@siddharthvp I just merged #658, which undoes the relevant bit as you requested (also covers the case of `auto()` which I'd missed).

I'd be curious to see how it looks broken into batches, but I don't think it's urgent.  I still think placing then removing links would be worse than the current/original situation.

The interface knows which pages are editable, or at least which pages are rollbackable, of which editable is one facet.  I don't think it's explicitly exposed, though, but it'd be nice if it were.

PR #633, opened 2019-05-12
protect: re-enable check for flaggedrevs/pending changes being enabled

Amorymeltzer:
As noted in #629, the [MediaWiki 1.34-wmf4](https://www.mediawiki.org/wiki/MediaWiki_1.34/wmf.4) unceremoniously removed `wgFlaggedRevsParams` (see https://phabricator.wikimedia.org/T219342#5116724 and https://gerrit.wikimedia.org/r/c/mediawiki/extensions/FlaggedRevs/+/508427), meaning that the changes in #551 meant the pending changes dropdown would *never* be shown.  This re-enables those changes using a check for the gadget extension, since it's what we've got (https://phabricator.wikimedia.org/T218479 would be helpful...).  Or, as git would say:

Revert "Revert "protect: Don't fail if flaggedrevs/pending changes aren't enabled"" (this reverts commit e6fd37bfc0a3b3faa976df84a274881dbba3051d)

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -123,6 +123,10 @@ Twinkle.protect.fetchProtectingAdmin = function twinkleprotectFetchProtectingAdm
 	});
 };
 
+// mw.loader.getState('ext.flaggedRevs.review') returns null if the
+// FlaggedRevs extension is not registered.  Previously, this was done with
+// wgFlaggedRevsParams, but after 1.34-wmf4 it is no longer exported if empty
+// (https://gerrit.wikimedia.org/r/c/mediawiki/extensions/FlaggedRevs/+/508427)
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
@MusikAnimal This was recommended to me by @Krinkle as I'm not very well versed in extension/`mw.loader` stuff, but some quick testing has this working — after all, it's just replacing one check with another (compare to the file in #551).  It's selfish, mostly useful atm to avoid breaking on testwiki, but could matter down the line for other wikis.  Still, I'm not sure if checking extensions is a road we want to go down.  Also of note is that, after #580, this would make only the second `mw.loader` in all of Twinkle.

Krinkle:
I have not reviewed the new code or whether it correctly restores the old logic.

But, I can confirm, the use of `mw.loader` here is correct 👍

PR #636, opened 2019-05-15
Speedy: Use fullurl when linking to file logs on CSD log

DannyS712:
No need to write out the full url and place multiple calls to get the server and log url when a parser function does the trick.

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1634,7 +1634,7 @@ Twinkle.speedy.callbacks = {
 			var extraInfo = '';
 
 			// If a logged file is deleted but exists on commons, the wikilink will be blue, so provide a link to the log
-			var fileLogLink = mw.config.get('wgNamespaceNumber') === 6 ? ' ([' + mw.config.get('wgServer') + mw.util.getUrl('Special:Log', {action: 'view', page: mw.config.get('wgPageName')}) + ' log])' : '';
+			var fileLogLink = mw.config.get('wgNamespaceNumber') === 6 ? ' ([{{fullurl:Special:Log|page=' + mw.config.get('wgPageName') + '}} log])' : '';
 
 			var editsummary = "Logging speedy deletion nomination";
 			appendText += "\n# [[:" + Morebits.pageNameNorm;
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I'll agree that `fullurl:` is probably more clear to someone editing the wikitext, but this isn't really a speed gain.  The `mw.config` and `mw.util` objects/functions are exported on every page, and while it may technically be slower than a simple string assignment, the difference would be dwarfed by comparison to even how long it would take you to click submit or for Twinkle to display even a single message.  Indeed, once the log is added, it's probably more expensive for the software to compute `fullurl` every time the csd log is loaded compared to using the direct link.  Neither should be a major factor here.

DannyS712:
@Amorymeltzer the primary motivation wasn't the speed, but rather the wikitext - its much harder to read through the source when there are long urls hard-coded in.

Amorymeltzer:
Yes, that's what I meant by
>I'll agree that `fullurl:` is probably more clear to someone editing the wikitext

modules/twinklespeedy.js
```diff
@@ -1634,7 +1634,7 @@ Twinkle.speedy.callbacks = {
 			var extraInfo = '';
 
 			// If a logged file is deleted but exists on commons, the wikilink will be blue, so provide a link to the log
-			var fileLogLink = mw.config.get('wgNamespaceNumber') === 6 ? ' ([' + mw.config.get('wgServer') + mw.util.getUrl('Special:Log', {action: 'view', page: mw.config.get('wgPageName')}) + ' log])' : '';
+			var fileLogLink = mw.config.get('wgNamespaceNumber') === 6 ? ' ([{{fullurl:Special:Log|page=' + mw.config.get('wgPageName') + '}} log])' : '';
```
Amorymeltzer: Needs to properly encode the page title for the URI.

```suggestion
			var fileLogLink = mw.config.get('wgNamespaceNumber') === 6 ? ' ([{{fullurl:Special:Log|page=' + mw.util.wikiUrlencode (mw.config.get('wgPageName')) + '}} log])' : '';
```

`encodeURIComponent` would work too, but using `mw.util` keeps the colon in the namespace intact.

PR #637, opened 2019-05-15
prod: Detect prior XfDs using templates on talk page rather than just existence of AfD discussion

Amorymeltzer:
Suggested by @siddharthvp.  Old XfD tags are reliably added upon closure of the XfD, whether automatically by XFDCloser or manually (certainly much more so than {{Old PROD}} ever was).  Different content may exist at the same title, so if the movie [[My Cousin Vinny]] is deleted at AfD and then later someone writes [[My Cousin Vinny]] about a family member, the mere existence of an AfD using that title shouldn't prevent the recreation from being PRODed.  It's a rare edge case, but does occur (https://en.wikipedia.org/w/index.php?oldid=896314665#Twinkle_will_now_try_to_detect_Template:Old_prod_and_old_AfDs,_and_will_place_the_talk_page_template_automatically) and is possibly the rare case where the user-managed template is more accurate.

Using the `templates` prop allows us to look for transcluded templates, which avoids the problem of template redirects.  Converting to this system also has the advantage of also allowing us to easily check for old file and bookprods, the former of which would have been impossible as FfD doesn't use individual subpages like AfD and MfD do.  {{Old AfD}} uses {{Old XfD multi}} while {{Old MfD}} and {{Oldffdfull}} are each their own templates; as of now, all XfD processes use {{Old XfD multi}} to handle multiple old nominations.  {{Oldpuffull}} and {{Multidel}} have also been included, as they are currently being used in PROD-able namespaces, as well as {{Olddrvfull}} and {{Olddelrev}}, which on their own would still prevent a PROD nomination.  {{Old VfD}} has no transclusions.

{{Article history}} can also handle old deletion discussions, but is not something we can use in this fashion (and is comparatively rare).  One option would be to include it but fall back to the api method this PR replaces if it is the only template found.  Nearly all uses are either article talk or portal talk, so we could check AfD and MfD.

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -180,31 +180,14 @@ Twinkle.prod.callbacks = {
 		var statelem = apiobj.statelem;
 		var params = apiobj.params;
 
-		// Check existence of AfD, would be easier if mw.Title.exists() worked [[phab:T184953]]
-		var exists;
-		if (namespace === 'article') {
-			var query = {
-				'action': 'query',
-				'prop': 'revisions',
-				'titles': 'Wikipedia:Articles for deletion/' + Morebits.pageNameNorm
-			};
-			var wikipedia_api = new Morebits.wiki.api('Checking for prior AfDs', query, function(apiobj) {
-				var xmlDoc = apiobj.responseXML;
-				exists = $(xmlDoc).find('rev').attr('revid');
-			});
-			// Wait for API call to finish
-			wikipedia_api.post({
-				async: false
-			});
-		}
-		if (exists) {
-			statelem.warn( 'Previous AfD for this page title found, aborting procedure' );
+		// Check talk page for templates indicating prior XfD
+		if($(xmlDoc).find('templates').length) {
+			statelem.warn( 'Previous XfD template found on talk page, aborting procedure' );
 			return;
 		}
 
 		// Check talk page for category indicating prior PROD
-		var cats = $(xmlDoc).find('categories');
-		if(cats.length) {
+		if($(xmlDoc).find('categories').length) {
 			if(params.blp) {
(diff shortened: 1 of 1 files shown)


modules/twinkleprod.js
```diff
@@ -445,14 +428,19 @@ Twinkle.prod.callback.evaluate = function twinkleprodCallbackEvaluate(e) {
 	Morebits.status.init( form );
 
 	var talk_title = new mw.Title(mw.config.get('wgPageName')).getTalkPage().getPrefixedText();
+	// Talk page templates for PROD-able discussions
+	var blocking_templates = 'Template:Old XfD multi|Template:Old MfD|Template:Oldffdfull|' + // Common prior XfD talk page templates
```
siddharthvp:
```suggestion
	var blocking_templates = 'Template:Old XfD multi|Template:Old AfD|Template:Old MfD|Template:Oldffdfull|' + // Common prior XfD talk page templates
```
lol, it looks like you forgot the most important one.

Amorymeltzer: That's the neat part — {{Old AfD}} just transcludes a single instance of {{Old AfD multi}}, which is a redirect to {{Old XfD multi}}, so there's no need to explicitly include it.  Check out [this query](https://en.wikipedia.org/wiki/Special:ApiSandbox#action=query&format=json&prop=templates&titles=Talk%3AOrder%2066%20(Star%20Wars)&tltemplates=Template%3AOld%20XfD%20multi%7CTemplate%3AOld%20MfD).

siddharthvp: Oh ok, didn't notice {{old AfD}} was implemented like that, though the other {{old [x]fd}} templates aren't.

PR #639, opened 2019-05-17
xfd: check "Wrap deletion tag with <noinclude>" if page carries {{sub…

siddharthvp:
Check "Wrap deletion tag with <noinclude>" by default if template page (or its documentation) carries {{[subst only](https://en.wikipedia.org/wiki/Template:Subst_only)}}.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -320,7 +320,8 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 							value: 'noinclude',
 							name: 'noinclude',
 							tooltip: 'Will wrap the deletion tag in &lt;noinclude&gt; tags, so that it won\'t get substituted along with the template.',
-							disabled: templateOrModule === 'module'
+							disabled: templateOrModule === 'module',
+							checked: !!$('.box-Subst_only').length // Default to checked if page carries {{subst only}} 
 						}
 					]
 			} );


siddharthvp:
Note: This will only work after the template has been [edited](https://en.wikipedia.org/w/index.php?title=Template:Subst_only&action=edit) to add `|name = Subst only`. I can't do it coz of TE-protection.

JJMC89:
> template has been edited to add `|name = Subst only`

Done

modules/twinklexfd.js
```diff
@@ -320,7 +320,8 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 							value: 'noinclude',
 							name: 'noinclude',
 							tooltip: 'Will wrap the deletion tag in &lt;noinclude&gt; tags, so that it won\'t get substituted along with the template.',
-							disabled: templateOrModule === 'module'
+							disabled: templateOrModule === 'module',
+							checked: !!$('.box-Subst_only').length // Default to checked if page carries {{subst only}}
```
Amorymeltzer: I haven't thought too deeply about it, but why is `!!` necessary, does `$('.box-Subst_only').length` not work?

siddharthvp: Type-correctness. It'd work without `!!` too.

PR #640, opened 2019-05-20
Protect: Hide on mediawiki pages

DannyS712:
Can't be protected or unprotected (neither can gadget / gadget description pages, but those are blocked by the title blacklist) - doesn't need to be shown.


Amorymeltzer:
Not inherently opposed, but they technically *can* be protected.  Clearly it never happens — completely unused atm on enwiki — but couldn't this be used to prevent interface editors from mucking about?  And, from a philosophical point of view, mediawiki pages could still be fully-protected during an edit war.

DannyS712:
@Amorymeltzer actually, interface editors also have `editprotected`, so this wouldn't stop them.

PR #641, opened 2019-05-24
warn: use jquery.chosen on select box

### LINKED ISSUE #631: Search user warnings
This probably a big one but it would be nice to have the ability to search the user warning dialog to find which warning you are looking for easier

siddharthvp:
Add jquery.chosen on select box to make templates (and their descriptions) searchable.
Partially revert #517, bringing back the `{{uw-` prefix, as chosen's search works properly despite the presence of that prefix.
Closes #631.

Have used `mw.loader.using('jquery.chosen')` here, until `jquery.chosen` gets added to the gadget definition. 

We should probably revisit the descriptions of templates to make them include relevant keywords that attract search hits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ jsl.conf
 stash*.txt
 stash*.diff
 *.jsX
+.vscode/
 
 # Windows
 Thumbs.db
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1130,7 +1130,7 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 		} );
 	}
 
-	$('select[name=sub_group]').chosen();
+	$('select[name=sub_group]').chosen({search_contains: true});
 
 	// Limit the max height of select dropdown to prevent dialog box from becoming scrollable
 	mw.util.addCSS('.chosen-container .chosen-results { max-height: 180px !important }');
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1133,7 +1133,7 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	$('select[name=sub_group]').chosen({search_contains: true});
 
 	// Limit the max height of select dropdown to prevent dialog box from becoming scrollable
-	mw.util.addCSS('.chosen-container .chosen-results { max-height: 180px !important }');
+	$('.chosen-results').css({'overflow': 'auto', 'max-height': '180px'});
 
 	// clear overridden label on article textbox
 	Morebits.quickForm.setElementTooltipVisibility(e.target.root.article, true);
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -62,7 +62,7 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 	var main_group = main_select.append( {
 			type: 'select',
 			name: 'main_group',
-			event:Twinkle.warn.callback.change_category
+			event: Twinkle.warn.callback.change_category
 		} );
 
 	var defaultGroup = parseInt(Twinkle.getPref('defaultWarningGroup'), 10);
@@ -110,7 +110,9 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 	// We must init the first choice (General Note);
 	var evt = document.createEvent( "Event" );
 	evt.initEvent( 'change', true, true );
-	result.main_group.dispatchEvent( evt );
+	mw.loader.using('jquery.chosen').then(function() {
+		result.main_group.dispatchEvent( evt );
+	});
 };
 
 // This is all the messages that might be dispatched by the code
@@ -1050,6 +1052,8 @@ Twinkle.warn.prev_article = null;
 Twinkle.warn.prev_reason = null;
 
 Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCategory(e) {
+	$("select[name=sub_group]").chosen("destroy");
+
 	var value = e.target.value;
 	var sub_group = e.target.root.sub_group;
 	sub_group.main_group = value;
@@ -1096,7 +1100,7 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 			// Slice out leading uw- from the menu display
 			var elem = new Morebits.quickForm.element( {
 				type: 'option',
-				label: (value === 'custom' ? "{{" + key + "}}" : key.slice(3)) + level + ": " + (level ? itemProperties[value].label : itemProperties.label),
+				label: "{{" + key + level + "}}: " + (level ? itemProperties[value].label : itemProperties.label),
 				value: key + level,
 				selected: selected
 			} );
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@ jsl.conf
 stash*.txt
 stash*.diff
 *.jsX
+.vscode/
```
Amorymeltzer: @siddharthvp: [Found it!](https://github.com/azatoth/twinkle/pull/642#discussion_r287580727)

modules/twinklewarn.js
```diff
@@ -1126,6 +1130,11 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 		} );
 	}
 
+	$('select[name=sub_group]').chosen();
```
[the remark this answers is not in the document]
Amorymeltzer: I like `include_group_label_in_selected: true` as well, but I think it makes the input tooooo long.

modules/twinklewarn.js
```diff
@@ -1126,6 +1130,11 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
...
+
+	// Limit the max height of select dropdown to prevent dialog box from becoming scrollable
+	mw.util.addCSS('.chosen-container .chosen-results { max-height: 180px !important }');
```
Amorymeltzer: This is super thoughtful, but I'm mixed on the end result.  For one, it's around `165px` for me that the background dialog becomes scrollbable/the chosen menu extends beyond the bottom of the warn popup.  Is there a way to replicate what the current menu does, namely extend above (and thus beyond) the dialog?  Some quick checking with `z-index` went nowhere.

Amorymeltzer:
@siddharthvp One minor improvement, one question/discussion point.

>Have used `mw.loader.using('jquery.chosen')` here, until `jquery.chosen` gets added to the gadget definition

Yup, thanks; that's a "me" job.  I'll take care of it after this gets merged

>We should probably revisit the descriptions of templates to make them include relevant keywords that attract search hits.

Indeed; I'll open a tracker and invite folks to discuss when this goes live.  I think it'll be hugely popular!

modules/twinklewarn.js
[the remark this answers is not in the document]
siddharthvp: > Match not just the beginning of words. Better, I think?

Is it? I'm not sure. When someone searches "he", all `the`s would get matched, for instance. Indeed, the [jQuery documentation for the "contains" selector](https://api.jquery.com/attribute-contains-selector/) notes that the "contains word" selector (default in chosen) is more appropriate in many cases.

I've no idea about `include_group_label_in_selected`. Where did you read about it?

modules/twinklewarn.js
siddharthvp: oh, i tested on 4 browsers, and the results were only slightly different (3-4 px), and settled with 180px. But it looks like it's gonna depend on more factors than just browser.

siddharthvp: And hey wait a minute, won't this css line affect _all_ chosen dropdowns, including the ones in afch and enterprisey's scripts? That's worrying, ideally chosen should have provided a way to set the max height as an option just like seach_contains.

modules/twinklewarn.js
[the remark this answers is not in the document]
Amorymeltzer: Well, I suppose it depends on how one searches, but as an example, searching for `biog` will currently not find `{{uw-autobiography}}` and `links` won't find `{{uw-redlink}}`.  Basically, you're right, but we're not overly verbose here, and I think aiding in discovery (I didn't know about `uw-redlinks`!) is worth finding a few erroneous `the`s and `an`s if people search poorly.

modules/twinklewarn.js
Amorymeltzer: >ideally chosen should have provided a way to set the max height

Yeah, kind of annoying, but hey, free software.

>it would affect other scripts only when the other script is used on the page after Warn was used and before the page is reloaded. Nothing to worry about.

Well, what if someone goes to warn a user but changes their mind?  Unlikely but unfriendly.

modules/twinklewarn.js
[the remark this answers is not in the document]
siddharthvp: I guess you're right. Will make the change.

modules/twinklewarn.js
Amorymeltzer: Thinking about it a bit, couldn't we just grab the drop?  That is:
```suggestion
	$('.chosen-drop').css({'overflow': 'auto', 'max-height': '180px'});
```
When written this way, `180px` doesn't overflow my screen either (not sure why, CSS is weird).

siddharthvp: This result in 2 scrollables within the chosen drop itself! https://ibb.co/C1bjcTS is what I'm seeing.

Amorymeltzer: @siddharthvp  Wha?!😖  I certainly don't get that!  Looks like we're both in chrome atm, have you tried in safemode/a different skin?  Maybe a mac vs pc?  Can you figure out what those two elements are?  Seems weird for both to have the `.chosen-drop` class??

Amorymeltzer: The duplicate bars issue notwithstanding, *my* scrollbar is weird with the above; it will disappear below the bottom of the dropdown.  I guessed on `.chosen-drop` but using `.chosen-results` fixes my issue.
```suggestion
	$('.chosen-results').css({'overflow': 'auto', 'max-height': '180px'});
```

modules/twinklewarn.js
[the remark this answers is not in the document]
Amorymeltzer: I was thinking it'd be handy for delsort as well, but everything is so simple there I don't think it'll be useful.  AFAICS the only cases would be for something like `people` in `Sportspeople`, but that's covered by being in the `People` category, so there's no advantage.

modules/twinklewarn.js
Amorymeltzer: If we can nail this down I'd like to put it in delsort as well, there's some funkiness going on there with the dropdown extending far below the dialog.

siddharthvp: `$('.chosen-results').css({'overflow': 'auto', 'max-height': '180px'});` works fine for me too, just like the original mw.util.addCSS fix.

> Can you figure out what those two elements are? Seems weird for both to have the .chosen-drop class??

fwiw the inner one is `div.chosen-results` and the outer is `ul.chosen-drop`

Amorymeltzer: So weird, but glad that worked!

Amorymeltzer: @siddharthvp Was playing around with adding this to xfd/delsort, but it's a little less clear what is good, though, since the delsort option is lower in the dialog, and a preview would extend the dialog.  135px or 155px seem okay but neither is quite ideal.  Curious what your thoughts are, if it'd even be helpful.

siddharthvp: @Amorymeltzer I'm getting `.chosen-results` to hit the bottom of the dialog at 147px (with preview off) and at about 167px with preview on. I think we should ignore the latter case, that is, set the max height with the assumption that preview is off. Are you sure its 135px for you?

Amorymeltzer: 141px actually, but yeah; that's in chrome, I get 142px in Safari and 143px in Firefox.  And for preview, I meant not just clicking the button but not scrolling, but to have the dropdown not overlap the text of the preview.  That for me is about 155px.  Agree with on the 167px case, though fwiw that's my number too.

siddharthvp: 145px is ok for me. But since you said its 141px for you, wouldn't that make the dialog get a scroll bar for you, which is what we're trying to avoid? IMO its okay for the dropdown to overlap the preview text. 

Or maybe its fine to not put in a css rule at all if we're unable to make out what's ideal. The user can scroll down in the dialog to have the full scroll-length of the dropdown visible. It's not much of a issue here as it was in Warn where the select was the most important field.

Amorymeltzer: Well, 145px wouldn't be ideal on my end but it certainly wouldn't be a step backward either.

I'm torn — the current view is annoying but context matters in delsort/multi-select and only displaying five could be limiting — so I guess doing nothing is a fine enough place to end up.

PR #642, opened 2019-05-25
tag: place new tags under {{AfDM}}, make \n optional in prod

siddharthvp:
Add {{AfDM}} to the regex. {{AfDM}} is used by Twinkle's XFD module instead of
{{Article for deletion/dated}} when it's a 2nd/3rd/nth nomination.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1528,7 +1528,7 @@ Twinkle.tag.callbacks = {
 					// capture template(s)
 					'(?:((?:\\s*' +
 					// AfD is special, as the tag includes html comments before and after the actual template
-					'(?:<!--.*AfD.*\\n\\{\\{Article for deletion\\/dated.*\\}\\}\\n<!--.*\\n<!--.*AfD.*(?:\\s*\\n))?|' + // trailing whitespace/newline needed since this subst's a newline
+					'(?:<!--.*AfD.*\\n\\{\\{(?:Article for deletion\\/dated|AfDM).*\\}\\}\\n<!--.*(?:\\n<!--.*)?AfD.*(?:\\s*\\n))?|' + // trailing whitespace/newline needed since this subst's a newline
 					// begin template format
 					'\\{\\{\\s*(?:' +
 					// CSD
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1534,7 +1534,7 @@ Twinkle.tag.callbacks = {
 					// CSD
 					'db|delete|db-.*?|speedy deletion-.*?|' +
 					// PROD
-					'(?:proposed deletion|prod blp)\\/dated\\n(?:\\s+\\|(?:concern|user|timestamp|help).*)+|' +
+					'(?:proposed deletion|prod blp)\\/dated(?:\\s*\\|(?:concern|user|timestamp|help).*)+|' +
 					// various hatnote templates
 					'about|correct title|dablink|distinguish|for|other\\s?(?:hurricaneuses|people|persons|places|uses(?:of)?)|redirect(?:-acronym)?|see\\s?(?:also|wiktionary)|selfref|the' +
 					// not a hatnote, but sometimes under a CSD or AfD


.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
```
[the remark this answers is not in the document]
siddharthvp: I remember adding .vscode/ to gitignore, but doesn't seem to have worked, deleted it now ...

modules/friendlytag.js
```diff
@@ -1528,13 +1528,13 @@ Twinkle.tag.callbacks = {
 					// capture template(s)
 					'(?:((?:\\s*' +
 					// AfD is special, as the tag includes html comments before and after the actual template
-					'(?:<!--.*AfD.*\\n\\{\\{Article for deletion\\/dated.*\\}\\}\\n<!--.*\\n<!--.*AfD.*(?:\\s*\\n))?|' + // trailing whitespace/newline needed since this subst's a newline
+					'(?:<!--.*AfD.*\\n\\{\\{(Article for deletion\\/dated|AfDM).*\\}\\}\\n<!--.*(\\n<!--.*)?AfD.*(?:\\s*\\n))?|' + // trailing whitespace/newline needed since this subst's a newline
```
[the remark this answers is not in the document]
siddharthvp: no, but isn't it simpler to leave as them as capturing? Anyway, added those `?:` coz I don't feel like seeing into whether it'd cause any errors ...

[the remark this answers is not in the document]
Amorymeltzer: Thanks — that's basically why.  Also, it's not a huge issue since we're not doing anything in bulk here, but capturing should be slower.

modules/friendlytag.js
```diff
@@ -1528,13 +1528,13 @@ Twinkle.tag.callbacks = {
...
 					// AfD is special, as the tag includes html comments before and after the actual template
-					'(?:<!--.*AfD.*\\n\\{\\{Article for deletion\\/dated.*\\}\\}\\n<!--.*\\n<!--.*AfD.*(?:\\s*\\n))?|' + // trailing whitespace/newline needed since this subst's a newline
+					'(?:<!--.*AfD.*\\n\\{\\{(?:Article for deletion\\/dated|AfDM).*\\}\\}\\n<!--.*(?:\\n<!--.*)?AfD.*(?:\\s*\\n))?|' + // trailing whitespace/newline needed since this subst's a newline
 					// begin template format
 					'\\{\\{\\s*(?:' +
 					// CSD
 					'db|delete|db-.*?|speedy deletion-.*?|' +
 					// PROD
-					'(?:proposed deletion|prod blp)\\/dated\\n(?:\\s+\\|(?:concern|user|timestamp|help).*)+|' +
+					'(?:proposed deletion|prod blp)\\/dated(?:\\s+\\|(?:concern|user|timestamp|help).*)+|' +
```
Amorymeltzer: Not sure which was intentional, but this didn't line up with your commit comment.

```suggestion
					'(?:proposed deletion|prod blp)\\/dated(?:\\s*\\|(?:concern|user|timestamp|help).*)+|' +
```

At any rate, what's the impetus behind making this change?  The templates should always be substituted, so would this change just cover the case of someone/something manually adding `{{proposed deletion/dated|concern=concern|timestamp={{subst:CURRENTTIMESTAMP}}}}` rather than `{{subst:prod|concern=concern}}`?

siddharthvp: `\\s*` is more correct, but of course as long as templates are substed, it doesn't matter. The impetus behind the change is that it makes the regex (slightly) simpler while also accounting for the case of someone manually adding {{proposed deletion/dated}}.

siddharthvp: Of course, here in tag module, all this hardly matters since the world won't freeze over if a tag gets added above a prod notice rather than below. But complicated regexes like this are likely to be reused/adapted by people in other more important applications (like I am doing in my rfud-helper script, where it's necessary that all deletion tags be removed from a page after it's been restored). So it's good to get the regex perfectly right.

Amorymeltzer: Okay, that's fine.  Generally I think it's good to be as parsimonious as reasonable.

PR #643, opened 2019-05-26
morebits: allow page to reload to a specific section of the current page

siddharthvp:
after action complete. Allows #609 to go forward.


siddharthvp:
this isn't working ...

siddharthvp:
fixed I think.... Boy, that was _extremely_ tricky. And what's really annoying is that this still sometimes works and sometimes doesn't work on the _same_ testcase.

PR #646, opened 2019-05-28
morebits: overhaul action complete status messages

siddharthvp:
(See #645, which adds docs for Morebits.status, if you're unfamiliar with the really convoluted Morebits.status system.)

The action complete status messages (the last line of status) in almost all Twinkle modules do not read correctly. Such as in Tag module, it says "Tagging complete: completed". In xfd, it says "Nomination complete, redirecting to discussion page: completed". None of the Twinkle modules make meaningful use of `Morebits.wiki.actionCompleted.postfix` (the part of the status message after the colon), allowing it to remain "completed" which is the default. There is no way to suppress the display of the colon and postfix.

This commit introduces a new function that creates a status line without a colon separator, and uses it for the action-completed message instead of using a normal Morebits.status object. This does away with `Morebits.wiki.actionCompleted.postfix`, which as described above, has no utility.

The default value of `Morebits.wiki.actionCompleted.notice` is changed from `Action`, which does not make much sense, to `null`. If it is not redefined in the module, the entire action-complete message is suppressed. This was not possible earlier, and was being achieved in RPP/PP through an ugly hack.

### DIFF
--- a/modules/twinklebatchdelete.js
+++ b/modules/twinklebatchdelete.js
@@ -199,8 +199,7 @@ Twinkle.batchdelete.callback = function twinklebatchdeleteCallback() {
 };
 
 Twinkle.batchdelete.callback.evaluate = function twinklebatchdeleteCallbackEvaluate(event) {
-	Morebits.wiki.actionCompleted.notice = 'Status';
-	Morebits.wiki.actionCompleted.postfix = 'batch deletion is now complete';
+	Morebits.wiki.actionCompleted.notice = 'Batch deletion is now complete';
 
 	var numProtected = $(Morebits.quickForm.getElements(event.target, 'pages')).filter(function(index, element) {
 		return element.checked && element.nextElementSibling.style.color === 'red';
--- a/modules/twinklebatchundelete.js
+++ b/modules/twinklebatchundelete.js
@@ -114,8 +114,7 @@ Twinkle.batchundelete.callback = function twinklebatchundeleteCallback() {
 };
 
 Twinkle.batchundelete.callback.evaluate = function( event ) {
-	Morebits.wiki.actionCompleted.notice = 'Status';
-	Morebits.wiki.actionCompleted.postfix = 'batch undeletion is now complete';
+	Morebits.wiki.actionCompleted.notice = 'Batch undeletion is now complete';
 
 	var numProtected = $(Morebits.quickForm.getElements(event.target, 'pages')).filter(function(index, element) {
 		return element.checked && element.nextElementSibling.style.color === 'red';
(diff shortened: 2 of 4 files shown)


Amorymeltzer:
Okay, wrapping my head around this.  Your summary was good, but it took me a bit to translate it to the code.  I'm not sure why I thought this was more involved?  A brief summary:

- `postfix` useless, possibly bad
- Remove it!
- Make special no colon line for `actionCompleted.notice`

Sounds good to me!  What happens to anything else out there?  It would just whatever was in `Morebits.wiki.actionCompleted.notice`, which up until this would be something like "Status" or whatever, right?  That's not ideal, but livable.  What would you think of, if there is a `Morebits.wiki.actionCompleted.postfix` present:
- concatenating the `actionCompleted.notice` and `actionCompleted.postfix` into `actionCompleted.notice`
- Log a notice in the console to stop using `actionCompleted.postfix`

Could remove in, say, 3-6 months?  Might be a little much for such a little thing.

siddharthvp:
> What happens to anything else out there? It would just whatever was in Morebits.wiki.actionCompleted.notice, which up until this would be something like "Status" or whatever, right? That's not ideal, but livable.

They would just show whatever is there in Morebits.wiki.actionCompleted.notice - which in [all external user scripts](https://en.wikipedia.org/w/index.php?title=Special:Search&limit=20&offset=0&ns2=1&search=%22Morebits.wiki.actionCompleted.notice%22&advancedSearch-current={}) is something along the lines of "... completed, reloading page in a second". So this is in fact ideal. 

User:Bellezzasolo/Scripts/arb.js, for instance, displays the following action complete msg now:
```
Notice complete; reloading talk page in a few seconds: completed
```
this would become
```
Notice complete; reloading talk page in a few seconds
```

>  What would you think of, if there is a Morebits.wiki.actionCompleted.postfix present:

I have taken care of the 3 modules that were using postfix. So it now is not present anywhere. Of the [8 search results](https://en.wikipedia.org/w/index.php?cirrusUserTesting=control&sort=relevance&search=%22Morebits.wiki.actionCompleted.postfix%22&title=Special:Search&profile=advanced&fulltext=1&advancedSearch-current=%7B%7D&ns2=1) for userspace scripts, 1 is a functional script by me that I'll fix, remaining 7 are unused and outdated copies of twinkle code.

The many things using `actionCompleted.notice`, which as mentioned above will all be _improved_ with this PR, rather than break.

Amorymeltzer:
>Of the 8 search results for userspace scripts...remaining 7 are unused...
This is basically the number I was looking for, places where information would be made worse.  Looks like nothing on enwiki will be borked by removing `postfix`, and I don't think anybody directly imports `morebits.js` from enwiki, so we should be all clear.  I did a global search for fun, all looks fine.

Amorymeltzer:
Merged in 9cef0d4

PR #647, opened 2019-05-28
tag: add quick filtering functionality

### LINKED ISSUE #621: Tag: Create collapsible subgroups
Background: there are way too many tag options, making it hard to find the one you are looking for. Many are already grouped with subheadings, but the groups should be collapsible

Request:
Create collapsible subgroups. Given that tag relies on checkboxes (not radio buttons), it is always possible to select multiple options. If a heading is selected, only then should the templates nested under that heading be visible to select. Selecting just the heading should not add any tags, and can either warn the user or do nothing.

Examples:
In the article tags, a "Lede" subgroup would contain:

- Lede missing
- Lede rewrite
- Lede too long
- Lede too short

all _within_ the current "Structure, formatting, and lead section" group. On the other hand, the "merge" subgroup would be the entirety of the current "Merging" section.

Another example:
Within the file tags, a "bad format" subgroup would contain

- Bad format
- Bad GIF
- Bad JPEG

while a "keep local" group would contain

- {{Do not move to Commons}} (PD issue)
- {{Do not move to Commons}} (other reason)
- {{Keep local}}

I hope this explanation makes sense.

siddharthvp:
This is gonna be popular! Now that finding tags has become so easy (closes #621), we can add more tags to Twinkle without fear of cluttering the interface. 

Tested on IE 11. Everything works. (Though it works a bit slow on IE.)

A question to be pondered on:  When there are no search results, should the work area be empty (as is now the case), or should it show all tags?

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -76,6 +76,49 @@ Twinkle.tag.callback = function friendlytagCallback() {
 				]
 			});
 
+			form.append({
+				type: 'input',
+				label: 'Quick filter: ',
+				name: 'quickfilter',
+				size: '30px',
+				event: function twinkletagquickfilter() {
+					var $form = $('#tagWorkArea');
+					// flush the DOM of all existing underline spans
+					$form.find('.search-hit').each(function(i,e) {
+						var label_element = e.parentElement;
+						var new_label = Morebits.htmlNode('label', label_element.textContent);
+						new_label.setAttribute('for', label_element.previousElementSibling.id);
+						$(label_element).replaceWith(new_label);
+					});
+					// allCheckboxDivs and allHeaders are defined globally, rather than in
+					// this function, to avoid having to recompute them on every keydown.
+					if (this.value) {
+						allCheckboxDivs.hide();
+						allHeaders.hide();
+						var searchString = this.value;
+						var searchRegex = new RegExp(mw.RegExp.escape(searchString), 'i');
+
+						$form.find('label').each(function() {
+							var label_text = this.textContent;
+							var searchHit = searchRegex.exec(label_text);
+							if (searchHit) {
+								var range = document.createRange();
+								var textnode = this.childNodes[0];
+								range.selectNodeContents(textnode);
+								range.setStart(textnode, searchHit.index);
+								range.setEnd(textnode, searchHit.index + searchString.length);
+								var underline_span = $('<span>').addClass('search-hit').css('text-decoration', 'underline')[0];
+								range.surroundContents(underline_span);
+								this.parentElement.style.display = 'block'; // un-hide
+							}
+						});
+					} else {
+						allCheckboxDivs.show();
+						allHeaders.show();
+					}
+				}
+			});
+
 			if (! Twinkle.tag.canRemove) {
 				var divElement = document.createElement('div');
 				divElement.innerHTML = 'For removal of existing tags, please open Tag menu from the current version of article';
@@ -164,6 +207,8 @@ Twinkle.tag.callback = function friendlytagCallback() {
 
 	if (Twinkle.tag.mode === "article") {
 
+		result.quickfilter.focus();  // place cursor in the Quick filter field as soon as window is opened
+
 		Twinkle.tag.alreadyPresentTags = [];
 
 		if (Twinkle.tag.canRemove) {
@@ -217,6 +262,9 @@ Twinkle.tag.callback = function friendlytagCallback() {
 		Morebits.quickForm.getElements(result, Twinkle.tag.mode + "Tags").forEach(generateLinks);
 	}
 };
+
+// Used in Quick Filter event function
+var allCheckboxDivs, allHeaders;
 
 Twinkle.tag.updateSortOrder = function(e) {
 	var sortorder = e.target.value;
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
>This is gonna be popular!

Hell yeah it is!

Amorymeltzer:
I'll try and take a look at this soon @siddharthvp but one thing to think about down the line is whether this could make keeping track of what has been selected a little harder for the user.  I think this addition is glorious, but I can imagine that having an option to display all checked options would be nice.

siddharthvp:
>  I can imagine that having an option to display all checked options would be nice.

hmm, where on the UI would you put such an option?

I have an easy option in mind - just display some <small> text adjacent to the Submit button telling what all is gonna happen when hit submit: like this - https://ibb.co/jHcHjZS, or if you prefer to be fully specific, like this - https://ibb.co/9GT0vKm . You can see the JS used to generate them on the console. This would work through click event listening function that would update the notice every time a checkbox is clicked. Quite easy to code ...

Amorymeltzer:
>I have an easy choice in mind

This is awesome!  I think the first display is perfect — *please* open a PR with that!

What was getting at (and missing) was a way to display all the items the user had selected. 
 It's a huge menu, and let's say I use this dope new quick filter input to find the five tags I want to add, some of which have additional inputs (merge, cleanup rewrite, etc.).  I now have a half-dozen boxes checked, but I gotta scroll and read 'em all to check I'm happy.  A minor improvement would be a way to filter only the options selected, so I could see what I'm doing very clearly.

Nobody has asked for it, and a mostly-unused checkbox wouldn't be a net-positive, so I'm not really sure how it could happen.  Moreover, your suggestion is close enough and far superior, so there's no need.

Amorymeltzer:
>A question to be pondered on: When there are no search results, should the work area be empty (as is now the case), or should it show all tags?

I think empty is better.  A message (à la `chosen`) would be nice, but if you filtered and found nothing, finding nothing is more clear/expected.

modules/friendlytag.js
```diff
@@ -76,6 +76,49 @@ Twinkle.tag.callback = function friendlytagCallback() {
...
+				name: 'quickfilter',
+				size: '30px',
+				event: function twinkletagquickfilter() {
+					var $form = $('#tagWorkArea');
+					// flush the DOM of all existing underline spans
+					$form.find('.search-hit').each(function(i,e) {
+						var label_element = e.parentElement;
+						// This would convert <label>Hello <span class=search-hit>wo</span>rld</label>
+						// to <label>Hello world</label> // eslint-disable-next-line no-self-assign
+						label_element.textContent = label_element.textContent;
```
siddharthvp: Figured out an optimisation. Instead of reconstructing the entire label element, it is sufficient to re-assign the `textContent`. So this may look like a most pointless line of code, and eslint is definitely not happy, but it does the trick. 

The alternative would be to do something like 
```
$(label_element).find('span').text(this.textContent);  // replace the span with a text node containing the span's textcontent
label_element.normalize();   // re-joins the adjacent text nodes
```
or even worse, what I had done before. Both are clearly more expensive.

With this, the thing works quite smoothly even on IE.

Amorymeltzer: @siddharthvp Maybe it's cheating a bit, but wouldn't this work?  Seems to on my end.
```suggestion
						label_element.innerHTML = label_element.textContent;
```

siddharthvp: Aha that is clever, but do we really need to let eslint become our master 🤔

Amorymeltzer: Thanks — I agree we don't need to let the automaton control us, but it's a good indication that something will be weird/confusing/etc.  Going with `label_element.innerHTML = label_element.textContent` actually makes a lot more sense to the reader, I think, since it's clear we're replacing any HTML with just text, where we probably could've had another comment explaining that `textContent = textContent` strips the spans.

Anyway, one simple line is much nicer!

Amorymeltzer:
@siddharthvp I'm happy to merge this (and #648) now, if you're all set

siddharthvp:
Just one last change: the search field is cleared when the viewing mode (alpha/category) is changed, since search results don't appear immediately after mode is changed. This fixes the awkward situation that would occur where there is stuff in the search box but the search results aren't showing.

Amorymeltzer:
>fixes the awkward situation

I think I ran into that once then forgot about it with the coolness of this.  Theoretically we should be able to re-rig `updateOnSortOrder` or something to re-initiate the prior search, but I'd imagine that'd require rethinking the structure of this.  Down the line, maybe, but nowhere near a priority, and I'm not even sure it'd be expected behavior.

This is so dope 🏆

siddharthvp:
@Amorymeltzer retrospectively all that was need was to dispatch a fake keypress event to the search field. But I think this is hardly worth thinking over much because why would anyone want to change view-mode in the middle of a search anyway?

Amorymeltzer:
Yeah, that's sort of what I was thinking above, that the filter works so well that you're bound to find what you like, so if you're going to move off the keyboard to the mouse/screen to change the order, it's likely you're done with that particular search.  I'm not sure which behavior is expected, so maybe we keep this as is, then if folks ask about the feature you're raring to go?

siddharthvp:
Yeah,  replacing `e.target.form.quickfilter.value = ''` with  `e.target.form.quickfilter.dispatchEvent(new Event('keyup'))` should be do it (untested), if folks ask for it which I doubt.

PR #648, opened 2019-05-31
tag: add status text after Submit button, fix untag bug

siddharthvp:
As discussed in #647.

The first commit is a bug fix, which I have included in this PR because the bug affects the working of status text update.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -220,10 +220,7 @@ Twinkle.tag.callback = function friendlytagCallback() {
 
 Twinkle.tag.updateSortOrder = function(e) {
 	var sortorder = e.target.value;
-	Twinkle.tag.checkedTags = e.target.form.getChecked("articleTags");
-	if (!Twinkle.tag.checkedTags) {
-		Twinkle.tag.checkedTags = [];
-	}
+	Twinkle.tag.checkedTags = e.target.form.getChecked("articleTags") || [];
 
 	var container = new Morebits.quickForm.element({ type: "fragment" });
 
@@ -475,13 +472,14 @@ Twinkle.tag.updateSortOrder = function(e) {
 		container.append({ type: "header", id: "tagHeader0", label: "Tags already present" });
 		var subdiv = container.append({ type: "div", id: "tagSubdiv0" });
 		var checkboxes = [];
+		var unCheckedTags = e.target.form.getUnchecked("alreadyPresentArticleTags") || [];
 		Twinkle.tag.alreadyPresentTags.forEach( function(tag) {
 			var description = Twinkle.tag.article.tags[tag];
 			var checkbox =
 				{
 					value: tag,
 					label: "{{" + tag + "}}" + ( description ? (": " + description) : ""),
-					checked: true
+					checked: unCheckedTags.indexOf(tag) === -1
 					//, subgroup: { type: 'input', name: 'removeReason', label: 'Reason', tooltip: 'Enter reason for removing this tag' }
 					// TODO: add option for providing reason for removal
 				};
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -207,6 +207,17 @@ Twinkle.tag.callback = function friendlytagCallback() {
 
 		}
 
+		// Add status text node after Submit button
+		var statusNode = document.createElement('small');
+		statusNode.id = 'tw-tag-status';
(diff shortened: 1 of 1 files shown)


modules/friendlytag.js
```diff
@@ -570,6 +578,24 @@ Twinkle.tag.updateSortOrder = function(e) {
...
+	$('[name=articleTags], [name=alreadyPresentArticleTags]').click(function() {
+		if (this.name === 'articleTags') {
+			if (this.checked) Twinkle.tag.status.numAdded++;
+			else Twinkle.tag.status.numAdded--;
+		} else if (this.name === 'alreadyPresentArticleTags') {
+			if (this.checked) Twinkle.tag.status.numRemoved--;
+			else Twinkle.tag.status.numRemoved++;
+		}
+
+		// update statusElement text
```
Amorymeltzer: I think this'd be better served above the `getElementById` and with a little expansion
```suggestion
		// tally tags added/removed, update statusElement text
```

modules/friendlytag.js
```diff
@@ -207,6 +207,16 @@ Twinkle.tag.callback = function friendlytagCallback() {
 
 		}
 
+		// Add status text element after Submit button
```
Amorymeltzer: This is incredibly silly perhaps, but the term `statusElement` is used heavily as `pageobj.getStatusElement()` — frequently shortened as `var statelem` — so, for clarity's sake, I think it'd be better to come up with a different term here.  I don't think "status" quite captures the essence of what this is displaying, but even moving to `statusText` would be clearer I think.  Something like `Twinkle.tag.changes` and `var changesNode` would be clearer as well.  I'm not exactly full of good ideas at the moment, but it's just additional clarity for maintenance down the line, so it need not be beautiful prose.

modules/friendlytag.js
```diff
@@ -570,6 +578,24 @@ Twinkle.tag.updateSortOrder = function(e) {
...
+		if (this.name === 'articleTags') {
+			if (this.checked) Twinkle.tag.status.numAdded++;
+			else Twinkle.tag.status.numAdded--;
+		} else if (this.name === 'alreadyPresentArticleTags') {
+			if (this.checked) Twinkle.tag.status.numRemoved--;
+			else Twinkle.tag.status.numRemoved++;
+		}
+
+		// update statusElement text
+		var firstPart = 'Added ' + Twinkle.tag.status.numAdded + ' tag' + (Twinkle.tag.status.numAdded > 1 ? 's' : '');
```
Amorymeltzer: Past tense seems weird.  Could also prefix the whole thang with "Will " to make "Will add 3 add remove 2" for a complete-ish sentence, but that seems excessive.
```suggestion
		var firstPart = 'Adding ' + Twinkle.tag.status.numAdded + ' tag' + (Twinkle.tag.status.numAdded > 1 ? 's' : '');
```

modules/friendlytag.js
```diff
@@ -570,6 +578,24 @@ Twinkle.tag.updateSortOrder = function(e) {
...
+		var secondPart = 'Removed ' + Twinkle.tag.status.numRemoved + ' tag' + (Twinkle.tag.status.numRemoved > 1 ? 's' : '');
```
Amorymeltzer:
```suggestion
		var secondPart = 'Removing ' + Twinkle.tag.status.numRemoved + ' tag' + (Twinkle.tag.status.numRemoved > 1 ? 's' : '');
```

modules/friendlytag.js
```diff
@@ -207,6 +207,16 @@ Twinkle.tag.callback = function friendlytagCallback() {
 
 		}
 
+		// Add status text element after Submit button
+		var statusElement = Morebits.htmlNode('small', '');
+		statusElement.id = 'tw-tag-status';
+		Twinkle.tag.status = {
+			// defined like this coz these need to available for reference in the click event handler
```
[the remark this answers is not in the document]
siddharthvp: "like this" seems more appropriate since we're talking abt why these are defined under `Twinkle.tag.` .

modules/friendlytag.js
siddharthvp: renamed to `statusNode` to avoid confusion with morebits-based `statusElement`s. I agree that status is not the correct word, but then, what is ...

Amorymeltzer: Perfect, thanks!

PR #653, opened 2019-06-01
tag: preserve checked status of custom tags over mode change

### LINKED ISSUE #652: tag: Custom tags unchecked when changing view (thus ignored statusNode)
Two phenotypes related to the same issue, but when changing between alphabetical and category listing of tags, any custom checked tags are unchecked.  This is a bug in the current version on enWiki.

Relatedly, that unchecking isn't detected by the `statusNode` from #648, so since the count update is all relative, changing the category will make permanent an off-by-n error.

I can look into this tomorrow/Monday when I have time, but cc @siddharthvp in case you've got a moment

siddharthvp:
Fixes #652.


Amorymeltzer:
Looks like this arose in b28bc422 nearly exactly five years ago!

PR #660, opened 2019-06-06
warn: fix display of specialized input prompts

### LINKED ISSUE #659: twinklewarn: single-issue notices: Wrong <label for="node_11">
While "Optional username that was reported (without User:)" is correct for "uw-aiv", that's pretty much the only warning message that requires such a label. All others should use "Linked page".

twinkle-wrongtext-20190606

Amorymeltzer:
Closes #659.  Changing selection from a `jquery.chosen` menu doesn't trigger `change_subcategory` since the event was attached to the original select.  We still need the extra `Twinkle.warn.callback.change_subcategory(e)` call from #432 to update display on simply `change_category` as well as the original, initial event in `sub_group` for users who opt-out of the new `jquery.chosen` display (#661)

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1151,20 +1151,22 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 		} );
 	}
 
-	if (!Twinkle.getPref('oldSelect')) {
-		$('select[name=sub_group]').chosen({width: '100%', search_contains: true});
-
-		// Limit the max height of select dropdown to prevent dialog box from becoming scrollable
-		$('.chosen-results').css({'overflow': 'auto', 'max-height': '180px'});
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
@siddharthvp You're the resident `chosen` expert now, does this ring right to you?

Amorymeltzer:
Pushed it live, seems to work fine, but I'm gonna hold off on merging to think about #661 first, in particular since removing the initial `event` will no longer be possible

siddharthvp:
> You're the resident chosen expert now

Oh, not at all. Since this and #661 conflict with each other, should i rewrite that to account for this? If so, do merge this one.

Amorymeltzer:
@siddharthvp No, don't bother; this is simple and easy, and I'll work this around that.  Don't sweat it.

PR #661, opened 2019-06-06
warn: make jquery.chosen based menu optional, via-opt-out

siddharthvp:
Per discussion at [[Wikipedia_talk:Twinkle#User_warning_dialog_dropdown_selector]]

The reason some folks don't like this is because chosen makes the select menu much less denser, requiring more scrolling to get to the templates, and this also seems to give undue focus to `{{uw-` part of the templates.

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -672,6 +672,14 @@ Twinkle.config.sections = [
 			type: "boolean"
 		},
 
+		// TwinkleConfig.oldSelect (boolean)
+		// if true, use the native select menu rather the jquery chosen-based one
+		{
+			name: "oldSelect",
+			label: "Use the non-searchable classic select menu",
+			type: "boolean"
+		},
+
 		{
 			name: "customWarningList",
 			label: "Custom warning templates to display",
@@ -1697,7 +1705,7 @@ Twinkle.config.writePrefs = function twinkleconfigWritePrefs(pageobj) {
 	text +=
 		";\n" +
 		"\n" +
-		"// </no" + "wiki>\n" + 
+		"// </no" + "wiki>\n" +
 		"// End of twinkleoptions.js\n";
 
 	pageobj.setPageText(text);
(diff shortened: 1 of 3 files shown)


Amorymeltzer:
Thoughts on making it opt-in rather than opt-out?

siddharthvp:
I guess only the old hands, who are used to their ways, aren't liking it. New users should love it. I think it's better to keep it opt-out.

Amorymeltzer:
@siddharthvp Issue there is that the old hands are the ones doing the majority of warnings and using Twinkle.

siddharthvp:
> Issue there is that the old hands are the ones doing the majority of warnings and using Twinkle.

Not inherently opposed, but its rather silly to have the feature-rich on by default. I don't know, it gets tough to decide things between us. Can others who are watching pitch in? Or should we open a discussion on this at WT:TW?

> Density could also be achieved with some simple CSS inre padding

Cool, have you tried that?

Amorymeltzer:
We've had only two folks with complaints, and I've got at least two on IRC who are all for it, so I'll hold off a bit to see if there's more input, but I'm leaning toward opt-out of the new style as well.  With a section at WT:TW and a linked comment at WP:VPT it should get noticed.

siddharthvp:
> Density could also be achieved with some simple CSS inre padding

I was playing around with that. See https://ibb.co/PT8b63Z. Removed the padding entirely off the top and bottom of the li elements. This looks much better and non-OOUI like, and may probably avoid the need to make the thing opt-out. Most people, including me, hate OOUI styles anyway. 

The trouble is the result was obtained by removing padding from chrome devtool's styles menu. I have no idea how to get that using code. I have wasted quite a silly amount of time trying it out with no success. Neither `$('.chosen-container .chosen-results li').css({'padding-top': '0px', 'padding-bottom': '0px'});` nor `mw.util.addCSS('.chosen-container .chosen-results li { padding-top: 0px !important; padding-bottom: 0px !important; }');` seems to work. Wanna have a go at it?

Amorymeltzer:
Yeah, I think I spoke too soon?  I came up against the same thing as well.  Even tried using `.on` to no avail.

>This looks much better and non-OOUI like... Most people, including me, hate OOUI styles anyway.

Preach.

Amorymeltzer:
No more input so gonna go ahead and merge this.  Thanks for being so quick on this @siddharthvp; it's not an ideal solution but so be it.

PR #662, opened 2019-06-06
warn: remove padding from chosen's li elements

siddharthvp:
Increases the density of the select menu. Use padding of 1px on search results for clarity.

@Amorymeltzer Don't know what was happening earlier, but this seems to be working fine now with `mw.util.addCSS`. Tested on multiple browsers. The jquery css solution won't work because the `li` elements are not added into DOM until the select is clicked the first time, so the jquery selection is of 0 length. 

If this is working for you too, can you deploy this live immediately, so that people can try it out (and probably change their opinion about the new interface😎) ?

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1161,11 +1161,15 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 
 	// Build select menu with jquery.chosen
 	if (!Twinkle.getPref('oldSelect')) {
-		$('select[name=sub_group]').chosen({width: '100%', search_contains: true}).change(function(e) {
-			Twinkle.warn.callback.change_subcategory(e);
-		});
+		$('select[name=sub_group]')
+			.chosen({width: '100%', search_contains: true})
+			.change(Twinkle.warn.callback.change_subcategory);
+
 		// Limit the max height of select dropdown to prevent dialog box from becoming scrollable
 		$('.chosen-results').css({'overflow': 'auto', 'max-height': '180px'});
+
+		// Remove padding
+		mw.util.addCSS('.chosen-drop .chosen-results li { padding-top: 0px; padding-bottom: 0px; }');
 	}
 };
 


modules/twinklewarn.js
```diff
@@ -1154,6 +1154,12 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	// Limit the max height of select dropdown to prevent dialog box from becoming scrollable
 	$('.chosen-results').css({'overflow': 'auto', 'max-height': '180px'});
 
+	// Remove padding
+	mw.util.addCSS('.chosen-drop .chosen-results li { padding-top: 0px; padding-bottom: 0px; }');
```
Amorymeltzer: I think let's just keep it simple and do 1px across the board, i.e. remove the fancy bit for search results
```suggestion
	mw.util.addCSS('.chosen-drop .chosen-results li { padding-top: 1px; padding-bottom: 1px; }');
```

Amorymeltzer:
Still not wild about `mw.util.addCSS` but `.chosen-drop .chosen-results li` is specific enough I suppose.

modules/twinklewarn.js
siddharthvp: why? The keyup handler may look a bit expensive but it doesn't cause any delays or anything. 0px for the default case is closest to the native select menu and packs in an extra template in the view without scrolling down.

Amorymeltzer: I meant it less as favoring 1px over 0px and more as it's unnecessary to do the dance to change the menu on the fly.  1 or 0 is fine with me (I probably favor 1px slightly) but I'd sooner not change the menu.  Going 0 everywhere is fine IMO.

modules/twinklewarn.js
```diff
@@ -1161,11 +1161,15 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 
 	// Build select menu with jquery.chosen
 	if (!Twinkle.getPref('oldSelect')) {
-		$('select[name=sub_group]').chosen({width: '100%', search_contains: true}).change(function(e) {
-			Twinkle.warn.callback.change_subcategory(e);
-		});
+		$('select[name=sub_group]')
+			.chosen({width: '100%', search_contains: true})
+			.change(Twinkle.warn.callback.change_subcategory);
```
Amorymeltzer: I swear, I'll never understand events

Amorymeltzer:
Thanks for this, @siddharthvp

PR #280, opened 2015-06-23
various updates to block module, styling tweak to Morebits

MusikAnimal:

PR #663, opened 2019-06-06
warn/xfd: force chosen select menu to display over the dialog; padding fix in xfd

siddharthvp:
Removes the need to set a max-height to it. Also enables more templates to be viewed without having to scroll.

Code credits: https://github.com/harvesthq/chosen/issues/1390#issuecomment-21397245

Would have added this to #662 hadn't it been merged by now.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1165,11 +1165,15 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 			.chosen({width: '100%', search_contains: true})
 			.change(Twinkle.warn.callback.change_subcategory);
 
-		// Limit the max height of select dropdown to prevent dialog box from becoming scrollable
-		$('.chosen-results').css({'overflow': 'auto', 'max-height': '180px'});
+		mw.util.addCSS(
+			// Force chosen select menu to display over the dialog while overflowing
+			// https://github.com/harvesthq/chosen/issues/1390#issuecomment-21397245
+			'.ui-dialog.morebits-dialog .morebits-dialog-content { overflow:visible !important;}' +
+			'.ui-dialog.morebits-dialog { overflow: inherit !important; }' +
 
-		// Remove padding
-		mw.util.addCSS('.chosen-drop .chosen-results li { padding-top: 0px; padding-bottom: 0px; }');
+			// Remove padding
+			'.morebits-dialog .chosen-drop .chosen-results li { padding-top: 0px; padding-bottom: 0px; }'
+		);
 	}
 };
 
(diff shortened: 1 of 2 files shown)


modules/twinklewarn.js
```diff
@@ -1165,8 +1165,8 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 			.chosen({width: '100%', search_contains: true})
 			.change(Twinkle.warn.callback.change_subcategory);
 
-		// Limit the max height of select dropdown to prevent dialog box from becoming scrollable
-		$('.chosen-results').css({'overflow': 'auto', 'max-height': '180px'});
+		// Force chosen select menu to display over the dialog when it overflows it
```
Amorymeltzer:
```suggestion
		// Force chosen select menu to display over the dialog when it overflows it
		// https://github.com/harvesthq/chosen/issues/1390#issuecomment-21397245
```

modules/twinklewarn.js
```diff
@@ -1165,8 +1165,8 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
...
+		mw.util.addCSS('.ui-dialog .ui-dialog-content { overflow:visible !important;} .ui-dialog { overflow: inherit !important; }')
```
Amorymeltzer:
```suggestion
		mw.util.addCSS('.ui-dialog.morebits-dialog .morebits-dialog-content { overflow:visible !important;} .ui-dialog.morebits-dialog { overflow: inherit !important; }');
```

Amorymeltzer:
Merge it fast, you say!  Wait, there's more, you say!  Sheesh, can't win! 😛 

Anyway, @siddharthvp, shouldn't this kind of thing be in `morebits.css`?  Shouldn't that reduce the need for the inherit?  That may have unintended consequences?  I *think* `overflow: hidden` is from mainly from the skin interface, although d9e6e52 from @MusikAnimal in 2016 (#280) might suggest otherwise?  I can't trigger that issue, so perhaps things have changed upstream?

siddharthvp:
> Merge it fast, you say! Wait, there's more, you say! Sheesh, can't win! 😛

lol, technically i had asked you to make them live, not merge (before the pref option was added). 

> Anyway, @siddharthvp, shouldn't this kind of thing be in morebits.css? Shouldn't that reduce the need for the inherit? That may have unintended consequences? I think overflow: hidden is from mainly from the skin interface, although d9e6e52 from @MusikAnimal in 2016 (#280) might suggest otherwise? I can't trigger that issue, so perhaps things have changed upstream?

meh, now you are analysing this? I was just excited to see that guy's code worked and copy-pasted it blindly. I have no idea what overflow property `ui.dialog` is inheriting and from where. If moving to `morebits.css` won't cause any side-effects, sure we'll move it.

Amorymeltzer:
@siddharthvp I agree this should go out soon, so let's do this: I'll merge this (done) and push it on-wiki (done) then I'll play around with morebits.css on testwiki when I get a moment and move 'em if everything seems copacetic.

PR #664, opened 2019-06-07
warn: increase height of chosen select

siddharthvp:
To match the height of the native select menu. 

Just this one more. Exact same number of templates will now be displayed as the earlier select menu did. This aims at making the transition to the new menu as smooth as possible. Now the reasons why anyone would dislike the new menu are almost null.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1167,9 +1167,13 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 
 		mw.util.addCSS(
 			// Force chosen select menu to display over the dialog while overflowing
-			// https://github.com/harvesthq/chosen/issues/1390#issuecomment-21397245
+			// based on https://github.com/harvesthq/chosen/issues/1390#issuecomment-21397245
 			'.ui-dialog.morebits-dialog .morebits-dialog-content { overflow:visible !important;}' +
 			'.ui-dialog.morebits-dialog { overflow: inherit !important; }' +
+
+			// Increase height to match that of native select
+			'.morebits-dialog .chosen-drop .chosen-results { max-height: 300px; }' +
+			'.morebits-dialog .chosen-drop { height: 338px; }' +
 
 			// Remove padding
 			'.morebits-dialog .chosen-drop .chosen-results li { padding-top: 0px; padding-bottom: 0px; }'


Amorymeltzer:
>match the height of the native select menu

This is going to be very specific @siddharthvp so hardcoding the height probably isn't good.  I'm all for increasing the height, maybe even to match the native, but it's not 300 pixels for everyone.  My old select, for example, goes to the bottom of my browser window.  Wouldn't `max-content` be the correct input here?  I don't know whether that'd be bad form or not, but I would think that'd get the desired results.

siddharthvp:
> My old select, for example, goes to the bottom of my browser window.

wow, does it? So this is the real reason why the two guys at WT:TW were moaning. I see it now that it varies between browsers. In Chrome and Firefox, it goes down to npov1 (though Firefox the length is much larger because it puts a lot of padding just like the original chosen), in Edge it goes all the way to blpprod-1. 

The trouble with `max-content` is that it goes _beyond_ the bottom of my browser window and to reach the lower options I have to scroll the _page_. Shall we call it something like 380px (for `.chosen-result`, then for `.chosen-drop` it'd be 418px) and be done?

> ETA: Would '.morebits-dialog .chosen-drop .chosen-results { max-height: max-content; }' be sufficient, i.e. would it still need '.morebits-dialog .chosen-drop { height: max-content; }'?

setting the rule only for chosen-result would make the text of the lower options appear disembodied without the white background.

Amorymeltzer:
Oy really?  That borks that.  I don't get the disembodied thing with `max-content`, so it seems this all requires a finer touch, esp re #665

Amorymeltzer:
@siddharthvp I've been playing around for a while, not getting much useful, but as regards:

>The trouble with `max-content` is that it goes *beyond* the bottom of my browser window and to reach the lower options I have to scroll the *page*.

I thought that was weird, but I'm on a 13-inch laptop, so I've got my browser fullscreen.  It turns out the old select menu does (and did) this in chrome and safari but not firefox.  So maybe it's not such an issue?

siddharthvp:
I am not sure I follow what you're saying, as replacing the code in this PR with max-content does not fix #665.  I have already described the solution (non-ideal but okay) to #665 over there though I didnt open a pr for that. Suggest making the change directly on the wiki, no need for all the paperwork here. I see that you reverted the whole edit on xfd, it was unnecessary to remove the padding fix. 

I guess there may be some way to get the ideal solution using z-index. https://github.com/harvesthq/chosen/issues/86 is also worth going over closely (though i didn't get any results from a cursory review).

Regarding max-content here: the old select menu reached nowhere close to the bottom on any browser I have. Regardless, do we want the menu to be _that_ large? I think anything between 320px - 380px would be reasonable. Since this is a very important menu, we may as well provide a preferences option for people to customise the size (like we do with CSD window size, and probably to replace the current option of turning chosen off), to account for people with unusual screen sizes.

Amorymeltzer:
My intent there was to respond to your issues with `max-content` here.  At least on my end (chrome/firefox/safari), using `max-content` instead of the hardcoded sizes in this PR doesn't produce any behavior that was any different from how the old select menu, so I was trying to ask you if it behaved appreciably different from the old select menu.  You've now answered yes so it's moot, but I brought up #665 because I was saying that perhaps `max-content` worked fine for me because of the CSS added in #663 that lead to #665, which is to say using `max-content` wouldn't have been fine/sufficient because #665 is indeed an important issue.

As for the chosen issue, yeah, I found that too; nothing too good as you say.  I briefly tried some `z-index` stuff but didn't much of anywhere.

Amorymeltzer:
Putting this up now as an improvement, but I'm gonna keep trying for a better way to do this all.  In the end, using `chosen` means that the advantage of having a large menu is diminished, so unless the goal is to undo #661, some of this mishegoss is likely unnecessary.

Amorymeltzer:
@siddharthvp Can we go over the heights here again?  I'm curious what you're seeing since it seems our browsers are behaving quite differently.  On my end, the current setup (in `master`, from this PR) results in a bunch of white space when there are only a few results.  This is because `'.morebits-dialog .chosen-drop { height: 338px; }'` uses `height` instead of `max-height`, but are you seeing the same thing?

Furthermore, does your browser need the heights to be different?  Everything shows up just fine for me AFAICT when they're the same.

siddharthvp:
>  On my end, the current setup (in master, from this PR) results in a bunch of white space when there are only a few results. This is because '.morebits-dialog .chosen-drop { height: 338px; }' uses height instead of max-height, but are you seeing the same thing?

That's a mistake, yeah. Should be changed to max-height, though the whitespace didn't initially strike to me as odd.

> Furthermore, does your browser need the heights to be different? Everything shows up just fine for me AFAICT when they're the same.

What heights are we talking about? The ones for chosen-drop and chosen-results? If so, yes. See https://ibb.co/fXdm80T. (when both are 300px)

Amorymeltzer:
And that's chrome, yeah?  Wild.

siddharthvp:
yeah. But it is to be expected, isn't it, since chosen drop is made up of chosen results plus the search box. The search box along with surrounding padding takes abt 38px, thus requiring the drop to be abt 38px higher.

Issue #665, opened 2019-06-07
Warning module: warning information frame doesn't resize with window lower edge


creffett:
The lower edge of the "warning information" frame in the warning module doesn't seem to be attached to the outer frame, so you can resize the outer frame left/right and it will grow or shrink appropriately, but resizing up/down does nothing. Normally this would just be a cosmetic issues, but longer previews (e.g. spam-2) cover up the submit button until the window is resized.

Steps to reproduce:
1. Open the warnings module
2. Select {{uw-spam2}}
3. Click `preview`
4. Notice that the submit button is no longer visible
5. Resize the window vertically, notice that the "warning information" frame does not resize with the rest of the frame.

siddharthvp:
Can be fixed by adding `max-height: max-content` to the existing css style for `.ui-dialog.morebits-dialog .morebits-dialog-content`. Not ideal, as it increases the height of the dialog rather providing scrolling, but can be incorporated temporarily until we can come up with something better. @Amorymeltzer

Amorymeltzer:
@siddharthvp Or we just revert #663, which introduced this, and reassess css more carefully.

siddharthvp:
> @siddharthvp Or we just revert #663, which introduced this, and reassess css more carefully.

We could revert the stuff in xfd, coz no one cares much abt the delsort stuff, and because afd nominations can be long and increased size of preview with no in-dialog scrolling will be annoying. But in Warn, I suspect reverting #663 would only drive more folks to turn it off altogether. Also the problem here is less severe since there are just a few long templates and that they are not too long to make the absence of in-dialog scrolling an issue.

Amorymeltzer:
I could've sworn your [proposed change](https://github.com/azatoth/twinkle/issues/665#issuecomment-499905637) was working for me earlier, but it's not now.

This issue will crop up upon resize or for smaller window sizes.

siddharthvp:
Trying refreshing the page a couple of times or clearing the cache. CSS changes aren't easily picked up by browser from my experience in working on this for the last few days.

Amorymeltzer:
@siddharthvp What about undoing the additions from #663 and adding this before the `addCSS`:

```javascript
$('select[name=sub_group]').on('chosen:showing_dropdown', function () {
	$(this).parents('div').css('overflow', 'visible');
});
$('select[name=sub_group]').on('chosen:hiding_dropdown', function () {
	$(this).parents('div').css('overflow', '');
});
```

[...]

PR #667, opened 2019-06-08
warn/xfd: chosen menu css fixes

siddharthvp:
Solves #665. Uses chosen:showing_dropdown and chosen:hiding_dropdown events to dynamically manipulate css properties.

As discussed: https://github.com/azatoth/twinkle/issues/665#issuecomment-500077949 and https://github.com/azatoth/twinkle/issues/665#issuecomment-500099207

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1163,15 +1163,18 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	if (!Twinkle.getPref('oldSelect')) {
 		$('select[name=sub_group]')
 			.chosen({width: '100%', search_contains: true})
-			.change(Twinkle.warn.callback.change_subcategory);
+			.change(Twinkle.warn.callback.change_subcategory)
+			.on('chosen:showing_dropdown', function () {
+				$(this).parents('div').css('overflow', 'visible');
+				$('#twinklewarn-previewbox').css({ 'max-height': '100px', 'overflow': 'hidden' });
+			})
+			.on('chosen:hiding_dropdown', function () {
+				$(this).parents('div').css('overflow', '');
+				$('#twinklewarn-previewbox').css({ 'max-height': '', 'overflow': '' });
+			});
 
 		mw.util.addCSS(
-			// Force chosen select menu to display over the dialog while overflowing
-			// based on https://github.com/harvesthq/chosen/issues/1390#issuecomment-21397245
-			'.ui-dialog.morebits-dialog .morebits-dialog-content { overflow:visible !important;}' +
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
@siddharthvp I get weird jumpiness of the select menu with this.  Previewing something like `uw-spam2` then scrolling down slightly before clicking on the menu again makes the content, including the menu jump, and can actually make it so I can't click on the menu to open it.

modules/twinklewarn.js
```diff
@@ -1163,15 +1163,18 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
...
 			.chosen({width: '100%', search_contains: true})
-			.change(Twinkle.warn.callback.change_subcategory);
+			.change(Twinkle.warn.callback.change_subcategory)
+			.on('chosen:showing_dropdown', function () {
+				$(this).parents('div').css('overflow', 'visible');
+				$('#twinklewarn-previewbox').css({ 'max-height': '100px', 'overflow': 'hidden' });
+			})
+			.on('chosen:hiding_dropdown', function () {
+				$(this).parents('div').css('overflow', '');
+				$('#twinklewarn-previewbox').css({ 'max-height': '', 'overflow': '' });
```
Amorymeltzer: Shouldn't this be `visible`?
```suggestion
				$('#twinklewarn-previewbox').css({ 'max-height': '', 'overflow': 'visible' });
```

Amorymeltzer:
I should add that the `$('#twinklewarn-previewbox')` bit only works assuming you don't have a different size dialog

Amorymeltzer:
Basically, if you're scrolled down enough such that clicking the chosen container makes the whole thing jump so much that your cursor is no longer within the boundary of the chosen container, it will close upon cursor release.  I don't know why `$(this).parents('div').css('overflow', 'visible');` is making the dialog jump.

siddharthvp:
I see the weird jumpiness persists even after removing the previewbox css changes that I added. I thought it'd be fixable by preventing the default event and scrolling the dialog to the top (`.scrollTo(0,0)`), and then triggering the original event. But `e.preventDefault` within the `chosen:dropdown_showing` handler is not blocking the dropdown from showing. So this expt goes nowhere. 

At this point, the next best fix is https://github.com/azatoth/twinkle/issues/665#issuecomment-499905637, or is it still not working for you? I forgot to mention earlier that `!important` would be needed, maybe that's why it wasn't working for u earlier. So the rule in the first line in mw.util.addCSS becomes: 
```
'.ui-dialog.morebits-dialog .morebits-dialog-content { overflow: visible !important; max-height: max-content !important; }'
```

PR #668, opened 2019-06-08
tag: quick filter: minor changes

siddharthvp:
- Prevent enter key from submitting form
- Disable browser input suggestions

The latter is annoying. The suggestions sometimes obscure search results. This setting is there by default in chosen's search inputs.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -207,6 +207,13 @@ Twinkle.tag.callback = function friendlytagCallback() {
 
 	if (Twinkle.tag.mode === "article") {
 
+		result.quickfilter.autocomplete = 'off'; // disable browser suggestions for this field
+		result.quickfilter.addEventListener('keypress', function(e) {
+			if (e.keyCode === 13) { // prevent enter key from submitting
+				e.preventDefault();
+				return false;
+			}
+		});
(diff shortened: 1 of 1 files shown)


modules/friendlytag.js
```diff
@@ -207,6 +207,13 @@ Twinkle.tag.callback = function friendlytagCallback() {
 
 	if (Twinkle.tag.mode === "article") {
 
+		result.quickfilter.autocomplete = 'off'; // disable browser suggestions for this field
+		result.quickfilter.addEventListener('keypress', function(e) {
+			if (e.keyCode === 13) { // prevent enter key from submitting
+				e.preventDefault();
+				return false;
```
siddharthvp: This couldn't have been added to the main event function because that's a `keyup` handler, and the form would have already been submitted by the time of `keyup`.

siddharthvp: This couldn't have been added to the main event function because that's a `keyup` handler, and the form would have already been submitted by the time of `keyup`.

siddharthvp:
Oh, totally forgot about this 4-month-old PR. This is redone in #724, as you've already seen. Sorry. I remembered planning to make these changes shortly after quick filter was added, but didn't remember actually having done it.

PR #670, opened 2019-06-09
ARV: Allow linking to redirects

DannyS712:
So that reports such as https://w.wiki/4mC can work properly. Also eliminates the need to process categories and images to ensure that they are linked. {{-r}} used rather than {{no redirect}} to simplify wikitext.

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -533,7 +533,7 @@ Twinkle.arv.callback.evaluate = function(e) {
 			if ( form.page.value !== '' ) {
 
 				// Allows linking to redirects, file and category links work as links by default
-				reason = 'On {{no redirect|' + form.page.value + '}}';
+				reason = 'On {{:no redirect|' + form.page.value + '}}';
 
 				if ( form.badid.value !== '' ) {
 					reason += ' ({{diff|' + form.page.value + '|' + form.badid.value + '|' + form.goodid.value + '|diff}})';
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I can fix it up for you, if you like

DannyS712:
I changed it to use the actual template. What do you mean by the need for an extra :?

Amorymeltzer:
To avoid a ping if it's to a user page, apparently this was missed when those updates were applied.

DannyS712:
But using {{no redirect}} means that it won't ping - see https://w.wiki/4vq - did you get pinged?

Amorymeltzer:
You have to sign for a ping to work

DannyS712:
Yeah, I realized - I've added the : - can you rebase? Thanks

siddharthvp:
For the sake of wikitext sanity, I think 
```javascript
reason = 'On {{-r|' + form.page.value.replace(/^User:/, ':User:') + '}}';
```
 would be ideal. Uses the concise the template shortcut as originally suggested by Danny instead of `no redirect` ,  the latter is anyway confusing as it leads one to think the page is a redirect. And adds the colon only if it's a user page.

Amorymeltzer:
Bikeshedding, as this is AIV so nearly nobody is looking at anyone else's wikitext and there shouldn't be much anyway, but `{{-r}}` is inscrutable, even to me as someone who regularly works with redirects.  And there's no real value in doing the work to only prepend the colon for userspace, it's fairly universally done elsewhere (inlcuding throughout Twinkle) with no ill affects.

PR #675, opened 2019-06-13
morebits: quickform input and textarea improvements

siddharthvp:
Apply classes, styles and id on the input/textarea rather than the container div, as this is more desirable in most situations. 

For input fields, fix the ancient bug that caused the container and the input field to get the same id; and enable support for custom id.

For textarea, nest a `<label>` within the h5, as requested in source TODO. This causes no visible changes AFAICT, other than that clicking on the label now will put cursor in the textarea.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -433,7 +433,7 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 			if (data.label) {
 				label = node.appendChild(document.createElement('label'));
 				label.appendChild(document.createTextNode(data.label));
-				label.setAttribute('for', id);
+				label.setAttribute('for', data.id || id);
 			}
 
 			subnode = node.appendChild(document.createElement('input'));
@@ -441,7 +441,6 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 				subnode.setAttribute('value', data.value);
 			}
 			subnode.setAttribute('name', data.name);
-			subnode.setAttribute('id', id);
 			subnode.setAttribute('type', 'text');
 			if (data.size) {
 				subnode.setAttribute('size', data.size);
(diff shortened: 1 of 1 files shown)


siddharthvp:
Strictly speaking these are breaking changes, as id, class and style are applied now to the input/textarea rather than its parent. However, I could find no usages at all of those attributes being applied to any `input` or `textarea` either in Twinkle or in the 51 enwiki userspace script pages (apart from my own scripts) that use Morebits.quickform.

Amorymeltzer:
>I could find no usages at all of those attributes being applied [snip] (apart from my own scripts)

What were you doing with them?  These unreliably named, aren't they?  I'm curious.

>This causes no visible changes AFAICT, other than that clicking on the label now will put cursor in the textarea.

Surprisingly cool!

Amorymeltzer:
Will merge after #729 solely to make my life moderately easier this coming week.

siddharthvp:
> What were you doing with them? 

classes and styles are being applied to `input`s and `textarea`s in my [DYK-helper](https://en.wikipedia.org/wiki/User:SD0001/DYK-helper.js) and [RFUD-helper](https://en.wikipedia.org/wiki/User:SD0001/RFUD-helper.js) scripts. Good that you asked, else I may have forgotten to update those scripts. 

>These unreliably named, aren't they? 

These attributes may be applied to the element itself or to its container div (depending on the type of element) so yeah it's a bit unreliable - at least you have to consult the morebits code every time to see what's going to happen. 

Once this and other morebits-related PRs are merged, I have plans to create an on-wiki documentation page for morebits, so that more people can learn to use it. It's a really cool library. It enabled me recently to create [easy-merge](https://en.wikipedia.org/wiki/User:SD0001/easy-merge) and [BDCS](https://en.wikipedia.org/wiki/User:SD0001/BDCS) scripts (per user requests) that each took _less than 2 hours_ to write! Not a single line of html code, not a single line of custom css, not even a single api query object - morebits handles everything. It feels a lot easier than OOUI (though in truth I haven't spent much time trying to figure out OOUI) - but OOUI scripts need larger amount of code, and produce ridiculously large number of rendered html elements.

PR #676, opened 2019-06-14
warn: Fix decade-old bug in regex parsing sig timestamps

Amorymeltzer:
The comma in the `HH:MM, DD Month YYYY` regex leads to this being parsed as an invalid date.  This has been unchanged since its inception in 2007, did this ever work?  This was masking a bug with the user-aborted warnings.

I'm tried running down how timestamps and signatures are created in mediawiki, I've read far too much of the ECMAScript language specifications, and I've spent a lot of time in various languages, but I still feel like I'm missing something obvious.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1276,16 +1276,17 @@ Twinkle.warn.callbacks = {
 	},
 	main: function(pageobj) {
 		var text = pageobj.getPageText();
+		var statelem = pageobj.getStatusElement();
 		var params = pageobj.getCallbackParameters();
 		var messageData = params.messageData;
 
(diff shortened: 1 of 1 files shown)


MusikAnimal:
You are correct that `HH:MM, DD Month YYYY` will parse as an invalid date. It's possible back in 2007 that it did, however (browser quirks, maybe?), and that's my best guess. I haven't done testing of any sort, but going by the code, I don't think you're missing anything. The old implementation simply shouldn't work.

Amorymeltzer:
Thanks for the sanity check.  That was my thought too, but I couldn't find anything in old ES specs before my eyes melted.  I'm guessing nobody noticed since it would just silently pass (yay javascript 👿 )

PR #679, opened 2019-06-18
morebits: pass api object to failure handlers wherever possible

siddharthvp:
Pass the `morebits.wiki.api` object (rather than the `morebits.wiki.page` object) to the failure callbacks wherever possible, except where the error is being handled before any API calls are made (such as when trying to protect without being admin or without setting the protection settings). The apiobj is more useful since it has the `errorCode`, using which the client script may choose to show customised error messages. The pageobj is anyway still accessible as it is set as the `parent` attribute of apiobj in all usages.

Regarding the success callbacks, `save()` is currently giving the pageobj (useful to allow further processing with the page) while the other functions are giving the apiobj. No changes made here.

See individual commit messages for more details. 

This does not break any existing code.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2650,7 +2650,7 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		// force error to stay on the screen
 		++Morebits.wiki.numberOfActionsLeft;
 
-		ctx.onSaveFailure(this);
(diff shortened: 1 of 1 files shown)


siddharthvp:
Closing in favour of a superior solution described in #783.

PR #682, opened 2019-06-20
morebits: fix handling of invalid token errors

siddharthvp:
Improve handling of token errors in `morebits.wiki.page`'s `save`, `deletePage` and `undeletePage`. 

I believe the 3rd commit fixes the probably decade-old bug (oh, again!) described [here](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_40#Twinkle_broken?). 

Though I am not sure whether these changes really work, as there's no easy way to test them.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2698,16 +2698,12 @@ Morebits.wiki.page = function(pageName, currentAction) {
 			}
 
 		// check for loss of edit token
-		// it's impractical to request a new token here, so invoke edit conflict logic when this happens
-		} else if (errorCode === 'notoken' && ctx.conflictRetries++ < ctx.maxConflictRetries) {
+		} else if ((errorCode === 'badtoken' || errorCode === 'notoken') && ctx.retries++ < ctx.maxRetries) {
 
 			ctx.statusElement.info('Edit token is invalid, retrying');
 			--Morebits.wiki.numberOfActionsLeft;  // allow for normal completion if retry succeeds
-			if (fnCanUseMwUserToken('edit')) {
-				this.load(fnAutoSave, ctx.onSaveFailure); // try the append or prepend again
-			} else {
-				ctx.loadApi.post(); // reload the page and reapply the edit
-			}
+			ctx.saveApi.query.token = fnGetToken.call(this);
+			ctx.saveApi.post();
 
 		// check for network or server error
 		} else if (errorCode === 'undefined' && ctx.retries++ < ctx.maxRetries) {
--- a/morebits.js
+++ b/morebits.js
@@ -2600,6 +2600,24 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		return true; // all OK
 	};
 
+	// helper function to get a new token on encountering token errors
+	// in save, deletePage, and undeletePage
+	// Being a synchronous ajax call, this blocks the event loop,
+	// and hence should be used sparingly.
+	var fnGetToken = function() {
+		var token;
+		var tokenApi = new Morebits.wiki.api('Getting token', {
+			action: 'query',
+			meta: 'tokens'
+		}, function(apiobj) {
+			token = $(apiobj.responseXML).find('tokens').attr('csrftoken');
+		}, null, function() {
+			this.getStatusElement().error('Failed to get token');
+		});
+		tokenApi.post({async: false});
+		return token;
+	};
+
 	// callback from saveApi.post()
 	var fnSaveSuccess = function() {
 		ctx.editMode = 'all';  // cancel append/prepend/revert modes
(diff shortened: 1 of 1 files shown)


siddharthvp:
I can confirm that this is working (at least for the `save` part). A page that I had left idle for a long time had got its edittoken expired resulting in failure of all morebits page actions. However, on pasting the morebits.js from this PR into the console, and retrying the morebits actions, they worked.

morebits.js
```diff
@@ -2681,15 +2695,12 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
 		// check for loss of edit token
 		// it's impractical to request a new token here, so invoke edit conflict logic when this happens
-		} else if (errorCode === 'notoken' && ctx.conflictRetries++ < ctx.maxConflictRetries) {
+		} else if ((errorCode === 'badtoken' || errorCode === 'notoken') && ctx.conflictRetries++ < ctx.maxConflictRetries) {
 
 			ctx.statusElement.info('Edit token is invalid, retrying');
 			--Morebits.wiki.numberOfActionsLeft;  // allow for normal completion if retry succeeds
-			if (fnCanUseMwUserToken('edit')) {
-				this.load(fnAutoSave, ctx.onSaveFailure); // try the append or prepend again
-			} else {
```
Amorymeltzer: I'm not entirely clear on this bit.  `fnCanUseMwUserToken` is [weird and not very well named](https://github.com/azatoth/twinkle/pull/616#discussion_r281037775) but wasn't the intent here to merely retry the pre/append again?  I don't see how that would help with badtoken issues, so changing it makes sense, I'm just a little confused on the path where the suggestion fully replaces the (assumed) previous behavior.

Amorymeltzer:
I'm sure this will be appreciated!🏅  These might be rare but they've got to be one of the more confusing errors for the average user.

Amorymeltzer:
Oh, I should add that, for testing purposes, logging in and out should invalidate these tokens, which should be somewhat easier than just waiting.  Going put my 2FA through the ringer I guess...

morebits.js
```diff
@@ -2600,6 +2600,20 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		return true; // all OK
 	};
 
+	var fnGetToken = function() {
```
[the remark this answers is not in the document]
siddharthvp: done, also added a note about this being a sync function. Also realized that the original commit message was wrong - async: false means that is _not_ an _async_ function.

morebits.js
siddharthvp: I don't understand your comment, or for that matter, the usage of `fnCanUseMwUserToken`, but I think this change here is clear. 

The call `this.load` requests a new copy of the latest stored revision from the server. Its callback `fnAutoSave` just calls `this.save` which constructs the `saveApi `query and posts the api request. _But all this was unnecessary_. We already had a copy of the latest stored revision, we had already constructed the `saveApi` query object, before encountering this error. Now, since this is a token error, the only change that needs to be made to the `saveApi.query` is object to update the token, and just repost it.  I don't see how it makes a difference whether we are prepending/appending/normal editing since the final step - creating `saveApi.query` and posting it - is the same across all of them.

So, this change is an optimisation that does away with the repeated execution of `this.load`, an unnecessary api call, and repeated execution of the `this.save`. 

Do I make sense? Or is there a fallacious gap in my reasoning?

Amorymeltzer: No, that matches with what I can reason (doesn't mean we're not *both* wrong, I suppose).  My concern with `this.load` -> `saveApi` would be handling any intervening edits/etc. to the page, but I can't seem to trigger that?

What I was trying to ask is basically what you yourself seem confused by, so at least we're on the same page.  Treating prepend/append differently makes sense, but the whole point AFAICT is to avoid loading the page unnecessarily, so I likewise am unsure what exactly `fnCanUseMwUserToken` was really doing here.  Something to do with redirect resolution, perhaps?

siddharthvp: > this.load -> saveApi would be handling any intervening edits/etc.

Yes. But if there were intermediate edits to the page, they will be handled anyway via edit conflict resolution mechanism of the next attempt. 

> Something to do with redirect resolution, perhaps?

No, since if editmode is prepend/append, redirect resolution is handled by the api server itself. (according to the comment inside `fnCanUseMwUserToken`). In `this.append` & `this.prepend`, the only way `if (fnCanUseMwUserToken('edit'))` can be true is if we need to fetch protection status. But it would already have been fetched (and the sysop would have had went past the confirmation) before we encountered the error. In the present setup, it looks as if sysops need to confirm the edit to a protected page a second time if the original attempt failed due to a token error. I don't think that is desirable.

siddharthvp:
The force push removes the `// it's impractical to request a new token here, so invoke edit conflict logic when this happens` comment (no longer applicable) and changes `if ((errorCode === 'badtoken' || errorCode === 'notoken') && ctx.conflictRetries++ < ctx.maxConflictRetries)` to `if ((errorCode === 'badtoken' || errorCode === 'notoken') && ctx.retries++ < ctx.maxRetries)`.

PR #686, opened 2019-08-08
Show G14 on redirects

galobtter:
G14 also applies to redirects per discussion at https://en.wikipedia.org/wiki/Wikipedia_talk:Criteria_for_speedy_deletion/Archive_74#Modified_proposal (and see https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#G14_speedy_deletion_criterion_(add_for_redirects?) )

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1056,8 +1056,7 @@ Twinkle.speedy.generalList = [
 	{
 		label: 'G14: Unnecessary disambiguation page',
 		value: 'disambig',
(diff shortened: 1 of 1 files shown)


galobtter:
Need to wait on [https://en.wikipedia.org/wiki/Wikipedia_talk:Criteria_for_speedy_deletion#RfC:_R5:_Redirects_ending_with_%22(disambiguation)%22_that_do_not_target_a_disambiguation_page](this discussion) about splitting off the portion of G14 that applies to redirects to another criterion.

Amorymeltzer:
@galobtter I don't think we need to wait; it's accurate as is, and if it's pulled out into R5, we can change then.  To close the loop, this was an oversight on my part in #684

galobtter:
Yeah, my thinking was that the discussion would be closed soon enough that there would be no point in making the change until it closed, but it has been nearly a month, so yeah, there's no need to wait.

Amorymeltzer:
Not sure it's going to go anywhere anyway (then again I'm biased)

PR #692, opened 2019-08-31
warn/xfd: Replace jquery.chosen menus with select2

### LINKED ISSUE #665: Warning module: warning information frame doesn't resize with window lower edge

siddharthvp:
Replace jQuery Chosen select menu with [select2](https://select2.org/). 

This fully fixes the bug [described here](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_41#Long_previews_overlap_and_block_%22Submit%22_button) (closes #665), also eliminating the stop-gap solution of removing the resize handles. The problems associated with using jquery.chosen within a modal dialog (like in all TW scripts) are not there in select2. Also, select2 is better both in terms of customizability and programmatic control, not to mention having a proper documentation (unlike chosen). Select2 unfortunately doesn't natively support highlighting of search results, so I have implemented that manually (as I did in Tag module).

Select2 is being loaded using the CDN, which is obviously non-ideal, especially since the support for loading of scripts from non-WMF sites may likely be removed in the future. Before that happens, we should get select2 added to the ResourceLoader, maybe as a replacement to chosen.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -15,10 +15,17 @@
  */
 
 Twinkle.warn = function twinklewarn() {
+
+	mw.loader.load('https://tools-static.wmflabs.org/cdnjs/ajax/libs/select2/4.0.10/js/select2.min.js');
+	mw.loader.load('https://tools-static.wmflabs.org/cdnjs/ajax/libs/select2/4.0.10/css/select2.min.css', 'text/css');
+
 	if (mw.config.get('wgRelevantUserName')) {
 		Twinkle.addPortletLink(Twinkle.warn.callback, 'Warn', 'tw-warn', 'Warn/notify user');
-		if (Twinkle.getPref('autoMenuAfterRollback') && mw.config.get('wgNamespaceNumber') === 3 &&
-				mw.util.getParamValue('vanarticle') && !mw.util.getParamValue('friendlywelcome') && !mw.util.getParamValue('noautowarn')) {
+		if (Twinkle.getPref('autoMenuAfterRollback') &&
+			mw.config.get('wgNamespaceNumber') === 3 &&
+			mw.util.getParamValue('vanarticle') &&
+			!mw.util.getParamValue('friendlywelcome') &&
+			!mw.util.getParamValue('noautowarn')) {
 			Twinkle.warn.callback();
 		}
 	}
@@ -61,7 +68,7 @@ Twinkle.warn = function twinklewarn() {
 
 Twinkle.warn.callback = function twinklewarnCallback() {
 	if (mw.config.get('wgRelevantUserName') === mw.config.get('wgUserName') &&
-			!confirm('You are about to warn yourself! Are you sure you want to proceed?')) {
+		!confirm('You are about to warn yourself! Are you sure you want to proceed?')) {
 		return;
 	}
 
@@ -1069,9 +1076,6 @@ Twinkle.warn.prev_article = null;
 Twinkle.warn.prev_reason = null;
 
 Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCategory(e) {
-	if (!Twinkle.getPref('oldSelect')) {
-		$('select[name=sub_group]').chosen('destroy');
-	}
 
 	var value = e.target.value;
 	var sub_group = e.target.root.sub_group;
@@ -1157,30 +1161,69 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	// Trigger custom label/change on main category change
 	Twinkle.warn.callback.change_subcategory(e);
 
-	// Build select menu with jquery.chosen
+	// Build select menu with select2
 	if (!Twinkle.getPref('oldSelect')) {
 		$('select[name=sub_group]')
-			.chosen({width: '100%', search_contains: true})
-			.change(Twinkle.warn.callback.change_subcategory);
+			.select2({ width: '100%' })
+			.change(Twinkle.warn.callback.change_subcategory)
+
+			// select2 natively doesn't support highlightening of search hits so we implement that here
+			.on('select2:open', function() {
+				$('.select2-search__field')[0].addEventListener('keyup', function() {
+					var $ul = $('.select2-results__options');
+					$ul.find('.search-hit').each(function(i, e) {
+						var li_element = e.parentElement;
+						// This would convert <li>Hello <span class=search-hit>wo</span>rld</li>
+						// to <label>Hello world</label>
+						li_element.innerHTML = li_element.textContent;
+					});
+
+					if (this.value) {
+						var searchString = this.value;
+						var searchRegex = new RegExp(mw.RegExp.escape(searchString), 'i');
+
+						var $selector;
+						// if main_group is level1/2/3/4/4im, the ul has option groups,
+						// select the group header (li > strong) and the list elements (li > ul > li),
+						// the odd way in which select2 organises the ul necessitates this hack
+						if ($('select[name=main_group]').val().indexOf('level') === 0) { // can probably write as this.value.indexOf('level') === 0
+							$selector = 'li > strong, li > ul > li';
+						} else {
+							$selector = 'li';
+						}
+
+						$ul.find($selector).each(function() {
+							var li_text = this.textContent;
+							var searchHit = searchRegex.exec(li_text);
+							if (searchHit) {
+								var range = document.createRange();
+								var textnode = this.childNodes[0];
+								range.selectNodeContents(textnode);
+								range.setStart(textnode, searchHit.index);
+								range.setEnd(textnode, searchHit.index + searchString.length);
+								var underline_span = $('<span>').addClass('search-hit').css('text-decoration', 'underline')[0];
+								range.surroundContents(underline_span);
+							}
+						});
+					}
+				});
+			});
 
 		mw.util.addCSS(
-			// Force chosen select menu to display over the dialog while overflowing
-			// based on https://github.com/harvesthq/chosen/issues/1390#issuecomment-21397245
-			'.ui-dialog.morebits-dialog .morebits-dialog-content { overflow:visible !important; }' +
-			'.ui-dialog.morebits-dialog { overflow: inherit !important; }' +
+			// prevent dropdown from appearing behind the dialog
+			'.select2-container { z-index: 10000; }' +
 
-			// Increase height to match that of native select
-			'.morebits-dialog .chosen-drop .chosen-results { max-height: 300px; }' +
-			'.morebits-dialog .chosen-drop { max-height: 338px; }' +
+			// Increase height
+			'.select2-container .select2-dropdown .select2-results > .select2-results__options { max-height: 350px; }' +
 
-			// Remove padding
-			'.morebits-dialog .chosen-drop .chosen-results li { padding-top: 0px; padding-bottom: 0px; }'
+			// Reduce padding
+			'.select2-results .select2-results__option { padding-top: 1px; padding-bottom: 1px; }' +
+			'.select2-results .select2-results__group { padding-top: 1px; padding-bottom: 1px; } ' +
+
+			// Reduce font size
+			'.select2-container .select2-dropdown .select2-results { font-size: 13px; }' +
+			'.select2-container .selection .select2-selection__rendered { font-size: 13px; }'
 		);
-		// (Hopefully) temporary fixes for visual issues caused by the above CSS, see #665 and #667
-		$('.ui-resizable-handle').remove();
-		// Jump to extend the dialog to include any previewing content
-		// without scrolling and prevent resizing
-		mw.util.addCSS('.ui-dialog.morebits-dialog .morebits-dialog-content { max-height: max-content !important; }');
 
 	}
 };
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1206,18 +1206,20 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	// Use select2 to make the select menu searchable
 	if (!Twinkle.getPref('oldSelect') && $.fn.select2) {
 		$('select[name=sub_group]')
-			.select2({ width: '100%' })
+			.select2({
+				width: '100%',
+				matcher: Morebits.select2.matcher,
+				templateResult: Morebits.select2.highlightSearchMatches,
+				language: {
+					searching: Morebits.select2.queryInterceptor
+				}
+			})
 			.change(Twinkle.warn.callback.change_subcategory);
 
-		// if main_group is level1/2/3/4/4im, the ul has option groups,
-		if ($('select[name=main_group]').val().indexOf('level') === 0) {
-			Morebits.select2SearchHighlights($('select[name=sub_group]'), true);
-		} else {
-			Morebits.select2SearchHighlights($('select[name=sub_group]'), false);
-		}
+		$('.select2-selection').keydown(Morebits.select2.autoStart);
 
 		mw.util.addCSS(
-			// prevent dropdown from appearing behind the dialog
+			// prevent dropdown from appearing behind the dialog, just in case
 			'.select2-container { z-index: 10000; }' +
 
 			// Increase height
@@ -1227,7 +1229,7 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 			'.select2-results .select2-results__option { padding-top: 1px; padding-bottom: 1px; }' +
 			'.select2-results .select2-results__group { padding-top: 1px; padding-bottom: 1px; } ' +
 
-			// Reduce font size
+			// Adjust font size
 			'.select2-container .select2-dropdown .select2-results { font-size: 13px; }' +
 			'.select2-container .selection .select2-selection__rendered { font-size: 13px; }'
 		);
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1158,7 +1158,7 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	// Trigger custom label/change on main category change
 	Twinkle.warn.callback.change_subcategory(e);
 
-	// Build select menu with select2
+	// Use select2 to make the select menu searchable
 	if (!Twinkle.getPref('oldSelect') && $.fn.select2) {
 		$('select[name=sub_group]')
 			.select2({ width: '100%' })
@@ -1166,11 +1166,10 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 
 		// if main_group is level1/2/3/4/4im, the ul has option groups,
 		if ($('select[name=main_group]').val().indexOf('level') === 0) {
-			mw.select2SearchHighlights($('select[name=sub_group]'), true);
+			Morebits.select2SearchHighlights($('select[name=sub_group]'), true);
 		} else {
-			mw.select2SearchHighlights($('select[name=sub_group]'), false);
+			Morebits.select2SearchHighlights($('select[name=sub_group]'), false);
 		}
-
 
 		mw.util.addCSS(
 			// prevent dropdown from appearing behind the dialog
@@ -1187,7 +1186,6 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 			'.select2-container .select2-dropdown .select2-results { font-size: 13px; }' +
 			'.select2-container .selection .select2-selection__rendered { font-size: 13px; }'
 		);
-
 	}
 };
 
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -16,10 +16,6 @@
 
 Twinkle.warn = function twinklewarn() {
 
-	if (!$.fn.select2) {
-		mw.loader.load('https://tools-static.wmflabs.org/cdnjs/ajax/libs/select2/4.0.10/js/select2.min.js');
-		mw.loader.load('https://tools-static.wmflabs.org/cdnjs/ajax/libs/select2/4.0.10/css/select2.min.css', 'text/css');
-	}
 	if (mw.config.get('wgRelevantUserName')) {
 		Twinkle.addPortletLink(Twinkle.warn.callback, 'Warn', 'tw-warn', 'Warn/notify user');
 		if (Twinkle.getPref('autoMenuAfterRollback') &&
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -16,9 +16,10 @@
 
 Twinkle.warn = function twinklewarn() {
 
-	mw.loader.load('https://tools-static.wmflabs.org/cdnjs/ajax/libs/select2/4.0.10/js/select2.min.js');
-	mw.loader.load('https://tools-static.wmflabs.org/cdnjs/ajax/libs/select2/4.0.10/css/select2.min.css', 'text/css');
-
+	if (!$.fn.select2) {
+		mw.loader.load('https://tools-static.wmflabs.org/cdnjs/ajax/libs/select2/4.0.10/js/select2.min.js');
+		mw.loader.load('https://tools-static.wmflabs.org/cdnjs/ajax/libs/select2/4.0.10/css/select2.min.css', 'text/css');
+	}
 	if (mw.config.get('wgRelevantUserName')) {
 		Twinkle.addPortletLink(Twinkle.warn.callback, 'Warn', 'tw-warn', 'Warn/notify user');
 		if (Twinkle.getPref('autoMenuAfterRollback') &&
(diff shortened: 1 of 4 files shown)


Amorymeltzer:
Just noting that I agree with @MusikAnimal on this.  Loading external code from non-WMF sites is something I think Twinkle users would be fairly uncomfortable with, and given the wide usage of TW, simply isn't safe.

siddharthvp:
@MusikAnimal any suggestion on how to move ahead with this, considering that we can all agree that is an improvement. (Also, the delsort stuff added in #143 is turning out to be widely popular among users - it's time we provide a more pleasant UX for that.)

Why isn't wmflabs.org a whitelisted domain, when the [Toolforge CDN homepage](https://tools.wmflabs.org/cdnjs/) says it's a "Collection of very useful JS / CSS libraries served in a fast, _Wikimedia Privacy Policy friendly way_." ?

Should I open a phab ticket for inclusion of select2 into ResourceLoader? Or would it be shot down with the standard "please use OOUI instead" response? I am not very sure whether OOUI can be used to do what we're doing with select2 here, but select2 definitely can be used for a lot of stuff that can't be done using OOUI at present.

Amorymeltzer:
>Why isn't wmflabs.org a whitelisted domain

[T223840: Can/should *.wmflabs.org be added to the default-src Content Security Policy?](https://phabricator.wikimedia.org/T223840)

siddharthvp:
> T223840

Didn't go over it thoroughly. Would the reasons why *.wmflabs.org is disallowed also apply to tools-static.wmflabs.org, noting that tools-static only hosts trustworthy libraries and no user-created tools?

Amorymeltzer:
There's arguments to be made for whitelisting certain domains, but to some folks * means * so ¯\_(ツ)_/¯.  I suppose, per @MusikAnimal, the interim use of the cdn is okay-ish?  I dunno, I don't think the CSP is going to get implemented for a while, as there are plenty of *.wmflabs things that are relied upon.

MusikAnimal:
I also don't think the CSP headers are going to be implemented in the very near future, or at least we'll have some sort of solution for whitelisting domains by then (and I concur /cdnjs should be whitelisted for everyone).

I should note tools-static.wmflabs.org does go down sometimes... but lots of things break when this happens. The only weird bit here is that Twinkle is an on-wiki gadget, so one wouldn't expect it to break when everything on-wiki is otherwise working fine.

I guess we _could_ upload the select2 assets to the wiki, and declare them as a dependency in the gadget definition. This would put them behind ResourceLoader. However I'm not sure how acceptable this practice is, as it is third-party code. I'm assuming the licensing is fine, so long as it's prominently stated at the top, otherwise it inherently is published under CC-BY-SA.

I think we can give the Toolforge CDN a try for now. My only real concern was pulling from Cloudfare or anything outside WMF.

I have not tested this PR.

siddharthvp:
> I should note tools-static.wmflabs.org does go down sometimes... but lots of things break when this happens. The only weird bit here is that Twinkle is an on-wiki gadget, so one wouldn't expect it to break when everything on-wiki is otherwise working fine.

See commit 2: The code where select2 acts on the interface is put into a `if ($.fn.select2) { }` block.  Since it is only required for making the menu searchable, Twinkle will still work in basic mode even if it doesn't load. (This is good enough for warn, though not for xfd as the browser's native multi-select menu is pathetic & unusable when the number of options is large, but I guess something is better than nothing.)

morebits.js
```diff
@@ -1134,6 +1135,52 @@ Morebits.array = {
 	}
 };
 
+/**
+ * select2 doesn't natively support highlightening of search hits so we implement that here
+ * @param {jQuery} $select
+ * @param {boolean} [hasOptionGroups=false] - does the select menu have optgroups?
+ * @requires {select2}
+ */
+mw.select2SearchHighlights = function select2SearchHighlights($select, hasOptionGroups) {
```
siddharthvp: Obviously this should be called `Morebits.select2SearchHighlights`, but when named that way, it _just doesn't work_. The bizarre console error "Morebits.select2SearchHighlights is not a function" appears. Even changing it to `var select2SearchHighlights` gives the same issue.  

But making it `mw.select2SearchHighlights` or even the technically incorrect `Twinkle.select2SearchHighlights` or just `select2SearchHighlights` (w/o `var` or anything) works fine. 

I can't for the life of me figure out what's wrong. It has been a long while since I came across a frustrating bug. Am I missing something obvious?

MusikAnimal: `Morebits.prototype.select2SearchHighlights = function ...`? Though I don't know if makes sense to namespace it under `Morebits.quickForm` or even `Morebits.quickForm.element`, as we do with the other public methods such as `Morebits.quickForm.element.prototype.append()`

siddharthvp: `Morebits.prototype.select2SearchHighlights` doesn't work (of course, as we aren't making any `Morebits` object).

Naming it `Morebits.quickForm.select2SearchHighlights` also doesn't work. 

I should note that `Morebits.userIsInGroup` and `Morebits.pageNameRegex` are both (public) functions and work fine.

MusikAnimal: Indeed Morebits is a normal object, not an object type or class. I was just guessing :) I don't have an answer for you. `Morebits.foo = function () { console.log('bar') }` for instance works in the JS console. I can try to help debug when I have time, but I probably wouldn't resort to putting this in `mw` where it doesn't belong.

siddharthvp: > I probably wouldn't resort to putting this in mw where it doesn't belong.

Yeah of course, this is just a makeshift arrangement for now. We could probably put this in twinkle.js file and call it `Twinkle.select2SearchHighlights`, coz that works.

siddharthvp: The issue could probably be on just my end. It works fine if I paste the whole of morebits.js into the console. But not when I am loading the file from my localhost import (which is how I usually test, it didn't give any problems thus far).

siddharthvp:
Ok, a new insanity has come up: as [reported on WT:TW](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#User_warning_type_drop-down_list_getting_cut_off), the behaviour of the jquery.chosen-based select menu has changed for the worse (at least on Chrome in Win10 - quite alarming as this is probably the most popular browser-OS combination). I haven't bothered to investigate the reason (could be due to an update in chosen or something else). 

Returning to this after 2 months, I've resolved the bug above (just a loader problem), and I think this is almost ready for merge. Before that, a couple of things need to be settled:

- FWIW I'd be in favour of loading select2 from an on-wiki gadget page, rather than loading from the CDN because of the two massive console error messages it produces. (I seem to have read somewhere that gadgets shouldn't give console errors.) There is some precedence for this - [MediaWiki:Gadget-moment.js](https://en.wikipedia.org/wiki/MediaWiki:Gadget-moment.js) (before moment was added to the RL), and also there's [User:Evad37/qunit-2.8.0.js](https://en.wikipedia.org/wiki/User:Evad37/qunit-2.8.0.js) and [User:Evad37/rater/lib/interact.js](https://en.wikipedia.org/wiki/User:Evad37/rater/lib/interact.js) recently uploaded by @evad37 (the latter is unused atm I guess). All of these, like select2, are released under MIT license.
- I say, let's deprecate the `oldSelect` preference option. It was added (#661) hurriedly before we started making CSS tweaks. It's rather unnecessary now and will be even more so after this is merged. Native and select2 almost look alike. It's only being [used by 5 users](https://en.wikipedia.org/w/index.php?sort=relevance&search=%22oldSelect%22+intitle%3Atwinkleoptions.js&title=Special:Search&profile=advanced&fulltext=1&advancedSearch-current=%7B%7D&ns2=1&searchToken=cvbrn3sxsc1fijlg3cr4gqo4u) - 4 of whom enabled it on 7 June 2019, before the CSS-based fixes were deployed. 
Having multiple modes would be difficult to maintain in the long run.  

@MusikAnimal and @Amorymeltzer

Amorymeltzer:
Apologies for taking so long to get around to this @siddharthvp and @MusikAnimal, I misread some emails and thought you two had agreed to wait.  I see now I was **very** wrong.  I can test, review, and help make changes in the next couple of days, but wanted to make some broad comments to the above beforehand:

* In general, a hearty +1 to this; you and I spent way too long on CSS crap with chosen, so anything that gets away from that the better AFAIAC!  Plus, select2 is way better maintained (it should probably replace chosen in RL).
* I don't like the idea of uploading select2 on-wiki, but I *really* don't the alternatives, so I think that's definitely the way to go.  License is fine.
* Let's leave `oldSelect` in place: the initial rollout in the warning menu could've been better, but the response was passionate.  Is it causing issues here?

siddharthvp:
Regarding `oldSelect`, it isn't causing any issues here but I had always thought of that as a temporary measure to give us time to get our shit right, not something to keep around long-term. With select2, the menu is almost alike the native one, except that it supports search. Also it isn't the case that select2 won't be loaded if this is enabled (getting rid of console errors), so there's like zero utility. 

Regarding users: 4 enabled it when the early chosen menu was very bad, and 1 now since the [new issues](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_42#User_warning_type_drop-down_list_getting_cut_off) started appearing. There's just 1 person who enabled it in between!

modules/twinklexfd.js
```diff
@@ -258,12 +259,31 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 			work_area = work_area.render();
 			old_area.parentNode.replaceChild(work_area, old_area);
 
-			$(work_area).find('[name=delsort]')
-				.attr('data-placeholder', 'Select delsort pages')
-				.chosen({width: '100%'});
+			if ($.fn.select2) {
+				$(work_area).find('[name=delsort]')
+					.attr('data-placeholder', 'Select delsort pages')
```
Amorymeltzer: So one kind of annoying thing is that apparently already selected items aren't removed/disabled from the dropdown if they're already in the multiselect pillbox.  I imagine there's a jquery-ish way to toggle an item disabled<->enabled/hidden<->visible when it moves in or out of the select, but unless it's straightforward, I doubt it's worth it here.

morebits.js
```diff
@@ -1152,6 +1152,52 @@ Morebits.array = {
...
+				var searchRegex = new RegExp(mw.RegExp.escape(searchString), 'i');
+
+				var selectorString;
+				// the odd way in which select2 organises the ul necessitates this hack
+				if (hasOptionGroups) {
+					selectorString = 'li > strong, li > ul > li';
+				} else {
+					selectorString = 'li';
+				}
+				$ul.find(selectorString).each(function() {
```
Amorymeltzer: There's a weird bug here.  Try this:

1. Open the warn menu, select single issue (or notice or custom or all templates)
2. Open the search menu; clicking is sufficient, even without search
3. Select level 1.  Try to search for something.

When you type, you get:
`VM765:1190 Uncaught DOMException: Failed to execute 'setStart' on 'Range': There is no child at offset`...

You can also trigger `setEnd`.

My initial guess was that the results here in `$ul` weren't being properly destroyed or recreated so there are counting issues, but it only issues the errors when on a menu with `optgroup`s after visiting any menu without `optgroup`s.  That would suggest otherwise, something to do with the `selectorString`?

modules/twinklewarn.js
```diff
@@ -1081,9 +1085,6 @@ Twinkle.warn.prev_article = null;
 Twinkle.warn.prev_reason = null;
 
 Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCategory(e) {
-	if (!Twinkle.getPref('oldSelect')) {
-		$('select[name=sub_group]').chosen('destroy');
```
Amorymeltzer: Do we really not need the `.destroy()` anymore?  I would've thought so, but seems select2 handles things much more smoothly!

Amorymeltzer:
@siddharthvp Would it make your life easier if I went and handled the adding of `select2`?  Would certainly make testing/developing easier.

@MusikAnimal Any qualms with this as an additional module/gadget?  Licensing doesn't seem to be an issue, it's MIT.

morebits.js
siddharthvp: Funny. The searching works fine, but the console is flooded with errors. 😑

siddharthvp: Fixed.

siddharthvp: It amazes me how you unearth such obscure and well-hidden bugs, even when they leave no visual trace. Thanks a lot for finding this one. What was happening is that multiple event listeners were building up on the $select (1 added every time the main_group is changed), so the entire highlighting code was executing multiple times. In one of these listeners, `hasOptionGroups` is false and hence the wrong `selectorString` is used - resulting in errors. But the searching itself worked correctly because there were other concurrent event listeners doing the right job!

Amorymeltzer:
Precipitous timing, the chosen folks just marked the latest as deprecated: https://github.com/harvesthq/chosen/commit/91041bc9dd6867f9a1668050a1b092d92027f13b

That shouldn't affect anything (cdnjs doesn't really cull) but all the more reason for this.  This PR is definitely going to go through, mutatis mutandis, so I've gone and opened #812 to take care of the work of adding select2 to the repo.

modules/twinklewarn.js
```diff
@@ -1202,31 +1203,34 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	// Trigger custom label/change on main category change
 	Twinkle.warn.callback.change_subcategory(e);
 
-	// Build select menu with jquery.chosen
-	if (!Twinkle.getPref('oldSelect')) {
+	// Use select2 to make the select menu searchable
+	if (!Twinkle.getPref('oldSelect') && $.fn.select2) {
 		$('select[name=sub_group]')
-			.chosen({width: '100%', search_contains: true})
```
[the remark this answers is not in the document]
siddharthvp: Can be done by using a custom function as the `matcher` in the select2 config ([stackoverflow](https://stackoverflow.com/questions/31571864/select2-search-match-only-words-that-start-with-search-term)). But the code for limiting to word searches doesn't appear to be trivial.

morebits.js
```diff
@@ -1152,6 +1152,52 @@ Morebits.array = {
...
+			});
+
+			if (this.value) {
+				var searchString = this.value;
+				var searchRegex = new RegExp(mw.RegExp.escape(searchString), 'i');
+
+				var selectorString;
+				// the odd way in which select2 organises the ul necessitates this hack
+				if (hasOptionGroups) {
+					selectorString = 'li > strong, li > ul > li';
```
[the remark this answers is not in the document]
siddharthvp: > I guess the thinking is that since it will show up as a match, users should know why?

Yes. wouldn't it be odd if a optgroup title shows up in the result without a highlight, and there are no options within that optgroup?

modules/twinklewarn.js
```diff
@@ -1202,31 +1203,34 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	// Trigger custom label/change on main category change
 	Twinkle.warn.callback.change_subcategory(e);
 
-	// Build select menu with jquery.chosen
-	if (!Twinkle.getPref('oldSelect')) {
+	// Use select2 to make the select menu searchable
```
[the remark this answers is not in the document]
siddharthvp: Hitting enter when the select menu is focussed does open up the dropdown. This might still be worth doing for maximum accessibility. 

> we can probably be a bit more concise since we know exactly what we're dealing with.

any ideas on how?

modules/twinklexfd.js
siddharthvp: Adding `'.select2-results__options [aria-selected=true] { display: none; }'` to the CSS does the trick. 

Clicking on the already selected item is another way to de-select it.  (Unlike in chosen where a single backspace deletes an option, here de-selecting an option can't be done by a single backspace - that just converts it to the text which then has to removed.) Would that have some utility? I rather guess not.

modules/twinklewarn.js
siddharthvp: select2 is always two steps ahead of chosen :)

modules/twinklewarn.js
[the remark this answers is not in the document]
Amorymeltzer: Don't know why we would put work into making it worse! 👍

twinkle.js
```diff
@@ -438,6 +438,10 @@ Twinkle.load = function () {
 	// Set custom Api-User-Agent header, for server-side logging purposes
 	Morebits.wiki.api.setApiUserAgent('Twinkle/2.0 (' + mw.config.get('wgDBname') + ')');
 
+	// Load select2 library from Toolforge CDN, used in warn and xfd
+	mw.loader.load('https://tools-static.wmflabs.org/cdnjs/ajax/libs/select2/4.0.10/js/select2.min.js');
+	mw.loader.load('https://tools-static.wmflabs.org/cdnjs/ajax/libs/select2/4.0.10/css/select2.min.css', 'text/css');
+
```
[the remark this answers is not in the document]
Amorymeltzer: #812 (but ugh)

morebits.js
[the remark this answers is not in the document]
Amorymeltzer: Maybe?  I find it distracting to have those highlighted as if I can select them.  They're not really proper search results after all.

modules/twinklewarn.js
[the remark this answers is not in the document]
Amorymeltzer: Anyway, I did some quick playing around and maybe not.  I thought we'd have a good id or other selector to latch on to, but I guess because select2 attaches weirdly it's not super easy (hence the `.closest` I guess)?  It's not too bad I suppose, put it somewhere in Morebits I guess?  enter is scary, and space is non-ideal.

(My initial comment was supposed to say

>It will open automatically when the user hits `<space>` (or `<enter>`!) 

But I left off the ` so they just didn't show!  Now fixed)

modules/twinklexfd.js
Amorymeltzer: The issue is if you're typing/using the keyboard to navigate and then hit enter, you get... nothing.  It's a weird situation, as least in my mind, to have the result polluted with non-op options, especially when they're already visible.  The CSS isn't sufficient since you'll still get "phantom" results (like the optgroup...) if it's the only result and the menu behaves like it's there.  I think the answer is "meh" and we can all live.

morebits.js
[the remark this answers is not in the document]
Amorymeltzer: Eh, I think you're probably right.  If you searching for "disruptive" in "common warnings" highlights the result, not common, which is good and normal, but typing "common" gives no results, so, yeah, I suppose it'd be weird.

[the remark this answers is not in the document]
Amorymeltzer: How crazy is something like

```javascript
$('li[role="group"]').each(function (idx, group) {
    if (!$(group).find('li[role="option"]').length) {
        $(group).hide();
    }
});
```

To hide empty optgroups?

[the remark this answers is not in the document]
Amorymeltzer: Actually, it's not too bad with a matcher, what do you think of the below?  Inspired by https://github.com/select2/select2/issues/3034#issuecomment-338286741, it removes optgroups with no options from the result.

```javascript
matcher: function(params, data) {
	var original_matcher = $.fn.select2.defaults.defaults.matcher;
	var result = original_matcher(params, data);
	if (result && (!result.children || !result.children.length)) {
		result = null;
	}
	return result;
}
```

modules/twinklewarn.js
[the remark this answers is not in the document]
Amorymeltzer: Maybe setting `dropdownParent` could make it easier?  Played around with it a few days ago with weird results so not sure if it'd be worth it.

modules/twinklexfd.js
siddharthvp: The CSS is hacky, yeah. Using jquery to toggle items hidden/visible also amounts to the same thing, and disabling/enabling doesn't seem to make sense as these are `li`s, not `option`s. So we have to use remove the items outright. This is something for which someone will need to do the dirty work of finding out how exactly which event to listen to, and/or what select2 config option to use. I can't say I'm volunteering.

modules/twinklewarn.js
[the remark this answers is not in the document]
siddharthvp: Dunno about dropdownParent, but do you want the [above hack](https://github.com/azatoth/twinkle/pull/692#discussion_r368061133) added?

morebits.js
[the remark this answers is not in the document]
siddharthvp: Actually the code doesn't fully replicate chosen. I just realised that what chosen does is that if the search term matches in the optgroup name, _all_ options in that optgroup are shown. That seems ideal. I think _this_ is the behaviour we should be emulating. Should be doable with a custom matcher.

modules/twinklewarn.js
```diff
@@ -1202,31 +1203,34 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
...
+			Morebits.select2SearchHighlights($('select[name=sub_group]'), false);
+		}
+
 		mw.util.addCSS(
-			// Force chosen select menu to display over the dialog while overflowing
-			// based on https://github.com/harvesthq/chosen/issues/1390#issuecomment-21397245
-			'.ui-dialog.morebits-dialog .morebits-dialog-content { overflow:visible !important; }' +
-			'.ui-dialog.morebits-dialog { overflow: inherit !important; }' +
+			// prevent dropdown from appearing behind the dialog
+			'.select2-container { z-index: 10000; }' +
```
Amorymeltzer: Forgot to mention this, but was this an issue for you or were you just being cautious?  I didn't see any issues without it in chrome.

modules/twinklewarn.js
```diff
@@ -1202,31 +1203,34 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
...
-		// (Hopefully) temporary fixes for visual issues caused by the above CSS, see #665 and #667
-		$('.ui-resizable-handle').remove();
-		// Jump to extend the dialog to include any previewing content
-		// without scrolling and prevent resizing
-		mw.util.addCSS('.ui-dialog.morebits-dialog .morebits-dialog-content { max-height: max-content !important; }');
+			// Reduce padding
+			'.select2-results .select2-results__option { padding-top: 1px; padding-bottom: 1px; }' +
+			'.select2-results .select2-results__group { padding-top: 1px; padding-bottom: 1px; } ' +
 
+			// Reduce font size
```
Amorymeltzer:
```suggestion
			// Increase font size
```

modules/twinklexfd.js
```diff
@@ -258,12 +259,31 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
+					// Remove black border
+					'.select2-container--default.select2-container--focus .select2-selection--multiple { border: 1px solid #aaa; }' +
 
-			// Reduce padding
-			mw.util.addCSS('.morebits-dialog .chosen-drop .chosen-results li { padding-top: 2px; padding-bottom: 2px; }');
+					// Reduce padding
+					'.select2-results .select2-results__option { padding-top: 1px; padding-bottom: 1px; }' +
+					'.select2-results .select2-results__group { padding-top: 1px; padding-bottom: 1px; } ' +
+
+					// Reduce font size
```
```suggestion
					// Increase font size
```

morebits.js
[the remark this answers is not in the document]
Amorymeltzer: Right you are!  I actually don't like that behavior, but I think it's probably expected.  Pretty sure that was the original intent of the comment I linked above, should work just fine?

```javascript
matcher: function(params, data) {
	var original_matcher = $.fn.select2.defaults.defaults.matcher;
	var result = original_matcher(params, data);
	if (result && data.children && result.children && data.children.length != result.children.length) {
		result.children = data.children;
	}
	return result;
}
```

modules/twinklewarn.js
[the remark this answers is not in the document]
Amorymeltzer: If you don't mind terribly, I think it'd be worth it 👍

modules/twinklewarn.js
siddharthvp: The dropdown did disappear behind the dialog at least on one instance. Can't replicate that now, but I guess this is good just to be safe?

modules/twinklewarn.js
siddharthvp: This must be platform-dependent, because I was seeing a default font size of a massive 16px. Let's change the word to "Adjust".

morebits.js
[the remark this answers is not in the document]
siddharthvp: not quite, that ends up showing everything in the optgroup if any option in the optgroup matches.

modules/twinklewarn.js
Amorymeltzer: We can revisit later if need be, but maybe just comment to that effect (‘just in case‘) for now, no need to spend time testing right now

modules/twinklewarn.js
Amorymeltzer: What percent was that?  I can't check  my system tonight but it was around 11.5px iirc.

siddharthvp:
@Amorymeltzer Dealt with mostly everything. Also did away with the hacky way of highlighting search hits with the more native-ish way using `templateResult`. All common is in morebits under a new `Morebits.select2` object. Wouldn't mind you double-checking that everything is ok.

Amorymeltzer:
Thus far it all looks good?  Would you mind explaining the `queryInterceptor` a bit?  You're feeding the default text for `searching` into it so `highlightSearchMatches` can access it, is that right?

Speaking of which, `highlightSearchMatches` seems really neat.  I'll play around with it and see if I can't break anything, but it all seems to work quite nicely!

Amorymeltzer:
@siddharthvp @MusikAnimal Regarding https://github.com/azatoth/twinkle/pull/692#issuecomment-542367663, is it worth removing the `$.fn.select2` ifs?  Shouldn't be necessary, yeah?

morebits.js
```diff
@@ -1152,6 +1152,72 @@ Morebits.array = {
...
+
+		if (result && params.term &&
+			data.text.toUpperCase().indexOf(params.term.toUpperCase()) !== -1) {
+			result.children = data.children;
+		}
+		return result;
+	},
+
+	/** Underline matched part of options */
+	highlightSearchMatches: function(data) {
```
Amorymeltzer: I said this already, @siddharthvp, but these really are quite well done.  Way more concise than I ever would have thought possible.

Amorymeltzer:
Merged as 4a1330c

morebits.js
siddharthvp: The idea comes from https://forums.select2.org/t/how-can-i-highlight-the-results-on-a-search/52/2. I just condensed down the code. 

That should also explain the `queryInterceptor` (they've used the implicitly global `query` variable). Too bad that `templateResult` doesn't expose the search query as a parameter. This hack is unlikely to break, unless select2 makes some deliberate breaking changes.

siddharthvp:
Thanks for all your hardwork and research on getting select2 to load! Just tried this out on enwiki. Smooth and amazing!

PR #693, opened 2019-08-31
tag: add {{short description}} to hatnotes regex

siddharthvp:
This makes the tags get placed below the hatnotes on pages where {{short description}} is present at the top.

Bug was reported at [TW talk page](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Bug:_Failure_to_place_maintenance_tags_below_hatnotes_when_there's_also_a_short_description_template).


DannyS712:
This isn't gerrit, but `+1` from me - thanks for following up on the bug report

PR #695, opened 2019-08-31
speedy/image: fix CSD log link for F5 replacement file

siddharthvp:
Fixes the bug reported at [TW talk page](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_41#CSD_F5_files).

### DIFF
--- a/modules/twinkleimage.js
+++ b/modules/twinkleimage.js
@@ -199,7 +199,7 @@ Twinkle.image.callback.evaluate = function twinkleimageCallbackEvaluate(event) {
 		reason = event.target.reason.value;
 	}
 	if (event.target.replacement) {
-		replacement = event.target.replacement.value;
+		replacement = event.target.replacement.value.replace(/^\s*(Image|File):/i, '');
 	}
 	if (event.target.derivative) {
 		derivative = event.target.derivative.checked;
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1556,14 +1556,19 @@ Twinkle.speedy.callbacks = {
 			}
 
 			var formatParamLog = function(normalize, csdparam, input) {
-				if ((normalize === 'G4' && csdparam === 'xfd') || (normalize === 'G6' && csdparam === 'page') || (normalize === 'G6' && csdparam === 'fullvotepage') || (normalize === 'G6' && csdparam === 'sourcepage')
-					|| (normalize === 'A2' && csdparam === 'source') || (normalize === 'A10' && csdparam === 'article') || (normalize === 'F5' && csdparam === 'replacement')) {
+				if ((normalize === 'G4' && csdparam === 'xfd') ||
+					(normalize === 'G6' && csdparam === 'page') ||
+					(normalize === 'G6' && csdparam === 'fullvotepage') ||
+					(normalize === 'G6' && csdparam === 'sourcepage') ||
+					(normalize === 'A2' && csdparam === 'source') ||
+					(normalize === 'A10' && csdparam === 'article')) {
 					input = '[[:' + input + ']]';
 				} else if (normalize === 'G5' && csdparam === 'user') {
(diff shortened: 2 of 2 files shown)


Amorymeltzer:
@siddharthvp and @MusikAnimal: I don't think this should have been committed as written. 
 Sorry for not being around to deal with this, but this has actually introduced a regression IMO.

The template added when nominating an F5 via `di` is `{{di-orphaned fair use}}`, and while it does its best via `ifexist`, the template can only detect the absence/presence of the `File:`/`Image:` namespace prefix for local image pages. That is, while `{{di-orphaned fair use|replacement=File:Interface-protection-shackle.svg}}` and `{{di-orphaned fair use|replacement=Interface-protection-shackle.svg}}` both work, only `{{di-orphaned fair use|replacement=File:Fée.svg}}` correctly links to the image while `{{di-orphaned fair use|replacement=Fée.svg}}` does not.  Fair use images being replaced by Commons images is rare, I suppose, but isn't that theoretically the ideal situation?  Indeed, perusing the images from the linked complaint (now deleted, so sysop-only) indicate that some of Begoon's such tags would have had broken links.  Some of their [more recent taggings](https://en.wikipedia.org/w/index.php?title=User:Begoon/CSD_log&diff=911756447&oldid=911465107&diffmode=source) (before this was merged) seem to have included the namespace prefix, which I think is typical behavior from other users?

More importantly, AFAICT https://github.com/azatoth/twinkle/pull/695/files#diff-60ef673bfda0c7bedf2918d1179ef0aeL202 now means that Twinkle will *never* place a correct template if the replacement image provided is on Commons.

I think a better way of doing this would have been to note (as we do in the CSD module) that the namespace is required by Twinkle, or even do the exact opposite of this PR and instead supplement user input by silently prepending `File:` if it isn't present.

siddharthvp:
You are correct. Though committing this is admittedly a mistake, I think
the issue needs to be fixed with the template. The template should
unconditionally prepend the "File:" prefix if there isn't one. If someone
gives Fee.jpg as the first parameter, [[Fee.jpg]] is definitely an invalid
link, but [[File:Fee.jpg]] could be valid or invalid; there is something to
gain (in the event that the file exists, either locally or on commons) and
nothing to lose by adding the File: prefix.

The template creator probably just ifexist as a cheap alternative to
checking that the string starts with "File:" which is tougher with
wiki-markup.

MusikAnimal:
Interesting, I had not thought about whether the template would work if there is not a local copy. I was actually only looking at the CSD log, the subject of this PR. I used testwiki and observed that before this change, the CSD log entry was [missing the namespace prefix](https://test.wikipedia.org/w/index.php?title=User:MusikAnimal/CSD_log&diff=401385&oldid=401382), and after this change it correctly [included it](https://test.wikipedia.org/w/index.php?title=User:MusikAnimal/CSD_log&diff=401389&oldid=401388). There {{di-orphaned fair use}} is not even implemented.

That all seems fine. I think it's the `.replace(/^\s*(Image|File):/i, '')` that's the problematic line, as you say.

> do the exact opposite of this PR and instead supplement user input by silently prepending File: if it isn't present.

This seems like the more proper solution.

JJMC89:
I've updated `{{di-orphaned fair use}}` to use `{{file title}}`, which will correctly deal with the namespace or lack thereof.

> 
> > do the exact opposite of this PR and instead supplement user input by silently prepending File: if it isn't present.
> 
> This seems like the more proper solution.

That would keep it consistent with F1 and F8. The tag menu should get updated to have the same treatment.

Amorymeltzer:
Musing out loud to nobody, but since we started logging parameters in the CSD log (#441, etc.), is it weird that we modify user input then log it?  Not really — it makes the logs more useful to everyone, and the only downside is to us, as if we need to debug, it's not immediately clear what exactly was entered by the user.  Still, sits somewhat weird.

Issue #178, opened 2013-09-12
Clean up default tag set (was: Please allow Twinkle to tag articles with {{news release}})


jasonspiro:
# Attention GCI students! See comment by atlight on 24 Nov 2015

Thank you very much for writing Twinkle.

friendlytag.js lets me add various "Style of writing" cleanup tags, including {{advert}} and {{tone}}.  But it does not let me add {{news release}} ("article reads like a news release, or is otherwise written in an overly promotional tone").

Please modify friendlytag.js so that it can tag articles with {{news release}}.

QEDK:
Well, same applies to welcome templates. Most are not used and since they're substed, we cannot really get a transclusion count. The only way's to run a wikitext search for the hidden template text.

jasonspiro:
@Technical-13 Ah. I actually didn't know that in my Twinkle preferences there was a place to adding custom tags. But now I do. Thank you.

Perhaps the "TW" menu, there should be a "Preferences" link, in order to make the preferences page more prominent. :)

atlight:
I'm going to "take the plunge" and put this forward as a Google Code-In task. Wikimedia is [participating](https://www.mediawiki.org/wiki/Google_Code-in_2015) in this coding event for 13- to 17-year-olds, so I thought I would throw our hat in the ring and use this task to see if GCI can benefit us.

### Additional information for GCI students

[Twinkle](https://en.wikipedia.org/wiki/WP:TW) is a popular JavaScript gadget used by thousands of Wikipedians to accelerate common maintenance tasks. The code is [hosted on GitHub](https://github.com/azatoth/twinkle).

Twinkle allows users to add "maintenance tags" to articles, to inform readers and editors about potential problems with the article. In order to save space, Twinkle only supports a subset of the maintenance tags available on Wikipedia. 

Your task is to undertake an audit of article maintenance tags to determine whether the set offered by Twinkle matches the needs of today's Wikipedia user.
* Create a spreadsheet that lists all the templates at [WP:TC](https://en.wikipedia.org/wiki/WP:TC), **excluding** the following:
  * "inline" type tags
  * tags with "section" in the template name
  * tags that are to be applied to talk pages.
* Use the ["Transclusion Count"](https://tools.wmflabs.org/templatecount/index.php) tool on Wikipedia to determine how many uses each maintenance tag has;
* If the number of transclusions is low (less than about 50), check to see how recently the template was created, and add this info to the spreadsheet. Recently created templates should be given more generous treatment;
[...]

PR #696, opened 2019-08-31
tag: re-add {{dead end}}, add {{sources exist}} per requests

siddharthvp:
Now that the interface supports search, clutter is no longer a reason for removing templates.
Adding back {{dead end}} was requested [here](https://en.wikipedia.org/wiki/Wt:Twinkle/Archive_41#{{Dead_end}}_has_disappeared). Also adding {{sources exist}}, requested [here](https://en.wikipedia.org/wiki/Wt:Twinkle/Archive_39#Can_we_please_add_the_{{Sources_exist}}_Tag?)

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -754,6 +754,7 @@ Twinkle.tag.article.tags = {
 	'Rough translation': 'poor translation from another language',
 	'Sections': 'needs to be divided into sections by topic',
 	'Self-published': 'contains excessive or inappropriate references to self-published sources',
+	'Sources exist': 'notable topic, sources are available that could be added to article',
 	'Technical': 'too technical for most readers to understand',
 	'Third-party': 'relies too heavily on sources too closely associated with the subject',
 	'Tone': 'tone or style may not reflect the encyclopedic tone used on Wikipedia',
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
We should probably put together some sort of criteria for inclusion; less important than #533 perhaps, but thinking of #178.

MusikAnimal:
> We should probably put together some sort of criteria for inclusion; less important than #533 perhaps, but thinking of #178.

I agree. In this case dead end has [sound rationale](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_41#%7B%7BDead_end%7D%7D_has_disappeared) for inclusion, and {{sources exits}} had [rough consensus](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_39#Can_we_please_add_the_%7B%7BSources_exist%7D%7D_Tag?). My concerns were alleviated with the new search feature, but if I was out of line in merging this I do apologize. Some sort of guideline would be helpful, noting that transclusion count doesn't always equate to usage, as is the case with short-lived tags like {{dead end}}.

Amorymeltzer:
Not at all!  Just noting that we've now moved from needing to be actively limited in our offerings to being able to be more inclusive.  I *do* think we should be cognizant of the position Twinkle is in — if something is listed, it's likely to get used more — but this is entirely in order.  We would want a way to assess items that are quickly removed, as the scripts in https://github.com/azatoth/twinkle/issues/178#issuecomment-174503908 and https://github.com/azatoth/twinkle/issues/178#issuecomment-474012574 obviously won't help.

PR #697, opened 2019-09-01
speedy: include xfd page name as parameter in db-repost-notice

siddharthvp:
Per request at [TW talk page](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Request_to_include_discussion_page_in_automatic_G4_notice).

Also improved the conditional statement for checking that the page given is in project namespace.


Amorymeltzer:
Oof, nice call on those substrings.  They weren't even parallel!

PR #703, opened 2019-09-06
tag: Update globalize template following TfD

### LINKED ISSUE #690: Globalize merger
The region specific sub-templates of Template:Globalize have been merged with the main template. Can you update the tag module to use the main template instead of the subpages?

Amorymeltzer:
Closes #690.  [TfD](https://en.wikipedia.org/wiki/Wikipedia:Templates_for_discussion/Log/2019_April_5#Template:Globalize) closed with consensus to do away with all the various region-specific subpages, massively simplifying things.  Specific entries are handled by Template:Globalize/name, so it should be sufficient for us to just offer an input box and let the template(s) sort it out.

----
See also https://en.wikipedia.org/wiki/Wikipedia:Templates_for_discussion/Holding_cell#Meta

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -362,33 +362,9 @@ Twinkle.tag.updateSortOrder = function(e) {
 				break;
 			case 'Globalize':
 				checkbox.subgroup = {
-					name: 'globalize',
-					type: 'select',
-					list: [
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
cc @briick03

Amorymeltzer:
Gonna push this live now since they've been listed in template jail for months.

Issue #182, opened 2013-10-07
Handle abuse filter warnings/errors


Amalthea:
An abuse filter warning is currently an irrecoverable error. API will return something like:

&lt;?xml version="1.0"?>&lt;api>&lt;edit code="abusefilter-warning-codespill" info="Hit AbuseFilter: Extraneous formatting" warning="&amp;lt;table [...]" result="Failure" />&lt;/api>

(warning HTML truncated).

I imagine Twinkle should display that warning and allow the user to try again (not sure what the mechanism for that is)

PR #707, opened 2019-09-13
morebits: Update handling of blacklist and abusefilter errors from API (breaking change for 1.34.0-wmf.23)

### LINKED ISSUE #704: Update API error handling for impending breaking changes
There's an [upcoming breaking change to the MediaWiki API](https://lists.wikimedia.org/pipermail/mediawiki-api-announce/2019-September/000148.html) which will impact our ability to display relevant error messages to users in the case of an abuse filter being triggered or the spam blacklist being hit.

We'll need to update [twinklefluff](https://github.com/azatoth/twinkle/blob/afcd932dd983d9f486f3f0457620abdfb6844479/modules/twinklefluff.js#L493) and [morebits](https://github.com/azatoth/twinkle/blob/28d50ac8b20fd474c00b29589360a08f0066189f/morebits.js#L2590) to reflect this breaking change.

Amorymeltzer:
Closes #704

As [recently announced](https://lists.wikimedia.org/pipermail/mediawiki-api-announce/2019-September/000148.html), a [breaking change](https://phabricator.wikimedia.org/T229539) is being made to how the API returns errors for the abuse filter and the spam blacklist; the differences can be seen in [T232583](https://phabricator.wikimedia.org/T232583), but the tl;dr is that instead of reporting a successful edit as "failure," these errors will finally (as of 1.34.0-wmf.23) be correctly reported as error codes.

This is, obviously, good, and simplifies things quite a bit on our end, in particular handling the two abusefilter option (`abusefilter-disallowed` and `abusefilter-warning`) in a reasonable fashion.  These need slightly different messages, and the full `info` response is a little too long.  On the other hand, while the default system response to `spamblacklist` is wordy, it seems to cover all the bases without being too excessive.  We're not currently retrying `abusefilter-warning` so I've folded all of these into the "hard error" section.

As for `twinklefluff.js`, the `Morebits.wiki.api` call to `Twinkle.fluff.callbacks.complete` doesn't include an `onError` call, so unless I'm mistaken, we won't need to handle them here anymore.  Theoretically, this should marginally help with #182.

----
I don't *think* there's a good way to test this beforehand, so, if all goes [according to plan](https://www.mediawiki.org/wiki/MediaWiki_1.34/wmf.23), my plan is to carve out some time Tuesday and Wednesday after this goes live on the testwiki (17 and 18 September) before the push to all wikis on the 19th.

Assuming I'm not wildly off-base here, I think what I'll *actually* do is push the additions to `fnSaveError` but **not** the removals from `fnSaveSuccess`; that should hopefully allow everything to transition in a way that should minimize disruption (i.e., I can push to enwiki ahead of time rather than sit there attempting to time the train perfectly).  Does that sound right?  I also intend on posting to WT:TW with technical instructions aimed at capturing incoming issues from non-enwiki wikis (such as meta, stuck in 2014(?)), in the hopes that we can be helpful.


Amorymeltzer:
cc @atlight and @MusikAnimal; I'd appreciate your eyes and thoughts on this, in particular:
- Did I have the dumb and this won't work?
- Language around the responses
- The removal of stuff in `fluff`

morebits.js
```diff
@@ -2691,6 +2671,21 @@ Morebits.wiki.page = function(pageName, currentAction) {
 			// non-admin attempting to edit a protected page - this gives a friendlier message than the default
 			if (errorCode === 'protectedpage') {
 				ctx.statusElement.error('Failed to save edit: Page is protected');
+			// check for absuefilter hits: disallowed or warning
+			} else if (errorCode.indexOf('abusefilter') === 0) {
```
siddharthvp: Unnecessary busywork. Just put two separate `else if`s for `abusefilter-disallowed` and `abusefilter-warning`.

siddharthvp:
> As for twinklefluff.js, the Morebits.wiki.api call to Twinkle.fluff.callbacks.complete doesn't include an onError call, so unless I'm mistaken, we won't need to handle them here anymore.

I think you're mistaken, theoretically at least. The removal means that abusefilter warnings/errors won't be handled at all, though they were being handled earlier. Either an error callback should be added in the Morebits.wiki.api call, or fluff should be upgraded to use `Morebits.wiki.page` like other TW modules. Granted, reverts of edits that _removed_ blacklisted links should be rare, so this isn't a priority. We can work on this later at leisure.

Amorymeltzer:
>abusefilter warnings/errors won't be handled at all, though they were being handled earlier

Won't they just be picked up like any other errors, say attempting to edit a protected page?  Fluff handles that just fine without us.  The (previously) duplicated stuff in `Twinkle.fluff.callbacks.complete` is only for successful calls, so why would these (being treated as errors) ever end up there anyway?

morebits.js
```diff
@@ -2691,6 +2671,21 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+				var desc = $(ctx.responseXML).find('error').find('abusefilter').attr('description');
```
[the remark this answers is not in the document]
Amorymeltzer: Indeed!  (Maybe leftover from a previous iteration I had?)

morebits.js
Amorymeltzer: Maybe it's just me, but I think cascading `if-else`s are generally easier to grok with relevant bits in one block, especially when they vary so much in size and depth, so readers can mentally say "okay this is the editconflict chunk, this is the blacklist chunk, this is the abusefilter chunk..."  And, at any rate, we'd have to duplicate the grabbing of `description` if separated.

siddharthvp:
> Won't they just be picked up like any other errors, say attempting to edit a protected page? 

Yeah. By "handling" I meant the dedicated handling given to them like in Morebits.wiki.page. But seeing that the server-generated error texts aren't much different or worse off than ours, I guess it isn't really necessary for now. But it would be necessary in order to resolve #182, which looks quite easy to me. (Just embed a "Proceed anyway" button within the morebits status message - morebits already allows for having putting html elements in status; it isn't limited to text. This doesn't add any UI dependency.)

morebits.js
```diff
@@ -2601,34 +2601,15 @@ Morebits.wiki.page = function(pageName, currentAction) {
 			link.appendChild(document.createTextNode(ctx.pageName));
 			ctx.statusElement.info(['completed (', link, ')']);
 			if (ctx.onSaveSuccess) {
-				ctx.onSaveSuccess(this);  // invoke callback
+				ctx.onSaveSuccess(this);  // invoke 1callback
```
MusikAnimal: `1callback`?

morebits.js
```diff
@@ -2691,6 +2671,21 @@ Morebits.wiki.page = function(pageName, currentAction) {
 			// non-admin attempting to edit a protected page - this gives a friendlier message than the default
 			if (errorCode === 'protectedpage') {
 				ctx.statusElement.error('Failed to save edit: Page is protected');
+			// check for absuefilter hits: disallowed or warning
+			} else if (errorCode.indexOf('abusefilter') === 0) {
+				var desc = $(ctx.responseXML).find('error').find('abusefilter').attr('description');
+				if (errorCode === 'abusefilter-disallowed') {
+					ctx.statusElement.error('The edit was disallowed by the edit filter rule "' + desc + '".');
```
MusikAnimal: Wouldn't `desc` be either the title of the filter, or the message it returns (doubtfully the latter since it's HTML)? In either case, you might want to add quotes around the message or append a colon: `'The edit was disallowed by the edit filter: ' + desc`. Otherwise it will read like the filter description is part of the sentence and be grammatically incorrect. It'd be really neat if we could link to the filter, too, but not a big deal if we can't or if it's too much work.

MusikAnimal:
I look forward to the day that all API responses in Twinkle are JSON with `formatversion=2`!

Anyway, I think your plan sounds great. Indeed we can't really test anything until it's on testwiki. I'll try to help out tomorrow/Wednesday and make sure it's working as it should. We of course will need to create a temporary test filter and spam blacklist entry, e.g. `equals_to_any(user_name, 'Amorymeltzer', 'MusikAnimal') & summary contains "TW"`. You might also try an edit that hits both the filter and the spam blacklist, to see what happens.

I have not gone through fluff in length, so not sure how it will be affected.

It is also my understanding that this isn't really a _breaking_ change to the API, in the sense that those errors just won't have special handling anymore. In our case it should still fall back to `ctx.statusElement.error('Failed to save edit: ' + ctx.saveApi.getErrorText())`, so we shouldn't need to worry too much about deploying at the same time as the train?

morebits.js
Amorymeltzer: Title, yeah, so 👍

I can play around with `id` once this is settled, see if we can't get something fancier in action.

morebits.js
Amorymeltzer: Well, you see, since it's just "one callback" being invoked I figured... nah, I must've missed a meta key when switching buffers in emacs 😆

Amorymeltzer:
@MusikAnimal: I spent some time yesterday cobbling together a series of test queries to be easily run on testwiki, both for `fluff` and `morebits` — check out filters 1 and 203, and of course http://areallyevilsite.com —  so I'll fix the issues here and push the hybrid to testwiki tomorrow morning; I'll report back after things transition, which gives all of Wednesday in case something is borked.  As you say, though, the fallbacks are largely sufficient, thankfully now that these are reported as errors.

Amorymeltzer:
OH and ❤️  `formatversion=2`

morebits.js
Amorymeltzer: @MusikAnimal The language was mostly (entirely?) copied from the what previously existed, but warning and disallowed didn't match, so lmk how it looks now.

siddharthvp:
> I look forward to the day that all API responses in Twinkle are JSON with formatversion=2!

@MusikAnimal A first step in that direction would be reviewing #493.

Amorymeltzer:
Everything seems fine on testwiki.  I'll do some more testing and poking around, but if nothing comes up I'll put the hybrid version up on enwiki later this afternoon.

Looking at it again, I don't love the default message from the sbl, I'm going to use what we had before.

Also, there's a little weirdness around `abusefilter-warning` with things like csd tagging.  AFAICT, after being warned, the resubmit works fine, but things like logging or notifying the creator can sometimes fail and sometimes succeed.  I don't think that's related to this change.

Amorymeltzer:
No complaints thus far, so will merge this.

siddharthvp:
@Amorymeltzer and @MusikAnimal This is what the spam blacklist error looks like:

Screenshot (1)

Something amiss? I wouldn't have expected such a long message.

Amorymeltzer:
@siddharthvp Thanks for the report, I found the source.  Testing then PR coming up...

PR #710, opened 2019-09-22
Correctly notify the creator of article, if it began as a redirect

siddharthvp:
I have tweaked Morebits.wiki.page.prototype.lookupCreation so that an option is now available which if set, the function would return the username/timestamp of the creation of the first non-redirect version of the article. 

I have then put this into use for AFD and PROD notifications, since in these cases we unambiguously want to notify the real article creator rather than the redirect creator. We should perhaps also try to incorporate this somehow to CSD, but I guess that it's a bit trickier. 

I guess this is long-requested feature, and should be very helpful to the project as now people who create articles from redirects would not remain in the dark when their work is sent to the gallows. Also, this should help us crush the upcoming competition from PageCuration!

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -204,6 +204,7 @@ Twinkle.prod.callbacks = {
 		var ts = new Morebits.wiki.page(mw.config.get('wgPageName'));
 		ts.setFollowRedirect(true);  // for NPP, and also because redirects are ineligible for PROD
 		ts.setCallbackParameters(params);
+		ts.setLookupNonRedirectCreator(true); // Look for author of first non-redirect revision
 		ts.lookupCreation(Twinkle.prod.callbacks.creationInfo);
 	},
 
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -716,6 +716,7 @@ Twinkle.xfd.callbacks = {
 			if (params.usertalk) {
 				var thispage = new Morebits.wiki.page(mw.config.get('wgPageName'));
 				thispage.setCallbackParameters(params);
+				thispage.setLookupNonRedirectCreator(true); // Look for author of first non-redirect revision
 				thispage.lookupCreation(Twinkle.xfd.callbacks.afd.userNotification);
 			}
 
--- a/morebits.js
+++ b/morebits.js
@@ -1648,6 +1648,9 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		protectCreate: null,
 		protectCascade: false,
 
+		// - creation lookup
+		lookupNonRedirectCreator: false,
+
 		// - stabilize (FlaggedRevs)
 		flaggedRevs: null,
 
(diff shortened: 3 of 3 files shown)


siddharthvp:
Code workflow: A regex test for `/^\s*#redirect/i' is used to check if a version is a redirect. This does not break internationalization, as the same syntax is used for redirects even in non-English language wikis. The tags ("Removing redirect"/"Redirect becoming article") are _not_ being used since they didn't aren't there in early revisions, nor on other wikis. 

At first, only the first revision is fetched, and that too only the the top section (to save bandwidth). If it turns out that it was a redirect and the `lookupNonRedirectCreator` option is enabled, the 50 earliest revisions are now fetched (again only rvsection=0) and we iterate to find the first non-redirect version.

modules/twinklexfd.js
```diff
@@ -716,6 +716,7 @@ Twinkle.xfd.callbacks = {
 			if (params.usertalk) {
 				var thispage = new Morebits.wiki.page(mw.config.get('wgPageName'));
 				thispage.setCallbackParameters(params);
+				thispage.setLookupNonRedirectCreator(true);
```
```suggestion
				thispage.setLookupNonRedirectCreator(true); // Look for first non-redirect author
```

modules/twinkleprod.js
```diff
@@ -204,6 +204,7 @@ Twinkle.prod.callbacks = {
 		var ts = new Morebits.wiki.page(mw.config.get('wgPageName'));
 		ts.setFollowRedirect(true);  // for NPP, and also because redirects are ineligible for PROD
 		ts.setCallbackParameters(params);
+		ts.setLookupNonRedirectCreator(true);
```
Amorymeltzer:
```suggestion
		ts.setLookupNonRedirectCreator(true); // Look for first non-redirect author
```

morebits.js
```diff
@@ -2708,18 +2717,53 @@ Morebits.wiki.page = function(pageName, currentAction) {
 			return; // abort
 		}
 
-		ctx.creator = $(xml).find('rev').attr('user');
+		if (!ctx.lookupNonRedirectCreator || !/^\s*#redirect/i.test($(xml).find('rev').text())) {
```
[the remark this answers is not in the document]
siddharthvp: > having the typical behavior first is nice

That's what I thought.

morebits.js
```diff
@@ -2708,18 +2717,53 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+
+		$(xml).find('rev').each(function(_, rev) {
+			if (!/^\s*#redirect/i.test(rev.textContent)) {
+				ctx.creator = rev.getAttribute('user');
+				ctx.timestamp = rev.getAttribute('timestamp');
+				return false; // break
+			}
+		});
+
 		if (!ctx.creator) {
```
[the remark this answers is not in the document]
siddharthvp: Good idea! Will do.

morebits.js
```diff
@@ -2041,6 +2044,21 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+	/**
+	 * @param {boolean} flag - if set true, the author and timestamp of the first non-redirect
+	 * version of the page is retrieved.
+	 *
+	 * Warning:
+	 * 1. If there are no revisions among the first 50 that are non-redirects, or if there are
+	 *    less 50 revisions and all are redirects, the original creation is retrived.
+	 * 2. Revisions that the user is not privileged to access (revdeled/suppressed) will be treated
+	 *    as non-redirects.
+	 */
```
siddharthvp: This makes sense as revdels in main namespace is usually for copyright violations, and only non-redirect revisions would have been revdeled as such.

morebits.js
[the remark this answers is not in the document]
siddharthvp: done.

morebits.js
Amorymeltzer: Doesn't really make a difference either way: not much else we can do, and it should be pretty rare regardless.  The reality is that early vandalism is going to be an issue no matter what, as articles that were initially redirects but had someone vandalize that redirect before becoming an article will still notify the wrong person.  In the end, though, this is basically taking a simple process and making it somewhat smarter, but that doesn't mean we need to account for every single contingency.

PR #717, opened 2019-10-01
tag: Indicate direction of merge in talk page notification header

Amorymeltzer:
[Requested at WT:TW](https://en.wikipedia.org/wiki/Special:Permalink/918911925#Merge_tags).  No reason not to match the edit summary, where we already did this, so let's just use the same language for everybody.


siddharthvp:
`Proposing to merge Foo into Bar` is good as an edit summary but not sensible as a section header. Please use some variation of `Proposed merge (of|from) Foo (in)?to (Bar|this article)`.

Amorymeltzer:
Trying to be efficient but that's fair; how's this?

PR #721, opened 2019-10-05
twinkle: remove bullets in vector menu; add monobook/modern menus

### LINKED ISSUE #713: Fix menus following recent change to Vector
https://gerrit.wikimedia.org/r/c/mediawiki/skins/Vector/+/537769/ removed the styling on `.vectorMenu ul`, so now Twinkle shows bullet points and margins in the dropdowns (temporarily fixed on enwiki with [Special:Diff/918221349](https://en.wikipedia.org/wiki/Special:Diff/918221349)).

`Twinkle.addPortlet()` is a little hard to decipher. Basically what we need to do for Vector is remove the outer div, and put `.menu` on the `ul` element itself.

Markup should change from:
```
<div class="menu">
    <ul>
        ...
    </ul>
</div>
```
to
```
<ul class="menu">
    ...
</ul>
```

The problem is the outer div is apparently used by other skins, and even on Vector for `.vectorTabs` (you can make Twinkle use tabs instead of menus...?).

I will try to fiddle around with a bit more when I have time, but if anyone feels inclined to help it'd be appreciated.

### LINKED ISSUE #716: Create menus for monobook and modern skins
I see that @MusikAnimal's MoreMenu provides menus in monobook and modern skins too, whereas Twinkle does so only for vector. The individual tabs take up a lot of space, and causes other scripts' buttons to be hidden if user has too many of them. Maybe the code can be copied over? Or is there something else to be wary about?

siddharthvp:
**Removal of bullets from vector menu:** (closes #713)
Per MusikAnimal's comments at https://github.com/azatoth/twinkle/issues/713#issue-499541613.

With this, `innerDivClass` becomes an unused variable, as it was being only applied to the innerdiv when `type === 'menu'`, and `type` is being normalised to be `menu` only if we're on vector skin (line 262). 

~~However, I've commented it out rather than removed because it may possibly be useful while working on  #716. (If not, I'll remove it then.)~~

**Addition of menus for monobook/modern skins:** (closes #716)
Added on October 9: see [comment below](https://github.com/azatoth/twinkle/pull/721#issuecomment-539843901).

### DIFF
--- a/twinkle.css
+++ b/twinkle.css
@@ -5,3 +5,42 @@
 .skin-vector #p-twinkle {
 	width: 3.24em;
 }
+
+/**
+ * Styles for menus in monobook and modern skins
+ * Adapted from [[User:Haza-w/dropdown-menus]], code at [[MediaWiki:Gadget-dropdown-menus.css]]
+ */
+#ca-twinkle-menu {
+	background-color: #EEEEEE;
+	border: 1px solid #AAAAAA;
+	border-collapse: collapse;
+	color: #638C9C;
+	font: 14px/22px Arial, Helvetica, sans-serif;
+	padding: 0.3em 0em .3em 0em;
+	position: absolute;
+	width: 78px;
+	z-index: 1000;
+}
+#ca-twinkle-menu ul {
+	list-style: none;
+	margin: 0 0 0 0;
+}
+#ca-twinkle-menu ul li {
+	display: list-item;
+	width: 70px;
+	font-size: 14px;
+	margin: 0em;
+	padding: 0 0 0 0.5em;
+	border: 0;
+	background: none;
+}
+#ca-twinkle-menu ul li a {
+	color: #002BB8 !important;
+	text-decoration: none !important;
+	background-color: #EEEEEE !important;
+	padding: 0 0 0 0;
+	text-transform: none;
+}
+#ca-twinkle-menu ul li:hover a {
+	text-decoration: underline !important;
+}
--- a/twinkle.js
+++ b/twinkle.js
@@ -236,6 +236,7 @@ Twinkle.getFriendlyPref = function twinkleGetFriendlyPref(name) {
  *
  * @return Node -- the DOM node of the new item (a DIV element) or null
  */
+
 Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
 	// sanity checks, and get required DOM nodes
 	var root = document.getElementById(navigation);
@@ -260,26 +261,23 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
 	var skin = mw.config.get('skin');
 	type = skin === 'vector' && type === 'menu' && (navigation === 'left-navigation' || navigation === 'right-navigation') ? 'menu' : '';
 	var outerDivClass;
-	var innerDivClass;
 	switch (skin) {
 		case 'vector':
+			// XXX: portal doesn't work
 			if (navigation !== 'portal' && navigation !== 'left-navigation' && navigation !== 'right-navigation') {
 				navigation = 'mw-panel';
 			}
 			outerDivClass = navigation === 'mw-panel' ? 'portal' : type === 'menu' ? 'vectorMenu' : 'vectorTabs';
-			innerDivClass = navigation === 'mw-panel' ? 'body' : type === 'menu' ? 'menu' : '';
 			break;
 		case 'modern':
 			if (navigation !== 'mw_portlets' && navigation !== 'mw_contentwrapper') {
 				navigation = 'mw_portlets';
 			}
 			outerDivClass = 'portlet';
-			innerDivClass = 'pBody';
 			break;
 		default:
 			navigation = 'column-one';
 			outerDivClass = 'portlet';
-			innerDivClass = 'pBody';
 			break;
 	}
 
@@ -329,13 +327,13 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
 	outerDiv.appendChild(h5);
 
 	var innerDiv = null;
+	var ul = document.createElement('ul');
 	if (type === 'menu') {
 		innerDiv = document.createElement('div');
-		innerDiv.className = innerDivClass;
 		outerDiv.appendChild(innerDiv);
+		ul.className = 'menu';
 	}
 
-	var ul = document.createElement('ul');
 	(innerDiv || outerDiv).appendChild(ul);
 
 	return outerDiv;
--- a/twinkle.js
+++ b/twinkle.js
@@ -456,12 +456,14 @@ Twinkle.addPortletLink = function(task, text, id, tooltip) {
 	}
 	var link = mw.util.addPortletLink(Twinkle.getPref('portletId'), typeof task === 'string' ? task : '#', text, id, tooltip);
 	$('.client-js .skin-vector #p-cactions').css('margin-right', 'initial');
-	if (typeof task === 'function') {
-		$(link).click(function (ev) {
+	$(link).click(function (ev) {
+		if (typeof task === 'function') {
 			task();
 			ev.preventDefault();
-		});
-	}
+		} else {
+			link.querySelector('a').click();
+		}
+	});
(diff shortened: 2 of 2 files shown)


Amorymeltzer:
@MusikAnimal This is essentially what I was suggesting at https://github.com/azatoth/twinkle/issues/713#issuecomment-536176200.  I did some further tests and it still seemed fine, but held off opening a PR, but it seems @siddharthvp came to the same conclusion!

Regarding the former stuff, I see no reason to keep it around in the meantime?  Nobody's going to use the code except you two (working on menus) and it'll make the diffs easier for me reviewing.

siddharthvp:
> Regarding the former stuff, I see no reason to keep it around in the meantime?

Removed.

siddharthvp:
Added the menus for monobook and modern skins. Tested on Chrome, Firefox, and IE 11. 

The functions `showMenu` and `hideMenu` are unmodified (except for changes made to satisfy our linter). It should be possible to trim these two functions to remove code that helps facilitate creation of submenus (as we don't have submenus in twinkle). But I'll leave that for later. 

To test this easily, just importScript [User:SD0001/twinkle.js](https://en.wikipedia.org/wiki/User:SD0001/twinkle.js) and see the magic! Not necessary to disable twinkle first, though you'd see two sets of buttons/menus if you don't. 

Screenshot (12) png
^Modern

Screenshot (13) png
^Monobook

The menu has been designed to look more like the vector menu than the original haza-w's dropdrown-menus. Feel free to suggest css changes, though I'd rather you make the css changes and test them yourself, because I've already spent enough time on getting them to work. 

There are some long items ("Und-batch" and "Since mine") that necessitate the menu to wide enough to accommodate them. When neither of these are there, I hope the menu doesn't look oddly left-shifted? 

The `selected` class is added to the button when menu is being shown. On monobook, this causes the button size to increase which means the elements right to it "jump". If you think this is poor UI/UX, this can be removed (for monobook).

Amorymeltzer:
@siddharthvp That was fast, looks sweet!  Can you split the menu addition into a separate PR?  The little one and I are getting over a fun infection, so I'd like to start up with some reviews in a bit, and I think it'd be better to deal with these items separately, if only so we can get proper feedback on your dope addition.

siddharthvp:
If commits 2-4 were to be split to another PR, they'd conflict with commit 1. Should I leave this PR with just the first commit, and open a second one that has all 4 (since I don't think it's possible to apply commits 2-4 over the current master, without applying commit 1 first), in which case I guess a merge conflict would again arise after the first PR is merged?  

I grow tired of doing all the paperwork (already spent about 2 hours for cleanly filing commits 2-4, cursing git and eslint all around), so would suggest keeping things as is.

Amorymeltzer:
@siddharthvp Sure thing.

@MusikAnimal I'm planning to get around to reviewing this tomorrow, but since you maintain MoreMenu and [are working on something similar](https://github.com/azatoth/twinkle/issues/716#issuecomment-537135285) I wouldn't mind your eyes on it at some point, just for confirmation it all gels.

siddharthvp:
I have put this up on testwiki. LGTM. Shout if anyone sees any visual issues.

siddharthvp:
Another thing: I realised that Twinkle appears to be doing some heavy-handed modifications to the site css in modern skin - lines 182-202 in `morebits.css`. I missed that earlier. Indeed, I wrote some menu css (in `twinkle.css`) that negates the css rules coming from morebits.css (which I had assumed were coming from the core skin code).  I _think_ this is not an issue to be solved, as people who continue to use buttons will need those css rules. The css I added to twinkle.css are all strictly applicable only to twinkle menu (`#ca-twinkle-menu`) so won't conflict with anything.

MusikAnimal:
This generally looks good! I know Haza-w's dropdown menus might have done the same thing, but I'm not super fond of boldening the "twinkle" label on hover, because this changes the size and shifts any menu items to the right of it.

I'm slowly finishing up the new version of MoreMenu, and it doesn't quite look the same. One thing I did for Monobook was inherit the apparently preferred tiny font size (95%, where Twinkle is using 14px). Not sure if want to do the same. I also used a white background, since the dark gray doesn't seem to match anything else in the skin. Finally, the left side of the dropdown aligns with the left side of the top menu item ("twinkle" in our case). I don't think there's any expectation for MoreMenu and Twinkle to look the same, as there is no native menu system for these skins anyway, but it'd be neat because many will be using both gadgets. That said I'm not claiming MoreMenu's implementation to be better, either! CSS is at https://github.com/MusikAnimal/MoreMenu/blob/08932e31b6924bc98eb2f7708fc9cda77ff2f2ce/src/MoreMenu.js#L285 should you be interested. `.mm-item` are items in the dropdown and `.mm-menu` is the menu as a whole.

siddharthvp:
Wow, your menus definitely look better @MusikAnimal! Mine was quite a rushed job, cobbled together in 3 days. It looked appealing to me when it was done, but no longer does after seeing yours. I wasn't much worried about the aesthetics and just tried to make it usable. I like your approach of having different sets of styles for each skin, as that really helps in making very specific skin-specific tweaks. I agree that the twinkle menus should similar to that of moremenu, as they look so much more native-ish. Do you want to adapt it for twinkle yourself in a new PR? 

The inherited font size is ok but a bigger `line-height` (say 2) would be better, to match that of vector and modern, and in consideration of usage on touch devices.

I see that you've also not bothered with the complicated code that makes the menu go away only after so-and-so milliseconds (`hovms`) after `mouseleave`. Should that be there? Haza-w's original had hovms=400, which looked too slow to me so I reduced it to 200. Now it's so fast that it almost looks instantaneous. I seem to think you're right in removing this outright. 

(BTW note that your menus don't work in monobook mobile layout, window.innerWidth < 550px. And the submenus aren't visible in timeless when 1100 < innerWidth <1340px.)

MusikAnimal:
> Wow, your menus definitely look better MusikAnimal!

Thank you :)

> I agree that the twinkle menus should similar to that of moremenu, as they look so much more native-ish. Do you want to adapt it for twinkle yourself in a new PR?

I could, but it might be a while. I really want to finish up with MoreMenu first. Then I'll have more stable code that we can use in Twinkle, if we want.

> The inherited font size is ok but a bigger `line-height` (say 2) would be better, to match that of vector and modern, and in consideration of usage on touch devices.

I will toy around with this. Right now it's just inheriting everything, which comes out to 1.5em line-height and 95% font size (for Monobook). I agree it is a bit small when there is a list of links.

> I see that you've also not bothered with the complicated code that makes the menu go away only after so-and-so milliseconds (`hovms`) after `mouseleave`. Should that be there? Haza-w's original had hovms=400, which looked too slow to me so I reduced it to 200. Now it's so fast that it almost looks instantaneous. I seem to think you're right in removing this outright.

Vector menus don't have any delay. I've found that once you get used to it, you can easily mouse over into a submenu without losing focus, and the same seems to hold true with the other skins.

> BTW note that your menus don't work in monobook mobile layout, window.innerWidth < 550px. And the submenus aren't visible in timeless when 1100 < innerWidth <1340px.

More work to do, indeed! Thanks for raising these issues.

The positioning of the menus also fails miserably on Monobook/Modern in RTL languages. I've been struggling to figure that one out.

MusikAnimal:
MoreMenu is now stable, I think. If you wanted to base your design off of it, the CSS can be found at https://github.com/MusikAnimal/MoreMenu/blob/b89892a889cac69f91f96ca88d9d77e9e746cbd2/src/MoreMenu.js#L345 (ignore the `.mm-submenu` stuff). This of course is purely my design so I again don't want to insist Twinkle use the same, but it would be cool if it did :)

siddharthvp:
> This of course is purely my design so I again don't want to insist Twinkle use the same, but it would be cool if it did :)

You should consider creating some library-like setup for the menus for easy reuse by other script developers.

<hr>

For now, I've opened #763 to handle the easy stuff - vector (and timeless). The cleanup from commit 3 has been incorporated over there. Monobook/Modern menus need a bit more time to handle which I don't have atm, and I don't know how much (if any) of the code in this PR will be reused while replicating MA's menus. Closing, as I think it's better to start over from 763.

PR #722, opened 2019-10-07
morebits: Use custom message for spamblacklist

Amorymeltzer:
Restores custom message removed in #707.  Custom messages as specified in MediaWiki:Spamprotectiontext and MediaWiki:Spamprotectionmatch get inserted into the response, which can lead to absurdly lengthy wikitext messages in Twinkle (see https://github.com/azatoth/twinkle/pull/707#issuecomment-539120184).

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2685,7 +2685,9 @@ Morebits.wiki.page = function(pageName, currentAction) {
 				}
 			// check for blacklist hits
 			} else if (errorCode === 'spamblacklist') {
-				ctx.statusElement.error(ctx.saveApi.getErrorText());
+				// .find('matches') returns an array in case multiple items are blacklisted, we only return the first
+				var spam = $(ctx.saveApi.getXML()).find('spamblacklist').find('matches').children()[0].textContent;
+				ctx.statusElement.error('Could not save the page because the URL ' + spam + ' is on the spam blacklist');
 			} else {
 				ctx.statusElement.error('Failed to save edit: ' + ctx.saveApi.getErrorText());
 			}


Amorymeltzer:
Thanks for reporting this, @siddharthvp

Amorymeltzer:
Live on testwiki in case anyone wants to test, but it seems to get me the response I want, so I'll push to enwiki in ~hour or so if no issues.

siddharthvp:
This is the message I see currently on testwiki:
Screenshot (10) png

Seems completely different from what you wrote. Didn't try investigating why, through.

Amorymeltzer:
@siddharthvp Did you use revert or something else from `fluff`?  This PR just tackles `morebits`, I believe what you're noticing is due to what you and I were talking about in #707, namely that since `fluff` uses `Morebits.wiki.api` and not `Morebits.wiki.page` nor does it provide an `onError` fallback, we end up with this.

We can either duplicate the `morebits` stuff in an additional `fluff` callback (quicker, worse) or convert to `Morebits.wiki.page` (better).

siddharthvp:
I used CSD, specifically tried to nominate the page for G12 deletion specifying copyvio source as a spamblacklisted page.

Amorymeltzer:
Try again?  Works just fine for me, same thing.

siddharthvp:
This time, tried to PROD the page with reason http://areallyevilsite.com . Same result: 
Screenshot (11) png

Also tried after clearing browser caches and cookies. Anyway, I guess this error text is also fine howsoever it is coming :)

But another bug is revealed here as you can see: placing of {{old prod}} succeeded w/o the prod being added :(

Amorymeltzer:
Do you have some customjs loading or something?  I get the correct message
when I try to prod a page...  And yes, same issue with the old prod :-1:

~A

PR #725, opened 2019-10-12
morebits: remove Morebits.queryString

siddharthvp:
Morebits.queryString is redundant to mw.util.getParamValue(), mw.util.getUrl() and jQuery.param(). All existing usages are easily replaced.
Specifically,
- Morebits.queryString.get() and Morebits.queryString.exists() are redundant to mw.util.getParamValue()
- Morebits.queryString.create() is redundant to mw.util.getUrl() and jQuery.param()

While queryString can be also used for constructing query string objects and getting/setting param values in them, this functionality is not used anywhere, nor are there any reasonable potential usecases.

This is untested, for now.

### DIFF
--- a/modules/twinklebatchprotect.js
+++ b/modules/twinklebatchprotect.js
@@ -282,7 +282,7 @@ Twinkle.batchprotect.callback = function twinklebatchprotectCallback() {
 	} else if (mw.config.get('wgCanonicalSpecialPageName') === 'Prefixindex') {
 		query.generator = 'allpages';
 		query.gapnamespace = mw.util.getParamValue('namespace') || $('select[name=namespace]').val();
-		query.gapprefix = Morebits.string.toUpperCaseFirstChar(mw.util.getParamValue('from') ? mw.util.getParamValue('from').replace('+', ' ') : $('input[name=prefix]').val());
+		query.gapprefix = mw.util.getParamValue('prefix') || $('input[name=prefix]').val();
 		query.gaplimit = Twinkle.getPref('batchMax'); // the max for sysops
 	} else {
 		query.generator = 'links';
--- a/modules/twinklebatchprotect.js
+++ b/modules/twinklebatchprotect.js
@@ -281,9 +281,8 @@ Twinkle.batchprotect.callback = function twinklebatchprotectCallback() {
 		query.gcmlimit = Twinkle.getPref('batchMax'); // the max for sysops
 	} else if (mw.config.get('wgCanonicalSpecialPageName') === 'Prefixindex') {
 		query.generator = 'allpages';
-		query.gapnamespace = Morebits.queryString.exists('namespace') ? Morebits.queryString.get('namespace') : $('select[name=namespace]').val();
-		query.gapprefix = Morebits.queryString.exists('from') ? Morebits.string.toUpperCaseFirstChar(Morebits.queryString.get('from').replace('+', ' ')) :
-			Morebits.string.toUpperCaseFirstChar($('input[name=prefix]').val());
+		query.gapnamespace = mw.util.getParamValue('namespace') || $('select[name=namespace]').val();
+		query.gapprefix = Morebits.string.toUpperCaseFirstChar(mw.util.getParamValue('from') ? mw.util.getParamValue('from').replace('+', ' ') : $('input[name=prefix]').val());
 		query.gaplimit = Twinkle.getPref('batchMax'); // the max for sysops
 	} else {
 		query.generator = 'links';
--- a/morebits.js
+++ b/morebits.js
@@ -1380,11 +1380,20 @@ Morebits.wiki.api.prototype = {
 
 		++Morebits.wiki.numberOfActionsLeft;
 
+		var queryString = $.map(this.query, function(val, i) {
+			if (Array.isArray(val)) {
+				return encodeURIComponent(i) + '=' + val.map(encodeURIComponent).join('|');
+			} else if (val !== undefined) {
+				return encodeURIComponent(i) + '=' + encodeURIComponent(val);
+			}
+		}).join('&').replace(/^(.*?)(\btoken=[^&]*)&(.*)/, '$1$3&$2');
+		// token should always be the last item in the query string (bug TW-B-0013)
+
 		var ajaxparams = $.extend({}, {
 			context: this,
 			type: 'POST',
 			url: mw.util.wikiScript('api'),
-			data: Morebits.queryString.create(this.query),
+			data: queryString,
 			dataType: 'xml',
 			headers: {
 				'Api-User-Agent': morebitsWikiApiUserAgent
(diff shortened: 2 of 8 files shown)


siddharthvp:
It's also worth noting that the whole `Morebits.queryString` class is redundant to the native JS `URLSearchParams` interface (not available in IE, though).

Amorymeltzer:
Haven't taken a look at the code yet, but wholeheartedly agree with this.  I think it was the fluff stuff I was doing earlier this year that made me really wonder why we still bothered with `queryString`.

morebits.js
```diff
@@ -1380,11 +1380,18 @@ Morebits.wiki.api.prototype = {
 
 		++Morebits.wiki.numberOfActionsLeft;
 
+		var queryString = $.param(this.query);
+		if (this.query.token) { // token should always be the last item in the query string (bug TW-B-0013)
+			queryString = queryString
+				.replace(/^(.*?)(&token=[^&]*)(.*)/, '$1$3$2') // shunt token from middle of query string to the end
+				.replace(/^(token=[^&]*)&(.*)/, '$2&$1'); // shunt token from beginning of query string to the end
+		}
+
```
siddharthvp: I wrote this initially as 
```js
	if (this.query.token) {
		var token = this.query.token;
		delete this.query.token;
		this.query.token = token;
	}
```
followed by $.param(this.query.token), but the `delete` operator is purported to cause performance impacts in V8, and isn't used anywhere else in Twinkle. Anyway, the two lines of ugly regex is much better than the 25 lines of ugly code that was there originally in `Morebits.queryString.create`!

Amorymeltzer:
Not sure I'll time before the holidays, but just some history as I start clearing the way:

- @azatoth first [added Jan 2007](https://en.wikipedia.org/w/index.php?title=User%3AAzaToth%2Fmorebits.js&diff=prev&oldid=102122570)
- Initially, seems like the motivation was for [fluff](https://en.wikipedia.org/w/index.php?title=User:AzaToth/twinklefluff.js&oldid=102122926)
- `mw.util.getParamValue` [available](https://phabricator.wikimedia.org/source/mediawiki/browse/master/resources/mediawiki/mediawiki.util.js;51e3fb9a716f3ccd3b114109e5a5045a09864ff0$70) in the [initial version](https://phabricator.wikimedia.org/rMW51e3fb9a716f3ccd3b114109e5a5045a09864ff0) of `mediawiki.util` in Oct 2010, circa 1.17
- `mw.util.getUrl` began life as `mw.util.wikiGetLink` [soon after](https://phabricator.wikimedia.org/rMWc28ba513342a9cbbcec8f4bbb580e833ee00a06a), and was [renamed](https://phabricator.wikimedia.org/rMW35970fb460060cab9a3984030241987a86146f09#change-vsGcyhoHyk4t) circa 1.23
- jQuery added to mediawiki in 1.16/1.17 at v1.3.2/1.4.2, respectively

Amorymeltzer:
Is it worth attempting to deprecate these first?  Maybe swapping in their equivalents would be easiest too start, and catch any edge cases. bI haven't yet checked usages elsewhere, but if there are none as you say @Siddharthvp, then maybe that's overkill/assigning too much importance.

siddharthvp:
[18 usages in userspace](https://en.wikipedia.org/w/index.php?sort=relevance&search=%22Morebits.queryString%22&title=Special%3ASearch&profile=advanced&fulltext=1&advancedSearch-current=%7B%7D&ns2=1):

- [User:Gigs/delsort](https://en.wikipedia.org/wiki/User:Gigs/delsort.js) - no one uses it + still uses sajax for API calls (so there's no way this can work)
- [User:Livitup/SimpleVote2012-code.js](https://en.wikipedia.org/wiki/User:Livitup/SimpleVote2012-code.js) ([4 users](https://en.wikipedia.org/w/index.php?sort=relevance&search=%22User%3ALivitup%2FSimpleVote2012.js%22&title=Special:Search&profile=advanced&fulltext=1&advancedSearch-current=%7B%7D&ns2=1)) - can be fixed by changing 
```
var query = new Morebits.queryString( anchors[i].href.split('?',2)[1].toString() );
var link = query.get('title');
``` 
to 
```
var link = anchors[i].match(/\btitle=([^&]*)/)[1]; 
``` 
if at all it worked before.
- [User:This,_that_and_the_other/SimpleVote2012-code.js](https://en.wikipedia.org/wiki/User:This,_that_and_the_other/SimpleVote2012-code.js) - fork of Livitup's version above. Unused. 
- Remaining are all unused copies of Twinkle code.

I don't think there's any usage worth giving deprecation warnings for.

modules/twinklebatchprotect.js
```diff
@@ -281,9 +281,8 @@ Twinkle.batchprotect.callback = function twinklebatchprotectCallback() {
 		query.gcmlimit = Twinkle.getPref('batchMax'); // the max for sysops
 	} else if (mw.config.get('wgCanonicalSpecialPageName') === 'Prefixindex') {
 		query.generator = 'allpages';
-		query.gapnamespace = Morebits.queryString.exists('namespace') ? Morebits.queryString.get('namespace') : $('select[name=namespace]').val();
-		query.gapprefix = Morebits.queryString.exists('from') ? Morebits.string.toUpperCaseFirstChar(Morebits.queryString.get('from').replace('+', ' ')) :
-			Morebits.string.toUpperCaseFirstChar($('input[name=prefix]').val());
+		query.gapnamespace = mw.util.getParamValue('namespace') || $('select[name=namespace]').val();
+		query.gapprefix = mw.util.getParamValue('prefix') || $('input[name=prefix]').val();
```
Amorymeltzer: Wow!  `from` is used to paginate, while `prefix` is the actual prefix.  That is, `prefix` defines the list (e.g. [`prefix=A`](https://en.wikipedia.org/wiki/Special:PrefixIndex?prefix=A&namespace=0)) while `from` defines what is shown on the user's screen (e.g. [`prefix=A&from=Amory_Lovins`](https://en.wikipedia.org/wiki/Special:PrefixIndex?prefix=A&from=Amory_Lovins&namespace=0)).  `batchprotect` has been busted on non-first-page results for years!  Very lame!

morebits.js
Amorymeltzer: Much nicer!  The original is certainly over-complicated, but I think casting into an array is indeed the generic way to do it (sans jQuery and `morebits` of course).  Either way, this should work for our specific purposes, so w00t!

Could we make it even tighter, though, by relying on the fact that tokens end in `+\`?  Would something like the following be sufficient?

```javascript
queryString = $.param(this.query).replace(/^(.*?)(token=[^%]+%2B%5C)&(.*)/, '$1$3&$2')
```

It:
* matches until `%`
* only matches if *something* is present (not sure what happens if there's an empty token parameter isn't last, but that shouldn't happen)
* removes the trailing `&` and places one before the terminal `token`:
    * if `token` is first, the `?` remains for the new first parameter
    * if `token` is in the middle, `token`'s `&` remains for the upcoming parameter
    * if `token` is last, there won't be a terminal `&` to match and there's nothing to do anyway!

morebits.js
```diff
@@ -3343,147 +3350,6 @@ Morebits.wikitext.page.prototype = {
...
- * var value = Morebits.queryString.get('key');
- * var obj = new Morebits.queryString('foo=bar&baz=quux');
- * value = obj.get('foo');
- */
-
-/**
- * @constructor
- * @param {string} qString   The query string
- */
-Morebits.queryString = function QueryString(qString) {
```
Amorymeltzer: This seems useful, maybe with the default of `location.search.substring(1)`

siddharthvp:
> The only thing I do think might be worth holding onto is the Morebits.queryString function itself, or at least something like it. I'm not familiar with an easy way using jQuery or mw.util to build an object with each parameter (e.g. new Morebits.queryString(string).params), am I wrong on that? It's not used by Twinkle, but that feels useful and fairly low overhead.

`new mw.Uri(string).query` does do that, and the `mw.Uri` dependency is pretty cheap. The only trouble with it is that it needs a full URL and won't accept just a querystring. 

Do you still think that this is needed @Amorymeltzer?

Amorymeltzer:
> `new mw.Uri(string).query` does do that, and the `mw.Uri` dependency is pretty cheap.

:100: TIL!  And what dependency, is it not in core?

>The only trouble with it is that it needs a full URL and won't accept just a querystring.

That's not an issue for me.  I was honestly more interested in the default behavior using the current location string, as I thought it could be handy to be able to easily iterate on each of the parameters.  It should only make a difference if trying to parse a pseudo link on the page that had the parameter string but not the full url (i.e. someone wrote "You can just append `?query=action&titles=Macbeth` to the `en.wikipedia.org` but use `/w/index.php` instead of `/wiki/" which is, uh, not a usecase.  Otherwise, it's a full url, the default from `location.href`, an object we already own (making it pointless), or we can just toss it behind `mw.util.getUrl`.  All easy.

> Do you still think that this is needed @Amorymeltzer?

Nope! 👍

morebits.js
siddharthvp: Great work in condensing it down to a single regex replace!

> (token=[^%]+%2B%5C)

not so sure whether this is an improvement. While a nice idea, it seems better to me to leave it as 

> (token=[^&]*)

$.param(this.query) encodes the URI components so `&` can't occur as part of the token itself. So encountering an `&` means that the token part has ended. So we aren't really tightening up anything right?

siddharthvp: NOTE: The code here doesn't account for the other features of `Morebits.queryString.create` - omitting parameters with value `undefined` and converting array parameters to `|`-delimited strings. 

This needs to be resolved before merge.

Amorymeltzer: So that'd be `queryString = $.param(this.query).replace(/^(.*?)(token=[^&]+)&(.*)/, '$1$3&$2')`, right? (`+` rather than `*`)

siddharthvp: meh, using `*` means an empty `token=` also gets moved to the end, which is better right?

Amorymeltzer:
@siddharthvp With https://github.com/azatoth/twinkle/compare/dde1faa4143ac42ce8b788034243186e89d1ae1a..5bd60477a2f829cedeaeaae741d628dce77cbc3b am I right that you're done working on this, and I can give it a final review?  Looking forward to it!

siddharthvp:
@Amorymeltzer yeah 👍 sorry if was being confusing above.

morebits.js
```diff
@@ -1380,11 +1380,20 @@ Morebits.wiki.api.prototype = {
 
 		++Morebits.wiki.numberOfActionsLeft;
 
+		var queryString = $.map(q, function(val, i) {
+			if (Array.isArray(val)) {
+				return encodeURIComponent(i) + '=' + val.map(encodeURIComponent).join('|');
+			} else if (val !== undefined) {
+				return encodeURIComponent(i) + '=' + encodeURIComponent(val);
+			}
+		}).join('&').replace(/^(.*?)(token=[^&]*)&(.*)/, '$1$3&$2');
```
Amorymeltzer: Okay there's actually another issue here @siddharthvp — for queries with `intoken` (e.g. editing) this leaves the `in` behind for another parameter.  Not good.  We're not going to overhaul the token system right now, so for now I think this should do it:

```suggestion
		}).join('&').replace(/^(.*?)((?:in)?token=[^&]*)&(.*)/, '$1$3&$2');
```

But I'm not 100% it's good enough.

Amorymeltzer: Actually, I think that's technically backwards; we *don't* want to move `intoken` (well, I suppose it doesn't really matter) so it'd be better to do something like either:

```suggestion
		}).join('&').replace(/^(.*?)[&\?](token=[^&]*)&(.*)/, '$1$3&$2');
```

or

```suggestion
		}).join('&').replace(/^(.*?)([^in]token=[^&]*)&(.*)/, '$1$3&$2');
```

siddharthvp: How about using `\b` ? It has zero-length match, so I _think_ that should be ideal.

Amorymeltzer: Mmm, yes, I think that's probably right/best/parsimonious

siddharthvp: appears to work fine, pushed.

morebits.js
```diff
@@ -1380,11 +1380,20 @@ Morebits.wiki.api.prototype = {
 
 		++Morebits.wiki.numberOfActionsLeft;
 
+		var queryString = $.map(q, function(val, i) {
```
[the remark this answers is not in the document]
siddharthvp: Haha, of course.

morebits.js
Amorymeltzer: Is there a difference between `\b(token...` and `(\btoken...`?  I'm not thinking there should be off hand.

siddharthvp: `\b` has a zero-length match, so no.

Amorymeltzer: Yeah - I was wondering more about ZWSP and exotics like that, I curious how they "counted."  Turns out they're `\S` and `\W!

Amorymeltzer:
I'll do some more testing later today, but I think this is good to go!

Amorymeltzer:
Merged as 37d6b670d38d2b3c6060490ad71eda5410561508 and bf1919fdd79ba7b9c342503d8acc2170568bba17

PR #724, opened 2019-10-12
tag: add quick filter for redirects and files too

siddharthvp:
Extend quick filtering (added in #647) to redirects and files too.

PR #728, opened 2019-10-16
tag: Move quick filter up, move patrol checkbox to bottom

Amorymeltzer:
The same quick filter line is called for each mode, but by putting it before the `switch (Twinkle.tag.mode)`, the quick filter will appear above the article sorting toggle, which should make more sense given its usefulness.  Also move the patrol checkbox to the bottom, near where the MI box would show up; birds of a feather and all that.

----
This occurred to me right after merging @siddharthvp's excellent changes in #724.  There's no need for `Window.setTitle` to be first, and the only other noticeable change is if the `switch` input is entirely broken, wherein the phantom Twinkle window would now have a quickfilter instead of being totally empty; not exactly a deal-breaker.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -43,7 +43,47 @@ Twinkle.tag.callback = function friendlytagCallback() {
 	Window.addFooterLink('Twinkle help', 'WP:TW/DOC#tag');
 
 	var form = new Morebits.quickForm(Twinkle.tag.callback.evaluate);
-	Twinkle.tag.quickFilter(form);
+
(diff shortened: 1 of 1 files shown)


siddharthvp:
This is nice. Thoughts on doing away with the `Twinkle.tag.quickFilter` function, since it's now used just once?

Amorymeltzer:
Hmm, I suppose that makes sense.  I kinda liked it out of the way, but I suppose it's short enough not to damage readability too much.  At the very least I should've added a comment!

Issue #414, opened 2018-03-20
Twinkle should use OOUI


Huji:
Currently, Twinkle uses jQuery.tipsy and other similar modules that are considered deprecated by MediaWiki's ResourceLoader. It also uses jQuery UI to draw its menus.

Since everything in MediaWiki is moving in the direction of OOUI, so should Twinkle.

atlight:
See #384 for some relevant discussion.

Huji:
Noted. My understanding of #384 is that, in essence, you had "We don't have any plans to move away from jQuery UI at the moment."

It is okay if you want to close this issue now; however, alternatively, we can keep it open for when the issue of movability of dialogues is addressed in OOUI. Is there a Phab ticket opened for it?

atlight:
Let's keep it open; I'm not sure why the other issue was closed. 

The relevant Phabricator task is https://phabricator.wikimedia.org/T51969. Bartosz's suggestion of a split-screen interface is interesting; I can imagine a fixed panel appearing on the right of the screen, which pushes the page content over to the left, as a very realistic possibility. As James Forrester observes, being able to move dialogs depends on your screen size, and does not necessarily allow the desired part(s) of the page to be viewed.

zchrykng:
Well... that phab thread is depressing.

Issue #384, opened 2017-05-10
jquery.ui


LunarTwilight:
Twinkle throws the following errors in console:

`This page is using the deprecated ResourceLoader module "jquery.ui.core".
Please use "mediawiki.ui.button" or "oojs-ui" instead.`

`This page is using the deprecated ResourceLoader module "jquery.ui.position".`

`This page is using the deprecated ResourceLoader module "jquery.ui.widget".`

atlight:
We don't have any plans to move away from jQuery UI at the moment. The fatal flaw of OOjs UI is that it doesn't allow the user to move dialogs around on the page, which is critical for Twinkle.

LunarTwilight:
I might be wrong, and I probably am, but wouldn't buttons and etc not affect the dialogs being movable?

atlight:
We use jQuery UI for Twinkle's dialogs. Although `jquery.ui.dialog` doesn't appear to have been explicitly deprecated, it depends on the three modules you've listed, so we would need to replace jQuery UI dialogs with OOjs UI dialogs in order to avoid the deprecation warnings.

PR #729, opened 2019-10-17
jquery.ui.dialog (and others) being moved within jquery.ui

Amorymeltzer:
See https://phabricator.wikimedia.org/T219604#5374960 and subsequent patches (e.g. https://gerrit.wikimedia.org/r/#/c/mediawiki/core/+/543736/)
Rearranging deck chairs before #414/https://phabricator.wikimedia.org/T49145

----
Assuming the train goes out on the 24th as planned, I'll merge this then.  Won't break anything not do so, just another console warning (*cough #384 cough*)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ There are two ways to upload Twinkle scripts to Wikipedia or another destination
 
 After the files are synced, [MediaWiki:Gadgets-definition][] should contain the following lines:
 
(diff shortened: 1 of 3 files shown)


siddharthvp:
Seems strange: this doesn't seem to have been discussed anywhere on phab (or was it)? I wonder what's the advantage in loading all the jquery modules when only a few are needed. If anything, wouldn't this just make it tougher for WMF to assess the usage of various jquery modules, now that each one isn't going to identified individually in gadget definitions?

Amorymeltzer:
That specific task goes into the genesis of it a bit, but with `jquery.ui` intended on being taken out entirely (T122014 and T100270 if you want to peruse a few hundred tasks) it's not unreasonable. 
 More to the point, the currently used `jquery.ui` is outdated, possibly permanently so, and so folks are manually dealing with/fixing issues, so I do understand steadily removing items.  I suppose consolidating would help down the line for later removals?  I'm sort of in agreement with you as regards tracking, but I think they just don't care since it's been "deprecated" for a while and its days are numbered anyway (although it's a big number...).

At the very least, we'll end up with one console warning instead of three.

PR #732, opened 2019-10-20
tag: fix bug when there are no available tags

Amorymeltzer:
If there are no available tags (all are already present or none provided) the module would die when trying to untag (like a printer low on cyan).  This is incredibly unlikely on enwiki, but I can theoretically imagine other wikis with fewer maintenance templates might run into an issue where all available templates are already transcluded.  Reported [on my testwiki talk](https://test.wikipedia.org/w/index.php?title=User_talk:Amorymeltzer&oldid=405366).

----

I spent some time delving into other tightenings when the various tag objects were empty or emptied, but it's only partially doable without some major restructuring, and I'm not sure it's even a good idea.


Amorymeltzer:
Yeah — intuitively that makes sense, not sure why `null` was chosen?  I suppose it's strictly more "correct" in some view?  I haven't looked at what/where that would require changing

PR #735, opened 2019-10-24
prod: old prod/xfd detection bug fix

siddharthvp:
First commit: fix bug reported on [Twinkle talk](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Article_PRODed_twice). 

2nd commit: remove redirected/deleted templates from the list. Also, the old drv templates were missing the "Template:" prefix, due to which they weren't working.

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -433,8 +433,8 @@ Twinkle.prod.callback.evaluate = function twinkleprodCallbackEvaluate(e) {
 	var talk_title = new mw.Title(mw.config.get('wgPageName')).getTalkPage().getPrefixedText();
 	// Talk page templates for PROD-able discussions
 	var blocking_templates = 'Template:Old XfD multi|Template:Old MfD|Template:Oldffdfull|' + // Common prior XfD talk page templates
-		'Template:Multidel|Template:Oldpuffull|' + // Uncommon/legacy prior XfD templates
-		'Olddrvfull|Olddelrev|' + // Prior DRV templates
+		'Template:Oldpuffull|' + // Legacy prior XfD template
+		'Template:Olddelrev|' + // Prior DRV template
 		'Template:Old prod';
 	var query = {
 		'action': 'query',
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -181,22 +181,25 @@ Twinkle.prod.callbacks = {
 		var statelem = apiobj.statelem;
 		var params = apiobj.params;
 
-		// Check talk page for templates indicating prior XfD
-		if ($(xmlDoc).find('templates').length) {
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Much better 👍!  Seems like a confluence of doing #566 then not thinking while doing #637.  I vaguely recall having a good reason for wanting to check the category, but can't think of it.  Whatever, churn don't care!

modules/twinkleprod.js
```diff
@@ -430,13 +430,13 @@ Twinkle.prod.callback.evaluate = function twinkleprodCallbackEvaluate(e) {
 	var talk_title = new mw.Title(mw.config.get('wgPageName')).getTalkPage().getPrefixedText();
 	// Talk page templates for PROD-able discussions
 	var blocking_templates = 'Template:Old XfD multi|Template:Old MfD|Template:Oldffdfull|' + // Common prior XfD talk page templates
-		'Template:Multidel|Template:Oldpuffull|' + // Uncommon/legacy prior XfD templates
-		'Olddrvfull|Olddelrev'; // Prior DRV templates
+		'Template:Oldpuffull|' + // Legacy prior XfD template
+		'Template:Olddelrev|' + // Prior DRV template
```
Amorymeltzer: Good catch on the `Template:` prefix

modules/twinkleprod.js
```diff
@@ -181,24 +181,24 @@ Twinkle.prod.callbacks = {
...
-			statelem.warn('Previous XfD template found on talk page, aborting procedure');
-			return;
-		}
-
-		// Check talk page for category indicating prior PROD
-		if ($(xmlDoc).find('categories').length) {
-			if (params.blp) {
-				if (!confirm('Previous PROD nomination found on talk page. Do you still want to continue applying BLPPROD? ')) {
-					statelem.warn('Previous PROD found on talk page, aborted by user');
+			var template = $(xmlDoc).find('templates tl')[0].getAttribute('title');
```
Amorymeltzer: Is the return order guaranteed here?  I seem to always get `Old XfD multi` before `Old prod`, and I think it sorts capitalized templates first, so if true, `Old prod` should actually *always* be last given our list of templates.  This works because of that, but at the very least should include a comment as a warning either here or below in case anything changes.

That being said, wouldn't it be better to check the length of `$(xmlDoc).find('templates tl')`?  That way, we show the PROD message iff there's one template and it's Old prod; for >1 templates or just one but *not* Old prod, show the XfD message.  That should cover our bases more thoroughly.

siddharthvp: > I think that's because it sorts capitalized templates first. 

Indeed, uppercase letters have lesser ASCII values are thus sorted ahead.

Regarding the 2nd para - yes, that sounds more robust. Changed.

PR #738, opened 2019-10-27
protect: Remove PC option entirely on inappropriate namespaces

Amorymeltzer:
Was disabled ~six months ago in #560 as a transition, but like #451, we should remove inappropriate options from the interface


Amorymeltzer:
Merged as 02375375af09903efb2e4993402c5adb0b2ecba9.

Issue #6, opened 2011-06-02
Avoid using the client clock for anything


Amalthea:
Twinkle uses the client clock for numerous tasks: section headers in user warnings and list pages for various XfD tasks come to mind.
Since client clocks are often a bit off (this comes up regularly), it would be beneficial to ask the server for the time, and use that one.

DannyS712:
What about, at least of section headers and adding the time _to a page_, using substituted time parser functions? See https://en.wikipedia.org/wiki/Help:Magic_words#Other_variables_by_type

siddharthvp:
[Relevant SO discussion ](https://stackoverflow.com/questions/38023193/javascript-date-now-consistency-across-machines-time-zones)

siddharthvp:
> What about, at least of section headers and adding the time _to a page_, using substituted time parser functions? See https://en.wikipedia.org/wiki/Help:Magic_words#Other_variables_by_type

In all the places where we need to add time to a page (CSD/PROD userspace logging and user talk page warning), we also need to check whether the section header of the current month is already there, so we must have the time beforehand.

siddharthvp:
> When you request a page for later editing from the api, you already get the starttimestamp in UTC. Can't that be reused ?

Indeed. Opened a PR for reusing this wherever possible.

PR #741, opened 2019-10-29
Use server time rather than client time wherever possible 

siddharthvp:
Largely addresses #6. Per https://github.com/azatoth/twinkle/issues/6#issuecomment-1304315, the server time is available to us whenever we are in a `.load()` callback, via the starttimestamp attribute in load API output which is stored in Morebits.wiki.page as `ctx.loadTime`, publicly exposed via `.getLoadTime()`.

In case of RFD, the curtimestamp parameter is added to the API request while finding the target of the redirect, so that the server time is available while determining the log page.

The only remaining usages of client clock is in:
- an3 - not important to fix as it is only used for deciding what revisions to retrieve from which user can select reverts.
- in determining daily log page name for CFD, FFD and TFD. The code that does is sitting outside of any load callback, because unlike in AfD, tagging of the page is occurring after the discussion is posted. Maybe the order can be changed in a future Great XfD De-Duplication Project.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -1232,7 +1232,8 @@ Twinkle.xfd.callbacks = {
 				var query = {
 					'action': 'query',
 					'titles': mw.config.get('wgPageName'),
-					'redirects': true
+					'redirects': true,
+					'curtimestamp': true
 				};
 				var wikipedia_api = new Morebits.wiki.api('Finding target of redirect', query, Twinkle.xfd.callbacks.rfd.findTargetCallback(callback));
 				wikipedia_api.params = params;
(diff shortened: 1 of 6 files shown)


siddharthvp:
Note: the `starttimestamp` parameter in load API output is added by `intoken=edit` (it adds the token and this timestamp). If we remove that or replace that by `meta=tokens` in the future, the current timestamp will need to be explicitly requested using the `curtimestamp` param.

Amorymeltzer:
I've yet to thoroughly review this, but it all looks very reasonable; I'm surprised we've gone so long without wanting `getLoadTime`!

> Note: the `starttimestamp` parameter in load API output is added by `intoken=edit` (it adds the token and this timestamp). If we remove that or replace that by `meta=tokens` in the future, the current timestamp will need to be explicitly requested using the `curtimestamp` param.

Could you add a comment to this effect?  #615 can be our open ticket, but a comment in the code would be nice.  I'll try to get to this in the new year, sorry for the lag.

modules/twinklexfd.js
```diff
@@ -1242,20 +1243,22 @@ Twinkle.xfd.callbacks = {
 		// This is a closure for the callback from the above API request, which gets the target of the redirect
 		findTargetCallback: function(callback) {
 			return function(apiobj) {
-				var xmlDoc = apiobj.responseXML;
-				var target = $(xmlDoc).find('redirects r').first().attr('to');
+				var $xmlDoc = $(apiobj.responseXML);
```
Amorymeltzer: Just noting that elsewhere in `xfd` we don't use the jquery naming convention; this is the only one that actually uses it for more than convenience, so it's reasonable to do here.

modules/twinklexfd.js
```diff
@@ -1398,7 +1401,7 @@ Twinkle.xfd.callback.evaluate = function(e) {
 	}
 
 	var query, wikipedia_page, wikipedia_api, logpage, params;
-	var date = new Date();
+	var date = new Date(); // XXX: avoid use of client clock
```
```suggestion
	// XXX: avoid use of client clock, still used by TfD, FfF, and CfD
	var date = new Date();
```

Just for extra clarity.

morebits.js
```diff
@@ -2136,6 +2136,13 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		return ctx.pageExists;
 	};
 
+	/**
+	 * @returns {string} ISO 8601 timestamp at which the page was last loaded.
+	 */
```
Amorymeltzer: Per https://github.com/azatoth/twinkle/pull/741#issuecomment-568593321, could you add a brief comment here?  Just something noting that it's only available in a `.load` and even then only because `intoken=edit` adds `starttimestamp`, and will need to be fixed if/when moving to `meta=tokens` via `curtimestamp`.

siddharthvp: > something noting that it's only available in a .load 

It already says " ... timestamp at which the page was **_last loaded_**" And it's available anytime after a load, not necessarily in `.load` callback.

> only because intoken=edit adds starttimestamp, and will need to be fixed if/when moving to meta=tokens via curtimestamp.

such a comment belongs at the place where `ctx.loadTime` is being assigned (as the public interface wouldn't be affected). Added it there.

modules/twinklexfd.js
siddharthvp: 👍

siddharthvp:
Looks like my git directory on windows has went bust (weird permission denied errors). Will push the latest changes when later when I log into ubuntu.

Issue #699, opened 2019-09-02
Interfacing with UAA


siddharthvp:
Feature request from Twinkle talk page archives:

> UAA is set up with buttons after each report so admins can initiate a block directly from UAA, which is helpful because the bot immediately sees this and removes those reports (UAA is perpetually backlogged so getting rid of reports that have been handled already is very helpful). However, all it does is block, so the admin then has to either use Twinkle or manually edit to add the block notice. Is there a way to integrate these two functions? Beeblebrox (talk) 20:44, 4 November 2016 (UTC) 

Basically, the request is to modify the block buttons on [WP:UAA](https://en.wikipedia.org/wiki/Wikipedia:UAA) (generated from {{user-uaa}}) so that clicking on them opens the TW block interface (with an appropriate preset selected on basis of the type of block button) rather than take the user to Special:Block.

Amorymeltzer:
Three years old, eh?  I can ask Zaphod if they still think it'd be useful, but I don't think there's anything for us to do here.  It's not difficult; we'd, what, add something like `[http://en.wikipedia.org/wiki/Special:Contributions/BadEditor?uaablock=1 TW]` to the template and then check for the param on loading, I suppose.  But there're a lot of links in templates we could try to usurp/glom on to, so I think the real solution is sysops shouldn't click on the block button if they want Twinkle, they should click on the user page/user talk/contribs buttons and block from there.

Amorymeltzer:
I suppose technically we could modify the `twinkleblock` callback to take input from a link on places where `wgRelevantUserName` isn't defined, which might be smoother, but is still a slippery slope of templates or places.  We'd still need to modify the templates.

Amorymeltzer:
I consider #699 to be a better resolution to this, plus I spoke to the Beebs about just that.

PR #743, opened 2019-10-31
twinkle.js: Add Special:Block to whitelisted special pages

Amorymeltzer:
Allows users to use Twinkle modules there if desired.  As of now, it loads: arv, warn, welcome, and talkback, as well as block if the user is a sysop.

Also updated/unified headers describing where some modules that use `wgRelevantUserName` that weren't changed (#208).

----
IMO this solves the concern expressed in #699.

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -9,7 +9,7 @@
  *** friendlytalkback.js: Talkback module
  ****************************************
  * Mode of invocation:     Tab ("TB")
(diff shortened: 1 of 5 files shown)


Amorymeltzer:
I'd've bet you good money this morning that this was already the case...

siddharthvp:
You'll have to update twinkle-pagestyles.css as well.

Amorymeltzer:
Good catch, thanks!

PR #754, opened 2019-11-20
Morebits.js: Minor cleanup

DannyS712:
PS1: Remove some whitespace, standardize some error messages, etc. Should be a no-op.


siddharthvp:
Don't remove all those newlines. Everyone is bound to have a different opinion on where they're appropriate. So the one who writes or edits the surrounding code gets to decide. Also, this will cause at least a couple merge conflicts with other PRs.

PR #756, opened 2019-11-25
twinkle/morebits/config: remove antique IE 9 stuff

siddharthvp:
JavaScript no longer loads in MediaWiki in IE <11, so there's no point in keeping old IE version checks or special cases.


Amorymeltzer:
Rather than amending the commit message:
*JS disabled for IE 10 in [T187869](https://phabricator.wikimedia.org/T187869) ~April 2018
*Fuller history at https://www.mediawiki.org/wiki/Template:Compatibility_browser

PR #757, opened 2019-11-25
warn: Note if user attempting to warn for an old edit 

Amorymeltzer:
<s>This is something that Huggle already does, with a [limit of 1 day](https://github.com/huggle/huggle3-qt-lx/blob/f44d4f7ea0f7b83fb16e4d14e9bedf3c18b09f15/src/huggle_core/warnings.cpp#L73) for the alert.  24 hours is also the window in which Twinkle warns users that they are attempting to provide an identical warning to one already present, so the parallel "staleness" is nice.

We can do this by adding a `vantimestamp` parameter to the user warning links opened by `twinklefluff` and merely checks that against the current date when issuing a warning to an editor.  I've also added it to the additional parameters we tack onto the user talkpage links on diffs; <s>doing so requires parsing the text of the "revision as of" text, since of the four formats users can select in Special:Preferences, only one is actually a valid `Date` format.  That text is also language-dependent, and this won't pass muster on RTL languages.</s>  doing so requires an API hit since the actual timestamps aren't available (T239439).

The time of the edit in question isn't exposed on the rollback success page, so we can't add add `vantimestamp` to that link.</s>

----

New commit message taking the below in account:

This will show a message in red to the user if they are considering issuing a warning older than 24 hours, and may thus be stale.  This is something that Huggle already does, with a [limit of 1 day](https://github.com/huggle/huggle3-qt-lx/blob/f44d4f7ea0f7b83fb16e4d14e9bedf3c18b09f15/src/huggle_core/warnings.cpp#L73) for the alert.  24 hours is also the window in which Twinkle warns users that they are attempting to provide an identical warning to one already present, so the parallel "staleness" is nice.

We can do this by using the `vanarticlerevid` from `twinklefluff` on diffs to query for the timestamp; ideally, diff timestamps would be available without an API call (T239439).  The rollback success page exposes neither the revision or time of the edit in question, nor does the url, so we can't check for old diff from the user talk link; this is comparatively rare anyway.

Also use .getTime() throughout when comparing date objects.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -30,6 +30,7 @@ Twinkle.warn = function twinklewarn() {
 			$vandalTalkLink.css('font-weight', 'bold');
 			$vandalTalkLink.wrapInner($('<span/>').attr('title', 'If appropriate, you can use Twinkle to warn the user about their edits to this page.'));
 
+			// Can't provide vanarticlerevid as only wgCurRevisionId is provided
 			var extraParam = 'vanarticle=' + mw.util.rawurlencode(Morebits.pageNameNorm);
 			var href = $vandalTalkLink.attr('href');
 			if (href.indexOf('?') === -1) {
@@ -41,9 +42,14 @@ Twinkle.warn = function twinklewarn() {
 	} else if (mw.config.get('wgDiffNewId') || mw.config.get('wgDiffOldId')) {
 		// Autofill user talk links on diffs with vanarticle for easy
 		// warning, but don't autowarn
-		var warnFromTalk = function(talkLink) {
+		var warnFromTalk = function(xtitle) {
+			var talkLink = $('#mw-diff-' + xtitle + '2 .mw-usertoollinks a').first();
 			if (talkLink.length) {
 				var extraParams = 'vanarticle=' + mw.util.rawurlencode(Morebits.pageNameNorm) + '&' + 'noautowarn=true';
+				// diffIDs for vanarticlerevid
+				extraParams += '&vanarticlerevid=';
+				extraParams += xtitle === 'otitle' ? mw.config.get('wgDiffOldId') : mw.config.get('wgDiffNewId');
+
 				var href = talkLink.attr('href');
 				if (href.indexOf('?') === -1) {
 					talkLink.attr('href', href + '?' + extraParams);
@@ -53,8 +59,8 @@ Twinkle.warn = function twinklewarn() {
 			}
 		};
 
-		warnFromTalk($('#mw-diff-otitle2 .mw-usertoollinks a').first());
-		warnFromTalk($('#mw-diff-ntitle2 .mw-usertoollinks a').first());
+		warnFromTalk('otitle'); // Older diff
+		warnFromTalk('ntitle'); // Newer diff
 	}
 };
 
@@ -109,6 +115,34 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 		value: Morebits.queryString.exists('vanarticle') ? Morebits.queryString.get('vanarticle') : '',
 		tooltip: 'A page can be linked within the notice, perhaps because it was a revert to said page that dispatched this notice. Leave empty for no page to be linked.'
 	});
+
+	form.append({
+		type: 'div',
+		label: '',
+		style: 'color: red',
+		id: 'twinkle-warn-olddiff-message'
+	});
+
+	// Confirm edit wasn't too old for a warning
+	var vanrevid = mw.util.getParamValue('vanarticlerevid');
+	if (vanrevid) {
+		var query = {
+			action: 'query',
+			prop: 'revisions',
+			rvprop: 'timestamp',
+			revids: vanrevid
+		};
+		new Morebits.wiki.api('Grabbing the revision timestamps', query, function(apiobj) {
+			var vantimestamp = $(apiobj.getResponse()).find('revisions rev').attr('timestamp');
+			var revDate = new Date(vantimestamp);
+			if (vantimestamp && !isNaN(revDate)) {
+				revDate.setUTCHours(revDate.getUTCHours() + 24);
+				if (revDate.getTime() < new Date().getTime()) {
+					$('#twinkle-warn-olddiff-message').text('Note: This edit was made more than 24 hours ago so a warning may be stale.');
+				}
+			}
+		}).post();
+	}
 
 	var more = form.append({ type: 'field', name: 'reasonGroup', label: 'Warning information' });
 	more.append({ type: 'textarea', label: 'Optional message:', name: 'reason', tooltip: 'Perhaps a reason, or that a more detailed notice must be appended' });
@@ -1332,10 +1366,10 @@ Twinkle.warn.callbacks = {
 
 		while ((current = history_re.exec(text)) !== null) {
 			var current_date = new Date(current[2] + current[3] + ' UTC');
-			if (!(current[1] in history) || history[current[1]] < current_date) {
+			if (!(current[1] in history) || history[current[1]].getTime() < current_date.getTime()) {
 				history[current[1]] = current_date;
 			}
-			if (current_date >= latest.date) {
+			if (current_date.getTime() >= latest.date.getTime()) {
 				latest.date = current_date;
 				latest.type = current[1];
 			}
@@ -1347,7 +1381,7 @@ Twinkle.warn.callbacks = {
 			var temp_time = new Date(history[params.sub_group]);
 			temp_time.setUTCHours(temp_time.getUTCHours() + 24);
 
-			if (temp_time > date) {
+			if (temp_time.getTime() > date.getTime()) {
 				if (!confirm('An identical ' + params.sub_group + ' has been issued in the last 24 hours.  \nWould you still like to add this warning/notice?')) {
 					statelem.error('aborted per user request');
 					return;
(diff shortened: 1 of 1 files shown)


modules/twinklewarn.js
```diff
@@ -1306,6 +1319,18 @@ Twinkle.warn.callbacks = {
 
 		var date = new Date();
 
+		var vantimestamp = parseInt(mw.util.getParamValue('vantimestamp'));
+		var revDate = new Date(vantimestamp);
+		if (vantimestamp && !isNaN(revDate)) {
+			revDate.setUTCHours(revDate.getUTCHours() + 24);
+			if (revDate < date) {
```
siddharthvp:
```suggestion
			if (revDate.getTime() < date.getTime()) {
```
type-correctness.

modules/twinklewarn.js
```diff
@@ -42,9 +42,22 @@ Twinkle.warn = function twinklewarn() {
...
 			if (talkLink.length) {
 				var extraParams = 'vanarticle=' + mw.util.rawurlencode(Morebits.pageNameNorm) + '&' + 'noautowarn=true';
+
+				// This relies upon the text shown above a diff, which depends upon the user's
+				// language selection and date format preference. While it could thus vary widely,
+				// once we remove the leading text, the date formats remain the same (fails of RTL).
+				// Of the four date formats users can select, only one is a
+				// valid Date format; the other three require removal of a comma.
+				var dateText = $('#mw-diff-' + xtitle + '1 a').first().text().replace(/^.*?(\d\d)/, '$1');
+				var vantimestamp = new Date(dateText.replace(/^(\d\d:\d\d),/, '$1')).getTime();
```
siddharthvp: Fails for RTL? This will fail on any language other than English, since JS Date() can't comprehend non-English month names. There are also languages that use non-Arabic numerals (such as Hindi) which aren't recognised by `\d`. 

This should be written as: 
```js
var diffTitleText = $('#mw-diff-' + xtitle + '1 a').first().text();
var dateobj = new Date(diffTitleText.replace(/^.*?(\d\d)/, '$1').replace(/^(\d\d:\d\d),/, '$1')); // language specific
var vantimestamp = dateobj.getTime();
```
so that the part that needs changing b/w wikis is all together.

modules/twinklewarn.js
```diff
@@ -54,8 +67,8 @@ Twinkle.warn = function twinklewarn() {
 			}
 		};
 
-		warnFromTalk($('#mw-diff-otitle2 .mw-usertoollinks a').first());
-		warnFromTalk($('#mw-diff-ntitle2 .mw-usertoollinks a').first());
+		warnFromTalk('otitle');
```
siddharthvp:
```suggestion
		warnFromTalk('otitle'); // older diff
```

And likewise for the latter. It's confusing to see passed nonsensical strings passed to the function, though i see there's no better alternative.

modules/twinklewarn.js
Amorymeltzer: Thanks, I dunno what I was smokin' when I wrote that.

In reality, this segment (hijacking the user talk links) isn't really tenable without either adding something like `wgDiffOldTimestamp` or an API call.  Relying on parsing text that is localized and formatted according to user preferences and customizable by wiki is just setting up for failure.  I think the former would be useful but I'd imagine the devs wouldn't be likely to add it.

siddharthvp:
Actually, there is a deeper issue here. The timestamp shown is per the time zone in the user preferences - which need not be the same as the system time zone (which JS Date() would consider it to be in). This is a problem because I'd imagine there'd be a lot of people who use the default UTC timezone on WP rather than the local zone (such as me until recently). 

If you really want to do this, you'll have to parse the user preference time zone (see function `getUserTimeZone()` in [T-Watch](https://en.wikipedia.org/wiki/User:SD0001/T-Watch.js)) , get the system time zone (`Date.prototype.getTimezoneOffset`) and then do some work to get the browser to understand the difference. The following will work (similar to what I did for [W-Ping](https://en.wikipedia.org/wiki/User:SD0001/W-Ping.js)):
```js
	var userzone = getUserTimeZone();
	var syszone = -new Date().getTimezoneOffset();
	var dateobj= new Date(timestamp);
        dateobj.setMinutes(dateobj.getMinutes() + syszone - userzone); 
       // now dateobj has the timestamp with it considered to be in the user preference time zone
```

Amorymeltzer:
Ugh yeah.  I may just say scrap the talk links for now — percentage-wise, I bet they are more likely to lead to such notifications but are a comparatively minuscule in usage.  If we really want this, I think it'd be better to just do the `rvprop=timestamp&revids=...` query on the diffIds.  I may still see how the devs feel about exposing that though in `wgDiffOldTimestamp` or something.

modules/twinklewarn.js
Amorymeltzer: Quite a few of these throughout...

Amorymeltzer: Think I got 'em all?  Eyes go blurry looking over `latest_date` and `history[current[1]]`

Amorymeltzer:
@siddharthvp I threw together a quick API hit, think it's worth it?

modules/twinklewarn.js
Amorymeltzer: Continued below

siddharthvp:
Don't have time for a detailed review atm but a couple of quick points: 

- Is there a reason why the api call is being put here? Why not just pass the revision id to the url and when the clicked upon - when we reach the usertalk page, do the api query to get the timestamp (of the revid in url) and show then show `alert` (if needed)? This should avoid any unnecessary api requests. 

- At any rate, don't use `async: false`. If/when the result becomes available, attach it at the end of the url.  

- actually it isn't _that_ difficult to get the user time zone and read the date string as if it's in that timezone. Thinking about, the user time zone is just `parseInt(mw.user.options.get('timecorrection').split('|')[0])`. And we can just convert this to string form ("UTC+0530") and append it at the end of the string given to the `Date` constructor.

Amorymeltzer:
Yeah, that's a ***much*** better idea @siddharthvp (although I still think these should be available in the interface it's free[cn]).  Merge conflicts aside, how's this second commit look?

siddharthvp:
Why is this taking place in `main`? That seems too late. Since the revid is there in the URL all along, I think the check should take place in `callback` itself. 

Note that in #765, I have removed all uses of "async: false" in TW. These are bad for performance, and don't work alongside promises (see [documentation](https://api.jquery.com/jquery.ajax/)). Adding a synchronous call here would be a step backwards. 

Doing this in `callback` does mean that theoretically a very fast user may submit the form before API call finishes. If this issue needs fixing, once #765 is merged, we could easily do something fancy using promises. (That is, make sure the promise has settled before `evaluate` is called, and don't call it if user doesn't ok the `confirm`.)

Amorymeltzer:
> Why is this taking place in `main`? That seems too late. Since the revid is there in the URL all along, I think the check should take place in `callback` itself.

That makes sense.  I had thought putting it at the end would give the user more time to get past the 24h window, but I suppose they'd just as like be annoyed to hit it after selecting a template, and might be more encouraged to give out a warning.

> Doing this in `callback` does mean that theoretically a very fast user may submit the form before API call finishes.

Sorry, I'm not really sure what you're getting at here.  If it's to be inside `callback`, I'd consider it inappropriate for the module's window to open before and unless receiving the confirmation.  This confirmation *should* hold things up.

siddharthvp:
> I'd consider it inappropriate for the module's window to open before and unless receiving the confirmation. This confirmation should hold things up.

Hmm. Maybe our ideas on UX differ. So you're saying that after the user has clicked on "Warn" option, the API call is to be sent and the window is not to open until the API returns, maybe with a "Loading..." dialog (like in the batch modules) being shown until then?  

It would be much simpler in code to just let the window load as usual. And then when the API returns and if user cancels, to do a `Window.close()` to close the warn dialog.

Amorymeltzer:
Oh I think that'd be awful.  Why open the window if they can't even interact with it?  I don't think there's a need for a "Loading...", just the `confirm`; I'm trying not to have the warn window open at all if inappropriate (sort of like a slow "You don't want to warn yourself").

siddharthvp:
My concern is the time delay between clicking on the menu button and some window (or `confirm`) appearing on screen. Ideally there should not be any delay. Not a huge problem if the API query is fast enough - I haven't checked. The ones for the batch modules definitely aren't. 

We could consider doing what talkback does. No confirm popup, but red text that gets added in the dialog when the API response arrives.

Amorymeltzer:
It's a pretty fast query - in theory noticeable, but in practice not really.

That's a good point on the red text, I think I like that idea better.  Less aggressive but also allows the user to change their mind.

Amorymeltzer:
Squashed and updated with the above consensus, pretty straightforward.

modules/twinklewarn.js
```diff
@@ -110,6 +116,34 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 		tooltip: 'A page can be linked within the notice, perhaps because it was a revert to said page that dispatched this notice. Leave empty for no page to be linked.'
 	});
 
+	form.append({
+		type: 'div',
+		label: '',
+		style: 'color: red',
```
siddharthvp: I have essentially zero experience with anti-vandalism so I wouldn't know, but is it really  wrong to issue warnings for edits that are more than 24 hours old? If the answer if anything other than an obvious yes, I'd rather the color be toned down to something less severe.

Amorymeltzer: It's actually comparatively soft to other tools: Huggle pops up an alert requiring confirmation (IIRC it also has a user preference to always skip issuing the warning entirely), and STiki has a popup as well.  As a sysop, I wouldn't say it's always wrong, but it should definitely be treated differently, so I think the red is appropriately eye catching.  The message is fairly tame.

PR #759, opened 2019-11-25
warn: Add option to show the single-notice messages in a single menu

### LINKED ISSUE #673: Warn: Combining single-issue notice and warn
Given that these are now searchable, the split between single-issue notices and single-issue warnings is not needed as much, and makes it harder to find less-used templates. I suggest that the two tabs be combined, either as a result of the user's preferences or by default. I'd like to work on this, but will hold off for now while waiting for feedback.

Amorymeltzer:
Uses a new config option, `combinedSingletMenus`, turned off by default.  The combined menu is sorted alphabetically.

Closes #673.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -90,8 +90,12 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 	main_group.append({ type: 'option', label: '3: Warning', value: 'level3', selected: defaultGroup === 3 });
 	main_group.append({ type: 'option', label: '4: Final warning', value: 'level4', selected: defaultGroup === 4 });
 	main_group.append({ type: 'option', label: '4im: Only warning', value: 'level4im', selected: defaultGroup === 5 });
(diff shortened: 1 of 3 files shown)


modules/twinklewarn.js
```diff
@@ -1139,6 +1143,13 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	if (value === 'singlenotice' || value === 'singlewarn') {
 		// no categories, just create the options right away
 		createEntries(Twinkle.warn.messages[value], sub_group, true);
+	} else if (value === 'singlecombined') {
+		var unSortedSinglets = $.extend({}, Twinkle.warn.messages.singlenotice, Twinkle.warn.messages.singlewarn);
```
Amorymeltzer: >should probably be written as Twinkle.warn.messages.singlenotice.

@siddharthvp Pinging you here rather than at https://github.com/azatoth/twinkle/pull/773#discussion_r352202964 since I'm resolving it here, but ✅ .

In fact, I just redid this section.  I initially waffled on whether this should be sorted, and only started figuring out sorting (and dealing with `messageData`) after `createEntries`, but just never went back and combined them, which should read clearer.

PR #762, opened 2019-11-25
Unify twinkleconfig and friendlyconfig

siddharthvp:
- Unify the Twinkle.defaultConfig.twinkle and Twinkle.defaultConfig.friendly to Twinkle.defaultConfig.
- Twinkle.getPref() can now retrieve friendly preferences also. Twinkle.getFriendlyPref() is removed.
- Now onwards, changes to WP:TWPREFS will save the preferences in twinkleoptions.js without separating them between twinkle and friendly. But the old format will still be acceptable.

A single config object is easier to work with, and would be helpful as far as i18n/l10n is concerned. This also simplifies a lot of code in twinkleconfig.

### DIFF
--- a/twinkle.js
+++ b/twinkle.js
@@ -176,12 +176,10 @@ if (mw.config.get('skin') === 'vector') {
 
 
 Twinkle.getPref = function twinkleGetPref(name) {
-	if (typeof Twinkle.prefs === 'object') {
-		if (Twinkle.prefs[name]) {
-			return Twinkle.prefs[name];
-		}
+	if (typeof Twinkle.prefs === 'object' && Twinkle.prefs[name]) {
+		return Twinkle.prefs[name];
 	}
-	// Old preferences format:
+	// Old preferences format, used before twinkleoptions.js was a thing
 	if (typeof window.TwinkleConfig === 'object' && window.TwinkleConfig[name]) {
 		return window.TwinkleConfig[name];
 	}
@@ -383,7 +381,7 @@ $.ajax({
 		try {
 			var options = JSON.parse(optionsText);
 			if (options) {
-				if (options.twinkle || options.friendly) {
+				if (options.twinkle || options.friendly) { // Old preferences format
 					Twinkle.prefs = $.extend(options.twinkle, options.friendly);
 				} else {
 					Twinkle.prefs = options;
--- a/twinkle.js
+++ b/twinkle.js
@@ -125,7 +125,7 @@ Twinkle.defaultConfig = {
 
 	// Formerly defaultConfig.friendly:
 
-  // Tag
+	// Tag
 	groupByDefault: true,
 	watchTaggedPages: true,
 	watchMergeDiscussions: true,
@@ -158,7 +158,7 @@ Twinkle.defaultConfig = {
 	// Shared
 	markSharedIPAsMinor: true
 };
-  
+
 // now some skin dependent config.
 switch (mw.config.get('skin')) {
 	case 'vector':
--- a/twinkle.js
+++ b/twinkle.js
@@ -20,7 +20,7 @@
 
 /* global Morebits */
 
-(function (window, document, $, undefined) { // Wrap with anonymous function
+(function (window, document, $) { // Wrap with anonymous function
 
 var Twinkle = {};
 window.Twinkle = Twinkle;  // allow global access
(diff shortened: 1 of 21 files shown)


Amorymeltzer:
Long overdue IMO.

twinkle.js
```diff
@@ -20,7 +20,7 @@
 
 /* global Morebits */
 
-(function (window, document, $, undefined) { // Wrap with anonymous function
```
siddharthvp: Does anyone know why on earth this was here? Since my changes removed the only two mentions of undefined in the file, this here was being flagged by the linter as an unused variable(!).

Amorymeltzer: See https://stackoverflow.com/a/15777832/2521092, and 2c64199f87fff2ba04970488542f199415e6b2e6 and 6d83b686496578d5fdc0beb0b2551a9d3986d8a7 for the reasoning for adding them to Twinkle.

I looked into the various anonymous wrapper functions when I was doing #500; I don't remember everything I learned, but it basically boils down to the fact that pre-ES5, `undefined` wasn't read-only.  It was good practice to define it to prevent bad people from being bad.

See also the eslint rules [no-undefined](https://eslint.org/docs/rules/no-undefined), [no-global-assign](https://eslint.org/docs/rules/no-global-assign), and [no-shadow-restricted-names](https://eslint.org/docs/rules/no-shadow-restricted-names).

I think I've got a todo somewhere to try to remove/tidy the `undefined` uses up, but it's buried beneath 87 others...

twinkle.js
```diff
@@ -174,40 +159,39 @@ Twinkle.defaultConfig.friendly = {
...
-Twinkle.getFriendlyPref = function twinkleGetFriendlyPref(name) {
-	var result;
-	if (typeof Twinkle.prefs === 'object' && typeof Twinkle.prefs.friendly === 'object') {
-		// look in Twinkle.prefs (twinkleoptions.js)
-		result = Twinkle.prefs.friendly[name];
-	} else if (typeof window.FriendlyConfig === 'object') {
-		// look in FriendlyConfig
-		result = window.FriendlyConfig[name];
+Twinkle.getPref = function twinkleGetPref(name) {
+	if (typeof Twinkle.prefs === 'object') {
```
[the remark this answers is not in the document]
siddharthvp: Hmm. This must be a holdover from a different structure of `Twinkle.getPref` I had earlier.

twinkle.js
```diff
@@ -174,40 +159,39 @@ Twinkle.defaultConfig.friendly = {
...
+		if (Twinkle.prefs[name]) {
+			return Twinkle.prefs[name];
+		}
 	}
-
-	if (result === undefined) {
-		return Twinkle.defaultConfig.friendly[name];
+	// Old preferences format:
```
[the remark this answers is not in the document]
siddharthvp: done.

twinkle.js
siddharthvp: Thanks for that. So it seems removal of `undefined` here is ok.

siddharthvp:
> * `twinkleconfig` (how very meta)

lol, didn't expect that there. config directive having config directives?

siddharthvp:
@Amorymeltzer Do you think we should add `optionsVersion: 2` to the object. This was recommended [by the comment in twinkle.js](https://github.com/azatoth/twinkle/pull/762/files#diff-551c5d596f19b4df338bbf4ab50919baL402-L412) (removed lines 402 - 412). This isn't necessary for now (since we are able to tell apart the old version by checking if `options.twinkle` exists). But I guess if our options structure were to change further in the future (such as loading the options from a json page, or from `mw.user.options`), I guess it would be helpful.

Amorymeltzer:
Yeah I pondered that — I vaguely recall referencing those comments somewhere a while ago.  I think what I arrived upon is that the current code is fairly sufficient at handling the older systems, and that putting in a version wouldn't change much, basically just shifting the explanations from code to comments.  I'd sooner keep version whatever for something where there's a major restructuring like you suggest (if it were made today, Twinkle'd probably use .json) but it doesn't seem necessary at the moment.

Still, doing away with the friendly object is significant, I suppose, in philosophy and history if not code, so maybe we call this version 1?  Seems appropriate for a unification.

tl;dr: Open but don't think it's strictly necessary.

siddharthvp:
Thrice approved, eh?🥰

Amorymeltzer:
You know that old saying: "Measure once, then create a bunch of merge conflicts and a couple of linting warnings so you have to cut twice"

PR #765, opened 2019-11-25
morebits.wiki.api: enable promise support

siddharthvp:
Morebits.wiki.api.prototype.post() now returns a jQuery promise object, on which done/then/fail/catch/always handlers can be attached, as an alternative to giving the success and failure callbacks as arguments in the constructor.

On the output of $.ajax(), `done` and `fail` are changed to `then`. This makes it possible to specify exactly what arguments (and `this` context) should be available to the handlers. We specify these to match the ones used in the callbacks, (`apiobj.parent` as `this` context, and the `apiobj` as the first and only argument). This makes it easy to transition existing code using callbacks to use promises.

Tested, confirmed working!

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1497,6 +1497,11 @@ Twinkle.speedy.callbacks = {
 				editsummary = 'Requesting speedy deletion ([[WP:CSD#' + params.normalizeds[0].toUpperCase() + '|CSD ' + params.normalizeds[0].toUpperCase() + ']]).';
 			}
 
+			// Set the correct value for |ts= parameter in {{db-g13}}
+			if (params.normalizeds.indexOf('g13') !== -1) {
+				code = code.replace('$TIMESTAMP', pageobj.getLastEditTime());
+			}
+
 			pageobj.setPageText(code + (params.normalizeds.indexOf('g10') !== -1 ? '' : '\n' + text)); // cause attack pages to be blanked
 			pageobj.setEditSummary(editsummary + Twinkle.getPref('summaryAd'));
 			pageobj.setWatchlist(params.watch);
@@ -1789,24 +1794,7 @@ Twinkle.speedy.getParameters = function twinklespeedyGetParameters(form, values)
 				break;
 
 			case 'afc':  // G13
-				var query = {
-						action: 'query',
-						titles: mw.config.get('wgPageName'),
-						prop: 'revisions',
-						rvprop: 'timestamp'
-					},
-					api = new Morebits.wiki.api('Grabbing the last revision timestamp', query, function(apiobj) {
-						var xmlDoc = apiobj.getXML(),
-							isoDateString = $(xmlDoc).find('rev').attr('timestamp');
-
-						currentParams.ts = isoDateString;
-					});
-
-				// Wait for API call to finish
-				api.post({
-					async: false
-				});
-
+				currentParams.ts = '$TIMESTAMP'; // to be replaced by the last revision timestamp when page is saved
 				break;
 
 			case 'redundantimage':  // F1
--- a/morebits.js
+++ b/morebits.js
@@ -2152,6 +2152,11 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		return ctx.revertUser;
 	};
 
+	/** @returns {string} ISO 8601 timestamp at which the page was last edited. */
+	this.getLastEditTime = function() {
+		return ctx.lastEditTime;
+	};
+
 	// Miscellaneous getters/setters:
 
 	/**
--- a/morebits.js
+++ b/morebits.js
@@ -1386,6 +1386,7 @@ Morebits.wiki.api.prototype = {
 	 * Carries out the request.
 	 * @param {Object} callerAjaxParameters Do not specify a parameter unless you really
 	 * really want to give jQuery some extra parameters
+	 * @returns {promise} - a jQuery promise object that is resolved or rejected with the api object.
 	 */
 	post: function(callerAjaxParameters) {
 
@@ -1402,8 +1403,9 @@ Morebits.wiki.api.prototype = {
 			}
 		}, callerAjaxParameters);
 
-		return $.ajax(ajaxparams).done(
-			function(response, statusText) {
+		return $.ajax(ajaxparams).then(
+
+			function onAPIsuccess(response, statusText) {
 				this.statusText = statusText;
 				this.response = this.responseXML = response;
 				if (this.query.format === 'json') {
@@ -1415,15 +1417,12 @@ Morebits.wiki.api.prototype = {
 				}
 
 				if (typeof this.errorCode === 'string') {
-
 					// the API didn't like what we told it, e.g., bad edit token or an error creating a page
-					this.returnError();
-					return;
+					return this.returnError();
 				}
 
 				// invoke success callback if one was supplied
 				if (this.onSuccess) {
-
 					// set the callback context to this.parent for new code and supply the API object
 					// as the first argument to the callback (for legacy code)
 					this.onSuccess.call(this.parent, this);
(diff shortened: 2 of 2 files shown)


siddharthvp:
New commit:  Added function `setStatusElement()` for setting an existing status element without using the constructor. While using the .then() and .catch() instead of the callbacks, this helps avoid having to put the third argument (`onSuccess`) as null or undefined, just so that the fourth can be used.

This "cleans up" the constructor usage. Going forward, new code using Morebits.wiki.api should be something like:

```js
var api = Morebits.wiki.api('currentaction', query);
api.setStatusElement(st); // mostly unnecessary
var action = api.post().then(doSomething).catch(doSomething);

action.then(doSomethingElse); // second handler (could be anywhere else in the code)
// wasn't possible earlier when using callbacks.

var api2 = new Morebits.wiki.api('currentaction', query);
var action2 = api2.post();

$.when(action, action2).then(function() {
    // do something when both api calls (sent parallelly) have finished, 
    // again this sort of thing wasn't possible earlier when using callbacks
});
```

Next up: make `Morebits.wiki.page` functions return promises as well.

siddharthvp:
I have reworked this PR by a bit. Actually it turns out that `async: false` API calls can't be made to work while we're returning chainable promises from `$.ajax` - the `.done()`  used to work with `async: false`, but `.then()` doesn't, and we need to use `.then()` to make the returning promise of `Morebits.wiki.api.prototype.post()` resolve with the `Morebits.wiki.api` object. 

I think the right way to resolve the problem is by simply getting rid of all `async: false` API actions, which I have done now.  It is also [not being used by any userspace script that uses Morebits](https://en.wikipedia.org/w/index.php?sort=relevance&search=%22async%3A+false%22+Morebits+contentmodel%3Ajavascript&title=Special%3ASearch&advancedSearch-current=%7B%7D&ns2=1). There were three usages in Twinkle: two right within morebits - in purging the page after an edit conflict and for getting a new edittoken after a token error. These were quite easy to replace with normal API calls. The other was in speedy, for getting the last page edit timestamp for db-g13 tag, [which has been removed](https://github.com/azatoth/twinkle/pull/765/commits/298cf19c3c4afb32b8bf5fb942e249a6ca826d05), which has been cleverly dealt with :)

Changes have been tested, and confirmed working. I have not been able to test the purge after editconflict feature, though. I can't seem to trigger the `editconflict` error. Pausing the execution after morebits.wiki.page#load, and editing the page from another tab in the meantime, still results in the page getting saved smoothly without an editconflict error (!). But it is pretty clear that the new code does exactly what the old code did. The `fnGetToken` change can be tested simply by commenting out the line in `this.save` where the token parameter is set, triggering a `notoken` error. 

Finally, the following code can be used to test promise callbacks are behaving as expected: 

```js
/* disable internet from the network tab to trigger the other type of error */
var apiobj = new Morebits.wiki.api(
	'',
	{
		action: 'query', /* change to something random to make the API return an error */
		titles: 'Main Page'
	},
	function onSuccess() { console.log('onSuccess', this, arguments); },
	null,
	function onFailure() { console.log('onFailure', this, arguments); }
).post();

apiobj.then(function() { console.log('then', this, arguments); });
apiobj.done(function() { console.log('done', this, arguments); });
apiobj.fail(function() { console.log('fail', this, arguments); });
apiobj.catch(function() { console.log('catch', this, arguments); });
apiobj.always(function() { console.log('always', this, arguments); });
```

modules/twinklespeedy.js
```diff
@@ -1497,6 +1497,11 @@ Twinkle.speedy.callbacks = {
 				editsummary = 'Requesting speedy deletion ([[WP:CSD#' + params.normalizeds[0].toUpperCase() + '|CSD ' + params.normalizeds[0].toUpperCase() + ']]).';
 			}
 
+			// Set the correct value for |ts= parameter in {{db-g13}}
+			if (params.normalizeds.indexOf('g13') === -1) {
```
DannyS712: Shouldn't this be checking if if `!== -1`?

morebits.js
```diff
@@ -1374,18 +1376,32 @@ Morebits.wiki.api.prototype = {
...
-	},  // keep track of parent object for callbacks
 	statelem: null,  // this non-standard name kept for backwards compatibility
 	statusText: null, // result received from the API, normally "success" or "error"
 	errorCode: null, // short text error code, if any, as documented in the MediaWiki API
 	errorText: null, // full error description, if any
 
+	/**
+	 * Keep track of parent object for callbacks
+	 * @param {*} parent
+     */
```
DannyS712: Inline with #780, shouldn't this be on one line? (same applies to `setStatusElement` below)

modules/twinklespeedy.js
siddharthvp: Indeed! Thanks. No idea how my [test](https://en.wikipedia.org/w/index.php?title=Draft:Sandbox&oldid=929730959&diff=prev&diffmode=source) worked despite this mistake. Probably it was using the old morebits :( I really need to come up with a fool-proof way of ensuring the updated morebits has loaded before tests are run.

morebits.js
siddharthvp: Done for the latter. Here there are two lines - one for description and one for param. While I guess putting them on the same line is still valid JSDoc, it looks clumsy.

morebits.js
```diff
@@ -2152,6 +2152,11 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		return ctx.revertUser;
 	};
 
+	/** @returns {string} ISO 8601 timestamp at which the page was last edited. */
+	this.getLastEditTime = function() {
```
Amorymeltzer: Weird that we've gone so long without something like this

Amorymeltzer: Still weird that it hasn't been an issue yet!

modules/twinklespeedy.js
```diff
@@ -1789,24 +1794,7 @@ Twinkle.speedy.getParameters = function twinklespeedyGetParameters(form, values)
...
-
-						currentParams.ts = isoDateString;
-					});
-
-				// Wait for API call to finish
-				api.post({
-					async: false
-				});
-
+				currentParams.ts = '$TIMESTAMP'; // to be replaced by the last revision timestamp when page is saved
```
Amorymeltzer, line 1797: Outstanding

morebits.js
```diff
@@ -1340,11 +1340,14 @@ Morebits.wiki.removeCheckpoint = function() {
  */
 
 /**
+ * In new code, the use of the last 3 parameters should be avoided, instead use setStatusElement() to bind the
```
Amorymeltzer, line 1343: This seems innocuous, but I don't really get why it helps?

morebits.js
```diff
@@ -2709,25 +2702,27 @@ Morebits.wiki.page = function(pageName, currentAction) {
 				titles: ctx.pageName  // redirects are already resolved
 			};
 
-			var purgeApi = new Morebits.wiki.api('Edit conflict detected, purging server cache', purgeQuery, null, ctx.statusElement);
-			purgeApi.post({ async: false });  // just wait for it, result is for debugging
-
-			--Morebits.wiki.numberOfActionsLeft;  // allow for normal completion if retry succeeds
+			var purgeApi = new Morebits.wiki.api('Edit conflict detected, purging server cache', purgeQuery, function() {
```
Amorymeltzer, line 2710: Nice catch!

morebits.js
```diff
@@ -2644,22 +2644,15 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
-		}, function(apiobj) {
-			token = $(apiobj.responseXML).find('tokens').attr('csrftoken');
-		}, null, function() {
-			this.getStatusElement().error('Failed to get token');
 		});
-		tokenApi.post({async: false});
-		return token;
+		return tokenApi.post().then(function(apiobj) {
+			return $(apiobj.responseXML).find('tokens').attr('csrftoken');
+		});
```
Amorymeltzer, line 2660: Don't we still want to catch the error?

Amorymeltzer:
🐢 🎩 🎉 

In all serious, I know we're all distracted/busy, but thanks for the persistence on this sort of thing @siddharthvp.  Really do appreciate it.

Issue #769, opened 2019-11-28
Should we remove moment dependency? (arv)


Amorymeltzer:
Added in a2c3f93a / 0beacc07 and originally loaded as a Twinkle module, it has since #249 been a simple dependency via the gadget definition.

Currently, it is used *only* for a single module, `twinklearv.js`, and even then only for single segment: reporting to WP:AN3.  In the past month, WP:AN3 has averaged about 5 reports every 3 days (i.e. there have been just under 50 reports in 31 days).  The formatting stuff is replaceable, so only the relative language from .` calendar` ("Today at 7PM") would be lost.  The specific display (see #770 for proposed implementation) isn't particularly important here, but the upshot is that momentjs is (1) barely used and (2) definitely not necessary.

I'm not sure how caches are treated on phones or tablets, but on my laptop it looks pretty generous.  Still, moment is somewhere around 8% of the full-sized Twinkle package (per inspect), and just under 10% as actually loaded (that's 86.8 and 23.4 kb, respectively).  It leads to fewer than two edits a day, so it's fairly unnecessary right now and just adds some weight and complexity.  There's probably a future for moment here as regards i18n, but it isn't now.

Amorymeltzer:
I first mentioned this in https://github.com/azatoth/twinkle/issues/723#issuecomment-547864351 where @siddharthavp said:
>Its a cheap library loaded efficiently via the resourceloader. And it makes date handling a bliss, as well as less error-prone, and is free of browser/es-version quirks potentially associated with JS Date. We could even think of replacing all instances of JS Date and use moment exclusively.

(also cc @MusikAnimal,  who commented there)

Another option would be to use the CDN (à la #692) for `dayjs`, which is smaller, but that's probably not worth it (no resourceloader).  There has been some [discussion of removing moment](https://phabricator.wikimedia.org/T146798) in core in favor of something else like dayjs.

siddharthvp:
> Another option would be to use the CDN (à la #692) for dayjs, which is smaller, but that's probably not worth it (no resourceloader). 

Or how about just writing a lightweight date library ourselves, customised for our needs? We could put it in Morebits, so that it can use the ResourceLoader.

MusikAnimal:
I was under the impression that dependencies aren't "downloaded" more than once, no matter how many gadgets use them. As you say, MediaWiki Core uses moment, so if my belief is true, then we shouldn't be adding much weight in using it ourselves. Or maybe I've had this wrong all along?

[...]

PR #770, opened 2019-11-28
arv: Remove moment dependency

Amorymeltzer:
See #769 for the rationale, but tl;dr the only uses of [momentjs](https://momentjs.com/) are in `twinklearv` and are cosmetic, so loading moment is wasteful.  There are two functions we need to replace, `.format()` and `.calendar()`; this uses `.toLocaleString()` and `.toLocaleDateString()`, respectively.

Regarding formatting, doing:

```javascript
var localeOptions = {
    hour: 'numeric',
    hour12: false,
    minute: 'numeric',
    day: 'numeric',
    month: 'long',
    year: 'numeric',
    timeZone: 'UTC',
    timeZoneName: 'short'
};
new Date().toLocaleString(undefined, localeOptions);
```

produces:
>November 25, 2019, 21:36 UTC

This doesn't fully recpitulate `moment().utc().format('HH:mm, D MMMM YYYY [(UTC)]')`, which produces:
>21:36, 25 November 2019 (UTC)

The difference is negligible, and although it was chosen (0beacc07) to match on-wiki formats, that's a user preference and, notably, *not* a valid date format (see #757).  A more exact reproduction can be achieved by combining some strings, doing something like:

```javascript
var localeTimeOptions = {
    timeStyle: 'short',
    hour12: false,
    timeZone: 'UTC'
};
var d = new Date();
d.toLocaleTimeString(undefined, localeTimeOptions) + ', ' + d.toLocaleDateString(undefined, {day: 'numeric', timeZone: 'UTC'}) + ' ' + new Date().toLocaleDateString(undefined, {month: 'long', year: 'numeric', timeZone: 'UTC'}) + ' (UTC)';
```

We could place this in a function, but hardly seems necessary, especially since, depending on their selected user preference, not all users are used to the same formats.

As for the relative dates from the `.calendar()` function, they simply can't be easily replaced.  This implementation uses `.toLocaleDateString()` to produce `Fri, Nov 22, 2019, 2:03 PM`, whereas `.calendar()` would have produced `Last Friday at 2:03 PM`.  The short month and weekday names are used to declutter the window as compared to a long string.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -16,7 +16,7 @@ There are two ways to upload Twinkle scripts to Wikipedia or another destination
 
 After the files are synced, ensure that [MediaWiki:Gadgets-definition][] contains the following lines:
 
-    * Twinkle[ResourceLoader|dependencies=mediawiki.notify,jquery.chosen,moment,ext.gadget.morebits|rights=autoconfirmed|type=general|peers=Twinkle-pagestyles]|Twinkle.js|twinkleprod.js|twinkleimage.js|twinklebatchundelete.js|twinklewarn.js|twinklespeedy.js|friendlyshared.js|twinklediff.js|twinkleunlink.js|friendlytag.js|twinkledeprod.js|friendlywelcome.js|twinklexfd.js|twinklebatchdelete.js|twinklebatchprotect.js|twinkleconfig.js|twinklefluff.js|twinkleprotect.js|twinklearv.js|twinkleblock.js|friendlytalkback.js|Twinkle.css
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
Opened as draft pending discussion in #769, and for feedback on display.

Also of note, I've left the various `toLocale...` checks using `undefined` as the `locale`, which defaults to the browser default.  We *could* use something like `mw.config.get('wgUserLanguage')`, although of course there's also `wgPageContentLanguage` and `wgContentLanguage`, and `language` and `variant` from `mw.user.options`, as well as some stuff in `wgVisualEditor` and `wgULSAcceptLanguageList`

siddharthvp:
Regarding formatting, please don't!  

Just use a function to create a wiki-styled timestamp:

```js
/* @param {Date} d */
var createTimestamp = function(d) { 
    var h = d.getUTCHours(), m = d.getUTCMinutes(), D = d.getUTCDate(), 
        M = d.getUTCMonthName(), Y = d.getUTCFullYear();
    return (h < 10 ? '0' + h : h) + ':' + (m < 10 ? '0' + m : m) + ', ' + 
        D + ' ' + M + ' ' + Y + ' (UTC)'; 
}
```

This is fewer lines in code, readable, guarantees cross-browser compatibility, and is readily internationalisable. No matter how many mw configs exists that give the language code, you simply can't trust `toLocaleString` to give the date in the given language. 

----

The relative dates are only used in the display, so I'm not bothered whatever way they turn out - the use of `toLocaleDateString()` with the system-default locale seems ok. Putting in some language code is superficial anyway since ARV is enwiki-specific, no one would be using this elsewhere.

Amorymeltzer:
Closed in favor of #814

PR #773, opened 2019-11-29
warn: Add option to show a list with all template warnings

Amorymeltzer:
Mentioned in https://github.com/azatoth/twinkle/issues/673#issuecomment-501858213.  While this makes use of 10 for `defaultWarningGroup`, I considered reusing 8, which was until the addition of the block module in #260 (well, also 430e2258) reserved for sysops to select blocking templates, to cut down on bloat.  There are a handful of users, some still active sysops, with the option set, but while that's a trivial barrier, I think it's better list it below the individual items, reinforcing the "all-of-the-above"-ness of the option.

This also tweaks the process by which `twinklewarn` tries to select the template that corresponds to what was selected in the previous menu.  As it has every item, the new kitchensink menu should always match exactly, but since the process was previously only relevant for levels 1-4im, the highest level would always be selected upon moving to the kitchensink menu (i.e., vandalism1 -> vandalism4im).  Now, in addition to doing an exact match in the kitchensink, it also won't keep searching once it has found the match.  Part of this is from #592.

Also removed a comment made incorrect by #641.

----
This should wait on #759, and in case folks want to discuss using 10 instead of 8, although after mulling it for a while, I really do think incrementing is better (especially when considering other projects).  Dunno if you have an opinion @MusikAnimal but ping given your involvement with the former option 8.

One other question is whether this menu should be sorted or not.  Unlike the combined singlet menu from #759, I don't think it's helpful.

This is rebased on the branch for #759, but may indeed need some conflict handling after that is processed.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -98,6 +98,7 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 	if (Twinkle.getPref('customWarningList').length) {
 		main_group.append({ type: 'option', label: 'Custom warnings', value: 'custom', selected: defaultGroup === 9 });
 	}
+	main_group.append({ type: 'option', label: 'All warning templates', value: 'kitchensink', selected: defaultGroup === 10 });
 
 	main_select.append({ type: 'select', name: 'sub_group', event: Twinkle.warn.callback.change_subcategory }); // Will be empty to begin with.
 
@@ -1090,19 +1091,25 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 	var old_subvalue = sub_group.value;
 	var old_subvalue_re;
 	if (old_subvalue) {
-		old_subvalue = old_subvalue.replace(/\d*(im)?$/, '');
-		old_subvalue_re = new RegExp(mw.util.escapeRegExp(old_subvalue) + '(\\d*(?:im)?)$');
+		if (value === 'kitchensink') { // Exact match possible in kitchensink menu
+			old_subvalue_re = new RegExp(mw.util.escapeRegExp(old_subvalue));
+		} else {
+			old_subvalue = old_subvalue.replace(/\d*(im)?$/, '');
+			old_subvalue_re = new RegExp(mw.util.escapeRegExp(old_subvalue) + '(\\d*(?:im)?)$');
+		}
 	}
 
 	while (sub_group.hasChildNodes()) {
 		sub_group.removeChild(sub_group.firstChild);
 	}
 
+	var selected = false;
 	// worker function to create the combo box entries
-	var createEntries = function(contents, container, wrapInOptgroup) {
+	var createEntries = function(contents, container, wrapInOptgroup, val) {
+		val = typeof val !== 'undefined' ? val : value; // IE doesn't support default parameters
 		// level2->2, singlewarn->''; also used to distinguish the
 		// scaled levels from singlenotice, singlewarn, and custom
-		var level = value.replace(/^\D+/g, '');
+		var level = val.replace(/^\D+/g, '');
 		// due to an apparent iOS bug, we have to add an option-group to prevent truncation of text
 		// (search WT:TW archives for "Problem selecting warnings on an iPhone")
 		if (wrapInOptgroup && $.client.profile().platform === 'iphone') {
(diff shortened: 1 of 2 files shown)


modules/twinklewarn.js
```diff
@@ -1144,20 +1149,21 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
...
+		var selectedVal = selectGroup.val(); // Preserve selection after sorting
 		selectGroup.children('option').detach().sort(function(a, b) {
 			return a.text > b.text ? 1 : -1;
 		}).appendTo(selectGroup);
-		selectGroup[0].selectedIndex = 0; // Select the first item
+		selectGroup.val(selectedVal);
 	} else if (value === 'custom') {
 		createEntries(Twinkle.getPref('customWarningList'), sub_group, true);
-	} else {
+	} else if (value !== 'kitchensink') { // Levels 1-4im
```
siddharthvp: better to put the if (value === kitchensink) part first. Or use a switch instead of this if-else-if ladder.

modules/twinklewarn.js
```diff
@@ -1171,6 +1177,17 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 			// create the options
 			createEntries(groupContents, optgroup, false);
 		});
+	} else { // All of the above
+		['level1', 'level2', 'level3', 'level4', 'level4im'].forEach(function(lvl) {
+			value = lvl;
+			Object.keys(Twinkle.warn.messages.levels).forEach(function(groupLabel) {
+				createEntries(Twinkle.warn.messages.levels[groupLabel], sub_group, true);
+			});
```
siddharthvp: Odd style. 
```js
$.each(Twinkle.warn.messages.levels, function(_, group) {
    createEntries(group, sub_group, true);
});
```
  
?

modules/twinklewarn.js
```diff
@@ -1144,20 +1149,21 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 		// no categories, just create the options right away
 		createEntries(Twinkle.warn.messages[value], sub_group, true);
 	} else if (value === 'singlecombined') {
-		createEntries(Twinkle.warn.messages['singlenotice'], sub_group, true);
-		createEntries(Twinkle.warn.messages['singlewarn'], sub_group, true);
+		var singletMessages = $.extend({}, Twinkle.warn.messages['singlenotice'], Twinkle.warn.messages['singlewarn']);
```
siddharthvp: should probably be written as `Twinkle.warn.messages.singlenotice`. This bracket notation is not used anywhere in TW except where it's necessary.

modules/twinklewarn.js
```diff
@@ -1171,6 +1177,17 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 			// create the options
 			createEntries(groupContents, optgroup, false);
 		});
+	} else { // All of the above
+		['level1', 'level2', 'level3', 'level4', 'level4im'].forEach(function(lvl) {
+			value = lvl;
```
siddharthvp: This is not the way to do it. Tweak the `createEntries` function to take the level as an argument. Removes the need for the `value = '';` kludge below. Using arguments is better than using closures.

Amorymeltzer: Quite right, off course.  I think now that was my intent back in June, but did this to get something working quick and just totally skipped over that memory five months later.

modules/twinklewarn.js
Amorymeltzer: Dunno about odd, at least no more odd than js vs jQuery.  I had it that way at first but changed it since I didn't like the empty variable in `$.each`.  I sort of like the visual difference from the level 1-4 section, but do agree this isn't particularly Twinkley.

modules/twinklewarn.js
Amorymeltzer: Dealt with at https://github.com/azatoth/twinkle/pull/759/files#r352365578

modules/twinklewarn.js
Amorymeltzer: I think I just liked the "all of the above"-ness of it?  A switch does that just fine.

modules/twinklewarn.js
siddharthvp: Actually it's better to use `Object.values` here (since we're only using the value) like I did [back there](https://github.com/azatoth/twinkle/pull/747#discussion_r359567205) wehre you laughed at me :)

Where both key and value are needed, I think `$.each` is the best way. The next best way would be to use [Object.entries() along with array destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries#Iterating_through_an_Object) but bah, that's ES6.

modules/twinklewarn.js
```diff
@@ -1081,25 +1086,31 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
...
 
 	while (sub_group.hasChildNodes()) {
 		sub_group.removeChild(sub_group.firstChild);
 	}
 
+	var selected = false;
 	// worker function to create the combo box entries
-	var createEntries = function(contents, container, wrapInOptgroup) {
+	var createEntries = function(contents, container, wrapInOptgroup, val) {
+		val = typeof val !== 'undefined' ? val : value; // IE doesn't support default parameters
```
siddharthvp: I like to use `val = val || value`. Or just pass in `value` wherever the function is being used. But I'm probably just nitpicking :)

modules/twinklewarn.js
```diff
@@ -1081,25 +1086,31 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 		$('select[name=sub_group]').chosen('destroy');
 	}
 
-	var value = e.target.value;
+	var value = e.target.value; // Gets destroyed in the kitchensink loop
```
siddharthvp:
```suggestion
	var value = e.target.value;
```
no longer happening.

modules/twinklewarn.js
```diff
@@ -1114,47 +1125,74 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 
 		$.each(contents, function(itemKey, itemProperties) {
 			// Skip if the current template doesn't have a version for the current level
-			if (!!level && !itemProperties[value]) {
+			if (!!level && !itemProperties[val]) {
 				return;
 			}
-			var key = typeof itemKey === 'string' ? itemKey : itemProperties.value;
+			var key = typeof itemKey === 'string' ? itemKey : itemProperties.val;
```
siddharthvp: Neither `itemProperties.val` nor `itemProperties.value` make sense since AFAICT `itemProperties` doesn't have a property named `val` or `value`. But it seems the code never reaches here as `typeof itemKey === 'string'` is always true.

modules/twinklewarn.js
Amorymeltzer: Changed.

>Where both key and value are needed, I think $.each is the best way

Bien sûr!

modules/twinklewarn.js
Amorymeltzer: Likewise, although here I did this to futureproof against other falsey values, as they could theoretically be made use of; I believe default parameters (when supported) are assigned `undefined` if not provided.  I'd sooner toss in this one line then add it everywhere just because one option needs it.

modules/twinklewarn.js
Amorymeltzer: This bit is for custom warning templates, which are brought in as something like:
```javascript
0: {
        "value": "Happy",
        "label": "Winter"
      },
1: ...
```
As such, they *do* have a different `typeof`, hence this line.  That being said, I apparently changed this overeagerly when previously refactoring above (`value` to `val`).  So... we were both wrong!  Fixed.

PR #777, opened 2019-12-02
warn: Add option to autoselect warning level 1-4

Amorymeltzer:
Edit April 2020: Very different, see https://github.com/azatoth/twinkle/pull/777#issuecomment-616898576

----

*Original message*

Full commit messages are below, but this will most certainly conflict with #759 and #773 so <s>opened as a draft</s> should wait on those; still, I wanted to discuss a few things here while testing/etc.:

* How long should elapse before a previous warning is considered stale and we reset the level rather than increment higher?
	* AFAICT, the default for Huggle is 30 days ([noted on enwiki](https://en.wikipedia.org/wiki/Wikipedia:Huggle/Feedback/Archive_21#Autowarn_and_stale_warnings), present in code at [1](https://github.com/huggle/huggle3-qt-lx/blob/13df16adf5744825f8a27d70b3bab998ad19f4b0/src/huggle_core/configuration.cpp#L187) and [2](https://github.com/huggle/huggle3-qt-lx/blob/13df16adf5744825f8a27d70b3bab998ad19f4b0/src/huggle_core/projectconfiguration.hpp#L172)) but on enWiki this is [currently set to 3 days](https://en.wikipedia.org/w/index.php?title=Wikipedia:Huggle/Config.yaml&diff=919417999&oldid=918243346). (I know I provided this to you @tobefree, but do check that I'm correct)
	* CBNG seems to [use 2 days](https://github.com/DamianZaremba/cluebotng/blob/4958e25d6874cba01c75f11debd2e511fd5a2ce5/bot/action_functions.php#L62).
* That default level is currently structured as a hidden ("advanced") preference, `autolevelStaleDays`, but should we make it open for user customization?  HG's is.
* <s>If they already have a level 4 warning, should saying **no** to the "do you really want to warn instead of report to AIV?" provide the user a link to the ARV module (current implementation) or just auto-open the ARV module for them?</s>
* <s>This uses the level 2 warnings for the previews; I wanted level 3 as that seemed an okay balance to ensure users know what the message could be, but there's one template without a level 3 warning (`{{uw-tempabuse3}]`).  That *does exist* as a redirect to `{{uw-disruptive3}}`, so if level 3 really is a better middle ground, we could add that redirect into the messages object.</s>
* <s>Is the display (no curlies, note that it's in reference to the series) a good idea?</s>

<s>I've left the hidden pref and the AIV link as separate commits in case that makes it easier to test/parse what I'm looking for feedback on.</s>  Maybe some of this should get discussed at WT:TW?  @musikanimal You do a lot of anti vandal work, so I wouldn't mind your input here.

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -682,7 +682,8 @@ Twinkle.config.sections = [
 					'7': 'Single-issue warnings',
 					// 8 was used for block templates before #260
 					'9': 'Custom warnings',
-					'10': 'All warning templates'
+					'10': 'All warning templates',
+					'11': 'Auto-select level (1-4)'
 				}
 			},
 
@@ -876,6 +877,13 @@ Twinkle.config.sections = [
 			// twinklefluff.js: defines how many revision to query maximum, maximum possible is 50, default is 50
 			{
 				name: 'revertMaxRevisions',
+				type: 'integer'
+			},
+			// twinklewarn.js: When using the autolevel select option, how many days makes a prior warning stale
+			// Huggle is three days ([[Special:Diff/918980316]] and [[Special:Diff/919417999]]) while ClueBotNG is two:
+			// https://github.com/DamianZaremba/cluebotng/blob/4958e25d6874cba01c75f11debd2e511fd5a2ce5/bot/action_functions.php#L62
+			{
+				name: 'autolevelStaleDays',
 				type: 'integer'
 			},
 			// twinklebatchdelete.js: How many pages should be processed maximum
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -46,17 +46,22 @@ Twinkle.warn = function twinklewarn() {
 	}
 };
 
+// Used to close window when switching to ARV in autolevel
+Twinkle.warn.dialog = null;
+
 Twinkle.warn.callback = function twinklewarnCallback() {
 	if (mw.config.get('wgRelevantUserName') === mw.config.get('wgUserName') &&
 		!confirm('You are about to warn yourself! Are you sure you want to proceed?')) {
 		return;
 	}
 
-	var Window = new Morebits.simpleWindow(600, 440);
-	Window.setTitle('Warn/notify user');
-	Window.setScriptName('Twinkle');
-	Window.addFooterLink('Choosing a warning level', 'WP:UWUL#Levels');
-	Window.addFooterLink('Twinkle help', 'WP:TW/DOC#warn');
+	var dialog;
+	Twinkle.warn.dialog = new Morebits.simpleWindow(600, 440);
+	dialog = Twinkle.warn.dialog;
+	dialog.setTitle('Warn/notify user');
+	dialog.setScriptName('Twinkle');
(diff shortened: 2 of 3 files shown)


siddharthvp:
Very nice idea! But the implementation could be made better:

- always using level2 for previews - ugh! Preview anyway needs an api call. So before doing that, just parse the page and make out the right template to show the preview of. This should be the behaviour people expect. To prevent repeated calls to the `revisions` api, save its output when first used (I guess things like this would be easier after #765).

- Use of `confirm` to ask the user whether they want to go to AIV is no good, as they can't interact with the page (which would be necessary to decide) without clicking ok or cancel first. It would be better to ask that using a button within `Morebits.status`. 

- The `'Will issue a {{' + warning + '}} template to the user, okay?'` thing would be mostly just annoying. It would be better if the user confirmation is determined without an annoying popup. How about determining the template as soon as the user changes options in the select menu? A small text of the form `Would use uw-xxxx3` could be put there. The page text should be fetched just once, regardless of how many times the user changes options. And once again on submit - if now the template to be used differs from what the user was expecting (because someone issued that level template in the meantime) then only seek a `confirm` via popup.

Amorymeltzer:
Totally forgot about this with everything else going on!  Still want to do it, although I think I'll wait until #814 since this will require a fair bit of refactoring around that.

The level2 hack is indeed awful.  Looping the logic into the preview call is fine idea, that should be doable now.  I think(?) your bullets one and three are somewhat at odds with one another?  I'll think about it; I'm inclined toward thinking that, as long as we're trying to correctly generate the preview, we might as well try to always do so, and account for the page to be updated in between subsequent previewing.

Amorymeltzer:
Woof.  Took a backseat, got around to it.  This has been largely rewritten; it could very well just be in a separate PR, but for the fact that I wanted to keep the comments here.

The new commit message summarizes it well enough I think, but this now:
- Show the actual level in the menu and in the preview
- Has a little note about what the level is
- Has a little note about when you should head to AIV
- Is a lot smarter than it was previously about templates, etc.

It's up as a separate commit for the ease of review/comparison if anyone cares, but I'll squash before merging.  cc @ToBeFree, @QEDK, and @siddharthvp since you all interacted previously.

siddharthvp:
Hi looks like stellar work. But on [User talk:SD0002](https://en.wikipedia.org/wiki/User_talk:SD0002) I'm getting "Will issue a level 1 template." even after having posted a uw-vandalism1 there.

Amorymeltzer:
@siddharthvp Can't check atm but did you load the preference, default or otherwise?  If you don't have `Twinkle.prefs.autolevelStaleDays` set it'll return `NaN`.

siddharthvp:
Ok, missed the change to twinkle.js.

Now, it says "Will issue a level 2 template" even when I am selecting a template other than uw-vandalism. Is this the intended behaviour?

ToBeFree:
Can't test, currently, but it makes me very happy to see this progress. Thank you very much for implementing this, and for the ping!

Amorymeltzer:
@siddharthvp Yup, that's right.  Figured repeating the template was redundant since it would have been manually chosen, but that it would be easy to miss a subtle change in number.

PR #779, opened 2019-12-06
XfD: Move RM technical request down

DannyS712:
Per https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=929502397#RM_requests, these should be added at the bottom of the section (directly above the next section header) rather than the top.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -1444,7 +1444,7 @@ Twinkle.xfd.callbacks = {
 			var params = pageobj.getCallbackParameters();
 			var statelem = pageobj.getStatusElement();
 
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
As noted by @siddharthvp Twinkle is currently following the instructions. I commented at [WT:TW](https://en.wikipedia.org/wiki/Special:Permalink/931043065#RM_requests), but unless there's a consensus to change from XfD-like to RfPP-like at [WT:RM](https://en.wikipedia.org/wiki/Wikipedia_talk:Requested_moves#New_technical_requests:_Place_at_the_top_or_bottom_of_the_list%3F), I don't think this should go forward.

Amorymeltzer:
Per comments at https://en.wikipedia.org/wiki/Wikipedia_talk:Requested_moves#New_technical_requests:_Place_at_the_top_or_bottom_of_the_list?

PR #782, opened 2019-12-09
morebits: batchOperation: minor improvements

siddharthvp:
1st commit: doc fixes.
2nd commit: execute `postFinish()` even when the `pageList` is empty. This situation could occur when the pageList used is variable.
3rd commit: allow individual status lines even if the `worker` function is not using `Morebits.wiki.page` or `Morebits.wiki.api`.

These will not have any effect on Twinkle.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -3968,35 +3968,50 @@ Morebits.batchOperation = function(currentAction) {
 		fnStartNewChunk();
 	};
 
-	this.workerSuccess = function(apiobj) {
-		// update or remove status line
-		if (apiobj && apiobj.getStatusElement) {
-			var statelem = apiobj.getStatusElement();
+	/**
+	 * To be called by worker before it terminates succesfully
+	 * @param {(Morebits.wiki.page|Morebits.wiki.api|string)} arg
+	 * This should be the `Morebits.wiki.page` or `Morebits.wiki.api` object used by worker
+	 * (for the adjustment of status lines emitted by them).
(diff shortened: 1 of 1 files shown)


morebits.js
```diff
@@ -3951,8 +3951,11 @@ Morebits.batchOperation = function(currentAction) {
 
 		var total = ctx.pageList.length;
 		if (!total) {
-			ctx.statusElement.info('nothing to do');
+			ctx.statusElement.info('no pages specified');
 			ctx.running = false;
+			if (ctx.postFinish) {
+				ctx.postFinish();
```
Amorymeltzer: I think this is better than modifying `fnDoneOne` or something, if only for simplicity's sake, but even `workerFailure` calls `fnDoneOne`.  It's a weird construct there overall.

morebits.js
```diff
@@ -3968,31 +3968,46 @@ Morebits.batchOperation = function(currentAction) {
...
 				statelem.unlink();
 			}
+
+		} else if (typeof arg === 'string' && ctx.options.preserveIndividualStatusLines) {
+			new Morebits.status(arg, ['done (', createPageLink(arg), ')']);
 		}
 
 		ctx.countFinishedSuccess++;
-		fnDoneOne(apiobj);
+		fnDoneOne();
```
Amorymeltzer: Why this even here in the first place?  Should presumably be likewise remove from `workerFailure` below.

siddharthvp: `fnDoneOne` doesn't take any argument, so yeah removed from workerFailure also.

PR #785, opened 2019-12-17
Morebits: Remove old outdated comment

DannyS712:
In fnProcessMove, remove the comment `XXX don't know whether this works for non-admins` after logic handling query.movesubpages - if the user cannot move subpages, this is enforced by the mediawiki api (https://gerrit.wikimedia.org/g/mediawiki/core/+/e398253aba0652afcf2a6eeb9fdb67e0a4ef8f0a/includes/api/ApiMove.php#210) and some non-admins (page movers) do indeed have the ability to move subpages.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2855,7 +2855,7 @@ Morebits.wiki.page = function(pageName, currentAction) {
 			query.movetalk = 'true';
 		}
 		if (ctx.moveSubpages) {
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Seems like we could do better error handling on this, but then again we could probably do better error handling on the whole `move` thing anyway.

PR #787, opened 2019-12-22
config: add two missing skins to 'Looking for Twinkle' notice

Amorymeltzer:
timeless and minerva(neue)

FWIW: cologne blue is still installed on enwiki despite not being automatically selectable


siddharthvp:
Twinkle doesn't work in minerva so I think there's little point in adding that notice there.

Amorymeltzer:
Meh.  It's for any of the skin js files, whether active (e.g., I use modern, it'd show up at `/timeless.js`) or even existing.  As long as it's not `[mw.config.get('skin'), 'common']`, I see no harm in being complete, at least as to what could reasonably be used by someone (i.e. not chick).

PR #789, opened 2019-12-30
Speedy: Allow users to request salting

DannyS712:
Creation protection can be requested via checkbox. Not available when sysop is in sysop mode (though a future task could allow setting creation protection while deleting)

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -119,6 +119,9 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
 						// enable/disable multiple
 						cForm.multiple.disabled = !cChecked;
 						cForm.multiple.checked = false;
+						// enable/disable requesting creation protection
+						cForm.salting.disabled = !cChecked;
+						cForm.salting.checked = false;
 
 						Twinkle.speedy.callback.modeChanged(cForm);
 
@@ -243,6 +246,21 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
 				disabled: isSysop && !Twinkle.getPref('deleteSysopDefaultToTag'),
 				event: function(event) {
 					Twinkle.speedy.callback.modeChanged(event.target.form);
+					event.stopPropagation();
+				}
+			}
+		]
+	});
+	tagOptions.append({
+		type: 'checkbox',
+		list: [
+			{
+				label: 'Request creation protection',
+				value: 'salting',
+				name: 'salting',
+				tooltip: 'When selected, tagging the page for deletion will be accompanied by a request for the deleting administrator to apply creation protection.',
+				disabled: isSysop && !Twinkle.getPref('deleteSysopDefaultToTag'),
+				event: function(event) {
 					event.stopPropagation();
 				}
 			}
@@ -1481,6 +1499,10 @@ Twinkle.speedy.callbacks = {
 				text = text.replace(/\{\{(mtc|(copy |move )?to ?commons|move to wikimedia commons|copy to wikimedia commons)[^}]*\}\}/gi, '');
 			}
 
+			if (params.requestsalt) {
+				code = '{{salt}}\n' + code;
+			}
+
 			// Generate edit summary for edit
 			var editsummary;
 			if (params.normalizeds.length > 1) {
@@ -2087,6 +2109,11 @@ Twinkle.speedy.callback.evaluateUser = function twinklespeedyCallbackEvaluateUse
 		});
 	}
 
+	var requestsalt = false;
+	if (form.salting.checked) {
+		requestsalt = true;
+	}
+
 	var csdlog = false;
 	if (Twinkle.getPref('logSpeedyNominations')) {
 		$.each(normalizeds, function(index, norm) {
@@ -2104,6 +2131,7 @@ Twinkle.speedy.callback.evaluateUser = function twinklespeedyCallbackEvaluateUse
 		usertalk: notifyuser,
 		welcomeuser: welcomeuser,
 		lognomination: csdlog,
+		requestsalt: requestsalt,
 		templateParams: Twinkle.speedy.getParameters(form, values)
 	};
 	if (!params.templateParams) {
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -258,7 +258,7 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
 				label: 'Tag for creation protection (salting) as well',
 				value: 'salting',
 				name: 'salting',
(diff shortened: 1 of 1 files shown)


DannyS712:
@siddharthvp @Amorymeltzer is this something that would need to be discussed onwiki first?

modules/twinklespeedy.js
```diff
@@ -1481,6 +1499,10 @@ Twinkle.speedy.callbacks = {
 				text = text.replace(/\{\{(mtc|(copy |move )?to ?commons|move to wikimedia commons|copy to wikimedia commons)[^}]*\}\}/gi, '');
 			}
 
+			if (params.requestsalt) {
+				code = '{{salt}}\n' + code;
```
Amorymeltzer:
```suggestion
				code = code + '\n{{salt}}';
```
>The template should be placed at the top of the page to be protected against creation, under any applicable speedy deletion template.

modules/twinklespeedy.js
```diff
@@ -2087,6 +2109,11 @@ Twinkle.speedy.callback.evaluateUser = function twinklespeedyCallbackEvaluateUse
 		});
 	}
 
+	var requestsalt = false;
```
Amorymeltzer: Couldn't this bit simply be replaced with `var requestsalt = form.salting.checked || false;` or a ternary or something?

modules/twinklespeedy.js
```diff
@@ -248,6 +251,21 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
...
 		]
 	});
+	tagOptions.append({
+		type: 'checkbox',
+		list: [
+			{
+				label: 'Request creation protection',
+				value: 'salting',
+				name: 'salting',
+				tooltip: 'When selected, tagging the page for deletion will be accompanied by a request for the deleting administrator to apply creation protection.',
```
Amorymeltzer: I think it'd be better to make this a little more forceful and clear *when* it should be used.  That is, something like "only use if the page or title has been repeatedly been created" but, you know, better.  As above, this could also be a little more clear as to what will happen (i.e. appending {{salt}}).

modules/twinklespeedy.js
```diff
@@ -248,6 +251,21 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
 			}
 		]
 	});
+	tagOptions.append({
+		type: 'checkbox',
+		list: [
+			{
+				label: 'Request creation protection',
```
Amorymeltzer: Getting wordy, but I'd like to make it clear this isn't going to post at RfPP or anything.  Might be better ways to phrase it but something like:

```suggestion
				label: 'Tag for creation protection (salting) as well',
```

modules/twinklespeedy.js
DannyS712: Done

modules/twinklespeedy.js
DannyS712: Done

modules/twinklespeedy.js
```diff
@@ -2087,6 +2109,8 @@ Twinkle.speedy.callback.evaluateUser = function twinklespeedyCallbackEvaluateUse
 		});
 	}
 
+	var requestsalt = form.salting.checked || false;
```
siddharthvp:
```suggestion
	var requestsalt = form.salting.checked;
```
and there's no need for this variable as it's used just once.

modules/twinklespeedy.js
```diff
@@ -248,6 +251,21 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
...
+				value: 'salting',
+				name: 'salting',
+				tooltip: 'When selected, the speedy deletion tag will be accompanied by a {{salt}} tag requesting that the deleting administrator apply creation protection. Only select if this page has been repeatedly recreated.',
+				disabled: isSysop && !Twinkle.getPref('deleteSysopDefaultToTag'),
+				event: function(event) {
+					event.stopPropagation();
+				}
+			}
+		]
+	});
```
siddharthvp: Put the three checkbox configs in one list. `tagOptions.append({ type: 'checkbox', list: [ { ... checkbox 1 ... }, { ... checkbox 2 ... } ]` rather than in 3 separate ones.

Also, I think the "Tag with multiple criteria" option should be at the bottom.

Amorymeltzer: >Also, I think the "Tag with multiple criteria" option should be at the bottom

OOS for this, but I think that'd be better discussed separately if it's a good idea; delete and tag are somewhat misaligned after #563.

siddharthvp: I don't know what "OOS" means. 

I meant the new checkbox should probably come above the the "Tag with multiple criteria" checkbox.

DannyS712: Moved above

modules/twinklespeedy.js
DannyS712: Removed variable

modules/twinklespeedy.js
Amorymeltzer: Ah gotcha, that makes sense.

modules/twinklespeedy.js
```diff
@@ -232,6 +235,21 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
 			}
 		]
 	});
+	tagOptions.append({
+		type: 'checkbox',
+		list: [
+			{
+				label: 'Tag for creation protection (salting) as well',
```
Amorymeltzer: I know I suggested this language, but it still feels awkward.  I don't have a suggestion — I think it's good to have something like "also" or "as well" but it's good to have "Tag" be the first word — but just wanted to note it.

DannyS712: Maybe "Request creation protection (salting) as well"?

Amorymeltzer: I think having the "tag" in there is helpful, so it's clear we're not making an RfPP request.  Maybe that's over the top, though?  It's probably fine as is, just making life tricky.

modules/twinklespeedy.js
```diff
@@ -1481,6 +1499,10 @@ Twinkle.speedy.callbacks = {
 				text = text.replace(/\{\{(mtc|(copy |move )?to ?commons|move to wikimedia commons|copy to wikimedia commons)[^}]*\}\}/gi, '');
 			}
 
+			if (params.requestsalt) {
+				code = code + '\n{{salt}}';
```
Amorymeltzer: Okay well I've gotten crickets regarding my [g10 question](https://github.com/azatoth/twinkle/pull/789#pullrequestreview-341554279), so I think we should try to get the tag to show up on g10 pages.  This should do it I believe:

```suggestion
				if (params.normalizeds.indexOf('g10') === -1) {
					code = code + '\n{{salt}}';
				} else {
					code = '{{salt}}\n' + code;
				}
```

PR #790, opened 2019-12-30
Protect: Add section link at rpp

DannyS712:
See onwiki request at https://w.wiki/EeR


siddharthvp:
While adding a section link is a good idea, the page name shouldn't redundantly appear again in the edit summary. Maybe `/* pageName */ Adding request for protectionType` looks better?

Amorymeltzer:
In theory I'd agree, but having the relevant page linked in the edit summary is nice and I wouldn't want to miss that.

PR #792, opened 2020-01-06
fluff: Restructure fluff to use Twinkle.fluff() instead of Twinkle.fluff.init()

Amorymeltzer:
`twinklefluff.js` and `twinkleconfig.js` are the only modules that are loaded via `Twinkle.module.init` rather than `Twinkle.module`.  This was sort of @atlight's work in 2c544c1 but goes all the way back to @UncleDouggie's restructuring in a79b7d9.  These are the only two modules that aren't added to a menu/tab and are instead just run, so something different did make sense.  Still, I don't see a reason for `Twinkle.fluff()` to be different from the other tool-based modules: it may not "add menu button" but it does "enable the module."  I'm fine with config being different since it's clearly very different in scope.

Oh, and we can't use `Object.assign` because IE.

----

Putting this live on testwiki for testing, especially since the diff is annoying.


siddharthvp:
Since the diff is anyway so messy, this might be a good time to add some organization here. Maybe group `contributions`, `diff`, and `oldid` functions into a `Twinkle.fluff.addRollbackLinks` object?

Amorymeltzer:
Nice idea — even makes the diff nicer!  I think I held off not with the accessory functions left around, but you're right that it's nice to have the link-adding functions together.  Done as `addLinks`.

PR #793, opened 2020-01-06
Add new option to disable individual Twinkle modules

Amorymeltzer:
Not everybody uses everything, and even in a menu, a lot of menus you never intend on using is annoying.  This adds two new config options, `Twinkle.config.disabledModules` and `Twinkle.config.disabledSysopModules`, and constructs an array of module names to be turned off.  The loading mechanism in `Twinkle.js` now only loads items not explicitly turned off.  The instances where we link from one to another (deli->spoeedy, speedy->unlink) work fine since they use the `callback`.  `config` cannot be turned off, and the default is, of course, to have everything on.

----

Currently live on testwiki for testing.  Depends on #792, but if that's not committed, it'd be trivial to replace `Twinkle[module]();` with:

```javascript
if (module !== 'fluff') {
   Twinkle[module]();
} else {
  Twinkle[module].init();
}
```

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -197,6 +197,24 @@ Twinkle.config.sections = [
 				name: 'dialogLargeFont',
 				label: 'Use larger text in Twinkle dialogs',
 				type: 'boolean'
+			},
+
+			// Twinkle.config.disabledModules (array)
+			{
+				name: 'disabledModules',
+				label: 'Turn off the selected Twinkle modules',
+				helptip: 'Anything you select here will NOT be available for use, so act with care. Uncheck to reactivate.',
+				type: 'set',
+				setValues: { arv: 'ARV', warn: 'Warn', welcome: 'Welcome', shared: 'Shared IP', talkback: 'Talkback', speedy: 'CSD', prod: 'PROD', xfd: 'XfD', image: 'Image (DI)', protect: 'Protect (RPP)', tag: 'Tag', diff: 'Diff', unlink: 'Unlink', 'fluff': 'Revert and rollback' }
+			},
+
+			// Twinkle.config.disabledSysopModules (array)
+			{
+				name: 'disabledSysopModules',
+				label: 'Turn off the selected admin-only modules',
+				helptip: 'Anything you select here will NOT be available for use, so act with care. Uncheck to reactivate.',
+				type: 'set',
+				setValues: { block: 'Block', deprod: 'DePROD', batchdelete: 'D-batch', batchprotect: 'P-batch', batchundelete: 'Und-batch' }
 			}
 		]
 	},
--- a/twinkle.js
+++ b/twinkle.js
@@ -8,9 +8,6 @@
  * Imported from github [https://github.com/azatoth/twinkle].
  * All changes should be made in the repository, otherwise they will be lost.
  *
- * To update this script from github, you must have a local repository set up. Then
- * follow the instructions at [https://github.com/azatoth/twinkle/blob/master/README.md].
- *
  * ----------
  *
  * This is AzaToth's Twinkle, the popular script sidekick for newbies, admins, and
@@ -52,6 +49,8 @@ Twinkle.defaultConfig = {
 	protectionSummaryAd: ' ([[WP:TW|TW]])',
 	userTalkPageMode: 'tab',
 	dialogLargeFont: false,
+	disabledModules: [],
+	disabledSysopModules: [],
 
 	// ARV
 	spiWatchReport: 'yes',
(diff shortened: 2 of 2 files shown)


siddharthvp:
I have always wondered why we initialise the modules from `twinkle.js` rather than from the module's file itself. It makes testing changes to `twinkle.js` and testing of custom Twinkle installations (like on other wikis) a whole lot more cumbersome as one must remember to comment out or change these initialisation lines every time. Maybe we can just convert the initialisation function to an IIFE? Or if that doesn't work, just call the function at the very end of the file, eg. `Twinkle.tag()` ? The order in which the menu options appear is still in our control through the gadget definition line (the files are executed in the order they appear in the gadget definition line). 

----
In any case, note that the `Morebits.userIsInGroup()` checks here aren't required as that gets checked anyway in the module. So, in case we bothering about the above in this PR, we could still do something more elegant like 
```
[ ... hard-coded list of all modules ... ].filter(function() {
    // filter out disabled ones
}).forEach(function(module) {
    Twinkle[module]();
});

**Comment by Amorymeltzer** [comment]
That's a fair point on the sysop-only modules, I hadn't considered that.  Is there really no advantage to including that here?  I can't think of any off the top of my head.  I initially had a few other options laid out in my initial post, including roughly what you suggest, but held off 'cause of the sysop stuff.  If there's really no need in `twinkle.js` then yeah, just listing in an array would be much nicer! 👍

File: modules/twinkleconfig.js [hunk 1]
```diff
@@ -200,6 +200,24 @@ Twinkle.config.sections = [
 				name: 'dialogLargeFont',
 				label: 'Use larger text in Twinkle dialogs',
 				type: 'boolean'
+			},
+
+			// Twinkle.config.disabledModules (array)
+			{
+				name: 'disabledModules',
+				label: 'Turn off the selected Twinkle modules',
+				helptip: 'Anything you select here will NOT be loaded by Twinkle, so act with care. Uncheck to reenable.',
```
siddharthvp:
```suggestion
				helptip: 'Anything you select here will NOT be available for use, so act with care. Uncheck to re-enable.',
```
Or something.

The original isn't technically true as everything is still loaded. It's just that the code isn't executed. Users shouldn't get the impression that disabling TW modules will result in faster page loads (or lesser bandwidth consumption).

[nitpick] `reenable` → `re-enable` maybe?

modules/twinkleconfig.js
```diff
@@ -200,6 +200,24 @@ Twinkle.config.sections = [
...
+				type: 'set',
+				setValues: { arv: 'ARV', warn: 'Warn', welcome: 'Welcome', shared: 'Shared IP', talkback: 'Talkback', speedy: 'CSD', prod: 'PROD', xfd: 'XfD', image: 'Image (DI)', protect: 'Protect (RPP)', tag: 'Tag', diff: 'Diff', unlink: 'Unlink', 'fluff': 'Revert and rollback' }
+			},
+
+			// Twinkle.config.disabledSysopModules (array)
+			{
+				name: 'disabledSysopModules',
+				label: 'Turn off the selected Admin-only modules',
```
siddharthvp:
```suggestion
				label: 'Turn off the selected admin-only modules',
```

modules/twinkleconfig.js
```diff
@@ -200,6 +200,24 @@ Twinkle.config.sections = [
...
+				helptip: 'Anything you select here will NOT be loaded by Twinkle, so act with care. Uncheck to reenable.',
+				type: 'set',
+				setValues: { block: 'Block', deprod: 'DePROD', batchdelete: 'D-batch', batchprotect: 'P-batch', batchundelete: 'Und-Bbatch' }
```
siddharthvp:
```suggestion
				setValues: { block: 'Block', deprod: 'DePROD', batchdelete: 'D-batch', batchprotect: 'P-batch', batchundelete: 'Und-batch' }
```

modules/twinkleconfig.js
Amorymeltzer: I think now is actually the time to commit Twinkle to the New Yorker standard: `reënable`

Amorymeltzer: The nuance is probably lost on most folks, and truth be told "available for use" isn't quite right either, but it's closer and I take the point.

How's reactivate?

Amorymeltzer:
Still not sure why; the sysop checks were in place for the sysop-only modules prior to this bit being written, so maybe just to save folks a half dozen/dozen `if` checks?  I suppose those do add up on every page load...

siddharthvp:
Haha those are theoretically supposed to quite speedy. However, we should consider storing the sysop check result in as `Morebits.userIsSysop`. It gets used too many times, no point in traversing the array every time.

Amorymeltzer:
{{done}} (in a separate commit for review)

modules/twinkleconfig.js
siddharthvp: > I think now is actually the time to commit Twinkle to the New Yorker standard: `reënable`

You're [kidding me](https://www.google.com/search?q=re%C3%ABnable+site%3Anewyorker.com)?

Amorymeltzer: [Diaresis all the things!](https://www.newyorker.com/culture/culture-desk/the-curse-of-the-diaeresis) 😀 😆 😀

PR #795, opened 2020-01-07
talkback: Don't parse html, always check user talk page for optout

Amorymeltzer:
1st commit: Template:No_talkback creates a link to `http://userjs.invalid/noTalkback?reason=`; that gets used by the talkback module to provide a custom reason to the user, but by using `.append`, it allows someone to provide and run code via `<script>`.  Changing to `.text` does away with that.  See phab:T241950.

In the future, there are other options here, such as querying `templates` (removes the ability for editors to provide a custom `reason`), querying `revisions` (regex to match page the content), or simple removing the ability to provide custom text.

2nd commit: Always check the talkpage for optout.  Also don't append a . if user provides a custom reason, allow them to pick their punctuation

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -115,7 +115,7 @@ Twinkle.talkback.callback.optoutStatus = function(apiobj) {
 
 	var $status = $('#twinkle-talkback-optout-message');
 	if ($status.length) {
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
cc @jimmyxu

Issue #723, opened 2019-10-08
Replace all uses of wgMonthNames and wgMonthNamesShort


Amorymeltzer:
Not super urgent yet, but these [will be removed](https://phabricator.wikimedia.org/T219340) from `mw.config` down the line.  We should discuss how we'd like to proceed.

Our best options seem to be either:
1. Revert #635, where we removed our own built-ins, or
2. Require loading `mediawiki.language.months` and then use the `mw.language.months.names` and `mw.language.months.abbrev` arrays

Amorymeltzer:
Option 2 is preferable if only for l10n reasons, so what's another dependency amongst friends?

siddharthvp:
Option 2 is not good because as noted on phab, mw.language uses the language of the user than that of the wiki. That would be quite counter-intuitive. 

We have a couple of more options: 

3. We are already loading `moment`. We could use `moment()._locale._months` and `moment()._locale._monthsShort`. But this is not guaranteed to give the local month names on all wikis. For it to work, the library needs to have locale registered. It works on fr and de WPs (both of which have both en and the local language locale registered), but not on es, zh, and hi, which only have the en locale registered. (Run `moment.locales()` to see list of registered locales.) 

4. Another option is to just [use the API](https://en.wikipedia.org/w/api.php?action=query&utf8=y&meta=allmessages&ammessages=january|february|march|april|may|june|july|august|september|october|november|december|jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec&amlang=fr) (fr example). A single api call can get all month names and short month names. This works correctly on every wiki. But then, that's a preemptive API call on every page load :(  But until recently, an api call was preemptively being used on every page load (a33a3c5) to check EC userright, added by @MusikAnimal. So it maybe it isn't an issue after all? 

I guess the best options are 1 and 4. I'm leaning on 1 as we aren't accounting for other wikis' usage anywhere else in Twinkle. But since this is morebits, it might be a worthy consideration here.

Amorymeltzer:
I suppose it depends on how (if?) they do it I suppose — I would hope the wiki-language V user-language issue to be something they resolve before removing these.  If not, then you're right and 2 is out and 1 is probably our best bet — i18n stuff is, as you say, not a concern atm and when (if?) that ever happens, month names will be the least of the worries.

Funny you should mention `moment`... I want to play around with some things and see if we can't remove it.

siddharthvp:
[...]

PR #635, opened 2019-05-12
morebits: internationalize Date, and remove unused functions

siddharthvp:
- remove `Date.monthNames` and `Date.monthNamesAbbrev` in favour of mw.config objects `wgMonthNames` and `wgMonthNamesShort`, which work across wikis and languages. A difference is that they are indexed from 0-11 while the config objects are indexed 1-12. These are unused except from `Date.prototype.getUTCMonthName`/ `Date.prototype.getUTCMonthNameAbrev`. 
- remove `Date.prototype.getMonthName` and `Date.prototype.getMonthNameAbbrev` as they are completely unused - not used in Twinkle or any other Morebits-based gadget/script. See searches [1](https://en.wikipedia.org/w/index.php?search=insource%3A%2FgetMonthNameAbbrev%2F&title=Special%3ASearch&profile=advanced&fulltext=1&advancedSearch-current=%7B%7D&ns0=1&ns1=1&ns2=1&ns3=1&ns4=1&ns5=1&ns6=1&ns7=1&ns8=1&ns9=1&ns10=1&ns11=1&ns12=1&ns13=1&ns14=1&ns15=1&ns100=1&ns101=1&ns108=1&ns109=1&ns118=1&ns119=1&ns446=1&ns447=1&ns710=1&ns711=1&ns828=1&ns829=1&ns2300=1&ns2301=1&ns2302=1&ns2303=1) and [2](https://en.wikipedia.org/w/index.php?title=Special:Search&limit=100&offset=0&ns0=1&ns1=1&ns2=1&ns3=1&ns4=1&ns5=1&ns6=1&ns7=1&ns8=1&ns9=1&ns10=1&ns11=1&ns12=1&ns13=1&ns14=1&ns15=1&ns100=1&ns101=1&ns108=1&ns109=1&ns118=1&ns119=1&ns446=1&ns447=1&ns710=1&ns711=1&ns828=1&ns829=1&ns2300=1&ns2301=1&ns2302=1&ns2303=1&search=insource%3A%2FgetMonthName%2F&advancedSearch-current=%7B%7D&searchToken=5hiz771molt4nbkx8bcva4k7t).

PR #796, opened 2020-01-07
Partially revert "morebits: internationalize Date, and remove unused functions"

Amorymeltzer:
This partially reverts commit 28d50ac from #635, restoring the hardcoded values for `Date.monthNames` and `Date.monthNamesAbbrev` while `Date.prototype.getMonthName` and `Date.prototype.getMonthNameAbbrev` remain removed (unused outside of ko.wiki).

Done in advance of [phab:T219340](phabricator.wikimedia.org/T219340); see discussion in #723, including #723 (comment) and #723 (comment).

As noted there, this is not the ideal situation, but it's simple, covers our current needs, and except for the last four months, was the norm for *years*.

cc @MusikAnimal @siddharthvp @Krinkle

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1249,13 +1249,19 @@ Morebits.unbinder.getCallback = function UnbinderGetCallback(self) {
  * is fairly unlikely that anyone will iterate over a Date object.
  */
 
+Date.monthNames = ['January', 'February', 'March', 'April', 'May', 'June',
+	'July', 'August', 'September', 'October', 'November', 'December' ];
+
+Date.monthNamesAbbrev = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+
 Date.prototype.getUTCMonthName = function() {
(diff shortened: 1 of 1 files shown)


morebits.js
```diff
@@ -1249,14 +1249,28 @@ Morebits.unbinder.getCallback = function UnbinderGetCallback(self) {
  * is fairly unlikely that anyone will iterate over a Date object.
  */
 
+Date.monthNames = ['January', 'February', 'March', 'April', 'May', 'June',
+	'July', 'August', 'September', 'October', 'November', 'December' ];
+
+Date.monthNamesAbbrev = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+
+Date.prototype.getMonthName = function() {
```
Krinkle: I suppose these could remain absent for now. Unless we need them back?

Amorymeltzer: Per #635, yeah.  A global search shows that only kowiki actually uses it (also ro but IIRC ko works) but otherwise unused.

PR #798, opened 2020-01-11
talkback: Don't show 'False' optout always

### LINKED ISSUE #797: Talkback optout gives negative feedback
If user has not opted out, we should not display an opt-out message at all. 
`if(Twinkle.talkback.optout)` should deal with this correctly but apparently is not.
@Amorymeltzer

Amorymeltzer:
Closes #797, introduced in 1f52094


Amorymeltzer:
cc @QEDK

PR #800, opened 2020-01-13
config: Add optionsVersion 2 to preferences now that they're unified

Amorymeltzer:
Was discussed in #762 but not done there.  The format has been pretty stable before the unification (occasional new options aside) so this is reasonable as a marker.

----

The intent here (see https://github.com/azatoth/twinkle/pull/762#issuecomment-566443649 and https://github.com/azatoth/twinkle/pull/762#issuecomment-566486414) is to just mark current preferences going forward.  Is this all there really is to it?  cc @siddharthvp (also @atlight and @MusikAnimal in case you're curious)

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -1536,13 +1536,13 @@ Twinkle.config.writePrefs = function twinkleconfigWritePrefs(pageobj) {
 	var form = pageobj.getCallbackParameters();
 
 	// this is the object which gets serialized into JSON
-	var newConfig = {};
+	var newConfig = {optionsVersion: 2};
 
 	// keeping track of all preferences that we encounter
 	// any others that are set in the user's current config are kept
 	// this way, preferences that this script doesn't know about are not lost
 	// (it does mean obsolete prefs will never go away, but... ah well...)
-	var foundPrefs = [];
+	var foundPrefs = ['optionsVersion'];
 
 	// a comparison function is needed later on
 	// it is just enough for our purposes (i.e. comparing strings, numbers, booleans,
--- a/twinkle.js
+++ b/twinkle.js
@@ -406,6 +406,8 @@ $.ajax({
 				} else {
 					Twinkle.prefs = options;
 				}
+				// v2 established after unification of Twinkle/Friendly objects
+				Twinkle.prefs.optionsVersion = Twinkle.prefs.optionsVersion || 1;
 			}
 		} catch (e) {
 			mw.notify('Could not parse twinkleoptions.js');


siddharthvp:
This patch doesn't achieve anything, right? The way to mark the version is to write the `optionsVersion` property into twinkleoptions.js files. 

It makes sense to either add optionsVersion=2, or to not add it at all. optionsVersion=1 is understood by the absence of the property; doesn't need to written to the page.

Amorymeltzer:
> This patch doesn't achieve anything, right? The way to mark the version is to write the `optionsVersion` property into twinkleoptions.js files.

No, `twinkleconfig` just processes the page as-is and takes the options from the `Twinkle.prefs` object (see https://github.com/azatoth/twinkle/blob/87ead8932245c433bf38a91a9aeadc81022ec0dd/modules/twinkleconfig.js#L1641).  Reading the `twinkleoptions.js` page is handled by `twinkle.js` (see the start of the block in question here).

> It makes sense to either add optionsVersion=2, or to not add it at all. optionsVersion=1 is understood by the absence of the property; doesn't need to written to the page.

Our conversation there got cut short, but the point of this is not to *do* anything, more just declare a marker in the sand.  I'm not wed to `1` versus `2`, just want to restart the conversation (truth be told, I misremembered and thought I was proposing version 2 and you proposed version 1, and I thought I was being conciliatory 😆).   #762 *will* restructure the page, so either way I think I'm convinced that it's okay to declare it so by writing to the page (I don't think anything else uses Twinkle's preferences but, in theory...).

I suppose the old-old could be considered v0, the long-time standard with separate twinkle and friendly objects as v1, and this unified structure as v2.  Or everything until being "unified" was sub v1.  Or maybe it's all too much, and until we changing the structure nothing matters so why bother.

siddharthvp:
Ok. Now I see that this patch indirectly causes the optionsVersion to be written to the page (right?). Took me several minutes to realise that.

I had considered doing it via twinkleconfig.js rather than twinkle.js: by changing [L1539](https://github.com/azatoth/twinkle/blob/87ead8932245c433bf38a91a9aeadc81022ec0dd/modules/twinkleconfig.js#L1539) from `var newConfig = {}` to `var newConfig = { optionsVersion: 2 }` and `var foundPrefs = [ 'optionsVersion' ]; below.

Don't you think that's a clearer way of doing it? This also means that 
- the optionsVersion property is placed right at the top of the JSON object, which I think is desirable. 
- If someone were to run `Twinkle.prefs.optionsVersion` in the console, they'd correctly get `2` for v2 and `undefined` for v1. But the patch here would give the same value for _both_ versions.

Amorymeltzer:
That's much better, yes!  I've just pushed that up as v2, works great; having it top of the prefs is a plus.  I've kept the bit in `twinkle.js` as I think `Twinkle.prefs.optionsVersion` is nicer as `1` than `undefined`.

siddharthvp:
Do we want to do something to avoid creating the twinkleoptions.js with just the optionsVersion property? something like:
```
// abort if there's only the optionsVersion prop
if (!pageobj.exists() && Object.keys(newConfig).length === 1) {
    pageobj.getStatusElement().info('No preferences changed from defaults. Nothing to save.');
    return;
}
```

Not really necessary because even now, twinkleoptions.js page can get created with an empty JSON object (if someone clicks on "Save changes" without changing any preferences).

Amorymeltzer:
Maybe (probably?), but I kind of see that as a plus?  Saving an entirely blank page is bad, but marking the version ain't bad.  I guess: if it weren't for this, I'd suggest adding it, and if it were already in there before this, I wouldn't suggest removing it, but I don't think it's wildly necessary.

PR #801, opened 2020-01-13
twinkle: fix bug from #762

### LINKED ISSUE #799: config doesn't handle empty values anymore
While you can still set various inputs to be empty – talkback section headings, default welcome template, prod log subpage, etc. — loading config again will lead to the default being re-inserted and replaced upon your next loading.  This means that it's possible to go to TW/PREF and just click save, and have a difference saved.

I presume this comes after #762 (cc @siddharthvp) but haven't looked into it yet.

siddharthvp:
Don't ignore falsy preference values (eg. '', false, 0).

Closes #799.

### DIFF
--- a/twinkle.js
+++ b/twinkle.js
@@ -186,14 +186,14 @@ switch (mw.config.get('skin')) {
 
 
 Twinkle.getPref = function twinkleGetPref(name) {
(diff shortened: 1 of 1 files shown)


siddharthvp:
Um, I think you're mis-reading it.

>

Issue #802, opened 2020-01-13
Add support for partial blocks


Amorymeltzer:
The [feature](https://en.wikipedia.org/wiki/Wikipedia:Community_health_initiative_on_English_Wikipedia/Partial_blocks) is now live nearly everywhere, should we attempt to support it?  It's a fair bit different in structure from the current block module, and there's no [policy yet in place](https://en.wikipedia.org/wiki/Wikipedia:Partial_blocks), but it'd be good to consider it.  cc @MusikAnimal as the original author

API-wise, the additional parameters are:
* `partial: 1`
* `pagerestrictions`
* `namespacerestrictions`

Also, it seems `allowusertalk` *must* be true for partial blocks unless user talk is a restricted namespace.

Amorymeltzer:
This happened rather quickly so there aren't yet any PB-specific templates.  I'd imagine we'd need different ones.

PR #803, opened 2020-01-13
Block: Tweak reblock message when user is partially blocked

Amorymeltzer:
This merely notes whether a user is partially blocked or sitewide blocked in the red message to sysops.  Also notes in a few minor places that the module is for sitewide blocks (#802).

A query to `list: 'blocks'` without any `bkprop`s defined (as we have here) apparently returns `partial: ''` if the user is partially blocked, and nothing if they're not, so relies on that.


Amorymeltzer:
Done in 6ed99f1

PR #806, opened 2020-01-15
batchdelete: fix bug in Select all/Deselect all buttons

siddharthvp:
The buttons didn't work when there were titles containing `"` in the page list.

### DIFF
--- a/modules/twinklebatchdelete.js
+++ b/modules/twinklebatchdelete.js
@@ -204,8 +204,8 @@ Twinkle.batchdelete.callback = function twinklebatchdeleteCallback() {
 			type: 'button',
 			label: 'Select All',
 			event: function dBatchSelectAll() {
-				result.getUnchecked('pages').forEach(function(e) {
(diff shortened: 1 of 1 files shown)


siddharthvp:
`.prop('checked', true)` doesn't trigger the click event, and thus the subgroups (which may have the list of subpages) won't be hidden/shown as necessary. So we have to use `.click()`, which in turn means that we have to specifically select only the checked (for deselect button) or unchecked (for select button) chekcboxes which isn't possible with `Morebits.quickForm.getElements`.

Amorymeltzer:
Ahhh right, I remember that now from when you added subpages.  A comment might be nice to add but otherwise 💯 from me.

Amorymeltzer:
Thank you for the contribution. It needed some fixes, so they were made in 778255e1a77491d9805999ec0c7a0549df87454e. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

Issue #405, opened 2017-12-25
{{No talkback}} category


JJMC89:
Per a [CfD for Category:Users who do not wish to receive talkbacks](https://en.wikipedia.org/wiki/Wikipedia:Categories_for_discussion/Log/2017_October_12#Category:Users_who_do_not_wish_to_receive_talkbacks), Twinkle should be modified to not depend on the category. If this is not possible without loss of functionality, the category will be renamed to `Wikipedians who wish to not receive talkbacks`.

The current category name is referenced:
https://github.com/azatoth/twinkle/blob/67e51a0754d29c78ef189d8731d8f2633285e3bd/modules/friendlytalkback.js#L89

JJMC89:
@MusikAnimal @atlight Is it possible to remove the category check without loss of functionality?

atlight:
Yes, it's possible; we'd have to run the API query unconditionally. Let me see what I can do.

atlight:
Hopefully fixed; someone like @MusikAnimal will need to synchronise this.

PR #807, opened 2020-01-15
talkback: Should we remove the custom optout message?

Amorymeltzer:
Opened to start some discussion, with some sample code for one implementation; I don't particularly care, but was on my mind.  It might be good to ask on-wiki as well.

It was first added in 2013 (99893a8) and relies upon creating a facade of a link with Template:No_talkback and then querying `extlinks`, but 6.5 years later, only three enWiki users (out of 26 or so) use the custom message, of which only one is actually active: @JJMC89, who has said they don't mind this being done away with.  In its place, we could just use the default message if Template:No_talkback is transcluded, simplifying things a bit.

<s>We could also just use a category, perhaps applied by the template to avoid migration.</s> See below.

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -86,10 +86,9 @@ Twinkle.talkback.callback = function() {
 	// Check whether the user has opted out from talkback
 	var query = {
 		action: 'query',
-		prop: 'extlinks',
+		prop: 'templates',
 		titles: 'User talk:' + mw.config.get('wgRelevantUserName'),
-		elquery: 'userjs.invalid/noTalkback',
-		ellimit: '1'
+		tltemplates: 'Template:No talkback'
 	};
 	var wpapi = new Morebits.wiki.api('Fetching talkback opt-out status', query, Twinkle.talkback.callback.optoutStatus);
 	wpapi.post();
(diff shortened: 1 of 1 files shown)


JJMC89:
No objections from me. [This CfD](https://en.wikipedia.org/wiki/Wikipedia:Categories_for_discussion/Log/2017_October_12#Category:Users_who_do_not_wish_to_receive_talkbacks) and #405 are relevant to the category option.

siddharthvp:
Whoa never knew such a thing existed! Seems a nice little feature that makes Twinkle look cool. I don't see any motivation for removing it? Relying on an external link and going off API:Extlinks is an ingenious idea. Much better than using API:Revisions (much bigger response size, hence greater BW usage) or API:Templates (as noted, can't support reason field as parameters are not returned).

That being said, there is a lot of bogus code here that can be greatly simplified. Wait I'll open a PR for that in a sec.

Amorymeltzer:
So that's a 👎  then 😀.  `revisions` definitely wouldn't work, it wouldn't support transclusion from a subpages/template.

Thanks for the links @JJMC89; I guess the opposition was to pointless categorization, which I agree with.  Opting out via template'd be better IMO, but I missed that there was a definite decision on that front.

PR #791, opened 2019-12-31
Merge pull request #791 from Amorymeltzer/xfd-fixrmtr-hiddencomment

Amorymeltzer:
See discussions at [[Special:Permalink/933352249#"Adding entry at WP:RM/TR: failed to find target spot for the entry"]] and [[Special:Permalink/933352294#New technical requests: Place at the top or bottom of the list?]]

----

I've already done a quick fix for some changes on-wiki, but this is slightly different.

PR #808, opened 2020-01-16
Merge pull request #808 from siddharthvp/tboptout

siddharthvp:
Alternative to #807.

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -95,28 +95,17 @@ Twinkle.talkback.callback = function() {
 	wpapi.post();
 };
 
-Twinkle.talkback.optout = null;
+Twinkle.talkback.optout = '';
 
 Twinkle.talkback.callback.optoutStatus = function(apiobj) {
-	var xml = apiobj.getXML();
-	var $el = $(xml).find('el');
-
+	var $el = $(apiobj.getXML()).find('el');
 	if ($el.length) {
 		Twinkle.talkback.optout = mw.config.get('wgRelevantUserName') + ' prefers not to receive talkbacks';
 		var url = $el.text();
-		if (url.indexOf('reason=') > -1) {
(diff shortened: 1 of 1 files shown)


modules/friendlytalkback.js
```diff
@@ -95,28 +95,17 @@ Twinkle.talkback.callback = function() {
...
-			Twinkle.talkback.optout += ': ' + decodeURIComponent(url.substring(url.indexOf('reason=') + 7));
-		} else {
-			Twinkle.talkback.optout += '.';
-		}
-	} else {
-		Twinkle.talkback.optout = false;
-	}
-
-	var $status = $('#twinkle-talkback-optout-message');
-	if ($status.length && Twinkle.talkback.optout) {
```
siddharthvp: both checks are unnecessary now: (i) trying to do `.text()` on empty jquery selection doesn't give any error. (ii) If not opted out, `Twinkle.talkback.optout` is just the empty string - no harm in appending it.

modules/friendlytalkback.js
```diff
@@ -95,28 +95,17 @@ Twinkle.talkback.callback = function() {
...
-		$status.text(Twinkle.talkback.optout);
+		var reason = mw.util.getParamValue('reason', url);
```
Amorymeltzer: Smart use of `mw.util.getParamValue`

siddharthvp:
That was quick!

> Amorymeltzer changed the title ~~talkback: simplify talkback optout code~~ Merge pull request #808 from siddharthvp/tboptout

This shouldn't be happening, right? Been using some broken script? Same with #794 and #791.

Amorymeltzer:
Thanks for catching that; I actually think there was something up with the GH interface, I recall seeing the "will rename when merging" when using the web interface on and off over the past week or so.  Possibly one of my chrome addons, I've got a few GH-specific ones.

PR #815, opened 2020-01-24
Hide space for portlet if unused

### LINKED ISSUE #809: Space for portlet added without portlet
As a non-admin
When I visit https://en.wikipedia.org/wiki/Special:PrefixIndex/User:DannyS712
I should not see the space for the twinkle portlet
Because no portlet is ever added
Because non admins have no functions accessible on special:prefixindex

DannyS712:
Closes #809

### DIFF
--- a/twinkle.js
+++ b/twinkle.js
@@ -483,7 +483,7 @@ Twinkle.load = function () {
 			'.morebits-dialog input, .morebits-dialog select, .morebits-dialog-content button { font-size: inherit !important; }');
 	}
 
-	if ($('#p-twinkle').length === 0) {
+	if (mw.config.get('skin') === 'vector' && Twinkle.getPref('portletType') === 'menu' && $('#p-twinkle').length === 0) {
 		$('#p-cactions').css('margin-right', 'initial');
 	}
 };
--- a/twinkle.js
+++ b/twinkle.js
@@ -482,6 +482,10 @@ Twinkle.load = function () {
 		mw.util.addCSS('.morebits-dialog-content, .morebits-dialog-footerlinks { font-size: 100% !important; } ' +
 			'.morebits-dialog input, .morebits-dialog select, .morebits-dialog-content button { font-size: inherit !important; }');
 	}
+
+	if ($('#p-twinkle').length === 0) {
+		$('#p-cactions').css('margin-right', 'initial');
+	}
 };
 
 }(window, document, jQuery)); // End wrap with anonymous function


Amorymeltzer:
@DannyS712 I asked this in #809, but is there a way to get a visual, on-screen change?

DannyS712:
> @DannyS712 I asked this in #809, but is there a way to get a visual, on-screen change?

What do you mean?

Amorymeltzer:
I mean, it's clear that Twinkle is affecting the `margin-right` of Vector's `p-cactions`, but, as someone who doesn't regularly use Vector, what's a good way to *visually* notice this, like by having other gadgets enabled or something?

twinkle.js
```diff
@@ -482,6 +482,10 @@ Twinkle.load = function () {
 		mw.util.addCSS('.morebits-dialog-content, .morebits-dialog-footerlinks { font-size: 100% !important; } ' +
 			'.morebits-dialog input, .morebits-dialog select, .morebits-dialog-content button { font-size: inherit !important; }');
 	}
+
+	if ($('#p-twinkle').length === 0) {
```
Amorymeltzer: This should work I suppose, but this is only an issue for vector menu users, right?  I don't think having unnecessary `initial` would muck anything up, but is there a reason not to be specific?

```suggestion
	if (mw.config.get('skin') === 'vector' && Twinkle.getPref('portletType') === 'menu' && $('#p-twinkle').length === 0) {
```

Amorymeltzer: I would suppose most anything touching this for non-vectormenu systems to be loaded after Twinkle, but given this whole thing is a result of not wanting the menu to jump, I guess it's more than possible if not entirely likely to interact.  Still, if something's touching this in non-vector, it's probably also touching it in vector, at which point there's been a collision and they've probably fixed it.

siddharthvp:
> @DannyS712 I asked this in #809, but is there a way to get a visual, on-screen change?

Get yourself desysopped <sub>or use a non-admin account</sub> and go to [PrefixIndex in vector skin](https://en.wikipedia.org/wiki/Special:PrefixIndex?useskin=vector).  There's an empty space next to the "More" button. Or am I misunderstanding your question?

siddharthvp:
@Amorymeltzer  raises a very valid point. Unsetting the CSS via JS causes the menu to jump, (which is ironically exactly what the css is supposed to prevent, though in the opposite way). We have to decide between which is a greater annoyance: an empty space where there should have the TW menu, or jumping back of the "More" to its original position?  

At PrefixIndex, the jump isn't much of an issue as there's just the More button. But after #793, the issue could arise on other pages as well (Read, Edit, View history would all jump) for people who choose to disable several modules. 

I wonder what @MusikAnimal thinks about this? He added the css originally.

Amorymeltzer:
Yes that's the reason I ask.  And yes, @siddharthvp, but I don't have *any* menu there in vector on my nonsysop account.  What's in it?  You two are obviously seeing something, but I've got a fairly clean/default gadget list for that account, so I'm assuming you both have put something in that menu that exposes the issue.

siddharthvp:
>  I don't have any menu there in vector on my nonsysop account.

Oh yeah, those items in "More" are from userscripts. You should be able to get one by running `mw.util.addPortletLink('p-cactions', '#', 'BAH');`, or by installing some random script.

Amorymeltzer:
As @siddharthvp noted and I suspected, this means a jump is back in play.  I suppose it's better than an empty space?

Amorymeltzer:
Circling back to this, what do folks think?  I guess jumping is better than a lingering/ghost space?

MusikAnimal:
In Twinkle, the only items that could show up at Special:PrefixIndex are for sysops, right? One possible solution would be to move the `.client-js > body.skin-vector.mw-special-Prefixindex #p-cactions` rule (and the other sysop-only rules, I guess) to a separate hidden peer gadget with `rights=sysop` in the definition.

At any rate, I can't imagine Special:PrefixIndex is super high-traffic, so it's not the end of the world to have empty space or jumping. If we have to chose between the two, I'd go with the latter. This is the way I always did it for [MoreMenu](https://meta.wikimedia.org/wiki/MoreMenu), and I never received complaints. Remember, before the peer gadget stuff went live, everyone had jumping when the TW menu first loaded. So if 99% of the time you don't see jumping (the other 1% being at Special:PrefixIndex), that's still a win.

Amorymeltzer:
Thanks MA — that's not a bad idea for the sysop-only bit, although as @SD0001 points out it'll become more pronounced with #793.  That roughly jives with what I figure, but then again I'm not on Vector so I really don't know how bad the jumping feels!

Amorymeltzer:
Thank you for the contribution. It needed some fixes, so they were made in 5cff2a80cb37a317fcef48af683a3bd45e63254e. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

Amorymeltzer:
Done in 5cff2a8 (though I forgot to fix the typo)

PR #817, opened 2020-01-25
[ARV] Confirm reporting blocked users

### LINKED ISSUE #816: Warn if reporting a blocked user to AIV/UAA
It would be helpful if Twinkle popped up a warning if you try to report a user to UAA or AIV when they're already blocked - I've made that mistake a few times. It's not a critical issue, since I'm pretty sure a bot will come along and clean up the filing, but it would be nice to have.

DannyS712:
Before proceeding to open the form, check if the relevant user is blocked or range blocked. If blocked, request confirmation.
The former "Twinkle.arv.callback" is now at "Twinkle.arv.callback.showform" - other than the portletLink, I didn't see any callers
Closes #816

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -62,7 +62,7 @@ Twinkle.arv.callback = function (username, isIP) {
 		if (confirm(confirmationMessage)) {
 			Twinkle.arv.callback.showform(username);
 		}
-	}, function(err) {
+	}, function() {
 		// If something went wrong, keep going
 		Twinkle.arv.callback.showform(username);
 	});
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -18,14 +18,58 @@ Twinkle.arv = function twinklearv() {
 		return;
 	}
 
-	var title = mw.util.isIPAddress(username) ? 'Report IP to administrators' : 'Report user to administrators';
+	var isIP = mw.util.isIPAddress(username);
+	var title = isIP ? 'Report IP to administrators' : 'Report user to administrators';
 
 	Twinkle.addPortletLink(function() {
-		Twinkle.arv.callback(username);
+		Twinkle.arv.callback(username, isIP);
 	}, 'ARV', 'tw-arv', title);
 };
 
-Twinkle.arv.callback = function (uid) {
+Twinkle.arv.callback = function (username, isIP) {
+	var params = {
+		format: 'json',
+		action: 'query',
+		list: 'blocks',
+		bkprop: 'range'
+	};
+	if (isIP) {
+		params.bkip = username;
+	} else {
+		params.bkusers = username;
+	}
+
+	new mw.Api().get(params).then(function(data) {
+		var blocklist = data.query.blocks;
+		if (blocklist.length === 0) {
+			// No blocks, proceed
+			Twinkle.arv.callback.showform(username);
+		}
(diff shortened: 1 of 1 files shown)


DannyS712:
Testing in developer tools:
* Delete the existing portlet
* Paste in the new code
* Call `Twinkle.load()`
Tested with user blocked, IP blocked, range blocked, and no blocks, works

modules/twinklearv.js
```diff
@@ -18,14 +18,57 @@ Twinkle.arv = function twinklearv() {
...
+		list: 'blocks',
+		bkprop: 'range'
+	};
+	if (isIP) {
+		params.bkip = username;
+	} else {
+		params.bkusers = username;
+	}
+
+	new mw.Api().get(params).then(function(data) {
```
Amorymeltzer: Presumably this would be nicer after #765 ?

DannyS712: Not sure - this is based on code in checking for blocks elsewhere in the repo (don't remember where exactly) but morebits api may not be needed

Amorymeltzer: [twinkleblock](https://github.com/azatoth/twinkle/blob/f1135f6304300eb55624e38ece6e921e209afc4b/modules/twinkleblock.js#L98-L107).  No, it's not needed, but I believe it's the kind of thing that #765 would enable `Morebits.wiki.api` to handle (note: I haven't really reviewed #765).  Not necessarily suggesting you change it, just noting its existence, especially given that it's an odd-one-out.

modules/twinklearv.js
```diff
@@ -18,14 +18,57 @@ Twinkle.arv = function twinklearv() {
...
 	}, 'ARV', 'tw-arv', title);
 };
 
-Twinkle.arv.callback = function (uid) {
+Twinkle.arv.callback = function (username, isIP) {
+	var params = {
+		format: 'json',
+		action: 'query',
+		list: 'blocks',
+		bkprop: 'range'
```
Amorymeltzer: Probably only needs to be added if `isIP`, right?

Amorymeltzer:
Thoughts on the above, @DannyS712?

Beyond that, do we really want to check for every option before loading?  Notifying before reporting to AIV and UAA is good, they'll just get autoremoved otherwise.  I guess AN3 as well, but do we definitely want to do this before SPI reporting?  Especially for the sock option.  I suppose it doesn't hurt much, but seems a waste, especially if it's going to be delaying opening the window.

Just thinking out loud: what about checking after the user has submitted the form?  We could cancel if AIV/UAA, then check/notify for AN3/SPI.  Does that make sense or am I mixed up?

modules/twinklearv.js
DannyS712: from what I recall, this is ignored if its not an IP

siddharthvp:
I too don't think it's good practise to put an API call that delays the opening of the form. One expects the something to open immediately when the they click the menu button. Using the red text option (as in #757) would be better IMO. 

Also, is there a reason why `mw.Api` is used instead of morebits.wiki.api? The latter now supports JSON, so you can simply replace 
```new mw.Api().get(params).then(successFunc, failureFunc)``` with
```new Morebits.wiki.api('Checking if already blocked', params, successFunc, null, failureFunc).post()```

Amorymeltzer:
Done in #1184

PR #819, opened 2020-01-26
twinkle.js: Only load on sysop-only special pages for sysops

Amorymeltzer:
wgRelevantUserName is defined on Special:Block, so it's theoretically useful for everyone (warn, tb, etc.)


Amorymeltzer:
Meh, order is unimportant, they'll each require a fixup from the other, but nothing particularly difficult.

Amorymeltzer:
Gonna wait on #820 until a bunch of others are merged (to reduce collateral conflicts), so will do this first.

PR #788, opened 2019-12-23
speedy/image: Silently append File: prefix on speedy parameters

Amorymeltzer:
The relevant templates (F1 and F8 in twinklespeedy; F5 in twinkleimage) each work either with or without the namespace prefix, but aligning them is a bit more neat, and as discussed in #695, this is somewhat more ideal.

PR #822, opened 2020-01-27
Image: Don't assume a replacement was specified

### LINKED ISSUE #821: Uncaught TypeError: Cannot read property 'value' of undefined
When attempting to tag for No fair use rationale (CSD F6), I get this error instead of actually tagging.
```
Uncaught TypeError: Cannot read property 'value' of undefined
    at HTMLFormElement.twinkleimageCallbackEvaluate (/w/load.php?lang=en&…or&version=kzzt9:27)
    at HTMLButtonElement.<anonymous> (<anonymous>:265:943)
```

DannyS712:
Check for `event.target.replacement` as well.
Follow up #788 
Closes #821

### DIFF
--- a/modules/twinkleimage.js
+++ b/modules/twinkleimage.js
@@ -198,7 +198,7 @@ Twinkle.image.callback.evaluate = function twinkleimageCallbackEvaluate(event) {
 	if (event.target.reason) {
 		reason = event.target.reason.value;
 	}
-	if (event.target.replacement.value && event.target.replacement.value.trim()) {
+	if (event.target.replacement && event.target.replacement.value && event.target.replacement.value.trim()) {
 		replacement = event.target.replacement.value;
 		replacement = /^\s*(Image|File):/i.test(replacement) ? replacement : 'File:' + replacement;
 	}


Amorymeltzer:
Thank you for the contribution. It needed some fixes, so they were made in f1b0573017362f2c36f383bd1e0b50b2f664af3e. Your contribution is still included (and still credited to you), with the appropriate modifications. Please feel free to ask about any of the changes.

DannyS712:
@Amorymeltzer what if replacement exists but not replacement.value?

Amorymeltzer:
Yeah, but I couldn't seem to trigger any errors?  I was admittedly testing somewhat quickly, are you able to?

DannyS712:
Just using the console:
```
var event = { target: { replacement: {} } }

event.target.replacement && event.target.replacement.value && event.target.replacement.value.trim()

event.target.replacement && event.target.replacement.value.trim()
```
First returns undefined, second is `Uncaught TypeError: Cannot read property 'trim' of undefined at <anonymous>:1:60`

Not sure if this could happen with the form, but just to be on the safe side...

Amorymeltzer:
Yeah, couldn't trigger it from the form.  I was surprised, tbh

PR #826, opened 2020-01-27
block: improve partial block menus

siddharthvp:
All this ideally should have been part of #813, but @Amorymeltzer was in a hurry :)

BTW, I think the partial block related changes, especially #825 and this, should be merged now and not in March. Don't wait for people to get used to the pathetic interfaces and [complain when we actually improve it](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&diff=prev&oldid=937707395).

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -430,12 +430,37 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 
 		$form.find('[name=pagerestrictions]').select2({
 			width: '100%',
-			tags: true,
-			placeholder: 'Enter pages to block user from',
+			placeholder: 'Select pages to block user from',
 			maximumSelectionLength: 10, // Software limitation [[phab:T202776]]
-			language: {
-				noResults: function() {
-					return 'No pages entered yet';
+			minimumInputLength: 1, // prevent ajax call when empty
+			ajax: {
+				url: mw.util.wikiScript('api'),
+				dataType: 'json',
+				delay: 100,
+				data: function(params) {
+					var title = mw.Title.newFromText(params.term);
+					if (!title) {
+						return;
+					}
+					return {
+						'action': 'query',
+						'format': 'json',
+						'list': 'allpages',
+						'apfrom': title.title,
+						'apnamespace': title.namespace,
+						'aplimit': '10'
+					};
+				},
+				processResults: function(data) {
+					return {
+						results: data.query.allpages.map(function(page) {
+							var title = mw.Title.newFromText(page.title, page.ns).toText();
+							return {
+								id: title,
+								text: title
+							};
+						})
+					};
 				}
 			}
 		});
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -462,6 +462,12 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 						})
 					};
 				}
+			},
+			templateSelection: function(choice) {
+				return $('<a>').text(choice.text).attr({
+					href: mw.util.getUrl(choice.text),
+					target: '_blank'
+				});
 			}
 		});
 
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -474,6 +474,7 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 
 		$form.find('[name=namespacerestrictions]').select2({
 			width: '100%',
+			matcher: Morebits.select2.matchers.wordBeginning,
 			language: {
 				searching: Morebits.select2.queryInterceptor
 			},
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -431,6 +431,11 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 		$form.find('[name=pagerestrictions]').select2({
 			width: '100%',
 			placeholder: 'Select pages to block user from',
+			language: {
+				errorLoading: function() {
+					return 'Incomplete or invalid search term';
+				}
+			},
 			maximumSelectionLength: 10, // Software limitation [[phab:T202776]]
 			minimumInputLength: 1, // prevent ajax call when empty
 			ajax: {
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -482,7 +482,6 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 			placeholder: 'Select namespaces to block user from'
 		});
 
-		// Reduce padding
 		mw.util.addCSS(
 			// prevent dropdown from appearing behind the dialog, just in case
 			'.select2-container { z-index: 10000; }' +
--- a/morebits.js
+++ b/morebits.js
@@ -1155,23 +1155,36 @@ Morebits.array = {
 /**
  * ************ Morebits.select2 ***************
  * Utilities to enhance select2 menus
- * See twinklewarn and twinklexfd for sample usages
+ * See twinklewarn, twinklexfd, twinkleblock for sample usages
  */
 Morebits.select2 = {
 
-	/**
-	 * Custom matcher in which if the optgroup matches, all options in that group are shown,
-	 * like in jquery.chosen
-	 */
-	matcher: function(params, data) {
-		var originalMatcher = $.fn.select2.defaults.defaults.matcher;
-		var result = originalMatcher(params, data);
+	matchers: {
+		/**
+		 * Custom matcher in which if the optgroup name matches, all options in that
+		 * group are shown, like in jquery.chosen
+		 */
+		optgroupFull: function(params, data) {
+			var originalMatcher = $.fn.select2.defaults.defaults.matcher;
+			var result = originalMatcher(params, data);
 
-		if (result && params.term &&
-			data.text.toUpperCase().indexOf(params.term.toUpperCase()) !== -1) {
-			result.children = data.children;
+			if (result && params.term &&
+				data.text.toUpperCase().indexOf(params.term.toUpperCase()) !== -1) {
+				result.children = data.children;
+			}
+			return result;
+		},
+
+		/** Custom matcher that matches from the beginning of words only */
+		wordBeginning: function(params, data) {
+			var originalMatcher = $.fn.select2.defaults.defaults.matcher;
+			var result = originalMatcher(params, data);
+			if (!params.term || (result &&
+				new RegExp('\\b' + mw.util.escapeRegExp(params.term), 'i').test(result.text))) {
+				return result;
+			}
+			return null;
 		}
-		return result;
 	},
 
 	/** Underline matched part of options */
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1242,7 +1242,7 @@ Twinkle.warn.callback.change_category = function twinklewarnCallbackChangeCatego
 		$('select[name=sub_group]')
 			.select2({
 				width: '100%',
-				matcher: Morebits.select2.matcher,
+				matcher: Morebits.select2.matchers.optgroupFull,
 				templateResult: Morebits.select2.highlightSearchMatches,
 				language: {
 					searching: Morebits.select2.queryInterceptor


Amorymeltzer:
Hah, beat me to the ajax, was gonna do that tonight!  But yes, and yes; I use the milestones as a marker mostly for what should/could be listed in the update, not necessarily when things go out.

modules/twinkleblock.js
```diff
@@ -433,9 +433,37 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
...
-					return 'No pages entered yet';
+			minimumInputLength: 1, // prevent ajax call when empty
+			createTag: function() { // prevent selection of arbitrary titles not returned by API
+				return null;
+			},
+			ajax: {
+				url: mw.util.wikiScript('api'),
+				dataType: 'json',
+				data: function(params) {
+					if (!params.term) {
```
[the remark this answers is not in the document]
Amorymeltzer: Actually, it *would* be good to check that it's not empty; `mw.Title` trims whitespace, but gets cranky if there's *only* whitespace.  I don't think there's a visual issue here, but logs an error.

```suggestion
					if (!params.term.trim()) {
```
Or something like it.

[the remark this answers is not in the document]
Amorymeltzer: Actually actually, it might be be better to just verify that `mw.Title` was successful below, as it also throws an error when you've got just the namespace entered (e.g. "User:").  Again, nothing visual in the module AFAICT, but a console error.

modules/twinkleblock.js
```diff
@@ -433,9 +433,37 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
...
+						return;
+					}
+					var title = new mw.Title(params.term);
```
Amorymeltzer: Per comment above, might be good to verify this is successful.

modules/twinkleblock.js
```diff
@@ -465,6 +465,12 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 						})
 					};
 				}
+			},
+			templateSelection: function(choice) {
```
Amorymeltzer: This is just cool as fuck! 😎

modules/twinkleblock.js
```diff
@@ -477,6 +477,15 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 
 		$form.find('[name=namespacerestrictions]').select2({
 			width: '100%',
+			matcher: function(params, data) {
```
Amorymeltzer: Very nice!  Much better, this has been bugging me.

modules/twinkleblock.js
```diff
@@ -433,9 +433,37 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
...
 			maximumSelectionLength: 10, // Software limitation [[phab:T202776]]
-			language: {
-				noResults: function() {
-					return 'No pages entered yet';
+			minimumInputLength: 1, // prevent ajax call when empty
+			createTag: function() { // prevent selection of arbitrary titles not returned by API
+				return null;
+			},
+			ajax: {
+				url: mw.util.wikiScript('api'),
```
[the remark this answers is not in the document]
Amorymeltzer: 75 or 100, maybe?

modules/twinkleblock.js
```diff
@@ -433,9 +433,37 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 			tags: true,
 			placeholder: 'Enter pages to block user from',
 			maximumSelectionLength: 10, // Software limitation [[phab:T202776]]
-			language: {
-				noResults: function() {
-					return 'No pages entered yet';
+			minimumInputLength: 1, // prevent ajax call when empty
+			createTag: function() { // prevent selection of arbitrary titles not returned by API
```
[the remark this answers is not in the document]
siddharthvp: Dunno what happened. I was trying to convert my [StubSorter](https://en.wikipedia.org/wiki/User:SD0001/StubSorter.js) to use select2 (currently implemented in chosen with One Big Hack for ajax lookups), but the ajax thing wasn't working at all. I left it. Later, on trying out the same here, it worked, and I had assumed that it was because of `tags: true` which I hadn't put in StubSorter. 

But you're right. It works even without it. `tags: true` and `createTag: function() { return null; }` appear to be cancelling each other out. Both appear to be unnecessary.

modules/twinkleblock.js
```diff
@@ -433,9 +433,37 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 			tags: true,
 			placeholder: 'Enter pages to block user from',
 			maximumSelectionLength: 10, // Software limitation [[phab:T202776]]
-			language: {
-				noResults: function() {
-					return 'No pages entered yet';
+			minimumInputLength: 1, // prevent ajax call when empty
```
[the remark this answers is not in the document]
siddharthvp: 1 is the pretty much the universal default on such input fields. And also since we aren't allowing selection of titles not returned by the API, that would make  it impossible to select pages with short names.

modules/twinkleblock.js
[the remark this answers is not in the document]
siddharthvp: > Is this necessary? When does select2 ever return an empty params?

When you first click on it, when there's nothing in the search field (I'd've expected the library to supply `params.term = ''` in such a case, but it doesn't). But ah, I did add `minimumInputLength: 1` to handle this better (doesn't give "Search results could not be loaded" text), so yeah this check is unnecessary now. You're right again 👌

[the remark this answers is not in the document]
siddharthvp: The trouble with `mw.Title` is that it throws. We've to use `mw.Title.newFromText` (which is exception-safe) to prevent console errors.

modules/twinkleblock.js
siddharthvp: Unlike above, this is the output returned by the API. Are there any legal titles that can't be parsed by `mw.Title` ? If so, I'd guess that needs to fixed MediaWiki-side. 

If there are such titles, I guess the fix would be return an empty object `{}` from the map (returning `null` or `undefined` results in errors).

modules/twinkleblock.js
siddharthvp: Looks like select2 has got an option for almost everything :)

modules/twinkleblock.js
siddharthvp: Should we put this in morebits? This is very much reusable.

siddharthvp:
More things to think about

- For the ajax lookups, should we use API:Search or API:Allpages? The former didn't occur to me earlier. But it should ignore case and do auto-correct and things like that. Results can be sorted by `relevance` or `just_match`. But API:Allpages makes it easy to get at the exact title if typing correctly. I guess most MediaWiki page selector menus use something like API:Search, so people would be used to that. 
- Should we add `mw.util.addCSS('.select2-results__option.loading-results { display: none; }');` ? This hides the "Searching..." text that appears on every keystroke in page selector field. Or is it useful?
- Across all multiselect fields, hitting backspace doesn't remove the last selected element. Rather it converts the option to text. This is useful in some applications (StubSorter) but I think all Twinkle fields are better with backspace removing the whole option? I've filed a patch in select2 to make this feature available as an option (https://github.com/select2/select2/pull/5752). But presently doing so  requires putting in the following hack: (https://github.com/select2/select2/issues/3354#issuecomment-132389291):
```js
$.fn.select2.amd.require(['select2/selection/search'], function (Search) {
    var oldRemoveChoice = Search.prototype.searchRemoveChoice;
    Search.prototype.searchRemoveChoice = function () {
        oldRemoveChoice.apply(this, arguments);
        this.$search.val('');
    };
});
```

modules/twinkleblock.js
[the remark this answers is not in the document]
Amorymeltzer: Fair enough, and I like the delay option better anyway.

modules/twinkleblock.js
Amorymeltzer: If we're not going to reuse it not a huge need, but couldn't hurt; should be broadly applicable enough.

modules/twinkleblock.js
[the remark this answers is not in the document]
Amorymeltzer: `tags` is just for user-generated input, and would be useful if nonexistent pages were allowed (hence, I mistakenly used it initially) but yeah, no need here.

modules/twinkleblock.js
Amorymeltzer: This comment was just in reference to https://github.com/azatoth/twinkle/pull/826#discussion_r371401798, although I thought I had a note about the one in `processResults` but 1. can't find it and 2. you dealt with it

Amorymeltzer:
Can't test until tonight but:

1. I was working on this myself when you posted this, so I stopped, but I was thinking to use Search, mainly for relevance sorting.  I was having trouble getting the generator to return results when there wasn't an exact match, but ymmv.  In theory I think it'd be better.  Why do you say `API:Allpages makes it easy to get at the exact title if typing correctly`?
2. Nah, I think it's quite helpful.

siddharthvp:
A quick try doesn't seem to be working. Native title searching (`srwhat=title`) isn't supported by CirrusSearch; we've got to use text search queries - either `prefix:"foo"` or `intitle:"foo"` but both seem to giving output that's inconsistent with what the Special:Search dropdown suggestions give. For example, [a search for something random like "jolly"](https://en.wikipedia.org/w/api.php?action=query&list=search&srsearch=prefix%3A%22jolly%22&srnamespace=*&srlimit=10&srinfo=&srprop=) doesn't give the page [Jolly](https://en.wikipedia.org/w/index.php?title=Jolly&redirect=no) as the first result, _or even among the first 10 results_!

> I was working on this myself when you posted this, so I stopped,

Ohhh gosh, why I did I have to put myself through this then!

Since you've write access to this branch and if you're able to get the search API to make sense, you can just push into this branch (either as a new commit or overwriting the first commit  - I don't mind).

Amorymeltzer:
Oh no I meant my comment to mean that I got stuck on the search API, much like it seems you did, around `gsrwhat`.  I do think search would indeed be better, but certainly not a priority atm with AllPages working quite well enough.

Regarding the backspace bit, I'm neither here nor there: I like the ability and find it useful, but I can imagine folks moving quickly might like the ability to "edit" previous entries.  It's possible to get weird results with AllPages, especially if pages don't exist or if results aren't intuitive (e.g., capitalization or punctuation); I've made plenty of "mistakes" while testing.

modules/twinkleblock.js
```diff
@@ -430,27 +430,67 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
...
+				}
+				return null;
+			},
 			language: {
 				searching: Morebits.select2.queryInterceptor
 			},
 			templateResult: Morebits.select2.highlightSearchMatches,
-			placeholder: 'Select namespaces to block user from'
+			placeholder: 'Select namespaces to block user from',
+			closeOnSelect: false
```
Amorymeltzer: Not entirely sure about this.  It's definitely a positive and makes things easier if you're using the cursor, but for someone using the keyboard, I think it's a bit of a step backward.  The text you entered to find a match lingers after you've selected that option: that means that you have to delete it before typing your next namespace.  It also means the best way to close the menu is to hit tab, or enter if you're on an option you already selected, which isn't particularly intuitive.

My instinct is that we're likely not to see many multiple-namespace blocks, or at least it won't be the norm, so could probably be left off.

modules/twinkleblock.js
```diff
@@ -430,27 +430,67 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 
 		$form.find('[name=pagerestrictions]').select2({
 			width: '100%',
-			tags: true,
-			placeholder: 'Enter pages to block user from',
+			placeholder: 'Select pages to block user from',
```
Amorymeltzer:
```suggestion
			placeholder: 'Select pages to block user from',
			language: {
				errorLoading: function() { return 'Incomplete or invalid search term'; }
			},
```

Or something like that.  Any page outside the article namespace will trigger this message; the delay should help, but if there's any pause after the colon, the user will see "The results could not be loaded", which (to me anyway) suggests an error on Twinkle's part.  It'd be good to cover the (slim) chance that an invalid search term gets entered (` `, `(`, `[`, etc.) as well.

modules/twinkleblock.js
siddharthvp: Done.

morebits.js
```diff
@@ -1155,23 +1155,36 @@ Morebits.array = {
...
-	 */
-	matcher: function(params, data) {
-		var originalMatcher = $.fn.select2.defaults.defaults.matcher;
-		var result = originalMatcher(params, data);
+	matchers: {
+		/**
+		 * Custom matcher in which if the optgroup name matches, all options in that
+		 * group are shown, like in jquery.chosen
+		 */
+		optgroupFull: function(params, data) {
```
siddharthvp: Pathetic name. But couldn't think of anything better.

Amorymeltzer:
Will deploy in a bit, folks will appreciate all this work, thanks @siddharthvp!

Xi-Plus:
https://github.com/azatoth/twinkle/pull/826#issuecomment-579188478
> * For the ajax lookups, should we use API:Search or API:Allpages? 

Why don't use `prefixsearch`. Special:Block use it. ([reference](https://gerrit.wikimedia.org/g/mediawiki/core/+/56fd901bc002182ae471f7d89157e58e4e91ebae/resources/src/mediawiki.widgets/mw.widgets.TitleWidget.js#199))

Amorymeltzer:
@Xi-Plus In my testing, `prefixsearch` was a lot less flexible for finding non-exact matches.  That is, it was good for finding the fourth and fifth letters if you typed the right first three, but not otherwise.  More importantly, though, `prefixsearch` can include non-existent pages, in particular those from Commons.

Xi-Plus:
Okay, I got it.

Issue #151, opened 2013-04-23
Make sure Twinkle cannot be used by non-autoconfirmed users


atlight:
Currently, new users can use the old Friendly modules, and possibly some others. Non-autoconfirmed/non-confirmed users should be denied access to all modules (except perhaps diff).

Should the handling of this be centralised to twinkle.footer.js? Do the Twinkle tabs even need to be displayed when a user can't use them?

Also, suggesting that MediaWiki:Gadgets-definition be updated to include `|rights=autoconfirmed` inside our square brackets. Amalthea, assigning you to this one :)

PR #828, opened 2020-01-27
Remove Twinkle.userAuthorized

Amorymeltzer:
It's not really serving much of a purpose: used only in `twinkle.js` and `twinklefluff.js`.  Consolidated in 6dc2a75 (2013) as part of #151, where the requirement was added to the gadget definition.  It used to be checked in each module (ad5af2f, 2011) but is now left just in `fluff`; there's some sense to that, though, as `fluff` is the only one that doesn't add a menu button (see also #792).

The reality, though, is that `Twinkle.userAuthorized = false` is not much of a "security" check, all the moreso because it came in `Twinkle.load`.  Moving the check higher up, before the Twinkle object is declared, makes things somewhat more tricky.  We could be slightly trickier, forgoing `Morebits` and doing:

`!(mw.config.get('wgUserGroups').indexOf('autoconfirmed')!==-1 || mw.config.get('wgUserGroups').indexOf('confirmed')!==-1)`

but the reality is that anyone trying to get past it will have no difficult doing that as well; indeed, anyone seriously attempting this far would probably just copy/write their own.  Anything more complicated is probably not worth doing.

I've removed the check in `fluff` but not replaced it with anything; thus it's possible for someone to somewhat easily get the links, but, as above not particularly difficult or protective.

I've also removed the message in `Twinkle.addPortlet` as I couldn't trigger the message, so not sure about that.

----

Opened as a draft as this is mostly me thinking out loud and not, like, a major issue; mostly I just think it's currently less effective and (slightly) more complicated than need be.  Still, I welcome other thoughts.  Another idea is to have Twinkle set a requirement, which is checked by Morebits for each API query.  That seems overkill by far.

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -17,40 +17,38 @@
  */
 
 Twinkle.fluff = function twinklefluff() {
-	if (Twinkle.userAuthorized) {
-		// A list of usernames, usually only bots, that vandalism revert is jumped over; that is,
-		// if vandalism revert was chosen on such username, then its target is on the revision before.
-		// This is for handling quick bots that makes edits seconds after the original edit is made.
-		// This only affects vandalism rollback; for good faith rollback, it will stop, indicating a bot
-		// has no faith, and for normal rollback, it will rollback that edit.
-		Twinkle.fluff.whiteList = [
-			'AnomieBOT',
-			'SineBot'
-		];
+	// A list of usernames, usually only bots, that vandalism revert is jumped over; that is,
+	// if vandalism revert was chosen on such username, then its target is on the revision before.
+	// This is for handling quick bots that makes edits seconds after the original edit is made.
+	// This only affects vandalism rollback; for good faith rollback, it will stop, indicating a bot
+	// has no faith, and for normal rollback, it will rollback that edit.
+	Twinkle.fluff.whiteList = [
+		'AnomieBOT',
+		'SineBot'
+	];
 
-		if (mw.util.getParamValue('twinklerevert')) {
-			// Return if the user can't edit the page in question
-			if (!mw.config.get('wgIsProbablyEditable')) {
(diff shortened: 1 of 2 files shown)


siddharthvp:
Woah that's a lot of words and research for something so trivial. All I'll say is since non-confirmed users can't even enable the gadget from their preferences (enforced by gadget definition),so the check isn't even necessary. The only way they can load Twinkle is to write an importScript statement for morebits, for the main file and for every module they want to use. Sounds a like lot more pain than just doing 10 edits and waiting for 4 days.

Amorymeltzer:
> Woah that's a lot of words and research for something so trivial.

I'll take that as a [compliment](https://fatbusinessman.com/2019/my-favourite-git-commit)!

> All I'll say is that since non-confirmed users can't even enable the gadget from their preferences (enforced by gadget definition), the check isn't even necessary.

Yes, it's mostly just a "make things slightly harder for your run-o-the-mill-vandal" check, but not much more than that.  The rest of your statement isn't actually true[0]; it's much easier than that right now, but making this change would largely necessitate doing that sort of trade-off.

[0]: No need to go into detail, but fairly trivial.

siddharthvp:
> The rest of your statement isn't actually true

Really? Fairly trivial? _How?_ What am I missing?

Amorymeltzer:
The gadget definition stops it from showing up in preferences, but doesn't disable the gadget itself.

siddharthvp:
Hmm yeah, I thought mw.loader.using('ext.gadget. ... ') would also check the gadget def, apparently it doesn't.

Amorymeltzer:
Yeah, I don't know much about how it works on the backend, but it does seem
incongruent for the `mw.loader` series not to check definitions.  I think
it's part of the fabled Gadgets 2.0.

~A

PR #829, opened 2020-01-28
 tag: fix broken discussion links in merge tags

siddharthvp:
Solves bug reported on [TW talk](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Automated_%22merge_from%22) - the talk page section title used in the template's discuss link should match the one actually produced. 

Also, adjusted the edit summary to be a section link to the discussion (kinda fake section=new API action). This seems to be a better alternative that no thought of in #717 where we discussed it earlier, @Amorymeltzer.

The only bug left is that TW unnecessarily adds `\n\n` (visible extra whitespace) if the talk page was empty (also there in some other modules - talkback I think). But I think there's no easy fix short of adding `section=new` support into morebits, or not adding `\n\n` at all (leads to untidy wikicode on non-new pages).

The 2nd commit does away with `params.mergeTarget` check which isn't required as the field is compulsory (since #545). 

Tests:
- Tagging {{merge}}: [source](https://test.wikipedia.org/w/index.php?title=SD0001test&diff=next&oldid=413427), [destination](https://test.wikipedia.org/w/index.php?title=SD0001test2&diff=next&oldid=413428), [source talk](https://test.wikipedia.org/w/index.php?title=Talk:SD0001test&diff=next&oldid=413425)
- Tagging {{merge from}}: [source](https://test.wikipedia.org/w/index.php?title=SD0001test&diff=413441&oldid=413439), [destination](https://test.wikipedia.org/w/index.php?title=SD0001test2&diff=413443&oldid=413440), [source talk](https://test.wikipedia.org/w/index.php?title=Talk:SD0001test&diff=413442&oldid=413430)
- Tagging {{merge to}}: [source](https://test.wikipedia.org/w/index.php?title=SD0001test&diff=413435&oldid=413432), [destination](https://test.wikipedia.org/w/index.php?title=SD0001test2&diff=413437&oldid=413434), [destination talk](https://test.wikipedia.org/w/index.php?title=Talk:SD0001test2&diff=413436&oldid=404636)

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1527,24 +1527,22 @@ Twinkle.tag.callbacks = {
 					case 'Merge to':
 					case 'Merge from':
 						params.mergeTag = tagName;
-						if (params.mergeTarget) {
-							// normalize the merge target for now and later
-							params.mergeTarget = Morebits.string.toUpperCaseFirstChar(params.mergeTarget.replace(/_/g, ' '));
+						// normalize the merge target for now and later
+						params.mergeTarget = Morebits.string.toUpperCaseFirstChar(params.mergeTarget.replace(/_/g, ' '));
 
-							currentTag += '|' + params.mergeTarget;
+						currentTag += '|' + params.mergeTarget;
 
-							// link to the correct section on the talk page, for article space only
-							if (mw.config.get('wgNamespaceNumber') === 0 && (params.mergeReason || params.discussArticle)) {
-								if (!params.discussArticle) {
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Gonna push this up now.

siddharthvp:
> Why do away with the links?

Ahh, only saw that now. The link syntax was breaking the section link in the edit summary. I somehow forgot that links were still needed for the section header themselves. Probably the fix would be to do a `.replace(/\[\[.*?\]\]/g, '$1')` within the edit summary.

PR #835, opened 2020-01-30
morebits: Allow addFooterLink to prepend rather than append

Amorymeltzer:
1st commit: Lets us use `addFooterLink` to prepend if desired.

2nd commit: Does that in block to link to Special:Unblock/User if the user is blocked.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -28,6 +28,10 @@ Twinkle.block.callback = function twinkleblockCallback() {
 		return;
 	}
 
+	Twinkle.block.currentBlockInfo = undefined;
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
In hindsight there are a lot of places I could have put the bit in the second commit, not sure I picked the most sensible one.  Maybe in `Twinkle.block.fetchUserInfo` would make more sense?

Amorymeltzer:
Another thought would be to put it inside the actual `Twinkle.block.fetchUserInfo` call.  That'd require an additional `if (Twinkle.block.currentBlockInfo)` but it would mean not relying on something like `Twinkle.block.Window = Window`, which I think would be better overall.

PR #836, opened 2020-02-01
xfd: Make target required for tfm/cfd/cfds, and specify action in cfd-notify

Amorymeltzer:
1. Adds the `required` attribute to `input` and `textarea` `quickForm` elements in `Morebits`
1. Uses that to require users to enter an input for TfM, CfM/CfR/CfS/CfC (i.e. WP:CFD but not deletion), and all CfDS.
1. Adds the `action` parameter to the {{cfd-notify}} template, and uses that language a bit more throughout

I opened this as a draft pending a response to my [question at WT:CFD](https://en.wikipedia.org/wiki/Wikipedia_talk:Categories_for_discussion#Should_Twinkle_require_nominators_to_input_new/other_category_names?) about point 2 above: I think it's a good idea, but I could theoretically imagine that some folks at CfD like being able to make nominations without the target specified.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -131,7 +131,7 @@ Morebits.sanitizeIPv6 = function (address) {
  *              - Attributes: name, list, event
  *              - Attributes (within list): name, label, value, checked, disabled, event, subgroup
  *   input     A text box.
- *              - Attributes: name, label, value, size, disabled, readonly, maxlength, event
+ *              - Attributes: name, label, value, size, disabled, required, readonly, maxlength, event
  *   dyninput  A set of text boxes with "Remove" buttons and an "Add" button.
  *              - Attributes: name, label, min, max, sublabel, value, size, maxlength, event
  *   hidden    An invisible form field.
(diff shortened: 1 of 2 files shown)


morebits.js
```diff
@@ -456,6 +456,9 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 			if (data.disabled) {
 				subnode.setAttribute('disabled', 'disabled');
 			}
+			if (data.required) {
+				subnode.setAttribute('required', true);
```
```suggestion
				subnode.setAttribute('required', 'required');
```
Or `subnode.setAttribute('required', '');` or `subnode.required = true`. Setting the value to true [is invalid in HTML5](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes).

morebits.js
```diff
@@ -646,6 +649,9 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
...
```
```suggestion
				subnode.setAttribute('required', 'required');
```

morebits.js
Amorymeltzer: Right you are, thanks!

PR #839, opened 2020-02-02
xfd: Add XfD log, parallel to CSD and PROD

Amorymeltzer:
There's not the same reason to log XfD nominations as there is for CSDs and PRODs, but this is useful/interesting regardless; a few folks have asked for it or though it would be useful.  It is also a lot more complicated.

The basics:
- Adds config options (logXfdNominations and xfdLogPageName)
- Lots of venue-specific log formatting (`toTLACase`, etc.)
- Like PROD/CSD, checks whether user was notified (although see below).  CfDS and RM have no user notification.
- Includes a link to the page log for files
- Stores common parameters and makes use of `.extend`

Nitty-gritty:
There are real complications around several venues, in particular confirmations on whether something actually happened and what information is available.  Most are dealt with, some aren't.  At a very basic level, we're just simply not going to check the success of each and every deletion sorting listing before adding to the XfD log, which gives some leeway for being lax on other parameters.  To wit:

- RfD logs the notification of just the initial contributer, not the target page (although it does log the redirect's target)
- MfDs of userspace pages created by a different user will log both names, but only checks if the notification to the creator was successful
- TfD merges (TfM) only log the creator of the first page, and can't currently note whether the second template's creator was notified or not (it *does* log the name of that second page, though)

----

Open to suggestions about better to ways to do some of these; this was put together fairly quickly and though I tested each option fairly well, there are bound to be mistakes.

For TfMs, to deal with the above, we could just assume the second creator (if there is one) was notified and add a generic note like '(notified creator)'.  <s>Additionally, after #836, CfD should change `'Target:'` to `params.action + ' to:'`.</s> Done

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -380,7 +380,7 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 					}
 				]
 			});
-			if (mw.config.get('wgNamespaceNumber') === 2 /* User: */ || mw.config.get('wgNamespaceNumber') === 3 /* User talk: */) {
+			if ((mw.config.get('wgNamespaceNumber') === 2 /* User: */ || mw.config.get('wgNamespaceNumber') === 3 /* User talk: */) && mw.config.exists('wgRelevantUserName')) {
 				work_area.append({
 					type: 'checkbox',
 					list: [
@@ -388,7 +388,7 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 							label: 'Also notify owner of userspace if they are not the page creator',
 							value: 'notifyuserspace',
 							name: 'notifyuserspace',
-							tooltip: 'If the user in whose userspace this page is located, is not the page creator (for example, the page is a rescued article stored as a userspace draft), notify the userspace owner as well.',
+							tooltip: 'If the user in whose userspace this page is located is not the page creator (for example, the page is a rescued article stored as a userspace draft), notify the userspace owner as well.',
 							checked: true
 						}
 					]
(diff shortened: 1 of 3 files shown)


modules/twinklexfd.js
```diff
@@ -689,6 +689,147 @@ Twinkle.xfd.callbacks = {
...
+		}
+		if (extraInfo) {
+			appendText += extraInfo;
+		}
+		appendText += ' ~~~~~';
+		if (params.reason) {
+			appendText += "\n#* '''Reason''': " + params.reason + '\n';
+		}
+
+
```
DannyS712: Shouldn't have two blank lines in a row

twinkle.js
```diff
@@ -110,6 +110,8 @@ Twinkle.defaultConfig = {
 	autoMenuAfterRollback: false,
 
 	// XfD
+	logXfdNominations: false,
+	xfdLogPageName: 'XFD log',
```
DannyS712: Suggest `XfD log` instead

Amorymeltzer: Went back and forth on it, never know how many folks prefer the (over-?)stylized version, but I do.  It doesn't help that Twinkle is somewhat inconsistent, although we do lean toward XfD.

siddharthvp:
I am very much in support of this. Indeed the only reason this feature request was "declined" in the past was apparently due to no one having the time to code it. Thanks for undertaking this!

modules/twinklexfd.js
```diff
@@ -1076,21 +1244,24 @@ Twinkle.xfd.callbacks = {
 			var initialContrib = pageobj.getCreator();
 			var params = pageobj.getCallbackParameters();
 
+			// Also notify the user who owns the subpage if they are not the creator
+			params.userspaceOwner = mw.config.get('wgTitle').indexOf('/') === -1 ? mw.config.get('wgTitle') : mw.config.get('wgTitle').substring(0, mw.config.get('wgTitle').indexOf('/'));
```
siddharthvp:
```suggestion
			params.userspaceOwner = mw.config.get('wgTitle').split('/')[0];
```

modules/twinklexfd.js
```diff
@@ -1076,21 +1244,24 @@ Twinkle.xfd.callbacks = {
...
+			if (params.notifyuserspace) {
+				if (params.userspaceOwner !== initialContrib) {
```
siddharthvp: `if`'s can be combined.

modules/twinklexfd.js
```diff
@@ -1076,21 +1244,24 @@ Twinkle.xfd.callbacks = {
...
-
-			// Also notify the user who owns the subpage if they are not the creator
-			if (params.notifyuserspace) {
-				var userspaceOwner = mw.config.get('wgTitle').indexOf('/') === -1 ? mw.config.get('wgTitle') : mw.config.get('wgTitle').substring(0, mw.config.get('wgTitle').indexOf('/'));
-				if (userspaceOwner !== initialContrib) {
-					Twinkle.xfd.callbacks.mfd.userNotificationMain(params, userspaceOwner, 'Notifying owner of userspace');
-				}
-			}
 		},
 		userNotificationMain: function(params, initialContrib, actionName) {
```
siddharthvp: parameter name `initialContrib` should be changed (maybe to `user`) as it could refer to the initial contributor or userspace owner.

modules/twinklexfd.js
```diff
@@ -1437,7 +1660,12 @@ Twinkle.xfd.callbacks = {
 			targettalkpage.setCreateOption('recreate');
 			Twinkle.xfd.setWatchPref(targettalkpage, Twinkle.getPref('xfdWatchRelated'));
 			targettalkpage.setFollowRedirect(true);
-			targettalkpage.append();
+			// Add to userspace log even if not notifying the creator
+			if (params.lognomination && !params.usertalk) {
+				targettalkpage.append(Twinkle.xfd.callbacks.addToLog(params, null));
```
siddharthvp: Syntax error.

modules/twinklexfd.js
```diff
@@ -689,6 +689,147 @@ Twinkle.xfd.callbacks = {
...
+					extraInfo += '; [[WP:AFDC|Category]]: ' + afdCats[params.xfdcat];
+				}
+				if (params.delsort_cats && params.delsort_cats.length) {
+					extraInfo += '; [[WP:DELSORT|Delsorting]]: ';
+					var delsort_cats = params.delsort_cats;
+					delsort_cats.forEach(function(cat, idx) {
+						delsort_cats[idx] = '[[Wikipedia:WikiProject Deletion sorting/' + cat + '|' + cat + ']]';
+					});
+					extraInfo += delsort_cats.join(', ');
+				}
```
siddharthvp: Pretty sure all this information isn't worth logging. These would make for a lot of clutter in a XfD log serving little purpose. 

Both of these (esp the latter) are frequently edited by delsort wikignomes after nomination is started (which won't get recorded in the log anyway).

As for `xfdcat`, roughly half of the users don't even bother with filling it (see the counts [here](https://en.wikipedia.org/wiki/Category:AfD_debates)). It is only used for a very broad categorisation of AfDs which I doubt anyone finds useful.

siddharthvp:
Can't the way in which `Twinkle.xfd.callbacks.addToLog` gets called be made less hacky? I see it's adapted from 80767a72 but I've never been fond of that. I think we ought to implement some TaskManager class (like in [XFDcloser](https://en.wikipedia.org/w/index.php?title=MediaWiki:Gadget-XFDcloser-core.js)) to handle sequences of operations like this without uglifying the code. XFDcloser does a zillion tasks but the code for each is completely separated.

modules/twinklexfd.js
Amorymeltzer: Even better: this was presumably added before `wgRelevantUserName` was available, we should just use that.

modules/twinklexfd.js
Amorymeltzer: Yeah, I don't really disagree.  Mostly put this all in as I'd rather have it all and remove than backfill it later.  It's only potentially interesting, but not hugely useful, especially given the XfD exists so there's a history of it.  Prob excessive here.

modules/twinklexfd.js
Amorymeltzer: Went for `userTarget`, but yeah, that'd be clearer.

modules/twinklexfd.js
Amorymeltzer: Also tossed in a `mw.config.exists('wgRelevantUserName')` for the `notifyuserspace` option, should now ensure that the user actually exists...

modules/twinklexfd.js
Amorymeltzer: Where?

Amorymeltzer:
I'll get to them eventually... big rewrites are hard to test, to put it mildly, in particular for off-target effects.

modules/twinklexfd.js
siddharthvp: Lol, it should be `targettalkpage.append(function() { Twinkle.xfd.callbacks.addToLog(params, null)); });`  

Otherwise `addToLog` is evaluated immediately, not after the `append` has succeeded. No console error is raised as `addToLog` returns undefined, so morebits just thinks `append` did not get any argument.

Amorymeltzer: Haha oh geez, silly to miss that 🤕

modules/twinklexfd.js
Amorymeltzer: Removed

twinkle.js
Amorymeltzer: Folks who got back to me felt the same as us, so changed!

siddharthvp:
The way the code is written, I guess you're also going to add the config option for selecting XfD boards in which to enable logging? This seems quite expected since CSD allows enabling/disabling it for every criteria.

I was going to suggest that since `params.lognomiation` (= `Twinkle.getPref('logXfdNominations')`) is globally accessible, the `if (params.lognomination)` check doesn't need to be written out 20 times, it can be handled, just once, within `addToLog`. But this is not applicable if that config option is there.

Amorymeltzer:
@siddharthvp Hadn't even considered that!  Seems crazy to me but yes, I think you're absolutely right that it ought to be included.

Amorymeltzer:
Updated to use `Morebits.userspaceLogger`

PR #842, opened 2020-02-09
xfd: Update technical RM hidden comment

DannyS712:
Onwiki comment updated in [1], per discussion at [2]; the current value results in the request being placed in the `Requests to revert undiscussed moves` section instead of the intended `Uncontroversial technical requests` section[3]
[1] Diff: https://w.wiki/GtM
[2] Discussion: https://w.wiki/GtN (permalink to current: https://w.wiki/GtP)
[3] Diff: https://w.wiki/GtQ

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -1460,7 +1460,7 @@ Twinkle.xfd.callbacks = {
 			var params = pageobj.getCallbackParameters();
 			var statelem = pageobj.getStatusElement();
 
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Thanks @DannyS712 — this is getting annoying.  I really don't want to leave a twinkle message, but if folks are gonna keep changing it (unlike every xfd...) then maybe we need to go for it.

PR #844, opened 2020-02-10
speedy: Adjust default tag mode for sysops

Amorymeltzer:
Requested by the community following discussion after an [arb case](https://en.wikipedia.org/wiki/Wikipedia:Arbitration/Requests/Case/RHaworth).  `deleteSysopDefaultToTag` is used by around 50 individuals, and there was some local consensus to have Twinkle default to tagging for sysops if there's no CSD tag already present.  Accordingly:

- `deleteSysopDefaultToTag` (default `false`) is replaced by `deleteSysopDefaultToDelete` (default false), reverting the default behavior for all sysops; sysops will now default to "tag" mode (if there's no CSD present, see below).  Users must manually opt for the previous default behavior, to default to "delete" mode.
- If there is a deletion tag present, Twinkle will now always default to "delete" mode.  This isn't technically required, but makes sense with the above, and is probably just a good idea in general.  We were using it for some behaviors already, so this makes use of `Twinkle.speedy.hasCSD` throughout since we're using `$('#delete-reason')` quite a bit now.

----

Is this the kind of thing that should update `optionsVersion`?  Doing 2.1 seems overly precious, I don't really feel like instituting semver there... :lol:

I had a broken url in my notes so I've lost the location of the discussion thread since I wrote this a week or two ago.

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -480,11 +480,12 @@ Twinkle.config.sections = [
 				type: 'boolean'
 			},
 
-			// TwinkleConfig.deleteSysopDefaultToTag (boolean)
-			// Make the CSD screen default to "tag" instead of "delete" (admin only)
+			// TwinkleConfig.deleteSysopDefaultToDelete (boolean)
+			// Make the CSD screen default to "delete" instead of "tag" (admin only)
 			{
-				name: 'deleteSysopDefaultToTag',
-				label: 'Default to speedy tagging instead of outright deletion',
+				name: 'deleteSysopDefaultToDelete',
+				label: 'Default to outright deletion instead of speedy tagging',
+				helptip: 'If there is a CSD tag already present, Twinkle will always default to "delete" mode',
 				adminOnly: true,
 				type: 'boolean'
 			},
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -36,6 +36,8 @@ Twinkle.speedy.callback = function twinklespeedyCallback() {
 
 // Used by unlink feature
 Twinkle.speedy.dialog = null;
+// Used throughout
+Twinkle.speedy.hasCSD = !!$('#delete-reason').length;
 
 // The speedy criteria list can be in one of several modes
 Twinkle.speedy.mode = {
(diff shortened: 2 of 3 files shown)


siddharthvp:
> Is this the kind of thing that should update optionsVersion? Doing 2.1 seems overly precious, I don't really feel like instituting semver there... :lol:

Not at all. Only need to bump the optionsVersion when making a change to the way in which options are being stored or retrieved, not when adding or removing individual options.

We can add some code to twinkleconfig to get the deprecated option removed from peoples' pref pages. I looked into this long back so here's exactly needs doing:

Add `var deprecatedPrefs = [ ... ];` before [L1649](https://github.com/azatoth/twinkle/blob/ec942334a684cfed0ca4a517138f1fbec685e56d/modules/twinkleconfig.js#L1649), and change the condition in L1652 to `if (foundPrefs.indexOf(tkey) === -1 && deprecatedPrefs.indexOf(tkey) === -1)`

Amorymeltzer:
(heh, Was mostly joking on that point, although theoretically the versioning system could handle them individually; I mooted on this in #520 but never did anything about it.  I think it'd be ideal not to have to list every deprecated option, but it should be infrequent I guess.  There's some funkiness around saving bespoke options I've been meaning to look into, not sure if it's related/would cause hardships.)

Amorymeltzer:
I finally found the [thread](http://en.wikipedia.org/wiki/Special:Permalink/940544517#Twinkle_and_speedies) and saw your summary @siddharthvp; sorry if I pre-empted your work on this!  I liked your suggestion to rename the buttons, though, that was good, so I've added a commit to do just that.

siddharthvp:
> sorry if I pre-empted your work on this! 

I didn't actually plan to do anything on this as I figured you'd be on it :) 

This looks fine to me. Only thing I didn't expect was the renaming of the option. I'd have just changed the default value of the existing option as it would have had the same effect, without having to worry about what to do with a deprecated option.

Amorymeltzer:
I didn't intend on it, but just really didn't like the grammar of `deleteSysopDefaultToTag`.  Former options aren't great, but they don't cause any harm, just litter the end of `twinkleoptions`.

PR #847, opened 2020-02-10
block: Add uw-aepblock

Amorymeltzer:
@QEDK Seem fine?

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -708,7 +708,7 @@ Twinkle.block.blockPresetsInfo = {
 		pageParam: true,
 		reason: '[[WP:Arbitration enforcement|Arbitration enforcement]]',
 		reasonParam: true,
-		summary: 'You have been blocked from editing for violating an [[WP:Arbitration|arbitration decision]] with your edits'
+		summary: 'You have been blocked from editing for violating an [[WP:Arbitration|arbitration decision]]'
 	},
 	'uw-bioblock': {
 		autoblock: true,
@@ -1040,6 +1040,14 @@ Twinkle.block.blockPresetsInfo = {
 		reasonParam: true,
 		reason: 'Email [[WP:Harassment|harassment]]',
 		summary: 'You have been [[WP:PB|blocked]] from emailing other editors for [[WP:Harassment|harassment]]'
+	},
+	'uw-aepblock': {
+		autoblock: true,
+		nocreate: false,
+		pageParam: false,
+		reason: '[[WP:Arbitration enforcement|Arbitration enforcement]]',
+		reasonParam: true,
+		summary: 'You have been [[WP:PB|blocked]] from editing for violating an [[WP:Arbitration|arbitration decision]]'
 	}
 };
 
@@ -1150,7 +1158,8 @@ Twinkle.block.blockGroupsPartial = [
 			{ label: 'Edit warring', value: 'uw-ewpblock' },
 			{ label: 'Misusing multiple accounts', value: 'uw-acpblock' },
 			{ label: 'Misusing multiple accounts - indefinite', value: 'uw-acpblockindef' },
-			{ label: 'Email harassment', value: 'uw-epblock' }
+			{ label: 'Email harassment', value: 'uw-epblock' },
+			{ label: 'Arbitration enforcement', value: 'uw-aepblock' }
 		]
 	}
 ];


modules/twinkleblock.js
```diff
@@ -1040,6 +1040,14 @@ Twinkle.block.blockPresetsInfo = {
...
 		reason: 'Email [[WP:Harassment|harassment]]',
 		summary: 'You have been [[WP:PB|blocked]] from emailing other editors for [[WP:Harassment|harassment]]'
+	},
+	'uw-aepblock': {
+		autoblock: true,
+		nocreate: false,
+		pageParam: false,
+		reason: '[[WP:Arbitration enforcement|Arbitration enforcement]]',
+		reasonParam: true,
+		summary: 'You have been [[WP:PB|blocked]] from editing for violating an [[WP:Arbitration|arbitration decision]] with your edits'
```
QEDK: Change to "...arbitration sanction." The last bit is hanging.

QEDK:
Looks good to go!

modules/twinkleblock.js
Amorymeltzer: @QEDK You mean:

```suggestion
		summary: 'You have been [[WP:PB|blocked]] from editing for violating an [[WP:Arbitration|arbitration sanction]]'
```

I'm fine with removal of "with your edits" — it was copied from the summary for uw-aeblock and agree it's weird — but I think it should stay as "decision": it's the language used by the actual template, and the block itself is the "sanction".  Yeah, technically, you can be sanctioned for violating a sanction, but it's a little jargony and unnatural English, and not 100% of cases.

modules/twinkleblock.js
```diff
@@ -1040,6 +1040,14 @@ Twinkle.block.blockPresetsInfo = {
 		reasonParam: true,
 		reason: 'Email [[WP:Harassment|harassment]]',
 		summary: 'You have been [[WP:PB|blocked]] from emailing other editors for [[WP:Harassment|harassment]]'
+	},
+	'uw-aepblock': {
```
QEDK: `forRegisteredOnly: true,`

modules/twinkleblock.js
```diff
@@ -1040,6 +1040,14 @@ Twinkle.block.blockPresetsInfo = {
...
+		autoblock: true,
+		nocreate: false,
+		pageParam: false,
+		reason: '[[WP:Arbitration enforcement|Arbitration enforcement]]',
```
QEDK: make this custom, AE is too vague.

modules/twinkleblock.js
```diff
@@ -1040,6 +1040,14 @@ Twinkle.block.blockPresetsInfo = {
...
+		reasonParam: true,
```
QEDK: should add a decisionParam to support DS topic codes?

modules/twinkleblock.js
QEDK: > @QEDK You mean:
> 
> I'm fine with removal of "with your edits" — it was copied from the summary for uw-aeblock and agree it's weird — but I think it should stay as "decision": it's the language used by the actual template, and the block itself is the "sanction". Yeah, technically, you can be sanctioned for violating a sanction, but it's a little jargony and unnatural English, and not 100% of cases.

That should do as well.

modules/twinkleblock.js
Amorymeltzer: Why?  `uw-aeblock` isn't

modules/twinkleblock.js
Amorymeltzer: Yeah, on my todo list, especially now that there will be two.  Couple ways to do it, will ponder.

modules/twinkleblock.js
Amorymeltzer: What to?  This is the same as `ue-aeblock`, and it's just the default in the box; sysops can of course change.

modules/twinkleblock.js
QEDK: > Why? `uw-aeblock` isn't
Not too sure about this? Do we issue AE blocks to IPs? 
I assumed the system goes like:
* Page under sanction
* Editor (not IP alerted)
* Blocked for violating sanction

modules/twinkleblock.js
QEDK: > What to? This is the same as `ue-aeblock`, and it's just the default in the box; sysops can of course change.
Add a tooltip here then? I stated this actually because I got a query regarding which AE did I protect the pages under (albeit my responsibility).

modules/twinkleblock.js
Amorymeltzer: It's rare, I imagine largely because of the extra paperwork and notice required, and because things like `uw-disruptblock` are just easier, but it does happen and there's no reason why not.  DS applies to editors, not just registered users.

modules/twinkleblock.js
Amorymeltzer: That'd be a good thing for the DS box to add, but yeah, a generic "You may want to change this from the default" tooltip couldn't hurt.

Issue #850, opened 2020-02-13
Add a block template for undisclosed paid editing


wiki-ST47:
To bridge the gap between uw-adblock (which is a block for advertising, it mentions COI but not PAID) and uw-spamublock (which does warn about PAID, but only applies to users with promotional usernames), coffee has created Template:Uw-upeblock:

https://en.wikipedia.org/wiki/Template:Uw-upeblock

Please add this template to Twinkle. The label in the block preset / template list can be "Undisclosed Paid Editing". The block reason can be `[[WP:PAID|Undisclosed paid editing]] in violation of the WMF [[WP:TOU|Terms of Use]]`. This template only applies to registered users, and should default to an indef block, with autoblock enabled, account creation blocked, and talk page editing and email both allowed. You can pass a "page" parameter to the template.

I'll send a pull request momentarily.

PR #851, opened 2020-02-13
Add uw-upeblock to block options

wiki-ST47:
Per #850, add a new Template:Uw-upeblock, for blocks related to undisclosed paid editing, to the list of available block presets.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -957,7 +957,7 @@ Twinkle.block.blockPresetsInfo = {
 		nocreate: true,
 		pageParam: true,
 		reason: '[[WP:PAID|Undisclosed paid editing]] in violation of the WMF [[WP:TOU|Terms of Use]]',
-		summary: 'You have been indefinitely blocked from editing because your account is being used only for [[WP:SPAM|spam or advertising]] in violation of [[WP:PAID|Wikipedia policy on paid advocacy'
+		summary: 'You have been indefinitely blocked from editing because your account is being used in violation of [[WP:PAID|Wikipedia policy on undisclosed paid advocacy]]'
 	},
 	'uw-vaublock': {
 		autoblock: true,
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -950,6 +950,14 @@ Twinkle.block.blockPresetsInfo = {
 		reason: '{{uw-ublock-double}} <!-- Username closely resembles another user, hard block -->',
 		summary: 'You have been indefinitely blocked from editing because your [[WP:U|username]] appears to impersonate another established Wikipedia user'
 	},
+	'uw-upeblock': {
+		autoblock: true,
+		expiry: 'infinity',
+	    	forRegisteredOnly: true,
+	    	nocreate: true,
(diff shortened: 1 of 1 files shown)


wiki-ST47:
I was testing this template in [my common.js](https://en.wikipedia.org/wiki/User:ST47/common.js) and it seems to work, however, I don't know how to properly test my "dev" version of Twinkle. Please have a look and ensure that I got everything right.

modules/twinkleblock.js
```diff
@@ -950,6 +950,15 @@ Twinkle.block.blockPresetsInfo = {
...
 		summary: 'You have been indefinitely blocked from editing because your [[WP:U|username]] appears to impersonate another established Wikipedia user'
 	},
+	'uw-upeblock': {
+		autoblock: true,
+		expiry: 'infinity',
+		forRegisteredOnly: true,
+		nocreate: true,
+		pageParam: true,
+		reason: '[[WP:PAID|Undisclosed paid editing]] in violation of the WMF [[WP:TOU|Terms of Use]]',
+		summary: 'You have been indefinitely blocked from editing because your account is being used only for [[WP:SPAM|spam or advertising]] in violation of [[WP:PAID|Wikipedia policy on paid advocacy'
```
Amorymeltzer: Missing brackets to close out the link:

```suggestion
		summary: 'You have been indefinitely blocked from editing because your account is being used only for [[WP:SPAM|spam or advertising]] in violation of [[WP:PAID|Wikipedia policy on paid advocacy]]'
```

modules/twinkleblock.js
```diff
@@ -950,6 +950,15 @@ Twinkle.block.blockPresetsInfo = {
 		reason: '{{uw-ublock-double}} <!-- Username closely resembles another user, hard block -->',
 		summary: 'You have been indefinitely blocked from editing because your [[WP:U|username]] appears to impersonate another established Wikipedia user'
 	},
+	'uw-upeblock': {
+		autoblock: true,
+		expiry: 'infinity',
+		forRegisteredOnly: true,
+		nocreate: true,
+		pageParam: true,
+		reason: '[[WP:PAID|Undisclosed paid editing]] in violation of the WMF [[WP:TOU|Terms of Use]]',
```
Amorymeltzer: I'm mixed on the use of WP:TOU here, given that it's just a soft redirect and otherwise sparse on information.  I suppose it's what's currently being used by the template so it's fine?

modules/twinkleblock.js
Amorymeltzer: That being said, I think it'd be good to have this be in closer alignment with the log entry, if only by specifically stating it's for UPE.  Maybe something like:

```suggestion
		summary: 'You have been indefinitely blocked from editing because your account is being used in violation of [[WP:PAID|Wikipedia policy on undisclosed paid advocacy]]'
```

Or something like that?

Amorymeltzer:
@wiki-ST47 I can help make some of these changes if you'd like

wiki-ST47:
I updated the block summary message based on your suggestions. [[m:TOU]] might be a better choice for the wikilink, if we want to avoid the soft redirect?

Amorymeltzer:
Is just saying UPE sufficient?  Maybe a little too succinct.  The template has changed a bit, but it's roughly in-line, so I say let's just give it a go, can change if need be.

PR #853, opened 2020-02-14
morebits: use innerHTML for printUserText message header, not textContent; use in ARV

### LINKED ISSUE #590: ARV: Add link to AIV when user has already been reported
Currently, when a user has already been reported to AIV, Twinkle refuses to report them, instead saying "The comments you typed are provided below, in case you wish to manually post them under the existing report for this user at AIV:" Ideally, the text "at AIV" should link to AIV. The text shown is defined [here](https://github.com/azatoth/twinkle/blob/master/modules/twinklearv.js#L578`), but the only links that are currently shown are footer links using [morebits' addFooterLink](https://github.com/azatoth/twinkle/blob/master/morebits.js#L4134), so I don't know how to approach this.

Amorymeltzer:
Swaps `textContent` out for `innerHTML` in the `p` element (the header).  Doing so allows the use of html (go figure), which menas we can link to AIV and UAA to close #590.  Shouldn't have any noticeable effects otherwise?

Alternatively, we could add some selector to either/both `printUserText` elements then work on that directly.

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -568,11 +568,12 @@ Twinkle.arv.callback.evaluate = function(e) {
 
 			aivPage.load(function() {
 				var text = aivPage.getPageText();
+				var $aivLink = '<a target="_blank" href="/wiki/WP:AIV">WP:AIV</a>';
 
 				// check if user has already been reported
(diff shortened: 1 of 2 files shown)


siddharthvp:
Alternatively we could allow wikilink syntax in the `message` and setup `addFooterLink` to render the wikilinks. This would be nice to have in the `Morebits.status` constructor as well.

Amorymeltzer:
That's an interesting idea for `Morebits.status`, what'd you have in mind?  I'd imagine that'd look something like having a `linkifier` function that pulls out (unbinder?) anything that looks like a wikilink and passes the rest as text?

PR #854, opened 2020-02-14
xfd: Use shorter comment and regex to avoid constantly updating RM/TR comment

Amorymeltzer:
The hidden message has been changing frequently, leaving us behind (e.g. #791, #842), so by using a short regex rather than an exact string match, we should be able to handle subsequent changes.  This change also only loads section number 2; that's possibly overkill, but could theoretically avoid some off-target effects with the page's other hidden comment.  Would cause issues in the unlikely event of the page being restructured.


Amorymeltzer:
Regarding section loading, see also #860

Issue #841, opened 2020-02-07
Improve Revert Conflict Handling


RhinosF1:
As Twinkle wants to be fast, The talk page is opened before rollback is complete. This means that if someone reverts at the same time then you may conflict without warning.

Twinkle should check before leaving a welcome (problem user/IP) or warning after reverting whether you actually did the revert.

RhinosF1:
17:12:34 <AntiComposite> RhinosF1, that's because Twinkle tries to be fast and opens the user talk page before the rollback has finished
17:13:56 <AntiComposite> The warn dialog could check if you're the next edit after the vanarticlerevid, if you want that submit a patch or file an issue

Amorymeltzer:
Curious @RhinosF1 did this happen to you?

Amorymeltzer:
I think this makes sense for warnings (#855), but less so for welcomes: warnings and welcomes aren't mutually exclusive, and it's not really an issue to "welcome" someone if you didn't do the revert.

I could be convinced, though.  I think it'd be better for welcoming to do something like checking to see if the page was recently edited in the last minute (or so), similar to what we do in `warn`.  Would that be useful?

RhinosF1:
Hi,

It did happen to me.

I sometimes use the ‘problem user welcomes’ and anon-unconstructive etc. Instead of a warning at first offence for unwelcomed users as it is more polite. Seeing if the talk page was edited would help.

PR #855, opened 2020-02-16
warn: Check if reverting user actually reverted

Amorymeltzer:
See #841: if a Twinkle user is on a usertalk page after trying to revert or rollback a revision (as opposed to clicking on the usertalk link on a diff, indicated by no `noautowarn` being present), check that the revision following the diff inquestion was made by the Twinkle user, otherwise add a note (building upon the message area added in #757) that they may not have been the person to have actually reverted the edit in question.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -102,13 +102,37 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 		type: 'div',
 		label: '',
 		style: 'color: red',
-		id: 'twinkle-warn-olddiff-message'
+		id: 'twinkle-warn-revert-messages'
 	});
 
-	// Confirm edit wasn't too old for a warning
 	var vanrevid = mw.util.getParamValue('vanarticlerevid');
 	if (vanrevid) {
-		var query = {
+		var message = '';
+		var query = {};
+
+		// If you tried reverting, check if *you* actually reverted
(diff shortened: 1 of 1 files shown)


modules/twinklewarn.js
```diff
@@ -102,13 +102,37 @@ Twinkle.warn.callback = function twinklewarnCallback() {
...
+				rvstartid: vanrevid,
+				rvlimit: 2,
+				rvdir: 'newer',
+				rvprop: 'user'
+			};
+
+			new Morebits.wiki.api('Checking if you successfully reverted the page', query, function(apiobj) {
+				var revertUser = $(apiobj.getResponse()).find('revisions rev')[1].getAttribute('user');
+				if (revertUser && revertUser !== mw.config.get('wgUserName')) {
+					message += ' Someone else besides you actually reverted the page, and may have already warned them.';
```
QEDK: Similar to Huggle, suggest changing this to:
"Someone else has already reverted the page and may have warned the user."

Amorymeltzer: Combined the two: "Someone else reverted the page and may have already warned the user"

siddharthvp:
It's worth noting that the API returns a `nochange: true` within the success response if an attempted edit didn't really change anything. But it doesn't seem possible to use that here since reversion is being done in another module (revert) or outside TW (rollback).

PR #856, opened 2020-02-16
morebits: Add getter (wiki.page) for time of last edit

Amorymeltzer:
Similar to 77242b9 but doesn't(?) rely on `intoken`.  Also moved `getCurrentID()` nearby, out of the revert-specific list (it's not a perfectly-organized list but w/e).

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2218,11 +2218,6 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		ctx.revertOldID = oldID;
 	};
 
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I'm sure we can make use of this...

Amorymeltzer:
Hmm, forgot you had this as part of [#765](https://github.com/azatoth/twinkle/pull/765/files#r360960854) @siddharthvp

PR #857, opened 2020-02-19
Slightly improve handling of Scribunto/non-wikitext content models

### LINKED ISSUE #572: csd/xfd: Tagging a module doesn't add it to watchlist
The CSD tag is (rightly) added to the documentation page as a result of #515. However, as a result the actual module page isn't watched according to user preferences. It should be.

Amorymeltzer:
**1st commit: morebits: lookupCreation won't query content and section unless lookupNonRedirectCreator is provided**

PR #710 added the ability for `fnLookupCreationSuccess` to find the creator of the first non-redirect revision via `fnLookupNonRedirectCreator` (see also #805) by adding `content` to the `rvprop`s list as well as `rvsection=0`.  That doesn't work for non-wikitext content models, who will returned a `rvnosuchsection` error in such cases.  Removing `content` would solve the error, as would `rvsection=0`, but the actual redirect check (`/^\s*#redirect/i`) doesn't make sense in those scenarios (e.g. Scribunto, JS, CSS), so there's no really point in even checking.

Instead, only attempt to use `content` and `rvsection` if `lookupNonRedirectCreator` is set.  Currently, only AfD and PROD do so, which aligns with how it should be limited.

We could probably just use `wgPageContentModel`, but in theory the ability to query pages other than the current one, even if the two have different content models, seems useful.  In practice, only `twinklexfd` attempts sch a thing, only for TfD merge nominations (`involvedpages`), and assumes that the target is in the same namespace as the current page, so Scribunto will match with Scribunto.

**2nd commit: xfd/speedy: Watch tagged modules**

Uses params.scribunto to check if we're dealing with a /doc subpage or not (easier and more portable than regex testing, etc.), and simply call the `watch` api.  Also removes some collateral with undesired `pageobj.setCreateOption('recreate')` for non-module taggings.  Closes #572; Sorry if you were actively looking at this @DannyS712 and I scooped you, I got thinking after the above.

**3rd commit: xfd: Fix tfm preview when namespace is included**

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1513,7 +1513,18 @@ Twinkle.speedy.callbacks = {
 			pageobj.setPageText(code + (params.normalizeds.indexOf('g10') !== -1 ? '' : '\n' + text)); // cause attack pages to be blanked
 			pageobj.setEditSummary(editsummary + Twinkle.getPref('summaryAd'));
 			pageobj.setWatchlist(params.watch);
-			pageobj.setCreateOption('recreate'); // Module /doc might not exist
+			if (params.scribunto) {
+				pageobj.setCreateOption('recreate'); // Module /doc might not exist
+				if (params.watch) {
+					// Watch module in addition to /doc subpage
+					var watch_query = {
+						action: 'watch',
+						titles: mw.config.get('wgPageName'),
+						token: mw.user.tokens.get('watchToken')
+					};
+					new Morebits.wiki.api('Adding Module to watchlist', watch_query).post();
+				}
+			}
 			pageobj.save(Twinkle.speedy.callbacks.user.tagComplete);
 		},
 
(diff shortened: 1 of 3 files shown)


Amorymeltzer:
Another way of doing the first commit would be to rely on `.setLookupNonRedirectCreator()`, to toggle on `ctx.lookupNonRedirectCreator === false` then note that `lookupNonRedirectCreator` should/must be set to `false` for all non-wikitext lookups.  It's a little more portable and probably makes more sense, and theoretically can cover more cases.  The downside is it's *theoretically* possible that we might have a lookup that could apply to both, and *that* would rely on toggling setting `lookupNonRedirectCreator` on `wgPageContentModel` or the page title like here.  Probably better to do that in code rather than the library.  Will ponder but I'm leaning toward that idea.

Amorymeltzer:
Now works as I mentioned in https://github.com/azatoth/twinkle/pull/857#issuecomment-591728590

PR #858, opened 2020-02-19
block: Use word boundary regex to match bot usernames

Amorymeltzer:
Requiring the username to end in 'bot' means we miss out on some major ones (e.g. AnomieBot III, ClueBot NG), but checking for a word boundary should improve things without much collateral (e.g. NinjaRobotPirate).


Amorymeltzer:
FWIW, out of 309 bot accounts, `/bot$/i` matches 261 while `/bot\b/i` match 286.  It's also more likely to match other accounts, which by definition will be off-target, but both neither has any offtarget effects on any autoconfirmed editors with edits in the past 30 days.

Issue #19, opened 2011-06-07
ARV glitches: "Failed to save edit: Edit conflict detected"


Amalthea:
From
- http://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=433020512#ARV
- http://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=433020512#Processing_AIV_request:_Failed_to_save_edit:_Edit_conflict_detected

UAA & AIV reports sometimes can't be saved due to an alleged edit conflict. It tries to save two or three times, and then gives up.
I could reproduce it once, but the problem dissolved while I was trying to debug it. It re-manifests from time to time though.

PR #860, opened 2020-02-20
xfd: reenable setPageSection for mfd and cfd

Amorymeltzer:
Disabled in c97229c due to bugs in the API returning edit conflicts; temporally, it seems related to #19, and I think 45ce088 may have been trying to fix the same thing.  I'm not sure what the error was but that was nine years ago, and testing seems fine, so I'm hoping this is okay, sort of a test for the rest.


Amorymeltzer:
Regarding section loading, see also #854

PR #862, opened 2020-02-21
xfd: Fix two bugs when RfDing soft redirects

### LINKED ISSUE #861: Failed to list a soft redirect at RFD
See the [history of Megathread](https://en.wikipedia.org/w/index.php?title=Megathread&action=history) and [the talk page](https://en.wikipedia.org/wiki/Talk:Megathread), and  [JzG's contributions for 11 Feb](https://en.wikipedia.org/w/index.php?target=JzG&namespace=all&tagfilter=&start=2020-02-11&end=2020-02-11&limit=50&title=Special%3AContributions). It seems Twinkle tried to list the discussion on a nonexistent subpage [[Wikipedia:Redirects for discussion/Log/NaN undefined NaN]] - it didn't create that page, but did add the nomination template with that link in the edit summary, and the same with the author notification. After adding a fake #REDIRECT to the page Twinkle worked as expected.

Amorymeltzer:
1. Since #741 (specifically, 3e21765), RfD was using a returned `curtimestamp` to generate the log page.  That query was only being done if the page wasn't a soft redirect, however, leading to NaN/undefined issues.  Closes #861 by falling back to the previous client clock method.
2. #527 (44c0e42) added the ability to notify a redirect's target, but returns an error if the title is invalid, such as with soft redirects via {{wi}}.  This now skips the notification in such cases.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -1325,7 +1325,8 @@ Twinkle.xfd.callbacks = {
 			};
 		},
 		main: function(params) {
-			var date = new Date(params.curtimestamp);
+			// Fallback to client clock for softredirects
+			var date = params.curtimestamp ? new Date(params.curtimestamp) : new Date();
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Pushing live.

PR #863, opened 2020-02-21
xfd: Avoid client clock for soft redirects

Amorymeltzer:
In doing #862, I undid the intent of #714 to avoid the client clock as much as possible for the specific case of soft redirects being nominated at RfD.  It never worked - thus #861 - but the idea is sound.  This has softredirects go through the same `curtimestamp` query and "typical" redirects, but simply skips processing of target or section.  It's not a free check anymore, but it should be incredibly infrequent and fairly quick.

Also added a note for softredirect failures from the history tab.  If run from history, the `Morebits.wiki.isPageRedirect` check for soft redirects won't work.  The module fails fine, but a note suggesting users retry from the content tab is helpful.


Amorymeltzer:
Not included in #862 since I wanted to do some extra testing/thought I'd be going to bed.

PR #875, opened 2020-03-02
arv: Combine AN3 queries, avoid errors, and handle revdel'd comments

Amorymeltzer:
1st commit: Combine the three largely identical AN3 queries, avoid errors

Don't print console errors if is provided: we shouldn't be unnecessarily spitting out console errors, either from null queries or mw.Title.  The warnings check is fine, though.

Has a more helpful API error messages, similar to #866.

2nd commit: Better text for revdel'd comments in AN3 diffs


Amorymeltzer:
I'm not sure if the `free_resolves` input needs to be dependent on a page actually loading, but it was previously so I kept it as such.

PR #876, opened 2020-03-06
block: better detect bots and turn off autoblock

siddharthvp:
Let's do #858 a bit systematically. While the hack used there is probably good enough for enwiki, not so for other language wikis.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -105,6 +105,7 @@ Twinkle.block.fetchUserInfo = function twinkleblockFetchUserInfo(fn) {
 		lelimit: 1,
 		bkusers: mw.config.get('wgRelevantUserName'),
 		ususers: mw.config.get('wgRelevantUserName'),
+		usprop: 'groupmemberships',
 		letitle: 'User:' + mw.config.get('wgRelevantUserName')
 	})
 		.then(function(data) {
@@ -124,6 +125,10 @@ Twinkle.block.fetchUserInfo = function twinkleblockFetchUserInfo(fn) {
 			Twinkle.block.hasBlockLog = !!data.query.logevents.length;
 			// Used later to check if block status changed while filling out the form
 			Twinkle.block.blockLogId = Twinkle.block.hasBlockLog ? data.query.logevents[0].logid : false;
+
+			Twinkle.block.userIsBot = userinfo.groupmemberships.map(function(e) {
+				return e.group;
+			}).indexOf('bot') !== -1;
 
 			if (typeof fn === 'function') {
 				return fn();
@@ -1294,7 +1299,7 @@ Twinkle.block.callback.update_form = function twinkleblockCallbackUpdateForm(e, 
 	data.hardblock = data.hardblock !== undefined ? data.hardblock : false;
 
 	// disable autoblock if blocking a bot
-	if (Twinkle.block.isRegistered && relevantUserName.search(/bot\b/i) > 0) {
+	if (Twinkle.block.userIsBot || relevantUserName.search(/bot\b/i) > 0) {
 		data.autoblock = false;
 	}
 


QEDK:
Uhhh, so there's a ton of bot accounts that may or may not be flagged as bot, or flagged as a bot but no "bot" in the username (AWB bots for example that should be autoblocked).
Also adding that most bots except the ones which run via Toolforge can be blocked, so a better way is to probably check the userpage for the Toolforge bot template.

siddharthvp:
> Uhhh, so there's a ton of bot accounts that may or may not be flagged as bot, or flagged as a bot but no "bot" in the username (AWB bots for example that should be autoblocked).

I am not sure what you're getting at. Which are the bots which aren't flagged as bots? If you mean mass edits done by scripts without approval, I don't we call them "bot" for our purposes, and autoblock=true for them would be ok. 

> Also adding that most bots except the ones which run via Toolforge can be blocked, so a better way is to probably check the userpage for the Toolforge bot template.

While blocking a bot that's malfunctioning, there's no need to block the botop's IP address, toolforge or awb or whatever. Regardless, an extra API call just for this would be overkill.

Amorymeltzer:
Agree that a check for the template is overkill, and likely not universally applicable either.

Anyway, the problem is that there are misnomers on both sides everywhere.  In particular for this PR, there are plenty of approved "bot" accounts that do not have the `bot` flag; AFAIK, they may or may not be run on the toolserver.  Actually, even if they're not, and are run locally, we probably wouldn't want to catch the user themself up in the autoblock.

QEDK:
> Agree that a check for the template is overkill, and likely not universally applicable either.
> 
> Anyway, the problem is that there are misnomers on both sides everywhere. In particular for this PR, there are plenty of approved "bot" accounts that do not have the `bot` flag; AFAIK, they may or may not be run on the toolserver. Actually, even if they're not, and are run locally, we probably wouldn't want to catch the user themself up in the autoblock.

Imo, it would be best just changing the regex to r"bot"/i and checking for a bot flag.

siddharthvp:
Okay, we can retain the poor man's regex check, along with the bot flag check.

modules/twinkleblock.js
```diff
@@ -1287,7 +1291,7 @@ Twinkle.block.callback.update_form = function twinkleblockCallbackUpdateForm(e,
 	data.hardblock = data.hardblock !== undefined ? data.hardblock : false;
 
 	// disable autoblock if blocking a bot
-	if (Twinkle.block.isRegistered && relevantUserName.search(/bot\b/i) > 0) {
+	if (Twinkle.block.userIsBot) {
```
Amorymeltzer, line 1294: Conversation stalled @siddharthvp @QEDK , but am I right that we arrived at just doing the following:

```suggestion
	if (Twinkle.block.userIsBot || relevantUserName.search(/bot\b/i) > 0) {
```

Do we need the `isRegistered` check/early fail?

```suggestion
	if (Twinkle.block.isRegistered && (Twinkle.block.userIsBot || relevantUserName.search(/bot\b/i) > 0)) {
```

siddharthvp: The isRegistered check is unnecessary as `relevanUserName` is an IP address in that case, which can't match the regex.

QEDK: I think the thinking was to avoid the regex check for IP addresses, it's what I'd do but I don't think it's much of a performance overhead. 🤔

Amorymeltzer: That was basically it, yeah; I agree it's not really worth it, just covering bases.

modules/twinkleblock.js
```diff
@@ -125,6 +126,10 @@ Twinkle.block.fetchUserInfo = function twinkleblockFetchUserInfo(fn) {
 			// Used later to check if block status changed while filling out the form
 			Twinkle.block.blockLogId = Twinkle.block.hasBlockLog ? data.query.logevents[0].logid : false;
 
+			Twinkle.block.userIsBot = userinfo.groupmemberships.map(function(e) {
```
Xi-Plus, line 129: `userinfo.groupmemberships` is undefined when you blocking IP.

Amorymeltzer: Word!  So, that'd be, what, `!!userinfo.groupmemberships && userinfo.groupmemberships.map...`?  Mobile, will push tomorrow.

Amorymeltzer: Thanks for the note @Xi-Plus; restructured it in c3916f4

PR #877, opened 2020-03-06
tag: add and update {{notability}} parameters

QEDK:
Requested at WT:TW.
Sport -> Sports mostly to keep it in line with the other parameters.
Add a missing list parameter.
Places -> Geographic in line with the Notability guideline that is called Geographic and include places (which is a geographic feature).


QEDK:
> LGTM 👍 and like the name updates, but as long as we're adding `Lists`, might as well add `Astro`, the other missing one!

Good catch! Thought I added all of em.

Issue #709, opened 2019-09-19
Log PROD/CSD tags from Page Curation


MusikAnimal:
See https://phabricator.wikimedia.org/T230455 for context.

[Page Curation](https://en.wikipedia.org/wiki/Wikipedia:Page_Curation) is a MediaWiki tool to help with new page patrolling, and offers some functionality similar to Twinkle. Users of the extension have requested that speedy/PROD nominations from Page Curation be logged to Special:MyPage/CSD_log, etc., just as Twinkle does. This was investigated and decided that this is solely a Twinkle invention, and especially given the changes to the logic over the years, it is better handled by Twinkle itself rather than MediaWiki.

https://gerrit.wikimedia.org/r/530977 introduces an "action queue" (similar to mw.hook) that allows integration with Page Curation. The difference from mw.hook being that the action queue is promised-based, which is necessary since Page Curation wants to reload the page after tagging. The proposal is for Twinkle to listen to the `delete` event and log the nominations just as Twinkle's speedy module does.

I don't think this will be a lot of work, and it'd really help out the NPP community. I'm happy to take it on, if no one is up for it, but I'd like to get the go ahead from you all.

The action queue functionality goes live next Thursday, September 26. Documentation for action queue is at https://www.mediawiki.org/wiki/Extension:PageTriage#Client-side_hooks

Amorymeltzer:
>I'm happy to take it on

Go for it!  I'll gladly review.

Are you thinking to place this in the csd module, or something separate?  It's a bit of a departure from most of Twinkle, but it seems like something that'd help (reviewing) NPP work, and anything we can do to help them is probably worth it.

MusikAnimal:
> Go for it! I'll gladly review.

Thanks :)

> Are you imagining this living in the csd module, prod, or something new/separate? It's a bit of a departure from most of Twinkle, but it seems like something that'd help (reviewing) NPP work, and anything we can do to help them is probably worth it.

Yes I was thinking the speedy module, where it currently handles the logging. We'd register the hook (action callback, technically is I guess is what it's called) somewhere towards the bottom, with clear comments as to why it's there. The NPP community wanted Page Curation to go off of twinkleoption.js, too, so it was obvious that Twinkle should be doing the work. I'm glad to hear approval :) `Twinkle.speedy.callbacks.user.addToLog` should work nicely. We _might_ need to adjust some code to have it return a promise, but that shouldn't be too hard either.

siddharthvp:
[...]

PR #878, opened 2020-03-06
morebits/speedy/prod: add Morebits.userspaceLogger

siddharthvp:
Class for handling the logic to add to a userspace log.

No need to duplicate the code across 2-3 modules. Makes the code more object-oriented.

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -360,56 +360,38 @@ Twinkle.prod.callbacks = {
 	},
 
 	addToLog: function(params) {
-		var wikipedia_page = new Morebits.wiki.page('User:' + mw.config.get('wgUserName') + '/' + Twinkle.getPref('prodLogPageName'), 'Adding entry to userspace log');
-		wikipedia_page.setCallbackParameters(params);
-		wikipedia_page.load(Twinkle.prod.callbacks.saveLog);
-	},
+		var usl = new Morebits.userspaceLogger(Twinkle.getPref('prodLogPageName'));
+		usl.initialText =
+			"This is a log of all [[WP:PROD|proposed deletion]] tags applied or endorsed by this user using [[WP:TW|Twinkle]]'s PROD module.\n\n" +
+			'If you no longer wish to keep this log, you can turn it off using the [[Wikipedia:Twinkle/Preferences|preferences panel]], and ' +
+			'nominate this page for speedy deletion under [[WP:CSD#U1|CSD U1]].\n';
 
-	saveLog: function(pageobj) {
-		var text = pageobj.getPageText();
-		var params = pageobj.getCallbackParameters();
-
-		// add blurb if log page doesn't exist
-		if (!pageobj.exists()) {
-			text =
-				"This is a log of all [[WP:PROD|proposed deletion]] tags applied or endorsed by this user using [[WP:TW|Twinkle]]'s PROD module.\n\n" +
-				'If you no longer wish to keep this log, you can turn it off using the [[Wikipedia:Twinkle/Preferences|preferences panel]], and ' +
-				'nominate this page for speedy deletion under [[WP:CSD#U1|CSD U1]].\n';
-		}
-
-		// create monthly header if it doesn't exist already
-		var date = new Morebits.date(pageobj.getLoadTime());
-		if (!date.monthHeaderRegex().test(text)) {
-			text += '\n\n' + date.monthHeader(3);
-		}
-
-		text += '\n# [[:' + Morebits.pageNameNorm + ']]';
+		var logText = '# [[:' + Morebits.pageNameNorm + ']]';
+		var summaryText;
 		// If a logged file is deleted but exists on commons, the wikilink will be blue, so provide a link to the log
-		text += namespace === 'file' ? ' ([{{fullurl:Special:Log|page=' + mw.util.wikiUrlencode(mw.config.get('wgPageName')) + '}} log]): ' : ': ';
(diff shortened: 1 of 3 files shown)


Amorymeltzer:
Haven't looked at it yet, but generally agree with this kind of thing; I've had a similar idea for a bit, so, :+1:.

Logistically, I was planning on just merging #839, but I suppose I can put that on hold and rework it to use this once it goes through.  More to the point, this would presumably benefit from #814, but then again there're plenty of conflicts there that need to be dealt with anyway.  I don't want to create more work, but what would work for you?

siddharthvp:
I wrote it using Morebits.date initially, before hitting on the dumb "Morebits.date is not a function" error - arrgh, that's in another branch! 

It's only 2 lines that need to be changed so the order doesn't matter. If #814 is merged first, I can tweak this PR to use that; if this is merged first, I can tweak _that_ to the same effect. 

Similarly for this and #839.

Amorymeltzer:
If it works for you, let's do #814 first; fewer commits, there's already plenty it's blocking, etc.

siddharthvp:
Now using Morebits.date in this PR

modules/twinkleprod.js
```diff
@@ -355,57 +355,38 @@ Twinkle.prod.callbacks = {
...
-		// create monthly header if it doesn't exist already
-		var date = new Date(pageobj.getLoadTime());
-		var headerRe = new RegExp('^==+\\s*' + date.getUTCMonthName() + '\\s+' + date.getUTCFullYear() + '\\s*==+', 'm');
-		if (!headerRe.exec(text)) {
-			text += '\n\n=== ' + date.getUTCMonthName() + ' ' + date.getUTCFullYear() + ' ===';
-		}
-
-		text += '\n# [[:' + Morebits.pageNameNorm + ']]';
+		var logText = '# [[:' + Morebits.pageNameNorm + ']]';
+		var summaryText;
```
DannyS712: for ease of reviewing the diff, can this stay as `summarytext` rather than `summaryText`? (same with `text` vs `logText`, though the reason for changing that is clearer so its not as important to keep)

morebits.js
```diff
@@ -3563,6 +3563,41 @@ Morebits.wikitext.page.prototype = {
...
+ * *********** Morebits.userspaceLogger ************
+ * Handles logging actions to a userspace log, used in
+ * twinklespeedy and twinkleprod.
+ */
+
+Morebits.userspaceLogger = function(logPageName) {
+	this.initialText = '';
+	this.headerLevel = 3;
+
+	this.log = function(logText, summaryText) {
```
DannyS712: Given that this only has one method, I would suggest instead having something like
```
Morebits.userspaceLogger = function(logPageName, logText, summaryText) {
	...
}
```
that handles everything instead of an object with only one needed function

morebits.js
```diff
@@ -3563,6 +3563,41 @@ Morebits.wikitext.page.prototype = {
...
+		var page = new Morebits.wiki.page('User:' + mw.config.get('wgUserName') + '/' + logPageName,
+			'Adding entry to userspace log'); // make this '... to ' + logPageName ?
+		return page.load(function(pageobj) {
+			var text = pageobj.getPageText();
```
DannyS712: It took me a bit to wrap my head around these lines, since you unconditionally use getPageText and then conditionally reset the text if the page doesn't exist. This should work, but can I suggest
```
var text;
if (!pageobj.exists()) {
	text = this.initialText;
} else {
	text = pageobj.getPageText()
}
```

modules/twinkleprod.js
siddharthvp: Look at the diff with whitespace disabled. It's quite clear.

morebits.js
siddharthvp: I just copied from twinklespeedy.js and twinkleprod.js. But it's worth noting that `getPageText` returns an empty string if page doesn't exist.
so 
```js
if (!text) {
	text = this.initialText;
}
```
may be even better.

morebits.js
siddharthvp: It isn't just these 3, there's also the public props `initialText` and `headerLevel` (though the latter is used always with the default value in TW). A function with 4-5 parameters is pretty bad design, akin to the pathetic `Morebits.wiki.api` constructor. But that's just my take.

Amorymeltzer:
Will try and get to this (mutatis mutandis) later this week but just wanted to link that it’d presumably help with #709

morebits.js
Amorymeltzer: No rush @siddharthvp but I am curious what you think of this approach.

siddharthvp:
Well, I am not understanding the motivation behind wanting to push this into Morebits.wiki.page - this is fairly niche and quite unlikely to be used outside TW. I think the 35 lines of compact code is much more maintainable than it would be if it's in MWP - which is what - a 1000 lines? The other problem with MWP is that it splits code for a task across the whole file (eg. `fnProcessDelete` is hundreds of lines away from `this.delete`) making the code flow difficult to follow. 

Regarding the error handling, I didn't even think of it coz of this being TW-only. But maybe it's a good idea to have nevertheless if this is in morebits.

Amorymeltzer:
That's fair.  Mainly the error handling stuff, but I did think it'd belong figuring as it's just taking advantage of `page`.  There wouldn't be too much distant for this, especially with my "meh" comment, but it's not a big deal.  Certainly clearer this way, without putting it in as an ugly duckling.

PR #880, opened 2020-03-08
Use unique class on proxy button to change button text

Amorymeltzer:
Added in #852, avoids overlapping with any other windows open

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -194,7 +194,7 @@ Twinkle.tag.callback = function friendlytagCallback() {
 			]
 		});
 	}
-	form.append({ type: 'submit' });
+	form.append({ type: 'submit', className: 'tw-tag-submit' });
 
 	var result = form.render();
 	Window.setContent(result);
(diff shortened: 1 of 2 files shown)


modules/friendlytag.js
```diff
@@ -267,7 +267,7 @@ Twinkle.tag.callback = function friendlytagCallback() {
 			numAdded: 0,
 			numRemoved: 0
 		};
-		$(Window.buttons[0]).after(statusNode);
+		$('.submitButtonProxy').after(statusNode);
```
siddharthvp: The purpose of using `Window.buttons[0]` was to specifically target the submit button of TW Tag window, even if there are other Morebits simpleWindows open. 

Since #852 allows for adding a class name to the quickform submit, that should be done here for uniqueness (maybe `tw-tag-submit`), rather than going with submitButtonProxy.

Amorymeltzer: Huh!  I hadn't even considered the possibility of someone doing that.  Updated with that in mind.

PR #882, opened 2020-03-09
config: Don't save deprecated options, properly save hidden options

Amorymeltzer:
They're barely used, but the various "hidden" preferences have been ignored by `twinkleconfig` for ages.  The unification of twinkleconfig and friendlyconfig from #762 (f0ebd3f) didn't actually solve that problem, and the check to catch them has only meant that deprecated options (present in [[twinkleoptions.js]], but different from the default by dint of no longer existing in the default) are retained instead (moved to the bottom).  The problem is that these were getting added to `foundPrefs` (`foundTwinklePrefs` and `foundFriendlyPrefs` prior to #762) without a value (skipped because they are hidden in the form), but since they were "found" the iteration over `Twinkle.prefs` to find missing items would by its very nature skip them.  Instead, since they can't be changed in the form, hidden prefs get passed straight through to the `newConfig` object.

This PR also: adjusts the processing loop to properly check for `adminOnly` options, and properly marks the `adminOnly` hidden options as such; removes the unhelpful completion notice upon saving; and removes the ages-ago deprecated `MinCutOff` options.

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -849,19 +849,14 @@ Twinkle.config.sections = [
 				name: 'revertMaxRevisions',
 				type: 'integer'
 			},
-			// twinklebatchdelete.js: How many pages should be processed at a time
-			{
-				name: 'batchdeleteChunks',
-				type: 'integer'
-			},
-			// twinklebatchdelete.js: How many pages left in the process of being completed should allow a new batch to be initialized
-			{
-				name: 'batchDeleteMinCutOff',
-				type: 'integer'
-			},
 			// twinklebatchdelete.js: How many pages should be processed maximum
 			{
 				name: 'batchMax',
+				type: 'integer'
+			},
+			// twinklebatchdelete.js: How many pages should be processed at a time
+			{
+				name: 'batchdeleteChunks',
 				type: 'integer'
 			},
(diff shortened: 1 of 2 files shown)


siddharthvp:
Nice, a superior alternative to https://github.com/azatoth/twinkle/pull/844#issuecomment-585398152 👍

modules/twinkleconfig.js
```diff
@@ -1586,72 +1569,67 @@ Twinkle.config.writePrefs = function twinkleconfigWritePrefs(pageobj) {
...
+
+						case 'customList':  // read from the jQuery data stored on the button object
+							userValue = $(form[pref.name]).data('value');
+							break;
+
+						default:
+							alert('twinkleconfig: unknown data type for preference ' + pref.name);
+							break;
+					}
+				} else if (Twinkle.prefs) {
```
Xi-Plus: Should be `Twinkle.prefs[pref.name]`? `Twinkle.prefs` always return `true`.

siddharthvp: `Twinkle.prefs` will be undefined if the twinkleoptions failed to load or parse (see how it's being set in twinkle.js file). 

Thanks anyway for keeping an eye out!

PR #883, opened 2020-03-09
morebits: don't handle notoken error

siddharthvp:
`notoken` error code has been removed from MediaWiki. Write queries without token now return the standard `missingparams` error.

It is anyway not necessary to handle this error since we do specify the token, so this just wouldn't ever occur.

In case of expired tokens, the `badtoken` error continues to be returned, which we do handle.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2822,7 +2822,7 @@ Morebits.wiki.page = function(pageName, currentAction) {
 			}
 
 		// check for loss of edit token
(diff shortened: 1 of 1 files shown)


siddharthvp:
I couldn't find a phab ticket for this, but this was announced on the Mediawiki-api-announce mailing list on 4 Feb 2020.

Amorymeltzer:
I'll look when I get a chance, but a link to the email in the archives would be fine.

siddharthvp:
https://lists.wikimedia.org/pipermail/mediawiki-api-announce/2020-February/000151.html

PR #885, opened 2020-03-12
morebits: Change status class to morebits_status from tw_status

Amorymeltzer:
Makes the library slightly more Twinkle-agnostic.  <s>Only proper use is User:SD0001/RFUD-helper.js and even that seems unnecessary given the comment; regardless, @siddharthvp is more than capable of managing any issues ;)</s> No real uses.

### DIFF
--- a/morebits.css
+++ b/morebits.css
@@ -10,19 +10,19 @@
 
 /* Morebits.status */
 
-.tw_status_status {
+.morebits_status_status {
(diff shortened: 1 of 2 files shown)


siddharthvp:
> Only proper use is User:SD0001/RFUD-helper.js and even that seems unnecessary given the comment

Removed.

PR #886, opened 2020-03-13
tag: Remove references to MI tag in edit summary

### LINKED ISSUE #718: tag: edit summaries are broken
Since #485, edit summaries generally are grammatically wrong, wherever {{multiple issues}} is being added. [Example edit](https://en.wikipedia.org/w/index.php?title=Army_Service_Component_Command&diff=prev&oldid=918863852). Curiously, in [this edit](https://en.wikipedia.org/w/index.php?title=Christopher_Appleton&diff=prev&oldid=918864777), "tag" is singular whereas in the the former edit, plural "tags" is used, though in both cases only 1 tag was added. Also, the "tag" word after "(within {{multiple issues}})" shouldn't really be there.

I made some efforts to ensure correct edit summaries in the initial few iterations of 485, but later lots of changes were made and I neglected to check up on the correctness of edit summaries.

As the codes for producing the edit summaries are quite involved and spread across multiple places, I propose simplifying them to remove mention of "(within {{multiple issues}})". I think that is a detail that isn't important enough to be in the edit summary, and this would simplify a lot of edit summary generation code.

Amorymeltzer:
Closes #718, which correctly notes that this made things (even more) overly-complicated, had some errors, and wasn't particularly meaningful.  Generally consolidates article tag generation of summaries.  Also fixes some pluralization issues and oxfordizes removal notes and extends edit summaries, as they now out at 500 not 255 (or 1000): phabricator.wikimedia.org/T188798

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1273,7 +1273,6 @@ Twinkle.tag.callbacks = {
 
 		// Remove tags that become superfluous with this action
 		var pageText = pageobj.getPageText().replace(/\{\{\s*([Uu]serspace draft)\s*(\|(?:\{\{[^{}]*\}\}|[^{}])*)?\}\}\s*/g, '');
-		var summaryText;
 		var params = pageobj.getCallbackParameters();
 
 		/**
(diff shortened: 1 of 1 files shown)


siddharthvp:
Would it make sense to centralise all the edit summary generation code in one place? Building up on your branch, I was thinking of something like (untested): 
```javascript
var generateEditSummary = function() {
	var makeSentence = function(arr) {
		if (arr.length < 3) {
			return arr.join(' and ');
		}
		var last = arr.pop();
		return arr.join(', ') + ' and ' + last;
	};

	var makeTemplateLink = function(tag) {
		var text = '{{[[';
		// if it is a custom tag with a parameter
		if (tag.indexOf('|') !== -1) {
			tag = tag.slice(0, tag.indexOf('|'));
		}
		text += tag.indexOf(':') !== -1 ? tag : 'Template:' + tag + '|' + tag;
		return text += ']]}}';
	};

	var removedTags = params.tagsToRemove.map(makeTemplateLink);
	var addedTags = params.tags.map(makeTemplateLink);

	var summaryText;
	if (addedTags.length) {
		summaryText = 'Added ' + makeSentence(addedTags) +
			(removedTags.length ? '; and removed ' + makeSentence(removedTags) : '');
	} else {
		summaryText = 'Removed ' + makeSentence(removedTags);
	}
	summaryText += ' tags to article';

	// avoid truncated summaries
	if (summaryText.length > (499 - Twinkle.getPref('summaryAd').length)) {
		summaryText = summaryText.replace(/\[\[[^|]+\|([^\]]+)\]\]/g, '$1');
	}

	return summaryText + Twinkle.getPref('summaryAd');
};
``` 

most code chunks copied from elsewhere, you'd especially recognise origin of the first one.

Amorymeltzer:
Yeah, I think something like that would be ideal.  I mostly tried to avoid breaking anything — it's wildly complex in there! — but agree that consolidating it would be best.

Amorymeltzer:
{{done}} as separate commit for review; will squash before merging.  Basically exactly what you wrote; also untested 'cause, uh, everyone's at home for some reason...

Amorymeltzer:
All commits squashed after merging in #889

PR #888, opened 2020-03-14
morebits: Use csrftoken obtained via meta=tokens

### LINKED ISSUE #615: morebits: Modernize token handling

Amorymeltzer:
There are a few connected commits here, but the meat of it is in the second commit.  Closes #615.

Using `intoken` was deprecated ages ago.  `curtimestamp` is used in `.load` now that we no longer use `intoken`.  As noted in #741, `starttimestamp` was only available because we were still using the outdated `intoken` method; with that gone, we need to explicitly provide `query.curtimestamp`, thus replicating the loading timestamp that we can feed it into `starttimestamp` and avoid edit conflicts.  Some references to token names in error messages were kept for ease of debugging.

Some token handling is changed in `block` (1st commit) and `fluff` (2nd commit).  The 3rd commit tweaks the querys to update the deprecated `watch`/`watchlist` parameter (excepting for PC).

4th commit: Use `fnCanUseMwUserToken` for move and stabilize.  First added in d37e981, reasonable to use for `move` and `stabilize` since we're not using any additional information.  Unlike with protection, flaggedrevs doesn't need any previous statuses; like protection, we don't issue alerts for acting on already-stabilized pages.  Thus, there isn't an additional need to check `wgStableRevisionId` or something.

The 5th commit is the one I'm really not sure of and wouldn't mind opinions on.  It uses a common query for non-load actions (delete, etc.) to consolidate/deduplicate the query.  `fnNeedTokenInfoQuery` takes the relevant `action`.  Of note, removes the `prop=info` check for non-sysop moves as well as all `prop=info|flagged` for `stabilize`; they were unused.

The query gets stuffed out of the way and although it should be easier to maintain, outside of token stuff like this these queries are unlikely to need a lot of future maintenance.  This is basically what Dan Abramov has written about: https://overreacted.io/goodbye-clean-code/

In theory, this means more `if`s for each action, in particular delete, and has the query even more removed from the proper context.  Another way to do this would be to set the median query but then have individual items tweak as necessary "locally" (`redirects` removed for undelete, stabilize and non-sysop move remove `prop` and `inprop`).  That keeps things somewhat closer to their relevant areas, but then again, `delete` is slow and it'd be is "harder" to "maintain."

It's live on testwiki (with an excessive amount of logging).  cc @MusikAnimal

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2418,19 +2418,7 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		if (fnCanUseMwUserToken('move')) {
 			fnProcessMove.call(this, this);
 		} else {
-			var query = {
-				action: 'query',
-				prop: 'info',
-				meta: 'tokens',
-				type: 'csrf',
-				titles: ctx.pageName
-			};
-			if (ctx.followRedirect) {
-				query.redirects = '';  // follow all redirects
-			}
-			if (Morebits.userIsSysop) {
-				query.inprop = 'protection';
-			}
+			var query = fnNeedTokenInfoQuery('move');
 
 			ctx.moveApi = new Morebits.wiki.api('retrieving token...', query, fnProcessMove, ctx.statusElement, ctx.onMoveFailure);
 			ctx.moveApi.setParent(this);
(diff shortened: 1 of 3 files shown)


morebits.js
```diff
@@ -2984,7 +2983,7 @@ Morebits.wiki.page = function(pageName, currentAction) {
 			query.noredirect = 'true';
 		}
 		if (ctx.watchlistOption === 'watch') {
-			query.watch = 'true';
+			query.watchlist = 'true';
```
```suggestion
			query.watchlist = 'watch';
```
nochange, preferences, unwatch, watch are the only allowed values. 

Here and in 4 more uses below.

Amorymeltzer: Ugh, braindumb.  Really I think it should be like in the `save`, just `watchlist: ctx.watchlistOption`

Amorymeltzer: Now done as in https://github.com/azatoth/twinkle/pull/888#discussion_r392944899.  Only exception is stabilization, which still uses the old system.

Amorymeltzer:
Tested fairly thoroughly before opening this, still haven't run into any bugs when using it.

siddharthvp:
Just noting that I'd been testing the RFUD-helper script some days back - it uses morebits.page. `load`, `save`, `prepend`, `move` and `undelete` - all of that was working smoothly.

Amorymeltzer:
@Xi-PLus Meant to ping you here earlier; I know you stay pretty on top of things for zh, but wanted to make sure you were aware of this both for there and meta.

Amorymeltzer:
Merged! 🏅 to me!

Joking aside, I'll keep testing and looking for bugs, but it'll honestly be easier to do so with it merged and not holding anything else up.

PR #890, opened 2020-03-18
tag/di: two new templates for files

siddharthvp:
Requested on [WT:TW](en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Template).


Amorymeltzer:
No worries, will do!  But do you want `s`?  'Cause that's how you get `s`! 😆

Amorymeltzer:
Not sure when I'll have a chance, but I'll reply on-wiki with my opposition to fails NFCC.  cc @JJMC89 fwiw

{{Bas SVG}} is :+1: from me.  I wonder if it'd be worth switching out jpeg/gif/svg depending on filename prefix?  Might be confusing.

Amorymeltzer:
@siddharthvp For now I've merged affb4603 while the discussion on {{di-fails NFCC}} is ongoing.

Amorymeltzer:
Closing given https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_42#Template

Issue #778, opened 2019-12-05
Twinkle-specific CSS in morebits.css


Amorymeltzer:
- [x] `morebits.js` [uses `.tw_status`...](https://github.com/azatoth/twinkle/blob/18798af0c500dc93ef40f65d5f3c8edccf026e0e/morebits.js#L3721) but shouldn't this more appropriately be named `mb_status`? (#885)
- [x] 511bd8f3 (re-)added Twinkle-specific CSS (see the `portlet` section), is there a good reason that shouldn't be in `twinkle.css`? (#893)

Amorymeltzer:
@enterprisey This was inspired by your recent WT:TW question, but I'm leaving for a few weeks so I figured I'd open this instead of looking into this further.

siddharthvp:
The `/* Portlet */` section isn't even Twinkle-specific. It affects the whole skin. It doesn't really belong anywhere. Previously noted in https://github.com/azatoth/twinkle/pull/721#issuecomment-551935531.

Amorymeltzer:
Opened PRs for both of the issues, as well as #892 for @enterprisey's specific request, so closing.

PR #893, opened 2020-03-21
morebits: Remove twinkle-specific CSS that is 100% redundant

Amorymeltzer:
See #778.  The stuff in the portlets section was all added in 511bd8f and is just for the Modern skin, but it is all already present in the main Modern CSS: https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/skins/Modern/+/master/resources/main.css.  Don't think it was even needed at the time, but regardless, does nothing and can be removed.

### DIFF
--- a/morebits.css
+++ b/morebits.css
@@ -177,49 +177,6 @@ div.morebits-previewbox .mw-editsection
 }
 
 
-
-
-/* Portlet */
-
-.skin-modern #mw_contentwrapper div.portlet {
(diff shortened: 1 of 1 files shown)


siddharthvp:
Neat! Didn't realise the fix for this would be so simple!

Amorymeltzer:
Hardest part was comparing different selector styles (well, and digging through old skin files from 2009-2011 for no reason).

PR #894, opened 2020-03-21
tag: Remove extra newline after section in merge discussions

Amorymeltzer:
This is a dumb PR, but I'm opening it rather than just doing it because I feel like I have to be missing something.  There's no visual change, but is there actually a reason the merge discussion has had two newlines for ages?  I would have thought someone would have noticed otherwise...


siddharthvp:
New sections added using the "New section" button (or using section=new API option) do have this "extra" newline. (All talk pages keep them, [example](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&action=edit)) It doesn't cause a visual change but makes the wikitext more readable.

Amorymeltzer:
:+1: Thanks I hate it

PR #895, opened 2020-03-21
lint: Enforce quote-props consistent-as-needed

Amorymeltzer:
Not done in #500 given the cosmetic nature and fairly mixed status of the project, but I've run into it a few times lately and have been mildly annoyed.  Adding `"keywords": true` seems reasonable, but it's not that common and I don't think in the end we really care.  Likewise with `numbers`.

https://eslint.org/docs/rules/quote-props

Separate commit for the actual application is just to make my life easier during the eventual rebasing.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -46,6 +46,7 @@
 		"no-trailing-spaces": "error",
 		"no-unneeded-ternary": "error",
 		"no-whitespace-before-property": "error",
+		"quote-props": ["error", "consistent-as-needed"],
 		"quotes": ["error", "single", { "avoidEscape": true }],
 		"semi": ["error", "always"],
 		"space-before-blocks": "error",
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -894,7 +894,7 @@ Twinkle.tag.redirectList = {
 		]
 	},
 	'Alternative names': {
-		'General': [
+		General: [
 			{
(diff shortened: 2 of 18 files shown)


siddharthvp:
This feels like adding rules for the sake of adding rules. At any rate I would prefer the 'consistent' rule rather than 'consistent-as-needed'.

Amorymeltzer:
I mean, isn't it all just rules for rules' sake?  I'll grant you this is beyond minor, but what was particularly bugging me was seeing, in the same module:

```js
var query = {
	'action': 'query',
	'title': '42'
};
...
var query = {
	action: 'query',
	title: '44'
};
```

Basically, going for consistency within an object and consistency throughout.  I also think it makes it obvious when the quotes are necessary:

```js
var param = {
	action: 'query',
	title: '46'
};
...
var params = {
	'Goldfinch': 'Long',
	'Lady-Astronaut': 'Short'
};
```
But, again, incredibly minor bikeshedding.

siddharthvp:
Is this eslint auto-fixable?

Amorymeltzer:
Yes

Amorymeltzer:
FWIW @siddharthvp, #1224 is (mostly) what I was waiting on for this.  I had thought to just fold it in there, would it mess you up too much?

siddharthvp:
@Amorymeltzer no

Issue #753, opened 2019-11-20
XfD: Only show valid options


DannyS712:
Currently, all XfD options are shown on pages. However,
* [?] Afd should only be shown for articles
* [?] Tfd should only be shown for templates and modules
* [x] Ffd should only be shown for files (#865)
* [x] Cfd & Cfds should only be shown for categories and templates (stubs) (#897, #926)
* [?] Rfd should only be shown for redirects (including category redirects)
* [x] RM should not be shown for categories (#865)
* [ ] <s>None</s> Only MfD should be shown for mediawiki namespace

Amorymeltzer:
So basically enforce the defaults, give or take?  I suppose I'm not opposed to it; I can see some value in leaving AfD/RfD/MfD vague-ish for edge cases, but I have no idea how often that happens.  I presume @azatoth's initial intent was ease plus entrusting the user, this goes back a decade.

>None should be shown for mediawiki namespace

MfD

DannyS712:
Okay, then leave MfD.

siddharthvp:
At least the AfD option should be set disabled if on a redirect, solving #379. And probably RM too in category namespace -- but that may be confusing and we can't put any tooltip next to a disabled option, so I think it would be better to allow the option but when selected, display some red text telling the user to use CfD/CfDS instead. We could use the red text option for former case too.

Amorymeltzer:
I'm fine with just disabled, at least on the 100% cases (which you've noted); can add a note to the dropdown tooltip that some items may be disabled.

Amorymeltzer:
#865 will do files and rm, as clearly appropriate low-hanging fruit.  I think it's worth discussing some of the others a bit, but I'm a little wary of disabling things like AfD/RfD on redirect status.  Well (or not so well) meaning users sometimes redirect/unredirect and but for a revert we occasionally end up with pages that are redirects that should head to AfD, or vice-versa.  Not opposed to it, but I think it's worth considering whether we need to be heavy-handed about it.  It's possible some of the ideas in #379 would be a better solution (should also be discussed).

PR #897, opened 2020-03-22
xfd: Enable CfD for stub templates, fix cfd(s)/tfd namespace previews

Amorymeltzer:
1st commit: Enable CfD for stub templates, disable CfD outside catspace and templatespace

Determination of a stub template made by simply asking if it is a template whose name ends in `-stub`.  Had to make {{sfd-t2}} and {{sfr-t2}} and add support for stubs to {{cfd-notify}}.  With this in place, TfD can now be disabled on such stub templates, and CfD can be disabled outside template or categor space (#753). (Leaving open for the entirety of templatespace allows [Rcat templates](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Redirects_for_discussion&oldid=949051818#What_should_be_the_venue_for_discussing_Rcat_templates?))

I'm not entirely sure how specific the `-stub` ending check is, and whether we should be disabling TfD on such a check.  AFAIK that's the only way to check based off the title.  Another option would be to check `wgCategories` for `Stub message templates`, I think that might be more accurate?  cc @siddharthvp and @DannyS712, stubs and categorization thereof aren't my area of expertise/interest.  Maybe it'd be better to just treat them both as fine anywhere in templatespace, for simplicty's sake?

2nd commit: Better fixes for namespaces in cfd/tfd/cfds previews and form replacements

Better and more complete way of doing 133d227 (#857), for both CfD and TfD, as well as some tidying for CfD/S.  Only replaces the current namepsace, which is basically what was being done anyway, but this way it's obvious when we're pigeonholing a page into an incorrect namespace.  For CfD/S, the behavior is basically inverted: instead of removing the namespace then adding it for just the discussion page, this removes it just for the tagging.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -71,7 +71,8 @@ Twinkle.xfd.callback = function twinklexfdCallback() {
 		type: 'option',
 		label: 'TfD (Templates for discussion)',
 		selected: [ 10, 828 ].indexOf(namespace) !== -1,  // Template and module namespaces
-		value: 'tfd'
+		value: 'tfd',
+		disabled: namespace === 10 && /-stub$/.test(Morebits.pageNameNorm) // Stub templates at CfD
 	});
 	categories.append({
 		type: 'option',
@@ -83,8 +84,9 @@ Twinkle.xfd.callback = function twinklexfdCallback() {
 	categories.append({
 		type: 'option',
 		label: 'CfD (Categories for discussion)',
-		selected: namespace === 14,  // Category namespace
-		value: 'cfd'
+		selected: namespace === 14 || (namespace === 10 && /-stub$/.test(Morebits.pageNameNorm)),  // Category namespace and stub templates
+		value: 'cfd',
+		disabled: [ 10, 14 ].indexOf(namespace) === -1 // Disabled outside category and templatespace
 	});
 	categories.append({
 		type: 'option',
@@ -299,7 +301,7 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 			});
 			work_area.append({
 				type: 'div',
-				label: 'Stub types and userboxes are not eligible for TfD. Stub types go to CfD, and userboxes go to MfD.'
+				label: 'Userboxes are not eligible for TfD; they go to MfD.'
 			});
 			var templateOrModule = mw.config.get('wgPageContentModel') === 'Scribunto' ? 'module' : 'template';
 			var tfd_category = work_area.append({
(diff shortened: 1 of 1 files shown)


siddharthvp:
Looks like a giant PITA that complicates the code a great deal and makes i18n even harder than it already was.

> I'm not entirely sure how specific the -stub ending check is, and whether we should be disabling TfD (or even CfD) on such a check. AFAIK that's the only way to check based off the title. Another option would be to check wgCategories for Stub message templates,

Yes the "-stub" ending is pretty damn specific. Checking for the category wouldn't catch non-standard templates like {{[Lepidoptera-stub](https://en.wikipedia.org/wiki/Template:Lepidoptera-stub)}}, which are more often are the ones that need to be sent to CfD. 

> With this in place, ... CfD can be disabled outside stub templates or categoryspace (#753).

[As you're aware](https://en.wikipedia.org/wiki/Wikipedia_talk:Redirects_for_discussion#What_should_be_the_venue_for_discussing_Rcat_templates?), there are a lot of editors who feel CfD is an appropriate venue for rcat templates. I don't like where #753 seems to be heading towards.  

----

I don't see an option for nominating stub templates for merging, but this is necessary. Not all stub templates have an associated category which could be nominated instead.

Amorymeltzer:
Indeed, I'm fairly skeptical of a good portion of #753, for exactly the fact that there's a lot of nuance in the edge cases; beyond even the rcat issue, I wouldn't complain about a userspace template being sent to TfD, for example.

>I don't see an option for nominating stub templates for merging, but this is necessary.

No template -> no option!

At any rate, I think i18n is largely a red herring here: anything that can handle the great variety in how projects handle deletion would be able to handle something as comparatively simple as this.  That is, XfD has much bigger problems.

Amorymeltzer:
@siddharthvp Thinking I might have misread your initial comment.  Is

>Looks like a giant PITA that complicates the code a great deal and makes i18n even harder than it already was.

Suggesting to decline or commiserating in the (increasingly complicated) state of things?

siddharthvp:
I'm not sure how bad the current situation is. I don't want to suggest declining considering you've already put so much work into this :) Another idea would be to internally treat stub tags as a separate XfD board - while this will get rid of the complexity, it will cause fair bit of duplication.

If this is going ahead, I still think there should be an option to nominate stub tags for merging, since this option is presently available (as CfM). I think {{sfm-t}} and {{sfm-t2}} can be created by copy-pasting from the sfr templates and simply changing "renaming" to "merging"?

Amorymeltzer:
Do folks really just say "ah screw it use cfm"?  Then yeah.  TBH I had thought the area was a little more developed, was surprised at lack of templates and something like merges.  Probably should bring that up at WT:CFD first since that would involve a change in behavior on the part of participants.

Amorymeltzer:
Didn't see further discussion, so brought 'em up at WT:CfD.

Also, changed the cfd disabling to be outside 10 and 14, to allow for rcats in addition to stub templates.

Amorymeltzer:
Only one comment in the last week or so, and that suggesting not worth it, so merged this, can add merges if/when necessary/feasible.

PR #900, opened 2020-03-24
morebits.css: Color red links red

siddharthvp:
Requested on [WT:TW](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Red_links)


Amorymeltzer:
Just seeing this now on mobile, looks like we were on this at the same time!

siddharthvp:
Lol, [Modi has just declared](https://www.nytimes.com/2020/03/24/world/asia/india-coronavirus-lockdown.html) everyone should stay at home so what else is there to do?

PR #901, opened 2020-03-24
morebits: Preserve link coloring in previewbox

Amorymeltzer:
Makes use of `:not()` to skip link coloring in `.preview-box`.  It looks like `.ui-dialog.morebits-dialog a` only had the `titlebar-close` elements, which don't need the coloring style, so we can remove that.

### DIFF
--- a/morebits.css
+++ b/morebits.css
@@ -294,7 +294,8 @@ body .ui-dialog.morebits-dialog .morebits-dialog-footerlinks a {
 	margin: .1em .4em -.2em 0;
 }
 
-.ui-dialog.morebits-dialog a, .ui-dialog.morebits-dialog .ui-widget-content a {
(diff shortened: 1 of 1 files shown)


siddharthvp:
Seems like a better way of doing it than #900, but from my [test edit on morebits.css](https://test.wikipedia.org/w/index.php?title=MediaWiki:Gadget-morebits.css&diff=cur&oldid=prev), it seems that no CSS rule is needed at all? I only see the "ridiculous nearly-black colour" on `unset`ting the original CSS rule (not when removing it altogether).

Amorymeltzer:
Same.  I want to double check all the mediawiki CSS but that's what I was seeing last night as well.  Being home with a toddler all day isn't helping productivity... 😂

Amorymeltzer:
Closed in favor of #904

PR #905, opened 2020-03-29
tag: Exclude {{Bare URLs}} from being put in {{multiple issues}}

QEDK:
Requested at WT:TW


Amorymeltzer:
[Archived without further comment](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&diff=951830933&oldid=prev), so I'm not sure what to make of it.

Amorymeltzer:
There wasn't any follow-up, so if you don't mind terribly, I'll close this.

QEDK:
> There wasn't any follow-up, so if you don't mind terribly, I'll close this.

That's probably the right course of action, if anyone complains, we can look into this again.

PR #907, opened 2020-03-29
morebits: Some tweaks to Morebits.date

Amorymeltzer:
As promised at https://github.com/azatoth/twinkle/pull/814#event-3176003347, opened for discussion.  Commit are separate for review, and ordered from what I consider most to least worthwhile:

1st commit: Use YYYY-MM-DD for Morebits.date in calendar localeData

`moment` on enWiki returns `DD/MM/YYYY` but that's just a locale specification I believe.  Given the US-centric-but-not-exclusive enWiki population, both `DD/MM` and `MM/DD` can lead to confusion (the MOS, for example, suggests avoiding these uses).  Doing `YYYY-MM-DD` should avoid any such confusion.

2nd commit: Allow adding/subtracting seconds

<s>3rd commit: Add `isSame` to Morebits.date to compliment `isAfter` and `isBefore`</s>

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1363,7 +1363,7 @@ Morebits.date.localeData = {
 		nextDay: '[Tomorrow at] h:mm A',
 		thisWeek: 'dddd [at] h:mm A',
 		pastWeek: '[Last] dddd [at] h:mm A',
-		other: 'DD/MM/YYYY'
+		other: 'YYYY-MM-DD'
 	}
 };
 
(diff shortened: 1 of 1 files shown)


siddharthvp:
The first commit is great. I'm mostly meh on the other two - especially the third one since `isSame` will only return true if the dates are same to the millisecond. Not likely to be of much use; even if it is, the functionality is already available by using the underlying date object (`_d`) directly.

Amorymeltzer:
Indeed.  Moreover, don't some browsers treat milliseconds differently?  No links as I'm on mobile but IIRC it's Firefox that fuzzes them so seeding for hashes is more secure.

Amorymeltzer:
1&2 squashed&merged

PR #909, opened 2020-03-31
Display G13 on Redirects

Thepenguin9:
G13 is still applicable in draft and userspace redirects, this will avoid having to tag with G6 or manually.

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1057,7 +1057,7 @@ Twinkle.speedy.generalList = [
 		label: 'G13: Page in draft namespace or userspace AfC submission, stale by over 6 months',
 		value: 'afc',
 		tooltip: 'Any rejected or unsubmitted AfC submission in userspace or any non-redirect page in draft namespace, that has not been edited for more than 6 months. Blank drafts in either namespace are also included.',
-		hideWhenRedirect: true,
+		hideWhenRedirect: false,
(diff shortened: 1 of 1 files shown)


siddharthvp:
[WP:G13](https://en.wikipedia.org/wiki/Wikipedia:Criteria_for_speedy_deletion#G13._Abandoned_Drafts_and_Articles_for_creation_submissions) clearly states "Redirects are exempt from G13 deletion".

Thepenguin9:
I spoke with an admin on the IRC who said that if the draft is stale regardless, its still eligible for G13, and all draft redirects have been deleted under G13. 

[This log](https://en.wikipedia.org/w/index.php?title=Special:Log&page=Draft:Jean_de_Forcade_de_Biaix) shows that the page was redirected, and now was deleted despite being a redirect.

Thepenguin9:
I have noticed the multiple criteria tag allows use of G13 in draft space also

Amorymeltzer:
Yeah, G13 doesn't apply to redirects.  Regarding the sysop you spoke with, I'm not sure what they said, but that deletion has been undone.

>I have noticed the multiple criteria tag allows use of G13 in draft space also

Do you mean [{{db-multiple}}](https://en.wikipedia.org/wiki/Template:Db-multiple)?  That doesn't include any sort of checks for whether a page is a redirect or in the right namespace.  Twinkle won't let you do it, though.

Issue #364, opened 2016-11-17
Twinkle should purge pages after filing an XfD.


super3588:
Whenever I start an XfD (in parrticular, an AfD), I use Twinkle to file it.  Unfortunately, one downside is that all the pages are modified at the same time (or so it seems), so, when the AfD template appears on the page I've nominated, the part that says **this article's entry** is red when it should be blue.  The same thing happens to the user talk notification Twinkle sends.  Currently, the only way I know of to fix this is to manually purge the page.  I usually do this using the _Add a "Purge" option to the top of the page, which purges the page's cache_ option in Special:Preferences under Gadgets.  It would be great if I didn't have to go manually purge pages every time I XfDed something.

Dinoguy1000:
This could probably be fixed without requiring extra purging just by having Twinkle be more careful about what order it saves pages in, and perhaps by requiring it to get confirmation that the AfD page has been created before adding notices.

Amorymeltzer:
I haven't been able to replicate this, does this still occur regularly?

PR #911, opened 2020-03-31
morebits: convert Morebits.wiki.page to use promises

### LINKED ISSUE #783: Morebits.wiki.page: overhaul callback arguments and enable use of promises
Arguments made available to callback functions in `Morebits.wiki.page` at present are pathetically inconsistent (noted [here](https://github.com/azatoth/twinkle/blob/master/modules/twinklebatchdelete.js#L533) by @atlight). The argument passed can be the the page object (`Morebits.wiki.page`) or api object (`Morebits.wiki.api`) depending on what function it is and whether it's the success or failure callback. In some cases, the `ctx`object (which is [supposed to be private](https://github.com/azatoth/twinkle/blob/master/morebits.js#L1628)) itself becomes available as the `this` reference in the callback! 

For the many cases where the page object is passed to the fail callback, it is not possible for client code to get to know what the error is (the errorCode being in the api object) - which makes it impossible to do anything on the basis of the type of error (such as show a customised error message or do a retry. There are usecases within Twinkle for this - such as #176). 

My plan to clean up the mess is: 

- Never pass the api object to any callback or as `this`. This is in line with `Morebits.wiki.page` implementation strategy of keeping private stuff private. This allows us to make internal changes to 
the class without fear of breaking existing code that could be making use of internal variables. 
- Rather, give the errorcode and the errortext as the 2nd, 3rd arguments, in addition to the page object which is _always_ the first argument. 

I earlier did some work in fixing this in #679 - the approach there being simply to try give the api object every time. I now think that this is the wrong way to go because (i) api object is supposed to be private, (ii) sometimes errors occur before we use any API call - such as when a non-admin tries to `protect`, or edit summary is not set before a `save`. In such cases, there is no api object to pass. Passing the page object every time - for success and all types of failures - gives a clean and consistent interface. Where no API call was made before returning, we make up our own error code.

It makes sense to deal with this at the same time as converting `Morebits.wiki.page` to use promises, because both require painful testing (especially in light of the incredibly unpredictable nature of `this` in javascript) I will be writing tests along the way. Outside Morebits, in the modules, there is very little that would need to be changed, because use of callbacks will still be supported alongside promises, and callback argument changes would require only require minimal tweaks here and there. 

I already have some work on this ([example commit](https://github.com/siddharthvp/twinkle/commit/50c1814bb38999b016271b1aeeb712b3a30e5cc6#diff-371d1913aa0a4d389d417d9bfda0e779R2381-R2384)), and plan to complete it and open a PR after #765 is merged. Opening the issue now in case people want to discuss the callback overhaul stuff - cc @atlight @MusikAnimal @Amorymeltzer.

siddharthvp:
Includes the commits of pre-requisite #765.

Closes #783. 

All functions in Morebits.wiki.page are updated to support promises. The existing callback scheme continues to work. Also, fixes the issues with inconsistency in callback arguments as mentioned in #783. 

It's worth noting that after @Amorymeltzer's #712, there is a remarkable structural similarity across most functions.

The `stabilize` function is untested, as FlaggedRevs is not there on the testwiki, so I have no way to test it.

~~I've added "tests" as morebits-test-promises.js (last commit). These tests aren't automated - you've copy paste each code chunk to the console to see that that chunk is working as mentioned. Suitable for running on the testwiki only. Converting these to unit tests shouldn't be that difficult, though I couldn't get it right in an hour and a half.~~ Unit tests added, see https://github.com/azatoth/twinkle/pull/911#issuecomment-617343034

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,7 +9,7 @@
 	"globals": {
 		"mw": "readonly"
 	},
-	"ignorePatterns": "select2*",
+	"ignorePatterns": ["select2*", "server.js"],
 	"rules": {
 		"no-console": "error",
 		"no-extra-parens": ["error", "all", { "nestedBinaryExpressions": false }],
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,6 @@
 		"eqeqeq": "error",
 		"guard-for-in": "error",
 		"no-caller": "error",
-		"no-else-return": "error",
 		"no-implicit-coercion": ["error", { "boolean": false }],
 		"no-lone-blocks": "error",
 		"no-multi-spaces": ["error", { "ignoreEOLComments": true }],
--- a/morebits-test-promises.js
+++ b/morebits-test-promises.js
@@ -1,0 +1,259 @@
+/* global Morebits */
+/* eslint-disable no-redeclare, brace-style, no-console */
+
+
+/**
+ * Snippets of testing code that you can copy-paste into the console.
+ * TODO: Convert to unit tests?
+ */
+
+(function() {
+
+// callbacks passed in constructor:
+var successCallback = function() {
+	console.log('callback-onSuccess', this, arguments);
+};
+var failureCallback = function() { console.log('callback-onFailure', this, arguments); };
+
+// promise callbacks:
+var doneCallback = function() { console.log('done', this, arguments); };
+var thenCallback = function() { console.log('then', this, arguments); };
+var failCallback = function() { console.log('fail', this, arguments); };
+var catchCallback = function() { console.log('catch', this, arguments); };
+var alwaysCallback = function() { console.log('always', this, arguments); };
+
+
+/** Creates a simple window that shows Morebits status messages */
+var initStatusWindow = function() {
+	if ($('#morebits-dialog-content').length && !Morebits.status.root) {
+		Morebits.status.init($('#morebits-dialog-content')[0]);
+	} else {
+		var w = new Morebits.simpleWindow(600, 600);
+		w.display();
+		Morebits.status.init(w.content);
+	}
+};
+
+
+// MOREBITS.WIKI.API
+
+initStatusWindow();
+
+/* disable internet from the network tab to trigger the other type of error */
+var apiobj = new Morebits.wiki.api(
+	'',
+	{
+		action: 'query', /* change to something random to make the API return an error */
+		titles: 'Main Page'
+	},
+	successCallback, null, failureCallback
+).post();
+
+apiobj.then(thenCallback);
+apiobj.done(doneCallback);
+apiobj.fail(failCallback);
+apiobj.catch(catchCallback);
+apiobj.always(alwaysCallback);
+
+/**
+ * Chainings: note on jQuery deferred handler types:
+ * .then() and .catch() return *new* promises, that may be resolved or rejected, according to
+ * the return value of the function in the handler
+ * .done() and .fail() simply return the original promise, regardless of what the function within
+ * the handler returns
+ */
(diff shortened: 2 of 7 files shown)


siddharthvp:
I at last figured out a nice scheme for doing unit testing. Will add the tests in a few days or weeks. Suggest not reviewing this until then, as this would be a lot easier to review then.

Amorymeltzer:
Neat!  I've been mulling some tests for `morebits`, it's the only bit that's reasonably amenable to it, but have been preoccupied.  I'll be curious to see what you did.

siddharthvp:
Ok, I've now added the unit tests. How to run:

- Run `npm run serve` to start up the server. It's a simple Node.js app, `server.js`.
- Add `mw.loader.load('http://127.0.0.1:5500/twinkle.testloader.js');` in your on-wiki common.js page. (Or just enter that into the console.)
- Click on the "Run Twinkle unit tests" button at the top. Do this while on a wiki-page that exists (some of the lazily-written tests depend on this.) 

Do this on the testwiki, though the tests which create pages also do cleanup after themselves by deleting the pages afterwards. Others edit the user's sandbox. I've created [Wikipedia:Twinkle/ExampleNonProtectedPage](https://test.wikipedia.org/wiki/Wikipedia:Twinkle/ExampleNonProtectedPage) and [Wikipedia:Twinkle/ExampleProtectedPage](https://test.wikipedia.org/wiki/Wikipedia:Twinkle/ExampleProtectedPage) used in a couple of tests. 

We are using browser-based unit testing with the content of workspace files made available in the browser through the localhost server. Any attempt to do node-based unit testing would require trying to replicate the MW environ (mw.config etc etc) which is extremely difficult.

Mocha is used as the testrunner and chai as the assertion library, loaded over CDNs for now. We might as well keep a copies in the repo and load using the server. 

All 33 of the tests are passing. Yippee!

Amorymeltzer:
Really cool!  I'm not particularly familiar with mocha, but this looks pretty great to me.  I want to get these out next month, so I'll comment more then, but love it.

I can say from personal experience that *trying to replicate the MW environ (mw.config etc etc)* is indeed *extremely difficult*. 😭 😆 😭

Amorymeltzer:
Rebased, merged, fixed some errors

Amorymeltzer (changes requested):
@siddharthvp I don't know what to say on this except it is really very excellent.  Promises are probably the hardest thing to wrap my head around, all the more so via jQuery, so I'm really very impressed.  The testing regime is great AFAICT, and is a really great start.  I'd really like to go further down this road and flesh it out, this is a really nice fire to get going with!  And wow does it speed things up!  We might need to finally do some purging (e.g. #364) to play catch up.

Unfortunately, I don't think I can merge this.  It suffers pretty heavily from the same issue as #1036, where it can redirect to `//wiki//wiki//wiki/PAGENAME`.  Maybe the intersection will help debug both of them, or at least lead us to figure out how to rewrite `Morebits.wiki.actionCompleted`.  Presumably related, but a single tagging can lead to several pairs of "Tagging completed" messages (e.g. three pairs).

siddharthvp:
This is hopeless. I haven't really tested again, but it seems the way to go ahead with this is to not use Morebits.wiki.actionCompleted.* The alternative to that is Morebits.taskManager(), which requires tasks to return promises in the first place. `$.Deferred()`s could be used for that. But still it involves refactoring every single module. And we're talking about refactoring within _this_ twinkle, as if all the refactoring needed for the twinkle-core effort wasn't enough. All the merge conflicts which have accumulated in the meanwhile aren't helping as well. The #1029 path appears much more appealing – no burden of backward compatibility.

Looks like this can be closed, unless @Amorymeltzer you have ideas on getting `actionCompleted` to work alongside promises.

Issue #627, opened 2019-05-09
Unable to see rollback links on diff pages


Gazoth:
Rollback links are not displaying on diff pages since the last update. I could see the links till around 16:00 UTC on 8 May, but I've not been able to see them since then. I've tried disabling all beta features, all custom JS scripts, all other gadgets except Twinkle, switching to a different browser, disabling all add-ons, but none of it made a difference. Explicitly adding showRollbackLinks pref in twinkleoptions.js also didn't help. Twinkle itself is loading, as I'm able to see "[welcome]" button for new users on the diff page. The rollback options show up just fine on contributions page, the issue is visible only on the diff page.

I don't know if the browser version will make a difference, but I can see the issue on both Chrome 74.0.3729.131 and Firefox 66.0.4.

Amorymeltzer:
@Gazoth Can you paste an example diff link where the agf/rollback/vandal links don't show up but should?  They show up fine for me.

Gazoth:
[Here is an example](https://en.wikipedia.org/w/index.php?title=BrahMos&diff=896161533&oldid=896158762&diffmode=source). I can see the welcome link as the editor of this revision is new, but I cannot see the rollback links. Also, I can see "restore this revision" link on oldid pages.

Amorymeltzer:
@Gazoth That diff doesn't include the latest revision of the page; the agf/roll/vandal links  only show up when you're looking at the latest revision on the right-hand side of the two-column diff.  Try [this link](https://en.wikipedia.org/w/index.php?title=BrahMos&diff=prev&oldid=896346291).

Gazoth:
I can see the issue in the link that you provided too. Even if the revision is not the newest, I should have been able to see "restore this revision" link, but I'm unable to see that too.

Amorymeltzer:
Okay then, can you copy the url of the latest diff of the article Macbeth and paste it here?  That might help.  Are you using the visual editor?

Additionally, when you have a moment, can you get a few pieces of information for me from [your preferences](https://en.wikipedia.org/wiki/Special:Preferences)?  It might help me replicate your issue.
- From the Appearance tab: your skin, and which boxes under **Diffs** you have checked
- From the Editing tab: whether you have the toolbar or "disable visual editor" boxes checked
- From the Gadgets tab: all boxes you have checked
- From the Beta tab: all boxes you have checked

Gazoth:
[...]

PR #914, opened 2020-04-07
fluff: Fix revert coming from contribs if diffonly is used

### LINKED ISSUE #913: Can't rollback from contributions with "Do not show page content below diffs" option enabled
Step to reproduce:
1. Enable "Do not show page content below diffs" in your Preferences.
2. Rollback edits from contributions.

Related code:
https://github.com/azatoth/twinkle/blob/b0fab3cddeb61284a9dd136c6b149d47e80795a4/modules/twinklefluff.js#L90-L95

This is due to [phab:T231744](https://phabricator.wikimedia.org/T231744). `mw.config.get('wgRevisionId')` returns 0 when page content is unavailable (`diffonly=1`).

Suggestion:
Append `&diffonly=0` in rollback links.

Amorymeltzer:
Fixes #913, same basic issue as #627/#628: if a user has "Do not show page content below diffs" selected in Special:Preferences, wgRevisionId is set to 0 on diffs.  While #628 took care of that for the links on diff pages themselves, trying to rollback from a contribs page was left unfixed.  Those links load the diff page then `auto`matically revert, so will run into the same issue.  This essentialy reverts back to the check before 3e5ec5b2 (#561), albeit by using `mw.util.getParamValue` rather than the no-longer-existing `Morebits.queryString.get` (#725).

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -88,7 +88,7 @@ Twinkle.fluff.restoreThisRevision = function (element, revType) {
 
 
 Twinkle.fluff.auto = function twinklefluffauto() {
-	if (mw.config.get('wgRevisionId') !== mw.config.get('wgCurRevisionId')) {
+	if (parseInt(mw.util.getParamValue('oldid'), 10) !== mw.config.get('wgCurRevisionId')) {
 		// not latest revision
 		alert("Can't rollback, page has changed in the meantime.");
 		return;


Amorymeltzer:
Had a quick moment during the little beastie's afternoon nap, @Xi-Plus, this should take care of it.  Just put up on enwiki.

Issue #912, opened 2020-04-01
When blocking a user whose talk page is a redirect, the block message is sent to the redirect target


wiki-ST47:
Greetings.

Scenario:
* "User A" forwards their talk page to "User talk:User B"
* An admin blocks User A using Twinkle with the default settings

Result:
* A block message is left on User talk:User B

Expected result:
* A block message is left on User talk:User A

I assume the default behavior is in place to follow redirects e.g. from a bot to the bot operator, or from an alt account to the master. Unfortunately, some LTAs have figured this out, and find it amusing to redirect their talk pages to those of legitimate users. I've noticed this a few times.

It appears that this can be solved by deleting the following line, from https://github.com/azatoth/twinkle/blob/master/modules/twinkleblock.js#L1476

    wikipedia_page.setFollowRedirect(true);

The same thing probably occurs for other warnings and notifications as well. I think that at minimum, warnings and block notices should never follow redirects, due to the abuse risk. Maybe the same for CSD/XFD/prod notices as well, but warnings and block notices are the biggest issue.

wiki-ST47:
Note: Previously disclosed here: https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_41#Twinkle_shouldn't_follow_redirects_when_delivering_warnings

And the recent examples are:
* https://en.wikipedia.org/w/index.php?title=User_talk:Suffusion_of_Yellow&action=history , due to https://en.wikipedia.org/w/index.php?title=User_talk:Suffusion_of_Yellew&oldid=948319452 and https://en.wikipedia.org/w/index.php?title=User_talk:Suffusien_of_Yellow&oldid=948435851
* https://en.wikipedia.org/w/index.php?title=User_talk:ToBeFree&action=history , due to https://en.wikipedia.org/w/index.php?title=User_talk:ToBoFree&oldid=948437644

Amorymeltzer:
Very lame.

I think I'm softening a bit, though I generally still think it likely that misuse is much rarer than correct use (see also #83).  Not sure there's a good way to get numbers on this.

Another thought is to introduce a `confirm`ation or something.  It could be specific to the block and warn modules, or even be part of the `fnCheckPageName` in morebits that generates "Info" status line.

siddharthvp:
> Not sure there's a good way to get numbers on this.

- [745 base user talk pages redirects to outside user talk space, excluding redirects to own user page](https://quarry.wmflabs.org/query/43743)
[...]

PR #915, opened 2020-04-09
deletion modules: Don't follow cross-ns redirect while notifying user

siddharthvp:
1st commit: add ability in morebits to detect cross-ns redirects and treat them separtely. 

2nd commit: apply that in xfd, speedy, prod, and image modules.

Partially addresses #912.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2005,6 +2005,7 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		maxConflictRetries: 2,
 		maxRetries: 2,
 		followRedirect: false,
+		followCrossNsRedirect: true,
 		watchlistOption: 'nochange',
 		creator: null,
 		timestamp: null,
@@ -2398,14 +2399,20 @@ Morebits.wiki.page = function(pageName, currentAction) {
 	 *     true  - a maximum of one redirect will be followed.
 	 *             In the event of a redirect, a message is displayed to the user and
 	 *             the redirect target can be retrieved with getPageName().
-	 *     false - the requested pageName will be used without regard to any redirect (default).
+	 *     false - (default) the requested pageName will be used without regard to any
+	 *             redirect.
+	 * @param {boolean} followCrossNsRedirect
+	 *      Not applicable if followRedirect is not set true.
+	 *      true - (default) follow redirect even if it is a cross-namespace redirect
+	 *      false - don't follow redirect if it is cross-namespace, edit the redirect itself
 	 */
-	this.setFollowRedirect = function(followRedirect) {
+	this.setFollowRedirect = function(followRedirect, followCrossNsRedirect) {
 		if (ctx.pageLoaded) {
 			ctx.statusElement.error('Internal error: cannot change redirect setting after the page has been loaded!');
 			return;
 		}
 		ctx.followRedirect = followRedirect;
+		ctx.followCrossNsRedirect = followCrossNsRedirect;
 	};
 
 	// lookup-creation setter function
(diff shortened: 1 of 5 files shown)


siddharthvp:
When a cross-ns redirect is detected, what this patch does is to ignore the redirect and edit the page anyway. Need to discuss whether to use other ways of handling the situation:
- should we abort the edit instead of editing the redirect
- or, wait for it, should we edit out the redirect and put the message there?

siddharthvp:
Also noting that this adds a new dependency - `mw.Title` - on morebits gadget.

Amorymeltzer:
> When a cross-ns redirect is detected, what this patch does is to ignore the redirect and edit the page anyway. Need to discuss whether to use other ways of handling the situation:
> 
> * should we abort the edit instead of editing the redirect
> * or, wait for it, should we edit out the redirect and put the message there?

My instinct would be abort, since editing the redirect page would mean not following the redirect, which the user/tool specifically requested be done.

morebits.js
```diff
@@ -2928,11 +2939,23 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+				}
+
+				// only notify user for redirects, not normalization
 				Morebits.status.info('Info', 'Redirected from ' + ctx.pageName + ' to ' + resolvedName);
+				ctx.pageName = resolvedName; // update to redirect target
+
+			} else {
+				ctx.pageName = resolvedName;  // update in case of normalization
 			}
-			ctx.pageName = resolvedName;  // always update in case of normalization
```
Amorymeltzer, line 2935: Couldn't this just stay here instead of being repeated above?  I suppose it's clearer this way.

Amorymeltzer:
Fixed up the conflicts in xfd from #839, but did some more testing and I think morebits needs to run `onFailure(this)` in https://github.com/azatoth/twinkle/pull/915/files#diff-371d1913aa0a4d389d417d9bfda0e779R3053-R3056.  I want to do some more testing with that, though.

morebits.js
```diff
@@ -3035,11 +3046,23 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
 			var resolvedName = $(xml).find('page').attr('title');
 
-			// only notify user for redirects, not normalization
 			if ($(xml).find('redirects').length > 0) {
+				// check for cross-namespace redirect:
+				var origNs = new mw.Title(ctx.pageName).namespace;
+				var newNs = new mw.Title(resolvedName).namespace;
+				if (origNs !== newNs && !ctx.followCrossNsRedirect) {
+					ctx.statusElement.error(ctx.pageName + ' is a cross-namespace redirect to ' + resolvedName);
+					return;
```
Amorymeltzer, line 3055:
```suggestion
					onFailure(this);
					return;
```

Yeah?

Amorymeltzer:
After [the rebase](https://github.com/azatoth/twinkle/compare/e239e8e4fba580dadc83638c7ed252765861b706..e888bda7b6623ded58924e493cbf0dffa2514179#diff-371d1913aa0a4d389d417d9bfda0e779), this no longer works?  I haven't examined too deeply @siddharthvp, maybe it's #765?  Not sure what else it would be.

morebits.js
siddharthvp: Yep

siddharthvp:
Fixed. It was checking for `ctx.followCrossNsRedirect` instead of `!ctx.followCrossNsRedirect` in fnCanUseMwToken.

morebits.js
siddharthvp: Yes it could. Removed the repetition.

morebits.js
```diff
@@ -3035,11 +3046,22 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		if ($(xml).find('page').attr('title')) {
 			var resolvedName = $(xml).find('page').attr('title');
 
-			// only notify user for redirects, not normalization
 			if ($(xml).find('redirects').length > 0) {
+				// check for cross-namespace redirect:
+				var origNs = new mw.Title(ctx.pageName).namespace;
+				var newNs = new mw.Title(resolvedName).namespace;
+				if (origNs !== newNs && !ctx.followCrossNsRedirect) {
+					ctx.statusElement.error('Aborted: ' + ctx.pageName + ' is a cross-namespace redirect to ' + resolvedName);
```
Amorymeltzer, line 3054:
```suggestion
					ctx.statusElement.error('ctx.pageName + ' is a cross-namespace redirect to ' + resolvedName + ', aborted edit');
```

I think something like this, closer to the original, would be better; otherwise, you get a double colon, e.g. `Notifying initial contributor (<username>): Aborted: User talk:<pageName> is a cross-namespace redirect to <resolvedName>`

Amorymeltzer:
Merged via b48f48e, awesome!

PR #916, opened 2020-04-12
sync.pl: Change how push/pull/deploy parameters are used

Amorymeltzer:
I'm the only one who really uses this, but I wanted to open a PR to at least document the fact that the way parameters should be given will change.

1. Rather than `--deploy` or `--push`, you now do `--mode=deploy`, etc.
1. The `base` parameter is now a prefix, so if you want to reference user subpage, you should do `--base=User:AzaToth/` (with the trailing `/`)
1. Duplicate parameters are now okay; the last one provided will be used

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -58,39 +58,40 @@ Each Twinkle module and dependency lives on the wiki as a separate file. The lis
 
 ### Synchronization using `sync.pl`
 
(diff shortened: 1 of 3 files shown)


QEDK:
It might be worthwhile automating this with Travis and versioning every release with a tag, which will trigger deployment.

Amorymeltzer:
@QEDK Yes, that's 100% on my todo list.  A little delayed, but after this next update, I'm hoping to get some time to discuss/move that way.

PR #917, opened 2020-04-12
Add pagetriage in place of patrol for NPP

Amorymeltzer:
A bunch of improvements growing out of the discussion at [Wikipedia talk:New pages patrol/Reviewers#Advice on deletions](http://en.wikipedia.org/wiki/Special:Permalink/950279840#Advice_on_deletions).

1. morebits: Fix `.patrol()` by grabbing/querying for the `revid` if `rcid` is unavailable.  Per @siddharthvp at WT:NPPR, this is a major source of missed patrols.  `revid` wasn't available as an option when this was added in cae9a73 (Apr 2011) and it hasn't been touched since.
1. morebits: `getPageID` getter for `pageid`, matches `wgArticleId`
1. morebits: Add triage/curation action to mark pages as reviewed with PageTriage extension.  Uses above getter, ignores errors like `.patrol`.  Requested at WT:NPPR so as to remove from both Special:NewPages and Special:NewPagesFeed.
1. xfd/csd/tag: Use `triage` instead of `patrol`
1. prod: Add option to `triage`, basically just mirroring CSD behavior (off by default).

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1858,6 +1858,8 @@ Morebits.wiki.api.setApiUserAgent = function(ua) {
  *
  * patrol(): Patrols a page; ignores errors
  *
+ * triage(): Marks page as reviewed using PageTriage, which implies patrolled; ignores errors
+ *
  * deletePage([onSuccess], [onFailure]): Deletes a page (for admins only)
  *
  * undeletePage([onSuccess], [onFailure]): Undeletes a page (for admins only)
@@ -2031,6 +2033,8 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		moveProcessApi: null,
 		patrolApi: null,
 		patrolProcessApi: null,
+		triageApi: null,
+		triageProcessApi: null,
 		deleteApi: null,
 		deleteProcessApi: null,
 		undeleteApi: null,
@@ -2647,6 +2651,38 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		}
 	};
 
+	/**
+	 * Marks the page as reviewed by the PageTriage extension
+	 * https://www.mediawiki.org/wiki/Extension:PageTriage
+	 *
+	 * Referred to as "review" on-wiki
+	 *
+	 * Will, by it's nature, mark as patrolled as well.
+	 *
+	 * Doesn't inherently rely on loading the page in question; simply
+	 * passing a pageid to the API is sufficient, so in those cases just
+	 * using Morebits.wiki.api is probably preferable.
+	 *
+	 * No error handling since we don't actually care about the errors
+	 */
+	this.triage = function() {
+		if (!Morebits.userIsSysop && !Morebits.userIsInGroup('patroller')) {
+			return;
+		}
+
+		// If on the page in question, don't need to query for page ID
+		if (new mw.Title(Morebits.pageNameNorm).getPrefixedText() === new mw.Title(ctx.pageName).getPrefixedText()) {
+			ctx.pageID = mw.config.get('wgArticleId');
+			fnProcessTriage(this, this);
+		} else {
+			var query = fnNeedTokenInfoQuery('triage');
+
+			ctx.triageApi = new Morebits.wiki.api('retrieving token...', query, fnProcessTriage);
+			ctx.triageApi.setParent(this);
+			ctx.triageApi.post();
+		}
+	};
+
 	// |delete| is a reserved word in some flavours of JS
 	/**
 	 * Deletes a page (for admins only)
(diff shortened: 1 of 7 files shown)


morebits.js
```diff
@@ -2625,6 +2601,45 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+	 * Marks the page as patrolled, using rcid (if available) or revid
+	 *
+	 * Patrolling as such doesn't need to rely on loading the page in
+	 * question; simply passing a revid to the API is sufficient, so in
+	 * those cases just using Morebits.wiki.api is probably preferable.
+	 *
+	 * No error handling since we don't actually care about the errors
+	 */
+	this.patrol = function() {
+		if (!Morebits.userIsSysop && !Morebits.userIsInGroup('patroller')) {
```
siddharthvp, line 2627: Most wikis don't have the patroller group. Would it be worth adding a `mw.config.get('wgDbName') === 'enwiki'` check here?

morebits.js
```diff
@@ -2625,6 +2601,45 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+			return;
+		}
+
+		// If patrolling current page, don't need to query for latest revID
+		if ($('.patrollink').length) {
+			var patrolhref = $('.patrollink a').attr('href');
+			ctx.rcid = mw.util.getParamValue('rcid', patrolhref);
+			fnProcessPatrol(this, this);
+		} else if (Morebits.string.toUpperCaseFirstChar(mw.config.get('wgPageName')).replace(/ /g, '_').trim() ===
+			Morebits.string.toUpperCaseFirstChar(ctx.pageName).replace(/ /g, '_').trim()) {
```
siddharthvp, line 2637: #915 adds `mw.Title` dependency to morebits. Might as well use that here.

morebits.js
Amorymeltzer: It's otherwise limited to `sysop` though, no?  `Morebits.userIsInGroup('patroller')` will just return `false` if the group doesn't exist, so there wouldn't be any difference.  The issue would be if a wiki **did** have a patroller group but it didn't have `patrol`...

morebits.js
Amorymeltzer: When I get a chance, I want to see what adding some of the `mw.Title` features into `morebits.wiki.page` for this sort of thing would look like.  Seems like it'd be useful, didn't want to embark on something *else* new for this, but, yes.

Amorymeltzer:
>It's 2020

Hmm?

morebits.js
siddharthvp: Looking at the code, looks like that's the case for most wikis, though I'm surprised. At enwiki, it was available to all autoconfirmed users before creation of the patroller group, but it turns out [that itself was an exception](https://phabricator.wikimedia.org/T14007) made many years back. 

There are a few wikis though which give patrol to autoconfirmed users, where this code will fail, but I guess that can be handled later along with i18n.

siddharthvp:
> > It's 2020
> 
> Hmm?

I mean why are you using XML for the APIs? While the rest of Twinkle does it, that's no reason to do that in new code. Everything in the MediaWIki world is now using JSON. Even popups which is as old as TW moved to JSON with formatversion 2 long back.

morebits.js
siddharthvp: I meant changing the "poor man's normalisation" to`new mw.Title(Morebits.pageNameNorm).toText() === new mw.Title(ctx.pageName).toText()`. Regarding the larger question of integrating mw.Title with morebits.wiki.page, what would probably make sense is to make morebits.wiki.page a sub-class of mw.Title (rather than trying to duplicate mw.Title code).

Amorymeltzer: Sorry, I was thinking out loud, orthogonal to this, about some advantages to having it around.  Something like (albeit nicer, with checks, etc.) could be helpful:

```js
    this.getTalkPageName = function() {
        return mw.Title.newFromText(this.getPageName()).getTalkPage().getPrefixedText();
    };
    this.getSubjectPageName = function() {
        return mw.Title.newFromText(this.getPageName()).getSubjectPage().getPrefixedText();
    };
```

As for here, yeah, sure.  I suppose it's simple enough to include in something off-topic.

morebits.js
Amorymeltzer: Oh yeah, the good ol' days!

The other option would be `mw.user.getRights().then(function (rights)...` but I really don't think that's an improvement.

Amorymeltzer:
I didn't think it'd be helpful when rewriting one function and adding another to make them look overwhelming different from everything else, possibly introducing new standard names/etc. by themselves; they're easier to check and review by comparison.

morebits.js
Amorymeltzer: Done here.  Without testing I'm not entirely sure what the `.trim()` was doing, I guess `ctx.pageName` would have whitespace?

PR #920, opened 2020-04-12
twinkle: Limit disabled sysop modules to sysops

Amorymeltzer:
Vaguely discussed in #793; ever-so-slightly speeds things up for non-sysops, slightly slower for sysops?  Less clear code for baiscally no gain?


siddharthvp:
> Less clear code for baiscally no gain?

Yep. I suspect the time to tokenise the extra code will balance out the benefits of doing a couple of less condition checks.

PR #921, opened 2020-04-13
morebits.date: Fix bug parsing parens around UTC timestamps

Amorymeltzer:
Not sure how this was missed in #814, but js date will see the parentheses in `(UTC)` and completely ignore the whole thing, setting the timezone local.  Should cause issues anywhere, but particularly problematic for the warn module.

Vaguely related discussion: https://github.com/azatoth/twinkle/pull/814/files#r374359937

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1346,8 +1346,8 @@ Morebits.date = function() {
 	this._d = new (Function.prototype.bind.apply(Date, [Date].concat(args)));
 
 	if (isNaN(this._d.getTime()) && typeof args[0] === 'string') {
-		// Try again after removing a comma, to get MediaWiki timestamps to parse
-		this._d = new (Function.prototype.bind.call(Date, Date, args[0].replace(/(\d\d:\d\d),/, '$1')));
+		// Try again after removing a comma and paren-wrapped timezone, to get MediaWiki timestamps to parse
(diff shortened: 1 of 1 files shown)


morebits.js
```diff
@@ -1346,8 +1346,8 @@ Morebits.date = function() {
 	this._d = new (Function.prototype.bind.apply(Date, [Date].concat(args)));
 
 	if (isNaN(this._d.getTime()) && typeof args[0] === 'string') {
-		// Try again after removing a comma, to get MediaWiki timestamps to parse
-		this._d = new (Function.prototype.bind.call(Date, Date, args[0].replace(/(\d\d:\d\d),/, '$1')));
+		// Try again after removing a comma and paren-wrapped timezone, to get MediaWiki timestamps to parse
+		this._d = new (Function.prototype.bind.call(Date, Date, args[0].replace(/(\d\d:\d\d),/, '$1').replace(/\(UTC\)/, 'UTC')));
```
Amorymeltzer, line 1350:
```suggestion
		this._d = new (Function.prototype.bind.call(Date, Date, args[0].replace(/(\d\d:\d\d),/, '$1').replace(/\(UTC\)$/, 'UTC')));
```

?

Amorymeltzer: Probably unnecessary, if it ever came up, it wouldn't be detected as a date anyway.

PR #922, opened 2020-04-13
warn: Ignore case when parsing warning history

Amorymeltzer:
Most templates have a hidden comment of the form `Template:uw-templatename`, but some have `Template:Uw-templatenme`; this adjusts the regex to account for those cases, as well as missing spaces.


Amorymeltzer:
I've abandoned the other change previously included here, it's not relevant or needed atm.  I'll add it as part of my #777 rewrite.

PR #902, opened 2020-03-25
talkback: Add preview

Amorymeltzer:
Repeatedly requested.  Restructured `evaluate` alongside a simple `getNoticeWikitext`.  Also makes used of `required: true` for relevant `page` inputs.

<s>2nd commit adds some heavy-handed processing to try to avoid errors.  Basically, any non-wikipedia non-talkpage is converted into its talkpage.  Not sure if this is good or just rude, so welcome feedback!</s> Have other plans where this makes more sense, will do there rather than muddy-up this PR.

PR #929, opened 2020-04-26
talkback: Restructure menu to simplify options, add whisperback

Amorymeltzer:
The layout of the talkback module window is a bit weird, with three talkback locations, then different options.  This collapses the three options into one so there's some coherence in the menu.  It builds on #902 and makes use of an object like c37b7c4/#758.  Includes a fairly heavy-handed normalization for likely mis-entered or erroneous page names to talkspace (project okay).

The first commit adds `{{whisperback}}` as a third talkback template option, as some prefer it and it's weird to have a dropdown with just two options.  The second commit restricts its usage by abusing the menus, and I'm not wild about it.  Should it be included?  If it is, should we let preview inform folks of weird/non-standard behavior?  If we're telling users where not to use it, should other talkback templates be included?

I think the restructure is worth it, but adding templates that don't do well outside of usertalk is a can of worms.

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -145,7 +145,8 @@ Twinkle.talkback.changeTarget = function(e) {
 			var templates = work_area.append({
 				type: 'select',
 				name: 'tbtemplate',
-				label: 'Which talkback template to use:'
+				label: 'Which talkback template to use:',
+				event: Twinkle.talkback.callback.change_template
 			});
 			$.each(Twinkle.talkback.templates.talkbacks, function(value, data) {
 				templates.append({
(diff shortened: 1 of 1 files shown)


siddharthvp:
Adding whisperback sounds like a great idea.

Amorymeltzer:
I'm just not sure about how to handle the ensuing complications without getting too in the weeds.  I suppose it's not likely to hurt too much.  I'll try to revisit that second commit.

Amorymeltzer:
Dealt with the merge conflicts, it was a doozy!  Tested a bit so seems to work, but this probably still has some kinks to work out.

Of note, this treats {{Please see}} as if `|section=` actually works, which AFAICT it currently does not.  I've engaged on the talk page there, but it's of note.

siddharthvp:
While adding whisperback and reorganising the module is great, the architecture of having "modes within modes" is too complicated of an interface for a user to navigate. In particular,
- {{Please see}} and {{whisperback}} aren't "talkback" templates. Even if they were, why not just make them radio buttons below the one for talkback? Not many would expect these nice options to be hidden under the select menu of the talkback section. 
- The "location of discussion" select and "other username"/"other page" input can be simplified down to a single input field, prefilled with self talk page name. 
  - When whisperback is selected, "location of discussion" can be changed to "Talk page of user" or something like that (as it only supports user talk pages).

This should make for a simpler UI. The max number of selects/inputs in the fieldset would be down from 4 to just 2.

Amorymeltzer:
Fair points, I suppose.  Somewhat more importantly: whisperback is [to be merged](https://en.wikipedia.org/wiki/Wikipedia:Templates_for_discussion/Log/2020_October_26#Template:Whisperback) with talkback itself.  Presumably a lot of this is salvageable, even with the above comment.

Amorymeltzer:
Close in favor of #1328, will reconsider whisperback alongside/after that.

PR #930, opened 2020-04-27
morebits: Only triage or patrol if page is available for doing so

Amorymeltzer:
Fallout from #917, in particular with the `triage` action.

`triage`: When AfD-ing an article that was already reviewed, Twinkle would throw a scary bold error saying that the article wasn't in the queue.  Which, of course, is true, <s>so we should check first.  There's an order-of-operations annoyance here, since `pagetriagelist` and `pagetriageaction` both require `pageid`, so we *have* to query for that first.  Basically, this introduces `triageProcessListApi` as an intermediary to check if the page is uncurated.</s>

We could check the API each time to see if the apge is uncurated (e.g. introduce `triageProcessListApi` as an intermediary), but there's an order-of-operations annoyance there, since `pagetriagelist` and `pagetriageaction` both require `pageid`, so we would *have* to query for that first and wait.  Instead, this just ignores the specific error (`bad-pagetriage-page`); it will still output a note in the console, but should be faster/smoother on the user end.

`patrol`: Doesn't throw up an error, but just notes the action as successfully done.  We can query the `recentchanges` list in the same query wherein we get a token/etc., although doing so removes the option for a query-less patrol action when we're on the page but the rcid link isn't available, e.g. history.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1877,7 +1877,7 @@ Morebits.wiki.api.setApiUserAgent = function(ua) {
  *
  * patrol(): Patrols a page; ignores errors
  *
- * triage(): Marks page as reviewed using PageTriage, which implies patrolled; ignores errors
+ * triage(): Marks page as reviewed using PageTriage, which implies patrolled; ignores most errors
  *
  * deletePage([onSuccess], [onFailure]): Deletes a page (for admins only)
  *
@@ -2647,13 +2647,10 @@ Morebits.wiki.page = function(pageName, currentAction) {
 			return;
 		}
 
-		// If patrolling current page, don't need to query for latest revID
+		// If a link is present, don't need to check if it's patrolled
 		if ($('.patrollink').length) {
 			var patrolhref = $('.patrollink a').attr('href');
 			ctx.rcid = mw.util.getParamValue('rcid', patrolhref);
-			fnProcessPatrol(this, this);
-		} else if (new mw.Title(Morebits.pageNameNorm).getPrefixedText() === new mw.Title(ctx.pageName).getPrefixedText()) {
-			ctx.revid = mw.config.get('wgRevisionId') || mw.config.get('wgCurRevisionId');
 			fnProcessPatrol(this, this);
 		} else {
 			var patrolQuery = {
@@ -2661,7 +2658,11 @@ Morebits.wiki.page = function(pageName, currentAction) {
 				prop: 'info',
 				meta: 'tokens',
 				type: 'patrol', // as long as we're querying, might as well get a token
-				titles: ctx.pageName
+				list: 'recentchanges', // check if the page is unpatrolled
+				titles: ctx.pageName,
+				rcprop: 'patrolled',
+				rctitle: ctx.pageName,
+				rclimit: 1
 			};
 
 			ctx.patrolApi = new Morebits.wiki.api('retrieving token...', patrolQuery, fnProcessPatrol);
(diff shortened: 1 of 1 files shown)


siddharthvp:
> Twinkle would throw a scary bold error saying that the article wasn't in the queue.

We can choose to just swallow the error, `apiobj.getStatusElement().unlink()` in case of that particular errorcode.

Amorymeltzer:
Hmm, hadn't though about that!  Will still result in a `console` warning, though, so maybe better to do this, awkward and wasteful though it may be?

morebits.js
```diff
@@ -2676,29 +2678,37 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
-		if (!Morebits.userIsSysop && !Morebits.userIsInGroup('patroller')) {
-			return;
-		}
-
-		// If on the page in question, don't need to query for page ID
-		if (new mw.Title(Morebits.pageNameNorm).getPrefixedText() === new mw.Title(ctx.pageName).getPrefixedText()) {
-			ctx.pageID = mw.config.get('wgArticleId');
-			fnProcessTriage(this, this);
+		// Fall back to patrol if not a valid triage namespace
+		if ([0, 2, mw.config.get('wgPageTriageDraftNamespaceId')].indexOf(mw.config.get('wgNamespaceNumber')) === -1) {
```
siddharthvp, line 2694: use `mw.config.get('pageTriageNamespaces')` instead.

siddharthvp:
> That will still result in a console warning, though, so maybe better to do this, awkward and wasteful though it may be?

Console warnings come all the time - edit conflicts, badtoken errors, etc - all of which we subsequently handle. No need to worry about that. Changing the status message seems to the more natural option than unlinking the status element though, roughly in line with the handling of the aforementioned errors.

morebits.js
Amorymeltzer: Hah yes!  I thought I was crazy not seeing it, serves me right for checking `wg`-prefixed items...

Amorymeltzer:
@siddharthvp Sorry, I'm not understanding what you're suggesting there.  I *think* you're saying this is fine?  Although maybe you're suggesting outputting something like "Page already curated" or something?  The latter isn't that bad, but it's a fairly silent treatment — only `tag` has it available for one-off toggling — so I think it'd just add needless clutter, especially for non-NPP folks. 

I didn't mean to suggest my opposition to console warnings, but rather that it's only a warning because I was sloppy, and we're explicitly trying not to treat it as one.

siddharthvp:
I was suggesting we don't need to check page triage status before reviewing, and handling the error that occurs if it was already reviewed - either by unlinking the statelem or changing its text. 

Sorry for the drive-by nature of my comments... I'm fairly busy nowadays and am only cursorily running through diffs of any new PRs...

Amorymeltzer:
No worries, you're all good!  I'm clearly only capable of dropping in and out as well, it's all volunteer so don't sweat it.

Amorymeltzer:
So, what, that'd be adding something like the below to `returnError`
```js
} else if (this.errorCode === 'bad-pagetriage-page') {
	this.statelem.info('Skipping');
```
Around https://github.com/azatoth/twinkle/blob/master/morebits.js#L1784

I guess that works; it's a unique message afaict, as opposed to if the ID doesn't exist, etc.  I'm wary of hampering `Morebits.wiki.api`; what if we/someone *did* want that error?

siddharthvp:
Sorry for the delay. I wasn't talking about tweaking Morebits.wiki.api - which is obviously a bad idea. I was saying we can add a failure callback to the api call: https://github.com/azatoth/twinkle/blob/5c201778eca0506a1dece08d8d97e03a2aa97dbf/morebits.js#L3339
```js
function onTriageFailure(apiobj) {
	if (apiobj.getErrorCode() === 'bad-pagetriage-page') {
		apiobj.getStatusElement().unlink(); // or .info('Skipping');
	}
}
```

Amorymeltzer:
Ahh I gotcha now, thanks.  I had dismissed that earlier, not wanting to emit an error when, in fact, there wasn't really one, but do agree it'd be tidier/more efficient.

Amorymeltzer:
After much waffling, agreed it's better to just <s>ignore</s> skip the error.

PR #939, opened 2020-05-06
friendlywelcome:  Add welcome-retro

BrandonXLF:
Add the option to use the {{Welcome-retro}} instead of {{Welcome}} since the welcome template was changed recently to exclude several links and to use buttons instead of the list of links.

### DIFF
--- a/modules/friendlywelcome.js
+++ b/modules/friendlywelcome.js
@@ -233,6 +233,11 @@ Twinkle.welcome.templates = {
 				linkedArticle: true,
 				syntax: '{{subst:welcome|$USERNAME$|art=$ARTICLE$}} ~~~~'
 			},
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
For further context, see:
- https://en.wikipedia.org/wiki/Wikipedia:Administrators%27_noticeboard/Incidents#Change_to_all_our_welcome_templates
- https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Courtesy_notice:_many_welcome_templates_now_wrappers
- https://en.wikipedia.org/wiki/User_talk:JzG#Using_rollback_on_all_of_Sdkb's_template_changes,_even_ones_unrelated_to_welcome_templates

Among others...

PR #941, opened 2020-05-10
fluff: fix two ancient bugs: typo and type

Amorymeltzer:
- Typo: `statitem` -> `statelem`.  This stems back to 2011 (54347f92/98240d17), although there was another instance present from the initial import that was fixed later that year in be5d8ffe
- Type: `parseInt` `revid`.  Fix potential type issue treating `revid` as a string, missed in 37cd0c68 (2011)


Amorymeltzer:
Note: I reverted the type fix in 16e077b; I must've been had some other changes present when testing, because that was literally the opposite of what I wanted.

PR #942, opened 2020-05-10
fluff: Add rollback links to Special:Recentchanges and history pages

Amorymeltzer:
Both off by default; these commits can be squashed, but are separate for perusal.

**1st commit:** Use the provided diff url to get oldid.  Avoids issues with arcane diffs urls, and with urls on recent changes where oldid/curid/diff are each provided, and oddly besides.

(this is just a setup necessary for the next commit, doesn't necessarily stand out in the diff)

**2nd commit:** Add ability to enable rollback links on Recentchanges(linked)

Adds new `recent` option, off by default.  Most of the complexity in the selectors comes from having to handle two preferences: using the new or old (no js) RC system, and the grouped or ungrouped preference.  Uses the `wikipage.content` hook to (re)generate links when new diffs are loaded.

Also expands `openTalkPageOnAutoRevert` to apply at RC/RCL, just like with contribs.  The text for that has been tweaked as well, as it was a misleading description around behavior for contributions pages.  The language stemmed from 0746c05 in 2011, but it wasn't an accurate description of how this preference is helpful.  It *does* define whether the user talk page is opened, but it doesn't help do multiple things on the contributions page, because a single revert will still redirect you away from the contributions page (and now recent changes).

**3rd commit:** Add links to history pages, off by default

----

One thing to note is that RC/RCL can be comparatively slow loading pages (albeit faster than the watchlist).  With Twinkle loading on them, that means they'll be even slower, and no other modules are active there.  In my testing it's not so bad, especially not compared to the new/fancy/slow JS on the page itself, but one thing we could do is introduce a non-standard check of the pref in twinkle.js so that recent changes isn't slowed down:

```js
if (Twinkle.getPref('showRollbackLinks').indexOf('recent') !== -1) {
        specialPageWhitelist = specialPageWhitelist.concat([ 'Recentchanges', Recentchangeslinked' ]);
}
```

Not sure it's net-positive in the broader picture, but wanted to put it out there.

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -395,13 +395,13 @@ Twinkle.config.sections = [
 
 			// TwinkleConfig.showRollbackLinks (array)
 			// Where Twinkle should show rollback links:
-			// diff, others, mine, contribs, recent
+			// diff, others, mine, contribs, history, recent
 			// Note from TTO: |contribs| seems to be equal to |others| + |mine|, i.e. redundant, so I left it out heres
 			{
 				name: 'showRollbackLinks',
 				label: 'Show rollback links on these pages',
 				type: 'set',
-				setValues: { diff: 'Diff pages', others: 'Contributions pages of other users', mine: 'My contributions page', recent: 'Recent changes and related changes special pages' }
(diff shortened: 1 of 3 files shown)


CanILogIn:
I actually wanted to make this suggestion, so I support this.

Amorymeltzer:
Still undecided on the `specialPageWhitelist` bit, but maybe no need to do ahead of time.

PR #944, opened 2020-05-11
add script for easily creating localhost test loaders

siddharthvp:
Script for automatically generating localhost test file configuration, coz I'm fed up of messing it up while doing it manually...

If you have permanently set up patch-test-loader.js to load up in the wiki (eg. I do that by permanently loading 127.0.0.1:5500/local.js from my common.js. When the server is turned on - using a switch in VS code, local.js in turn loads patch-test-loader.js), then you all you need to do to test the current branch is to run `npm run patchtest`, and switch on the server (root dir: twinkle, host: 127.0.0.1, port: 5500). One way to do that easily is to use `server.js` (`npm run server`) after PR #911 is merged.

### DIFF
--- a/patch-test.js
+++ b/patch-test.js
@@ -1,0 +1,90 @@
+/* eslint-env node, es6 */
+/* eslint-disable no-console */
+/* eslint-disable es5/no-destructuring */
+/* eslint-disable es5/no-template-literals */
+/* eslint-disable es5/no-es6-methods */
+/* eslint-disable es5/no-shorthand-properties */
+/* eslint-disable es5/no-block-scoping */
+
+/**
+ * Creates a file patch-test-loader.js with the necessary import statements
+ * to test the changes made in a pull request or in the current working state
+ *
+ * How to use:
+ * - Run `npm run patchtest` which generates patch-test-loader.js
+ * - Set up a localhost server (such as by using server.js or by running
+ *   php -S localhost:5500) and load patch-test-loader.js in the wiki environment,
+ *   using the browser console or the common.js page.
+ */
+
+const fs = require('fs');
+const {execSync} = require('child_process');
+
+const server = 'http://127.0.0.1:5500';
+
+// find the last common commit between this branch and master, and get the list
+// of files changed since that commit
+try {
+	var stdout = execSync('git diff $(git merge-base $(git rev-parse --abbrev-ref HEAD) master) --name-only').toString();
+	let changedFiles = stdout.split(/\r?\n/);
+	createTestLoader(changedFiles);
+} catch (err) {
+	console.log(err.toString());
+}
+
+/** @param {string[]} changedFiles */
+function createTestLoader(changedFiles) {
+
+	let importsCount = 0; // for the message written to the console in the ends
+	let importLine = function(file) {
+		if (!changedFiles.includes(file)) {
+			return '';
+		}
+		if (file.endsWith('.js')) {
+			importsCount++;
+			if (file.startsWith('modules/')) {
+				return `mw.loader.getScript('${server}/${file}');`;
+			}
+			return `return mw.loader.getScript('${server}/${file}');`;
+		} else if (file.endsWith('.css')) {
+			importsCount++;
+			return `mw.loader.load('${server}/${file}', 'text/css');`;
+		}
+	};
+
+	let jsLoaderSource = `// Wait for Twinkle gadget to load, so that we can then overwrite it
+	mw.loader.using('ext.gadget.Twinkle').then(function() {
+		${importLine('morebits.css')}
+		${importLine('morebits.js')}
+
+	}).then(function() {
+		${importLine('twinkle.css')}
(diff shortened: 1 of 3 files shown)


patch-test.js
```diff
@@ -0,0 +1,86 @@
...
+ * - Set up a localhost server (such as by using server.js) and load patch-test-loader.js
+ *   in the wiki environment, using the browser console or the common.js page.
+ */
+
+const fs = require('fs');
+const {exec} = require('child_process');
+
+// find the last common commit between this branch and master, and get the list
+// of files changed since that commit
+exec('git diff $(git merge-base $(git rev-parse --abbrev-ref HEAD) master) --name-only', (err, stdout) => {
```
siddharthvp, line 19: Breaking this down:
- `git rev-parse --abbrev-ref HEAD` gives the name of current branch
- `git merge-base brName master` gives the last common commit b/w this branch and master
- `git diff commitID` gives the diff of that commit with the current working directory (changes from subsequent commits, and any staged and unstaged changes). With the `--name-only` flag it just lists names of files changed.

siddharthvp:
Known issue: this doesn't reliably work if `twinkle.js` is modified. The Twinkle[module]() is not a function may occur if because of some weird loading race conditions. There is no easily automated way to fix this other than moving out the module initialisations to the module files' themselves.

siddharthvp:
Tweaked to work alongside #946.

patch-test.js
```diff
@@ -0,0 +1,90 @@
+/* eslint-env node, es6 */
+/* eslint-disable no-console */
+/* eslint-disable es5/no-destructuring */
+/* eslint-disable es5/no-template-literals */
+/* eslint-disable es5/no-es6-methods */
+/* eslint-disable es5/no-shorthand-properties */
+/* eslint-disable es5/no-block-scoping */
```
siddharthvp, line 7: Ugly. The `eslint-env` setting ought to automatically disable these es5 rules. Reported issue upstream: https://github.com/nkt/eslint-plugin-es5/issues/40

patch-test.js
```diff
@@ -0,0 +1,90 @@
...
+ * - Run `npm run patchtest` which generates patch-test-loader.js
+ * - Set up a localhost server (such as by using server.js or by running
+ *   php -S localhost:5500) and load patch-test-loader.js in the wiki environment,
+ *   using the browser console or the common.js page.
+ */
+
+const fs = require('fs');
+const {execSync} = require('child_process');
+
+const server = 'http://127.0.0.1:5500';
```
Amorymeltzer, line 23: Why 5500?  I always thought 8000 or 8080 were more standard for one-off local hostings?  Is this a V$code thing?

(I thought I asked this yesterday, but must've gotten gobbled up somewhere.)

siddharthvp: I don't know about 8000, but 8080 is the port used by Apache Tomcat servers and so it frequently ends up being used for HTTP applications. In short, they're used for serious stuff :)

Amorymeltzer: Well... under 1024 are the reserved, "serious" ones.  80 is http, which I believe is the reason why historically 8000 and 8080 are both used for various unofficial hostings and one-offs, and as an http alternative (8008 too).  None of it matters, really, unless the person is testing *and* serving something else, and even then that's on them!  There's thousands to choose from, so it's largely a moot consideration, but 8000/8080 have been used for hosting and casting for *decades*.  As an example, it's the default for python's `SimpleHTTPServer`.

siddharthvp: Well, using a standard port like 8000/8080 does not make sense as these are the ones that are most likely to be already in use. They are used for serious hostings (my college's wifi login page used to on 8080, until they switched to sophos). This is I believe because usually the decision isn't really in their hand -- it's determined by the server software or the service provider. Eg. on Toolforge, the nginx proxy decides what port to use and puts that out as an environment variable [which the tool devs use](https://github.com/siddharthvp/toolforge-node-app-base/blob/master/server.js#L21). 

Arguing over what port to use for a service that would be used only by the person hosting it is one of the most pointless things we can do. Since we can choose anything, it probably makes sense to use a completely random port, eg. [xfdcloser uses](https://github.com/evad37/xfdcloser/blob/master/server.js) 8125 and [redwarn uses](https://gitlab.com/redwarn/redwarn-web/-/blob/master/CONTRIBUTING.md) 9696.

Amorymeltzer: Hey man sorry if it came off like I was trying to argue or bikeshed, I really wasn't trying to.  We're talking past each other a bit here, but I really was just curious where 5500 came form.  It's a nice round number and I'm used to seeing 8000for personal localhostings, so I was wondering what led you to use it.  You vaguely referenced VScode, so I thought maybe it was connected to that?  I've [no issue](https://github.com/azatoth/twinkle/pull/944#pullrequestreview-446904635) with this as is, and sorry for making you feel like I did.  Thanks for putting this together, it's neat!

PR #946, opened 2020-05-12
Decentralise module initialisations

siddharthvp:
Initialise the modules from the module files' themselves rather than centrally from twinkle.js. Doing it this way makes testing of twinkle.js and of custom Twinkle installations (like on other wikis) less cumbersome as one no longer needs to comment out or modify the initialisation lines in twinkle.js (as mentioned [here](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_42#How_to_use_as_a_user_script?) and [here](https://github.com/azatoth/twinkle/pull/462#issuecomment-462495588)). This also fixes the issue https://github.com/azatoth/twinkle/pull/944#issuecomment-627359133 allowing for automated setup for a test file for changes that affect twinkle.js.

To do this, we re-purpose Twinkle.addInitCallback() function ~~(presently unused anywhere within TW or from user scripts), such that it takes (part of) the name of the function (rather than the function itself)~~ <ins>such that it takes the name of the function as well</ins> - so that we can check if the module was disabled. twinkleconfig and any "custom modules" (usually none) can't be disabled, so will always be loaded.

The order in which the menu buttons appear is still within our control through MediaWiki:Gadgets-definition - the files are executed in the order in which they appear in the gadget definition line. **The gadget definition on-wiki will need to be tweaked as done in DEVELOPER.md file.**

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -17,7 +17,7 @@ There are two ways to upload Twinkle scripts to Wikipedia or another destination
 
 After the files are synced, ensure that [MediaWiki:Gadgets-definition][] contains the following lines:
 
-    * Twinkle[ResourceLoader|dependencies=ext.gadget.morebits,ext.gadget.select2,mediawiki.api,mediawiki.language|rights=autoconfirmed|type=general|peers=Twinkle-pagestyles]|Twinkle.js|twinkleprod.js|twinkleimage.js|twinklebatchundelete.js|twinklewarn.js|twinklespeedy.js|friendlyshared.js|twinklediff.js|twinkleunlink.js|friendlytag.js|twinkledeprod.js|friendlywelcome.js|twinklexfd.js|twinklebatchdelete.js|twinklebatchprotect.js|twinkleconfig.js|twinklefluff.js|twinkleprotect.js|twinklearv.js|twinkleblock.js|friendlytalkback.js|Twinkle.css
+    * Twinkle[ResourceLoader|dependencies=ext.gadget.morebits,ext.gadget.select2,mediawiki.api,mediawiki.language|rights=autoconfirmed|type=general|peers=Twinkle-pagestyles]|Twinkle.js|Twinkle.css|twinklearv.js|twinklewarn.js|twinkleblock.js|friendlywelcome.js|friendlyshared.js|friendlytalkback.js|twinklespeedy.js|twinkleprod.js|twinklexfd.js|twinkleimage.js|twinkleprotect.js|friendlytag.js|twinklediff.js|twinkleunlink.js|twinklefluff.js|twinkledeprod.js|twinklebatchdelete.js|twinklebatchprotect.js|twinklebatchundelete.js|twinkleconfig.js
     * morebits[ResourceLoader|dependencies=mediawiki.user,mediawiki.util,mediawiki.Title,jquery.ui,jquery.tipsy|hidden]|morebits.js|morebits.css
     * Twinkle-pagestyles[hidden|skins=vector]|Twinkle-pagestyles.css
     * select2[ResourceLoader|hidden]|select2.min.js|select2.min.css
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -1320,6 +1320,7 @@ Twinkle.config.init = function twinkleconfigInit() {
 		}
 	}
 };
+Twinkle.addInitCallback(Twinkle.config.init);
 
 // custom list-related stuff
 
--- a/twinkle.js
+++ b/twinkle.js
@@ -27,10 +27,16 @@ if (!Morebits.userIsInGroup('autoconfirmed') && !Morebits.userIsInGroup('confirm
 var Twinkle = {};
 window.Twinkle = Twinkle;  // allow global access
 
-// for use by custom modules (normally empty)
 Twinkle.initCallbacks = [];
-Twinkle.addInitCallback = function twinkleAddInitCallback(func) {
-	Twinkle.initCallbacks.push(func);
+
+/**
+ * Adds a callback to execute when Twinkle has loaded.
+ * @param {function} func
+ * @param {string} [name] - name of module used to check if is disabled.
+ * If name is not given, module is loaded unconditionally.
+ */
+Twinkle.addInitCallback = function twinkleAddInitCallback(func, name) {
+	Twinkle.initCallbacks.push({ func: func, name: name });
 };
 
 Twinkle.defaultConfig = {};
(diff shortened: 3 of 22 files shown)


siddharthvp:
Tested, confirmed working on the testwiki.

Amorymeltzer:
Nice!!  I'll take a deeper run later but I don't think there are any downsides?

siddharthvp:
No, AFAICT.

Amorymeltzer:
Yeah I can't think of anything; I think #828 probably took care of any potential issues around access.

twinkle.js
```diff
@@ -27,7 +27,6 @@ if (!Morebits.userIsInGroup('autoconfirmed') && !Morebits.userIsInGroup('confirm
 var Twinkle = {};
 window.Twinkle = Twinkle;  // allow global access
 
-// for use by custom modules (normally empty)
 Twinkle.initCallbacks = [];
 Twinkle.addInitCallback = function twinkleAddInitCallback(func) {
```
Amorymeltzer, line 31: Is it not a problem to reuse this?  This basically just changes the standard loading system to the one for "custom modules" afaict, but since we've got `Twinkle.addInitCallback = function` twice now I would think this would break the "custom modules".

modules/twinkleconfig.js
```diff
@@ -19,7 +19,8 @@
  */
 
 
-Twinkle.config = {};
+Twinkle.config = twinkleconfigInit; // defined below
```
Amorymeltzer, line 22: I suppose it's silly to add the init bit to line 900+, but is there a good reason not to move the actual init function up here?  I say yes, having the objects that are 95% of what gets edited up top makes things easier for folks, especially new folks, but just wanted to mention.

Amorymeltzer:
No reason not to do this before #944, yeah?  I can more quickly verify this, so no need to wait while I collect thoughts and time re: tests?

twinkle.js
siddharthvp: I don't think there's a problem as no such "custom modules" exist. Twinkle.addInitCallback is completely unused (only search results are old copies of Twinkle code).

modules/twinkleconfig.js
siddharthvp: Yeah i think it's good to keep configurable data at the top of the file. It's a bit weird to take advantage of JS' variable hoisting but seems worth it.

siddharthvp:
> No reason not to do this before #944, yeah? I can more quickly verify this, so no need to wait while I collect thoughts and time re: tests?

Nah. Also #944 will need a tweak given that this one changes the way modules are loaded.

twinkle.js
Amorymeltzer: @siddharthvp Okay but from a code point of view, it should be renamed or the former call be removed.  I *think* it's used on other projects, so a simple rename of this new function should suffice?

Amorymeltzer:
Conflicts removed!

siddharthvp:
Terrible mess-up from forgetting to pull after your rebase @Amorymeltzer. Will fix later. In the meantime, can you look at f0f2fca0c. This changes the structure and IMO this is better for the reasons mentioned in commit message. Lemme know what you think.

Amorymeltzer:
Sorry, that's probably on me.

Just looking over it quickly, seems like a marked improvement — and I was happy with it before! — especially backward-compat.  I like the idea of having it be agnostic.  Seems wasteful to have `Twinkle.module, 'module'` everywhere, but, again, reasonable, and probably far more so than any alternative.  I'm still not clear why we need to define `Twinkle.addInitCallback` twice?  It seems like the second one could just be folded into the `forEach`?  I've not tested so maybe I'm just missing something.

siddharthvp:
Squashed it in and updated the commit message.

> Seems wasteful to have Twinkle.module, 'module' everywhere

That's what prevented me from thinking of this earlier. That's the only downside.

> I'm still not clear why we need to define Twinkle.addInitCallback twice?

This is not new. `Twinkle.addInitCallback` has always had 2 definitions - one to handle the modules loading before the twinkleoptions has loaded and one to handles the ones loading after that. Added this as a comment.

Amorymeltzer:
Yeah, sorry for not being clear after I realized that; I was really just asking if there was a reason we needed it written that way?  Not an issue blocking this ofc, but I'm missing why.

@Xi-Plus is this something that'd mess you up?  We don't have a great way for indicating changes to the gadget.

siddharthvp:
> Yeah, sorry for not being clear after I realized that; I was really just asking if there was a reason we needed it written that way? Not an issue blocking this ofc, but I'm missing why.

Not sure what you're missing. This originally existed for custom modules and of course we don't have any control on when the custom modules would load. So basically this took care of both the scenarios.

siddharthvp:
As if this didn't already have enough labels! 😁

twinkle.js
```diff
@@ -446,33 +452,19 @@ Twinkle.load = function () {
...
-	Twinkle.initCallbacks.forEach(function (func) {
-		func();
-	});
-	Twinkle.addInitCallback = function (func) {
-		func();
+	// Redefine addInitCallback so that any modules being loaded now on are directly
+	// initialised rather than added to initCallbacks array
+	Twinkle.addInitCallback = function(func, name) {
+		if (!name || Twinkle.disabledModules.indexOf(name) === -1) {
+			Twinkle[func]();
```
Amorymeltzer, line 461: Right off the bat, this kills it; should still be `func()`.  As an example, arv adds `{func: Twinkle.arv, name: 'arv'}` to `initCallbacks`.  That passes `twinklearv`, so `Twinkle.twinklearv()` isn't a function (`Twinkle.arv = function twinklearv() {`, but either way, `Twinkle['Twinkle.arv']` isn't a function).

```suggestion
			func();
```

Amorymeltzer:
Haven't done much testing since the rewrite (which was fine), but visually looks good!  Concern is mostly downstream, but not much that I know of (as above).  The most likely candidate ([arb.js](https://en.wikipedia.org/wiki/User:Bellezzasolo/Scripts/arb.js)) doesn't do anything with this.

twinkle.js
siddharthvp: Ufoos. Fixed.

Amorymeltzer:
@Xi-Plus, @Huji: One last check/ping, since this is a big change to the gadget.  I've got a nice way of checking changes in #1035, but want to clear this first.

Xi-Plus:
> @Xi-Plus is this something that'd mess you up? We don't have a great way for indicating changes to the gadget.

No. I have no problem.

DEVELOPER.md
```diff
@@ -17,7 +17,7 @@ There are two ways to upload Twinkle scripts to Wikipedia or another destination
...
+    * Twinkle[ResourceLoader|dependencies=ext.gadget.morebits,ext.gadget.select2,mediawiki.api,mediawiki.language|rights=autoconfirmed|type=general|peers=Twinkle-pagestyles]|Twinkle.js|Twinkle.css|twinklearv.js|twinklewarn.js|twinkleblock.js|friendlywelcome.js|friendlyshared.js|friendlytalkback.js|twinklespeedy.js|twinkleprod.js|twinklexfd.js|twinkleimage.js|twinkleprotect.js|friendlytag.js|twinklediff.js|twinkleunlink.js|twinklefluff.js|twinkledeprod.js|twinklebatchdelete.js|twinklebatchprotect.js|twinklebatchundelete.js|twinkleconfig.js
     * morebits[ResourceLoader|dependencies=mediawiki.user,mediawiki.util,mediawiki.Title,jquery.ui,jquery.tipsy|hidden]|morebits.js|morebits.css
```
Huji, line 21: Is there a reason the order of these is changed? It makes for a confusing diff on version control and when I compare it on https://www.diffchecker.com/e0QzkyMf I see no reason why the order of modules should be changed.

Amorymeltzer: @Huji See the note in the top message:

>The order in which the menu buttons appear is still within our control through MediaWiki:Gadgets-definition - the files are executed in the order in which they appear in the gadget definition line. The gadget definition on-wiki will need to be tweaked as done in DEVELOPER.md file.

This PR changes how and when the files are loaded.  Currently, it is `twinkle.js` which loads each module, via `Twinkle.load`.  Rather than hard-coding the order of modules in `twinkle.js`, this leverages the gadget definition to establish the order.  The difference is mostly for testing Twinkle.js itself, but it does make sense overall.  After this PR, it probably makes alternative installations easier to maintain.

Amorymeltzer: That being said, I think there's probably an argument to having `twinkleconfig` ahead of the others, although it shouldn't matter?  `fluff` could be moved to the very end.

Huji: Yeah, that is what I mean. Don't move `fluff` or `welcome` or `xfd` or `arv` , etc. Only move things that need to be moved. Or am I missing the point?

siddharthvp: IIRC I copied the sequence from the original code (which was based on sound logic) to the gadget def (which was a random jumble). 

In any case, it's clear from the diff that net lengths of the old and new strings are the same -- so nothing was added or removed. I am not sure what's confusing. 

> That being said, I think there's probably an argument to having twinkleconfig ahead of the others, although it shouldn't matter? fluff could be moved to the very end.

Isn't it the other way round? We want things to load/execute fast. Functional modules like fluff should go first, not things like twinkleconfig which is loaded only on one page that's rarely visited.

Huji: Well, if nothing is supposed to be changed, then there should not be any difference in this line of code; yet there is.

Do you need to rebase? Do you need to first, in a separate patch, reorder module names to what you think is desirable (and explain why), and then use it here without change?

In any case, I'm going to stop pushing for this. I am not a primary maintainer of Twinkle, so it is not my place to push any harder. My entire point is this: each commit should ONLY change what is needed for that commit; it should not include any other changes.

Amorymeltzer: @Huji This matches what was in the `twinkleModules` list in twinkle.js, see https://github.com/azatoth/twinkle/pull/946/files#diff-551c5d596f19b4df338bbf4ab50919baL450-L459, which is how the loading order is currently determined.  This PR supersedes and does away with that list in favor of using the order in the gadget, so it's a one-time move to establish the proper loading order via the gadget.  It should work fine without changing the order, but the order of the items in the menu/tabs will be different.

@siddharthvp I meant more from a "base things you can't/shouldn't turn off" but good point!  Especially on fluff.

Huji: @Amorymeltzer oh, now I get it. In that case, I think a new order is justified.

Amorymeltzer:
Merged!  Thanks everybody, especially @siddharthvp!

PR #948, opened 2020-05-13
fluff: Don't open the talkpage unless/until the reversion is complete

Amorymeltzer:
This is a pain to test without #942 and #947 but should preempt issues like #855 when there's an edit conflict detected, which should be more likely after #942.  This kind-of relates to the duplication of behaviors in `morebits`, but these perceived "edit conflicts" are still handled there.


Amorymeltzer:
Hmm, forgot about that first commit, meant to merge/squash that elsewhere.

PR #956, opened 2020-06-03
speedy: Align A7 criteria labels with the actual criteria

mdaniels5757:
CSD A7 "applies to any article about a [list of things omitted] that does not indicate why its subject is important or significant". However, currently Twinkle summarizes A7 as "Unremarkable", which isn't accurate -- there are plenty of unremarkable things that do not satisfy the very low bar of A7 (which "does not apply to any article that makes any credible claim of significance or importance even if the claim is not supported by a reliable source or does not qualify on Wikipedia's notability guidelines.")

I propose changing "Unremarkable" to "No indication of importance or significance of subject".

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -696,49 +696,49 @@ Twinkle.speedy.articleList = [
 		}
 	},
 	{
-		label: 'A7: No indication of importance or significance (people, groups, companies, web content, individual animals, or organized events)',
+		label: 'A7: No indication of importance (people, groups, companies, web content, individual animals, or organized events)',
 		value: 'a7',
 		tooltip: 'An article about a real person, group of people, band, club, company, web content, individual animal, tour, or party that does not assert the importance or significance of its subject. If controversial, or if a previous AfD has resulted in the article being kept, the article should be nominated for AfD instead',
 		hideWhenSingle: true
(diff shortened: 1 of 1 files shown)


mdaniels5757:
I agree it's a bit too wordy, and have pushed a change that removes the "of subject" bit (so they are now `No indication of importance or significance (person)`.

I'm not sold dropping "significance" is the best way to shorten it further: "importance" and "significance" are slightly different (at least, different enough to warrant including both words in the actual criteria). If we had to drop one word, I'd argue keeping "significant" and removing "important" stays truer to the original meaning: at least to me, something important is normally significant, but something significant is often not important. (In other words, importance normally implies significance, but significance often does not imply importance)

If you wanted to keep both words and consolidate slightly more, and were ok with using slashes, `No indication of importance/significance (person)` would be an easy change to make that drops 3 characters.

Amorymeltzer:
I suggested "No indication of importance" because that's the actual header used for the criterion on the CSD policy page, viz https://en.wikipedia.org/wiki/Wikipedia:Criteria_for_speedy_deletion#A7._No_indication_of_importance_(people,_animals,_organizations,_web_content,_events).

mdaniels5757:
Yup, you're right. Done.

PR #957, opened 2020-06-03
talkback: Add Template:EFFPReply

mdaniels5757:
Sometimes, I want to give talkback messages to IP users from [WP:EFFP/R](https://en.wikipedia.org/wiki/Wikipedia:Edit_filter/False_positives/Reports). I made [this template](https://en.wikipedia.org/wiki/Template:EFFPReply) for it, and would like to use it from Twinkle. Cheers, Michael (User:Mdaniels5757)

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -338,6 +338,11 @@ Twinkle.talkback.noticeboards = {
 		label: 'WP:OTRS/N (OTRS noticeboard)',
 		text: '{{OTRSreply|1=$SECTION|2=~~~~}}',
 		editSummary: 'You have replies at the [[Wikipedia:OTRS noticeboard|OTRS noticeboard]]'
+	},
+	'effp' {
+		label: 'WP:EFFP/R (Edit filter false positive report)',
+		text: '{{EFFPReply|1=$SECTION|2=~~~~}}',
+		editSummary: 'You have replies at the [[WP:EFFP/R|Edit filters false positive report]] page'
 	}
 };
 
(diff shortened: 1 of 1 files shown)


modules/friendlytalkback.js
```diff
@@ -338,6 +338,11 @@ Twinkle.talkback.noticeboards = {
 		label: 'WP:OTRS/N (OTRS noticeboard)',
 		text: '{{OTRSreply|1=$SECTION|2=~~~~}}',
 		editSummary: 'You have replies at the [[Wikipedia:OTRS noticeboard|OTRS noticeboard]]'
+	},
+	'effp': {
+		label: 'WP:EFFP/R (Edit filter false positive report)',
+		text: '{{EFFPReply|1=$SECTION|2=~~~~}}',
+		editSummary: 'You have replies at the [[WP:EFFP/R|Edit filters false positive report]] page'
```
Amorymeltzer, line 345:
```suggestion
		editSummary: 'You have replies at the [[Wikipedia:Edit filter/False positives/Reports|Edit filters false positive report]] page'
```

Might as well use the full link.
----
It'd be nonstandard, but is there a reason not to say something like "You have replies to your edit filter false positive report"?  It's already somewhat nonstandard, and makes a tad more sense.  Regardless, the language around this feels off.

Amorymeltzer:
I think this would be a good idea!  I'd be curious if the usual suspects find it useful, does anyone else use it?  Tucked away in the menu it's not much harm or clutter.

It should probably be placed in order, though?

mdaniels5757:
Agreed on both counts. Summary changed, and should be in order now.

Amorymeltzer:
Awesome, thanks!

Amorymeltzer:
(Bah tried to squash, GH interface froze)

PR #958, opened 2020-06-06
Update Template:W-short

MJL-wiki:
Previously wasn't transcluding/substituting properly

### DIFF
--- a/modules/friendlywelcome.js
+++ b/modules/friendlywelcome.js
@@ -240,7 +240,7 @@ Twinkle.welcome.templates = {
 			},
 			'welcome-short': {
 				description: 'a shorter welcome message',
-				syntax: '{{subst:welcome-short|$USERNAME$}} $EXTRA$ ~~~~'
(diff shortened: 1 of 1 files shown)


modules/friendlywelcome.js
```diff
@@ -240,7 +240,7 @@ Twinkle.welcome.templates = {
 			},
 			'welcome-short': {
 				description: 'a shorter welcome message',
-				syntax: '{{subst:welcome-short|$USERNAME$}} $EXTRA$ ~~~~'
+				syntax: '$HEADER$ {{subst:w-short}} $EXTRA$ '
```
Amorymeltzer, line 243: This has a `heading` parameter, shouldn't we just use that?
```suggestion
				syntax: '{{subst:w-short|heading=yes}} $EXTRA$ '
```

MJL-wiki: > This has a `heading` parameter, shouldn't we just use that?

Done

Amorymeltzer:
Merged as a6e95ee, thanks!

PR #962, opened 2020-06-08
twinkle: Update vector classes (vectorMenu -> vector-menu)

Amorymeltzer:
The work is a little spread out, but see [T253329](https://phabricator.wikimedia.org/T253329), [T254798](https://phabricator.wikimedia.org/T254798), and [T254797](https://phabricator.wikimedia.org/T254797).

Basically, for our purposes:

- `vectorMenu` replaced with `vector-menu`
- `vectorMenuCheckbox` replaced with `vector-menu-checkbox`
- `menu` replaced with `vector-menu-content-list`, although this isn't unique

----

Draft pending some feedback and because I haven't tested this just yet.


Amorymeltzer:
AFAICT `vectorTabs` can be replaced with `vector-menu-tabs` as well, but that's not changing right now.  Plus, it's a mess and I'm not sure what the final state is supposed to be; right now, it looks like they're `vector-menu vectorTabs vector-menu-tabs`.

Amorymeltzer:
\*heavy sigh\* There are like a half dozen things that are in flux right now, and AFAICT the wikis themselves are not ready.  I'd like to avoid a lot of future maintenance, and so I've got some additional work for future-proofing, but I think most of it will have to wait.  I *think* the barebones will be functional once again this week so I can test it more easily on testwiki after tomorrow.

Amorymeltzer:
Okay @Xi-Plus @MusikAnimal I've pushed and tested this on testwiki, as far as I can tell everything looks good!  As above, things are in a weird place right now, so DO let me know if you can find any issues.  I reworked things a bit (last modified by @siddharthvp in #763) but looks like I fixed a bug or two.

Amorymeltzer:
I *think* this (and #963) can be merged and pushed live, so if I hear no complains I'll do that tonight.  @Xi-Plus, would it be easier for you if I just merged these/this now?

Xi-Plus:
When should I deploy the patch to the site?

Amorymeltzer:
@Xi-Plus I think right away.  When testwiki got updated (last night EDT) to the latest (1.35-wmf.37), that messed up the Vector menus, so it should be right now.  Wikipedias will be updated Thursday, so unless there's a major change, I'd imagine it'd need to be out by then.

I just did a quick test, and it looks like not all of the updated classes are present in wmf.36, so I had to re-add `menu` and `vectorMenuCheckbox` for safety.  With some quick testing during naptime, I *think* this latest version (8f7b2f3) should be safe to put up any time starting now.

Amorymeltzer:
I'll plan on merging this tonight (EDT) and pushing live to enwiki ahead of the update later this week.

Amorymeltzer:
Thanks for the :+1: @MusikAnimal — your comments there cover it pretty well!

I do worry that the disparate Twinkle installations are going to run into some serious problems, especially the older ones, so hopefully that's a good roadmap.  I'll post at en's WT:TW in case anyone ends up there.

PR #964, opened 2020-06-10
fluff: Actually use revertMaxRevisions, ignored for 13 years

Amorymeltzer:
It seems to have been removed from use in [September 2007](https://en.wikipedia.org/w/index.php?title=User:AzaToth/twinklefluff.js&diff=156141985&oldid=146497464).

I'm not sure we even need this, since it's only *used* once, so I'm open to a discussion on that end.  The value is referenced in an error message, and in theory hardcoding a value in two places is bad, but so is a mostly-unused preference, even if hidden.


Amorymeltzer:
I'm probably in favor of actually just removing this, but I suppose at this point it doesn't really hurt

Amorymeltzer:
Good candidate for removal if we feel like doing an audit at some point, but for now, I suppose it's fine.

PR #965, opened 2020-06-10
config: Remove obsolete torev option from openTalkPage

Amorymeltzer:
When reverting to a revision, by definition there isn't a straightforward way to define who was in error, at least not if there more than one revision is being undone (and even then, it's not straightforward).  This hasn't been a functional option for well over a decade (https://en.wikipedia.org/w/index.php?title=User:AzaToth/twinklefluff.js&diff=104232943&oldid=103392000); I vaguely recall realizing this before I became a sysop.  There are plenty of users who have opted for it in their preferences, but it hasn't worked for anyone since well before the current preferences system was created!

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -347,7 +347,7 @@ Twinkle.config.sections = [
 				name: 'openTalkPage',
 				label: 'Open user talk page after these types of reversions',
 				type: 'set',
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I suppose we could try our best and just open the talk page for the most recent user, but that's not information that fluff has right now, so we'd need to adjust or make a new query.

PR #966, opened 2020-06-12
fluff/warn: Pass timestamp from fluff to warn, allow skipping stale API check

Amorymeltzer:
`timestamp` and `comment` have been in fluff's lookup queries since the module began, but they have been entirely unused since nearly the beginning.  I don't think the `comment` was ever used, while `timestamp` was removed just a few days after [inception](https://en.wikipedia.org/w/index.php?title=User:AzaToth/twinklefluff.js&diff=102706684&oldid=102702487).  It was then added to `revertToRevision` a [few days later](https://en.wikipedia.org/w/index.php?title=User:AzaToth/twinklefluff.js&diff=104232943&oldid=103392000) but was likewise not used there.  We can make use of it when rollbacking, though, if only to pass it to the warn module (a la `vanarticlerevid`, etc.), which will allow warn to avoid an API lookup when checking if the diff is stale (#757).

`timestamp` has been removed from the `revertToRevision` query, and `comment` has been removed from both; they're not unuseful, per se, we're just not using them.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -138,14 +138,7 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 		}
 
 		// Confirm edit wasn't too old for a warning
-		query = {
-			action: 'query',
-			prop: 'revisions',
-			rvprop: 'timestamp',
-			revids: vanrevid
-		};
-		new Morebits.wiki.api('Grabbing the revision timestamps', query, function(apiobj) {
-			var vantimestamp = $(apiobj.getResponse()).find('revisions rev').attr('timestamp');
+		var checkStale = function(vantimestamp) {
 			var revDate = new Morebits.date(vantimestamp);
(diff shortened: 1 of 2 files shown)


modules/twinklewarn.js
```diff
@@ -138,22 +138,32 @@ Twinkle.warn.callback = function twinklewarnCallback() {
...
 				}
 			}
-		}).post();
+		};
+
+		var vantimestamp = mw.util.getParamValue('vantimestamp');
+		// Provided from a fluff module-based revert, no API lookup necessary
+		if (vantimestamp) {
+			checkStale(vantimestamp);
+		} else {
```
Amorymeltzer, line 155: I vaguely remember there being something else that would or could provide `vanarticlerevid`, so I figured leaving the API lookup would be helpful, but I'm not so sure of that.  Perhaps it'd be fine to just do away with this and rely on ourselves?

PR #968, opened 2020-06-13
Add ESLint plugin to warn on ES6+

Amorymeltzer:
I stumbled on https://github.com/nkt/eslint-plugin-es5 about a year ago but it had too many false-positives when used with jQuery.  Most of that (https://github.com/nkt/eslint-plugin-es5/issues/12) appears to have been resolved with v1.5.0, specifically https://github.com/nkt/eslint-plugin-es5/pull/36.  This won't catch everything but I *think* it shouldn't have many false positives, as long as things like jQuery's `.find` are used on the same line as `$`, whether `jQuery` or as a variable name; we're inconsistent on that front, but it's not a bad idea to encourage it.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,6 @@
 {
-	"extends": ["eslint:recommended"],
+	"plugins": ["es5"],
+	"extends": ["eslint:recommended", "plugin:es5/no-es2015"],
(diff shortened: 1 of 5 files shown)


Amorymeltzer:
This will fail on the fixes in #967

PR #969, opened 2020-06-13
Add support for change tags, remove `summaryAd`

### LINKED ISSUE #265: Use tag instead of edit summary suffix to identify Twinkle edits
Now that we can define custom tags in MediaWiki and slap them on when editing, is it better to use that instead of the (TW) suffix? It's probably better for stats purposes, but I should say I have no idea how much people customize the suffix, or how they need the customizability.

Currently it [looks like this](https://zh.wikipedia.org/w/index.php?title=Special:Recentchanges&tagfilter=Twinkle&uselang=en) on zhwiki, and only [one line of code](https://github.com/jimmyxu/twinkle/commit/20fbe1518043e4aaa7fd6ef4fa0f2f0dc3929ad3) needs to be added.

Amorymeltzer:
See individual commits for details, but basically:

- Defines `Twinkle.changeTags = 'twinkle'` in twinkle.js and applies that across the board
    - Applied to `Morebits.wiki.page` via `setChangesTags()`, and manually to `Morebits.wiki.api` via `tags`
- Excludes PageTriage ([T252980](https://phabricator.wikimedia.org/T252980)) and FlaggedRevs ([T247721](https://phabricator.wikimedia.org/T247721)) (and other 
- Remove customization of `summaryAd` preferences (they options remain hidden as backups, including for use in the above)

There's some precedence for at least being polite and asking before creating a tag like this; I'll post at VPT at some point.  cc @MusikAnimal since this would be a notable change.  Closes #265

### DIFF
--- a/modules/friendlyshared.js
+++ b/modules/friendlyshared.js
@@ -156,7 +156,8 @@ Twinkle.shared.callbacks = {
 
 		var summaryText = 'Added {{[[Template:' + params.template + '|' + params.template + ']]}} template.';
 		pageobj.setPageText(text + pageText);
-		pageobj.setEditSummary(summaryText + Twinkle.getPref('summaryAd'));
+		pageobj.setEditSummary(summaryText);
+		pageobj.setChangeTags(Twinkle.changeTags);
 		pageobj.setMinorEdit(Twinkle.getPref('markSharedIPAsMinor'));
 		pageobj.setCreateOption('recreate');
 		pageobj.save();
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1194,12 +1194,12 @@ Twinkle.tag.callbacks = {
 			}
 
 			// avoid truncated summaries
-			if (summaryText.length > (499 - Twinkle.getPref('summaryAd').length)) {
+			if (summaryText.length > 499) {
 				summaryText = summaryText.replace(/\[\[[^|]+\|([^\]]+)\]\]/g, '$1');
 			}
 
 			pageobj.setPageText(pageText);
-			pageobj.setEditSummary(summaryText + Twinkle.getPref('summaryAd'));
+			pageobj.setEditSummary(summaryText);
 			pageobj.setWatchlist(Twinkle.getPref('watchTaggedPages'));
 			pageobj.setMinorEdit(Twinkle.getPref('markTaggedPagesAsMinor'));
 			pageobj.setCreateOption('nocreate');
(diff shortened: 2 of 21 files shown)


Amorymeltzer:
This could be nicer to situations where no tag is desired, thus `summaryAd` would be.  I'm open to ideas for handling that nicely.  One way I suppose would be to change `setChangeTags` to adjust `ctx.editSummary` if given a second parameter, so that it'd look something like:

```js
this.setChangeTags = function(tags, summary) {
	if (!summary) {
		ctx.changeTags = (!!tags && (typeof tags === 'string' ? tags : tags.join('|'))) || '';
	} else {
		ctx.editSummary = ctx.editSummary + summary;
	}
};

pageobj.setChangeTags(Twinkle.changeTags);
pageobj.setChangeTags(null, Twinkle.getPref('summaryAd'));
```

That's really not ideal.  It'd also require having `setChangeTags` after `setEditSummary` every time.

Amorymeltzer:
@Xi-Plus IIRC you already have a tag, but I wonder how other folks (@Huji?) feel about the above idea re: adding appending to the edit summary if no tag is defined: https://github.com/azatoth/twinkle/pull/969#issuecomment-646033039

MusikAnimal:
Just now seeing this. Exciting! Change tags are much easier to query than edit summaries :) Especially because users can alter or blank the `summaryAd`, making it difficult or impossible to identify when Twinkle was used.

Change tags don't apply to logged actions like deletion, right? I think we should continue to use `summaryAd` on those. You might consider removing the ability to configure it too, if people don't care. The point is to be able to identify semi-automation, and if we allow the "ad" to be configurable then we could lose that.

Amorymeltzer:
@MusikAnimal Thanks, and they do!  You've basically identified all the reasons for it, but inclusion for actions and in log entries is another big one.  Noted in the commits, but, for things we care about, only PageTriage and FlaggedRevs don't allow them yet.

I do kind of like the idea of hardcoding it, even beyond presence in `prefs`.

Amorymeltzer:
Having thought about [it](https://github.com/azatoth/twinkle/pull/969#issuecomment-646033039) for a bit, I don't think we should try to auto-edit edit summaries.  It's a lot of extra overhead and using a changetag simplifies and improves a lot of things greatly.

Rebased, etc.  I've incorporated @MusikAnimal's suggestion above, the two remaining uses of `Twinkle.getPref('summaryAd')` (FlaggedRevs and PageTriage, which don't currently allow for revision tags) now use the hardcoded `Twinkle.summaryAd` (for emotional continuity, although I'd prefer clarity `taglessAd`).  The deprecated `summaryAd`, `deletionSummaryAd`, and `protectionSummaryAd` remain in `Twinkle.defaultConfig` but *not* in `twinkleconfig.js`, so they should be available for anyone relying on them for a bit (atm, I think just Belle's arb.js) but not customizable.

Any objections?  I'll be notifying WT:TW/WP:VPT tomorrow.

Amorymeltzer:
Leaving morebits tag blank per feedback, but the option remains.

Amorymeltzer:
🎉 🏷️ 🎉

PR #970, opened 2020-06-13
Remove useragent 2.0 specification

Amorymeltzer:
First added in c236127 but the 2.0 specification hasn't been updated or considered in nearly a decade and its not something we're using.  These are used for logging purposes, all that really matters is that there's a unique and identifiable string, for which the tool and project suffice.

Also move to `wgWikiID` rather than `wgDBname` per https://phabricator.wikimedia.org/rMWd45baf7f0734a2506c77925064f42bc13e1bff5e

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1836,14 +1836,14 @@ Morebits.wiki.api.prototype = {
 
 // Custom user agent header, used by WMF for server-side logging
 // See https://lists.wikimedia.org/pipermail/mediawiki-api-announce/2014-November/000075.html
-var morebitsWikiApiUserAgent = 'morebits.js/2.0 ([[w:WT:TW]])';
+var morebitsWikiApiUserAgent = 'morebits.js ([[w:WT:TW]])';
 
 /**
  * Sets the custom user agent header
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
Sort of related, but @MusikAnimal what would you think about me tagging milestones?  I'd rather be a bit more continuous than some sort of semver, but something like YYYY-MM-DD for bulk on-wiki syncs seems reasonable.  Not particularly useful, but just to have something more recent...

MusikAnimal:
> Sort of related, but @MusikAnimal what would you think about me tagging milestones? I'd rather be a bit more continuous than some sort of semver, but something like YYYY-MM-DD for bulk on-wiki syncs seems reasonable. Not particularly useful, but just to have something more recent...

I didn't know we were using milestones! But any sort of tagging to a release seems like a good idea. You could also use https://github.com/azatoth/twinkle/releases and just put in the date rather than a version number.

Amorymeltzer:
Great!  That's basically my plan.  I've got a little bash script to help me collate commits, gonna incorporate it at some point once I add a few more features like this.

PR #971, opened 2020-06-14
morebits: Remove tipsy dependency, use jQuery UI for tooltips

Amorymeltzer:
This harkens back to the homegrown tooltips that were replaced by tipsy (a14f46f, 2011), mainly because we've still been defining the CSS!  Some tweaks were made: with a decade of black, I figured blue would be too bright, so mediumblue was used (we could, of course, keep it black).  There should only be minor differences in appearance.  The classes are changed, but there's nobody using them (in fact, there are more instances of old morebits pages with the *old*, pre-tipsy style).

https://api.jqueryui.com/tooltip/

### DIFF
--- a/morebits.css
+++ b/morebits.css
@@ -114,18 +114,18 @@ form.quickform .quickformSubgroup
 	margin-left: 3em;
 }
 
-form.quickform .tooltipButtonContainer
+/* The tooltip button and the content itself */
+form.quickform .morebits-tooltipButton
 {
-	position: relative;
-	width: 100%;
-}
-
-form.quickform .tooltipButton
-{
-	color: blue;
+	color: mediumblue;
 	font-weight: bold;
 	cursor: help;
-	padding: .2em;
+	padding: .3em;
+}
+.morebits-ui-tooltip
+{
+	padding: 4px 6px 4px 6px;
+	font-size: 13px;
 }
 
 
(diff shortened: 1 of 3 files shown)


morebits.css
```diff
@@ -114,18 +114,18 @@ form.quickform .quickformSubgroup
...
+	color: mediumblue;
 	font-weight: bold;
 	cursor: help;
-	padding: .2em;
+	padding: .3em;
+}
+.morebits-ui-tooltip
+{
+	padding: 4px 6px 4px 6px;
+	font-size: 1.2em;
```
siddharthvp, line 128: The font size within the tooltip is ENORMOUS at my end - looks like about 18px, not easily measurable.

siddharthvp:
Also, the tooltip is hiding behind the select2 dropdowns when they're open.

morebits.css
Amorymeltzer: 19.2, yeah.  This is annoying skin shit.  I was aiming for 12 px, but they behave differently.  For a quick example:

- Vector: `0.8em` of `100%` = 12.8px
- Timeless: `1.1em` of 0.95em` = 16.72px
- Modern: `1.1em` of `x-small` = 11px

There's a second issue, though.  12px looks a little small on Vector, because it has `font-family: sans-serif` whereas 13px is a little big on Modern and Timeless, because they both have `font-family: Verdana,Arial,sans-serif`.

YMMV, but IMO `13px` looks better on Modern and Timeless than `12px` does on Vector (and we should be favoring Vector anyway).  `rem` and `%` are no help AFAICT.

```suggestion
	font-size: 13px;
```

Amorymeltzer:
Well, we do have `.select2-container { z-index: 10000; }` everywhere.  This was discussed in https://github.com/azatoth/twinkle/pull/692#discussion_r369859551, although I'm not sure how necessary those are atm.  `.tipsy` was set to 100000, so we were off by a factor of ten.

We could just set the z-index to something similarly asinine (100000 or 10001), or really investigate whether we really still/ever needed the select2 z-index.

morebits.css
siddharthvp: The default (without any `font-size` set) looks good to me on vector, monobook and modern. For timeless, it's huge but easily restored back to normal by putting something like
```css
.skin-timeless .morebits-ui-tooltip {
	font-size: 13px;
}
```

siddharthvp:
> We could just set the z-index to something similarly asinine (100000 or 10001), or really investigate whether we really still/ever needed the select2 z-index.

Yeah, can't trigger any issues after removing the z-index on the select2 dropdowns; feel free to remove that.

Amorymeltzer:
There do appear to be a number of scripts that assume `jquery.tipsy` is available, which is probably a pretty good sign that folks were relying on twinkle/morebits using it.  That being said, a quick survey shows:

- Many are ancient and stale
- Most are derived from/clones of/inspired by one script, [User:Writ Keeper/Scripts/teahouseUtility.js](https://en.wikipedia.org/wiki/User:Writ_Keeper/Scripts/teahouseUtility.js) (~100 active users import, ~300 total)
- The tipsy bit isn't and hasn't been working there anyway

I left a note for WK on the talkpage, but I don't perceive this to be a problem or hindrance to removal.

PR #972, opened 2020-06-14
prod: Don't include empty reason in log

Amorymeltzer:


siddharthvp:
Shouldn't reason be required in prod module?

Amorymeltzer:
Eh, maybe if I had my druthers, but it's not required by policy or in the template.  PROD is theoretically only for uncontroversial cases where no opposition is expected; that's why a reason isn't required when either adding or removing the template (although specifically encouraged for removal).  I'd be in favor of adding a little "Hey this was empty did you mean to leave it blank?" alert, though.

PR #973, opened 2020-06-14
prod: Fix a number of bugs around endorsing

Amorymeltzer:
Most prominently, this moves the check for `{{old prod}}` to after the endorsement check, which allows proper endorsing (on the fritz since #735).  This means that if there's a prod already present, this will allow endorsing a (regular) PROD even if the talk page has been tagged (as should be the case).

Additional fixes:
- <s>Fix case sensitivity bug when endorsing/checking endorsing</s> Done in #975 
- Correctly note prior tag type when endorsing
- Move article age check into new prod check, where it makes the most sense.

Additionally, as written, this changes things so it *won't* add another old prod template to the talk page when BLPPRODing.  I'm on the fence about that, so would welcome feedback.


Amorymeltzer:
Actually, that first bullet point should probably be split out and pushed immediately... will try to take care of that later tonight.

PR #975, opened 2020-06-15
prod: Fix caps matching for {{Proposed deletion endorsed}}

Amorymeltzer:
Twinkle has been adding it with a lowercase p but only case-insensitively matching the uppercase P


Amorymeltzer:
Splitting this out from #973 to push bugfix.

PR #981, opened 2020-06-18
xfd: Fix xfd log numbering

mdaniels5757:
Issue: The log numbering at, e.g., [User:Mdaniels5757/XfD log at 963070422](https://en.wikipedia.org/wiki/Special:Permalink/963070422) is wrong (repeating 1s) because there is an extra newline when a reason is provided.

Solution: Removed that newline (line 817, twinklexfd.js).

Cheers,
Michael (User:Mdaniels5757)


Amorymeltzer:
Since I had access, I went and just made the change; will merge this now.  Thanks for the ID!

PR #988, opened 2020-06-19
Pass numbers through convertNumber

### LINKED ISSUE #987: Pass numbers through convertNumber
When a number is dynamically added to an interface message, pass it through `mw.language.convertNumber()` first.

In English, it helps by adding commas far large numbers. For example, in the batchdelete module, the size of a fill is [currently](https://github.com/azatoth/twinkle/blob/d3e40f37457f3fdb42fc2340d543b15d0a4ab8cc/modules/twinklebatchdelete.js#L191) inserted into a message using `metadata.push(size + ' bytes');` and changing it to `metadata.push(mw.language.convertNumber(size) + ' bytes');` will result in "143908 bytes" to read as "143,908 bytes" instead.

It also helps with #129 for non-English languages (e.g. those that use digits other than 123, or those that use a different "thousands separator")

Huji:
Fixes #987

### DIFF
--- a/modules/twinkleunlink.js
+++ b/modules/twinkleunlink.js
@@ -171,7 +171,7 @@ Twinkle.unlink.callbacks = {
 					if ($(xmlDoc).find('query-continue').length) {
 						apiobj.params.form.append({
 							type: 'div',
-							label: 'First ' + list.length.toString() + ' file usages shown.'
+							label: 'First ' + mw.language.convertNumber(list.length) + ' file usages shown.'
 						});
 					}
 					apiobj.params.form.append({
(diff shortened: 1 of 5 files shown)


Amorymeltzer:
@Huji Would you mind editing the Gadget definition in https://github.com/azatoth/twinkle/blob/master/DEVELOPER.md to include `mediawiki.language`?

modules/twinkleunlink.js
```diff
@@ -171,7 +171,7 @@ Twinkle.unlink.callbacks = {
 					if ($(xmlDoc).find('query-continue').length) {
 						apiobj.params.form.append({
 							type: 'div',
-							label: 'First ' + list.length.toString() + ' file usages shown.'
+							label: 'First ' + mw.language.convertNumber(list.length.toString()) + ' file usages shown.'
```
Amorymeltzer, line 174: Doesn't this negate the need for `toString()`?  AFAICT `mw.language.convertNumber` returns a string unless we pass a truthy second parameter.  Below as well.

Amorymeltzer:
Merged as cc32530!

PR #990, opened 2020-06-21
DEVELOPER: Document mediawiki.api dependency

DannyS712:
C.f. https://en.wikipedia.org/wiki/Special:Diff/963059130


DannyS712:
Dupe of #983

PR #991, opened 2020-06-21
friendlytalkback: minor cleanup to getNoticeWikitext

DannyS712:
Use guard clauses instead of needing `if` blocks

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -416,27 +416,27 @@ Twinkle.talkback.preview = function(form) {
 };
 
 Twinkle.talkback.getNoticeWikitext = function(tbtarget, page, section, message) {
-	var text;
 	if (tbtarget === 'notice') {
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
This is quite welcome!  I can rework #929 to give this precedence, but this is gonna conflict pretty heavily with #908.  I want to get those massive changes though, so let's revisit once that's merged?

Amorymeltzer:
I *think* this got invalidated by the addition of `switch`es in #908, is that right?  Sorry for unspent effort

DannyS712:
> I _think_ this got invalidated by the addition of `switch`es in #908, is that right? Sorry for unspent effort

Nope

Amorymeltzer:
Are the `if`s better than the `switch`, or, sorry, am I missing a different issue?

siddharthvp:
Likewise, not seeing any improvements here after the rebase.

PR #992, opened 2020-06-21
Friendly Tag: Separate file tag data and rendering

DannyS712:
Move the data for the different groups of file tags to `Twinkle.tag.file.groups`. This will make it easier for projects that fork Twinkle to update it to their local tags by separating what is project-dependent (the tag data) from what is project-independent (the handling of adding the form checkboxes)

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -162,20 +162,10 @@ Twinkle.tag.callback = function friendlytagCallback() {
 		case 'file':
 			Window.setTitle('File maintenance tagging');
 
-			form.append({ type: 'header', label: 'License and sourcing problem tags' });
-			form.append({ type: 'checkbox', name: 'tags', list: Twinkle.tag.file.licenseList });
-
-			form.append({ type: 'header', label: 'Wikimedia Commons-related tags' });
-			form.append({ type: 'checkbox', name: 'tags', list: Twinkle.tag.file.commonsList });
-
-			form.append({ type: 'header', label: 'Cleanup tags' });
-			form.append({ type: 'checkbox', name: 'tags', list: Twinkle.tag.file.cleanupList });
-
-			form.append({ type: 'header', label: 'Image quality tags' });
-			form.append({ type: 'checkbox', name: 'tags', list: Twinkle.tag.file.qualityList });
-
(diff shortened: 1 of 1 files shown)


DannyS712:
If this is approved, I'll send a larger patch for similar changes across the repo - there are other places where the same logic applies.

DannyS712:
Not sure how to properly add reviewers with github, but @Amorymeltzer can you take a look?

Amorymeltzer:
These definitely do need tightening IMO.  I was mulling over some similar/related changes a while ago, but is there a reason not to go one further and throw all the file tags into a single object, like we have in `Twinkle.tag.article.tagCategories`?  We could do something like we have at https://github.com/azatoth/twinkle/blob/1e616f5ad6ba02b59f1975033f9524b238a8e534/modules/friendlytag.js#L569-L581 for articles, yeah?

DannyS712:
> These definitely do need tightening IMO. I was mulling over some similar/related changes a while ago, but is there a reason not to go one further and throw all the file tags into a single object, like we have in `Twinkle.tag.article.tagCategories`? We could do something like we have at
> 
> https://github.com/azatoth/twinkle/blob/1e616f5ad6ba02b59f1975033f9524b238a8e534/modules/friendlytag.js#L569-L581
> 
> for articles, yeah?

I was trying to be conservative in the changes, because I was hoping to do something like this on a larger scale and wanted to ease review. Keeping it simple like this makes it easier to follow the changes - perhaps convert to a single object as a follow-up?

Amorymeltzer:
Well, I suppose it depends on what and where you intend on doing it.  These are brief enough that, for this anyway, it'll be fine to review, so I don't think a half-measure is needed on my end.

I found my old redirect stuff, basically doing the same thing as this; I've got some descriptions to add but I can push it up so our work is aligned.

Amorymeltzer:
@DannyS712 Example at #1002

Amorymeltzer:
@DannyS712 I dealt with the conflicts and restructured it slightly further to match, can I push it here for your review?

DannyS712:
> @DannyS712 I dealt with the conflicts and restructured it slightly further to match, can I push it here for your review?

Sure

PR #993, opened 2020-06-21
fluff: Fail nicely on and don't reveal revdel'd usernames

Amorymeltzer:
Sometimes we use or reference usernames in fluff, whether the user being reverted or occasionally the user of the revision to which we are reverting.  This would cause issues if the username was hidden; more alarmingly, if the twinkler was a sysop, the username could be inadvertantly revealed (same for oversight although we're pretty careful about clean up).  This basically checks the `userhidden` property on the selected revisions, then uses that to determine whether to display the user name or a placeholder text (present in the new (as of this commit) `Twinkle.fluff.hiddenName`, set to `an unknown user`).  There is some slightly extra infrastructure around `params.user`, given that the selectors mean we're passing a falsey value for the username, but I think it's clearer this way, and opens the door for future works around cases like diffs viewed with `&unhide=1`.

The main annoyance is on Special:Contributions, where someone with the appropriate group would be able to see revdel'd edits.  The only difference for those edits is a `strong` element with no span or class selector available, so in order to selectively remove these, we'd have to do something like searching for all `strong` elements, then check the content of each one, as there are other bolded elements and nearby items (rollback, tags) are not reliably present.  We could manually add something to [[MediaWiki:Rev-deleted-user-contribs]], but that's far from ideal, especially given how rare this likely is.  Thus, it's not done, at least until [T255903](https://phabricator.wikimedia.org/T255903).

I also improved the `formatSummary` function a bit.  It now checks if the `$USER` string was present before attempting to replace it.  It likewise now checks that a username was supplied before constructing a link; if not, it will use `Twinkle.fluff.hiddenName`.  I also bumped the edit summary limit (done for tag in 009c80e/#886, see https://phabricator.wikimedia.org/T188798)

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -61,6 +61,8 @@ Twinkle.fluff = function twinklefluff() {
 Twinkle.fluff.trustedBots = ['AnomieBOT', 'SineBot'];
 Twinkle.fluff.skipTalk = null;
 Twinkle.fluff.rollbackInPlace = null;
+// String to insert when a username is hidden
+Twinkle.fluff.hiddenName = 'an unknown user';
 
 // Consolidated construction of fluff links
 Twinkle.fluff.linkBuilder = {
@@ -82,6 +84,7 @@ Twinkle.fluff.linkBuilder = {
 
 	/**
 	 * @param {string} vandal - Username of the editor being reverted (required)
+	 * Provide a falsey value if the username is hidden
 	 * @param {boolean} inline - True to create two links in a span, false
 	 * to create three links in a div (optional)
 	 * @param {number|string} [rev=wgCurRevisionId] - Revision ID being reverted (optional)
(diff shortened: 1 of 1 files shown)


modules/twinklefluff.js
```diff
@@ -64,6 +64,8 @@ Twinkle.fluff = function twinklefluff() {
 
 Twinkle.fluff.skipTalk = null;
 Twinkle.fluff.rollbackInPlace = null;
+// String to insert when a username is hidden
+Twinkle.fluff.hiddenName = 'an unknown user';
```
siddharthvp, line 68: Weren't we using `<username hidden>` at some point?

siddharthvp:
This seems like a lot of extra complexity. Instead of keeping separate properties like `params.userHidden` why not just store the username as a placeholder value (a string which can't be valid username), and use that?

modules/twinklefluff.js
Amorymeltzer: In arv, yeah, in #875.  I thought this would be clearer in a sentence-like edit summary

Amorymeltzer:
>why not just store the username as a placeholder value

I did that at first, but this ended up being clearer (there were a fair bit of `if - else`s going on).  I also wanted to keep the option to show and process information for sysops who can actually see it.  I don't particularly mind either way, tbh.

PR #998, opened 2020-06-22
unlink: Alert/return if no pages selected, show more accurate interface message

Amorymeltzer:
Draft as it will conflict with #908, want to wait on that.

### DIFF
--- a/modules/twinkleunlink.js
+++ b/modules/twinkleunlink.js
@@ -102,19 +102,24 @@ Twinkle.unlink.callback.evaluate = function twinkleunlinkCallbackEvaluate(event)
 		return;
 	}
 
-	Morebits.simpleWindow.setButtonsEnabled(false);
-	Morebits.status.init(form);
-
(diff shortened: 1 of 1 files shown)


modules/twinkleunlink.js
```diff
@@ -102,19 +102,24 @@ Twinkle.unlink.callback.evaluate = function twinkleunlinkCallbackEvaluate(event)
 		return;
 	}
 
-	Morebits.simpleWindow.setButtonsEnabled(false);
-	Morebits.status.init(form);
-
 	input.backlinks = input.backlinks || [];
 	input.imageusage = input.imageusage || [];
 	var pages = Morebits.array.uniq(input.backlinks.concat(input.imageusage));
+	if (!pages) {
```
Xi-Plus, line 108: `![]` equals false

Amorymeltzer: Arg, yes of course, thanks.  ede3a22

PR #999, opened 2020-06-23
[ARV] Cleanup: Don't save single-use api variables

DannyS712:
If an api object is only used once, no need to store it in a variable


Amorymeltzer:
🎆

PR #1000, opened 2020-06-25
tag: Disallow conflicting file tags and use functions to validate

Amorymeltzer:
1st commit disallows some conflicting file tags, as requested at [WT:TW](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_42#Mutually_exclusive_tags).  There are probably more we could do, this covers:
- Only one of {{Bad GIF}}, {{Bad JPEG}}, {{Bad SVG}}, and {{Bad format}} is used
- Only one of {{Should be PNG}}, {{Should be SVG}}, and {{Should be text}} is used
- Only one of {{Bad SVG}} and {{Vector version available}} is used
- Only one of {{Bad JPEG}} and {{Overcompressed JPEG}} is used
- Only one of {{PNG version available}} and {{Vector version available}} is used
- The correct extension is used for Bad GIF/JPEG/SVG and Should be PNG/SVG
- Only allows JPEGs to be tagged with {{Overcompressed JPEG}}
- Only allows SVGs to be tagged with {{Bad trace}} and {{Bad font}}

2nd commit uses `checkIncompatible` and `checkParameter` to make the repetitive validation measures more straightforward.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1984,6 +1984,35 @@ Twinkle.tag.callback.evaluate = function friendlytagCallbackEvaluate(e) {
 			el.type === 'checkbox' ? form[el.name].checked : form[el.name].value;
 	});
 
+
+	// Validation
+
+	// Given an array of incompatible tags, check if we have two or more selected
+	var checkIncompatible = function(conflicts, extra) {
+		var count = conflicts.reduce(function(sum, tag) {
+			return sum += params.tags.indexOf(tag) !== -1;
+		}, 0);
+		if (count > 1) {
+			var message = 'Please select only one of: {{' + conflicts.join('}}, {{') + '}}.';
+			message += extra ? ' ' + extra : '';
+			alert(message);
+			return true;
+		}
+	};
+	// Given a tag, ensure an associate parameter is present
+	// Maybe just sock this away in each function???
+	var checkParameter = function(tag, parameter, description) {
+		description = description || 'a reason';
+		if (params.tags.indexOf(tag) !== -1 && params[parameter].trim() === '') {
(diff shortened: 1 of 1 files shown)


modules/friendlytag.js
```diff
@@ -2009,47 +2035,97 @@ Twinkle.tag.callback.evaluate = function friendlytagCallbackEvaluate(e) {
...
+			if (checkIncompatible(['Bad JPEG', 'Overcompressed JPEG'])) {
+				return;
+			}
+			if (checkIncompatible(['PNG version available', 'Vector version available'])) {
+				return;
+			}
+
+			var extension;
+			if ((extension = Morebits.pageNameNorm.match(/\.(\w+)$/)) !== null) {
+				extension = extension[1];
```
Amorymeltzer, line 2072: Per a [discussion with the requester](https://en.wikipedia.org/wiki/User_talk:Smjg#Twinkle_pull_request_for_your_file_incompatibility_checks), it should be possible to use MIME types for jpg/gif/png, with a fallback to this for svgs.  Something stupid like `((extension = $('.mime-type').text()) && extension.split(/\//)[1]) || ((extension = Morebits.pageNameNorm.match(/\.(\w+)$/)) !== null) && (extension = extension[1])` works 😆

siddharthvp: To get the extension, we can use `mw.Title.newFromText(Morebits.pageNameNorm).getExtension()` as well.

Amorymeltzer: Done

Amorymeltzer:
Unrelated, but woo, 1000 issues/PRs!

PR #1002, opened 2020-06-27
tag: Restructure redirect list, expand list by ~60

Amorymeltzer:
Restructures the list of redirects to something like the article category list, and greatly expands the list of redirects available for tagging.

The latter has been requested frequently, and since the introduction of the quickfiler, it's much more feasible.  Using [Template:R template index](https://en.wikipedia.org/wiki/Template:R_template_index) as the model, this categorizes redirects with those headers and adds every missing redirect with >500 transclusions, which is about 60 new items.  If we limited it to those with >1000, it'd be about 40.

In theory, this is one of many steps toward unifying the various tagging mechanisms.  The headers created here are a little wonky, and with #908 the quickfilter should be able to more readily help, making use of `$allCheckboxDivs` and `$allHeaders`

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -823,23 +823,34 @@ Twinkle.tag.article.tagList = {
 };
 
 // Tags for REDIRECTS start here
+// Not by policy, but the list roughly approximates items with >500
+// transclusions from Template:R template index
 Twinkle.tag.redirectList = {
 	'Grammar, punctuation, and spelling': {
 		'Abbreviation': [
 			{ tag: 'R from acronym', description: 'redirect from an acronym (e.g. POTUS) to its expanded form' },
-			{ tag: 'R from initialism', description: 'redirect from an initialism (e.g. AGF) to its expanded form' }
+			{ tag: 'R from initialism', description: 'redirect from an initialism (e.g. AGF) to its expanded form' },
+			{ tag: 'R from MathSciNet abbreviation', description: 'redirect from MathSciNet publication title abbreviation to the unabbreviated title' },
+			{ tag: 'R from NLM abbreviation', description: 'redirect from a NLM publication title abbreviation to the unabbreviated title' }
 		],
 		'Capitalisation': [
 			{ tag: 'R from CamelCase', description: 'redirect from a CamelCase title' },
-			{ tag: 'R from other capitalisation', description: 'redirect from a title with another method of capitalisation' }
+			{ tag: 'R from other capitalisation', description: 'redirect from a title with another method of capitalisation' },
+			{ tag: 'R from miscapitalisation', description: 'redirect from a capitalisation error' }
 		],
 		'Grammar & punctuation': [
 			{ tag: 'R from modification', description: 'redirect from a modification of the target\'s title, such as with words rearranged' },
-			{ tag: 'R from plural', description: 'redirect from a plural word to the singular equivalent' }
+			{ tag: 'R from plural', description: 'redirect from a plural word to the singular equivalent' },
+			{ tag: 'R to plural', description: 'redirect from a singular noun to its plural form' }
+		],
+		'Parts of speech': [
+			{ tag: 'R from verb', description: 'redirect from an English-language verb or verb phrase' },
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I left off `R from move` since it's now added automatically, but I suppose we could add it

modules/friendlytag.js
```diff
@@ -168,14 +168,25 @@ Twinkle.tag.callback = function friendlytagCallback() {
...
+				var subdiv = form.append({ type: 'div', id: 'tagSubdiv' + i++ });
+				$.each(content, function(subtitle, array) {
+					subdiv.append({ type: 'div', label: [ Morebits.htmlNode('b', subtitle) ] });
 
-			form.append({ type: 'header', label: 'Alternative name templates' });
-			form.append({ type: 'checkbox', name: 'redirectTags', list: Twinkle.tag.alternativeList });
+					var checkboxes = [];
+					$.each(array, function(k, item) {
+						checkboxes.push({ value: item.tag, label: '{{' + item.tag + '}}: ' + item.description });
+					});
```
siddharthvp, line 181: This can be more efficient using `map` as `checkboxes = array.map(item => ({ ... });`.

modules/friendlytag.js
```diff
@@ -168,14 +168,25 @@ Twinkle.tag.callback = function friendlytagCallback() {
 		case 'redirect':
 			Window.setTitle('Redirect tagging');
 
-			form.append({ type: 'header', label: 'Spelling, misspelling, tense and capitalization templates' });
-			form.append({ type: 'checkbox', name: 'redirectTags', list: Twinkle.tag.spellingList });
+			var i = 1;
+			$.each(Twinkle.tag.redirectList, function(title, content) {
+				form.append({ type: 'header', id: 'tagHeader' + i, label: title });
+				var subdiv = form.append({ type: 'div', id: 'tagSubdiv' + i++ });
+				$.each(content, function(subtitle, array) {
```
siddharthvp, line 175: [minor] Can you rename the variable names so that they are more intuitive? `content` -> `group`, `array` -> `subgroup`, `title` -> `groupName`, `subtitle` -> `subgroupName`.

modules/friendlytag.js
Amorymeltzer: Word!  Done.

modules/friendlytag.js
Amorymeltzer: Was copied from article, but yeah, much better.  See https://github.com/azatoth/twinkle/pull/1003#discussion_r446517069 as well.

mdaniels5757:
> I left off `R from move` since it's now added automatically, but I suppose we could add it.

That'd be nice; I use {{R from move}} manually a fair bit for moves after round-robin swaps.

Amorymeltzer:
Yeah, that's fair — added.

Amorymeltzer:
One thing to consider is whether there is/are any validations that should be done.

The short answer is yes: there are plenty that conflict (e.g. `{{R printworthy}}` and `{{R unprintworthy}}`) but the long answer is it's complex and we should only try to handle the obvious cases.

modules/friendlytag.js
```diff
@@ -186,14 +186,21 @@ Twinkle.tag.callback = function friendlytagCallback() {
...
+			$.each(Twinkle.tag.redirectList, function(groupName, group) {
+				form.append({ type: 'header', id: 'tagHeader' + i, label: groupName });
+				var subdiv = form.append({ type: 'div', id: 'tagSubdiv' + i++ });
+				$.each(group, function(subgroupName, subgroup) {
+					subdiv.append({ type: 'div', label: [ Morebits.htmlNode('b', subgroupName) ] });
+					subdiv.append({
+						type: 'checkbox',
+						name: 'redirectTags',
+						list: subgroup.map(function (item) {
+							return { value: item.tag, label: '{{' + item.tag + '}}: ' + item.description };
```
Xi-Plus, line 199: Bug: Lost your extra input data stored in subgroup
Suggest: `return { value: item.tag, label: '{{' + item.tag + '}}: ' + item.description, subgroup: item.subgroup };`

Amorymeltzer: Thanks for keeping me honest @Xi-Plus, done (plus another from a different PR ) in #1055

PR #1009, opened 2020-06-28
tag: Add {{paid contributions}}

unforgettableid:
{{Paid contributions}} is quite a useful template, but it has only about 100 mainspace transclusions. Why so few? Perhaps the explanation is: It's not in Twinkle yet, and therefore only a small number of editors know that it exists. :)

{{Paid contributions}} is different from {{undisclosed paid}}.

{{Paid contributions}} is for disclosed paid edits which may need cleanup. {{Undisclosed paid}} is for undisclosed edits which look like they were probably paid.


Amorymeltzer:
Thanks for offering this!  There's some history/discussion in #178 and #696 around inclusion criteria (tl;dr there's no hard-and-fast criteria).  As you note, there's a chicken-egg concern with templates (more likely to be used if in Twinkle), but we also can't really account for tempo, since it's unclear how quickly some tags are removed.  100 looks really low and unused, but it could also just be well-monitored and removed.

At any rate, I think this would benefit from some on-wiki discussion/consensus first.  The particular concern is a serious one, and, especially given the low transclusion count, I think it wouldn't hurt to have robust support.  Especially since things like {{COI}} exist, which are broader and encompassing.

siddharthvp:
This recently came up at the [AFC talk page](https://en.wikipedia.org/wiki/Wikipedia_talk:WikiProject_Articles_for_creation#Tagging_confirmed_paid_editor_pages). I think we should merge this as basically the thing is that people are going to erroneously apply {{undisclosed paid}} or {{COI}} if they don't see this template in Twinkle.

Since it can only be applied on _confirmed_ paid editing cases, 100 doesn't sound that low a figure.

Amorymeltzer:
Yeah, not opposed *per se*, just think inclusion criteria are worth keeping in mind.  Anyway, this has been around for a while, so fine by me.

Amorymeltzer (approved):
Thanks for the submission and patience!

PR #1010, opened 2020-06-29
block: fix IE bug

siddharthvp:
In IE, fieldset elements are not available as properties on the HTMLFormElement object. The loop was silently being skipped as `$(undefined).find(...).each(...)` doesn't throw. This bug has been here since the block module was added in 465eb53.


Amorymeltzer:
Gonna merge live

PR #1014, opened 2020-06-30
warn: Fix bug where autolevel message wouldn't display

Amorymeltzer:
The red notice area was added as `olddiff-messages` in #757 and renamed to `revert-messages` in #855.  The autolevel option in #777 tried to change this to `warning-message` but utterly failed to do so.  Thus, the (non-red) autolevel has not been showing which option was being used.


Amorymeltzer:
Manually pushed this live

PR #1020, opened 2020-07-06
protect: Don't suggest asking bots to unprotect

Amorymeltzer:
Uses `Twinkle.protect.trustedBots`, just two on enwiki (MusikBot II and TFA Protector Bot)


siddharthvp:
I'm thinking that we don't need another `trustedBots` array. Since only adminbots can protect pages, they're all trusted. What I would do is to use the bot username regex check to tell apart the bots.

In fluff, hardcoding `trustedBots` was necessary because otherwise it would be possible for anyone to register an account with a bot-like username and vandalise, and Twinkle won't be able to revert those edits. Such a situation doesn't occur here in this module.

Amorymeltzer:
Fair enough, although I figured this way would likely be more helpful for other projects.

PR #1022, opened 2020-07-06
Add `insertAfterTemplates` to Morebits.wikitext.page, respect MOS:ORDER when tagging, update hatnotes

Amorymeltzer:
As noted [on WT:TW](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=964243708#Twinkle_violating_MOS:ORDER_-_needs_correction), Twinkle isn't great about [MOS:ORDER](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Layout#Order_of_article_elements) when tagging things.  Accordingly, this:

- 1st commit adds a function to Morebits.wikitext.page to insert a string after a provided template regex.  This is basically copied from the handling in `tag`.  Takes either a string or an array and allows for custom flags.
- 2nd commit uses that in xfd/protect/prod/csd/tag, with tag getting the extra benefit of finally learning about protection tags (`{{pp}}`, `{{pp-move}}`, etc.).  A new hidden preference, `hatnoteRegex`, is added to store the "various hatnotes templates` regex in a unified location.
- 3rd commit expands thats regex with most hatnotes from [Template:Hatnote templates](https://en.wikipedia.org/w/index.php?title=Template:Hatnote_templates) with more than a hundred or so transclusions.

There are some more notes in the individual commits.  The main annoyance is that AfD has a comment *before* the actual template, so having to support that as well is a drag.  Maybe the option to specify a flag for the regex for `insertAfterTemplates` is unnecessary?

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -4009,6 +4009,72 @@ Morebits.wikitext.page.prototype = {
 		return this;
 	},
 
+	/**
+	 * Smartly insert a tag atop page text but after specified templates,
+	 * such as hatnotes, short description, or deletion and protection templates.
+	 * Notably, does *not* insert a newline after the tag
+	 *
+	 * @param {string} tag - The tag to be inserted
+	 * @param {string|string[]} regex - Templates after which to insert tag,
+	 * given as either as a (regex-valid) string or an array to be joined by pipes
+	 * @param {string} [flags=i] - Regex flags to apply. Optional, defaults to /i
+	 * @param {string|string[]} preRegex - Optional regex string or array to match
+	 * before any template matches (i.e. before `{{`), such as html comments
+	 *
+	 * @returns {Morebits.wikitext.page}
+	 */
+	insertAfterTemplates: function(tag, regex, flags, preRegex) {
+		if (!tag) {
+			throw new Error('No tag provided');
+		}
+
+		// .length is only a property of strings and arrays so we
+		// shouldn't need to check type
+		if (!regex || !regex.length) {
+			throw new Error('No regex provided');
+		} else if (Array.isArray(regex)) {
+			regex = regex.join('|');
+		}
+
+		flags = flags || 'i';
+
+		if (!preRegex || !preRegex.length) {
+			preRegex = '';
+		} else if (Array.isArray(preRegex)) {
+			preRegex = preRegex.join('|');
+		}
+
+
+		// Regex is extra complicated to allow for templates with
+		// parameters and to handle whitespace properly
+		this.text = this.text.replace(
+			new RegExp(
+				// leading whitespace
+				'^\\s*' +
+				// capture template(s)
+				'(?:((?:\\s*' +
+				// Pre-template regex, such as leading html comments
+				preRegex + '|' +
+				// begin template format
+				'\\{\\{\\s*(?:' +
+				// Template regex
+				regex +
+				// end main template name, optionally with a number
+				// Probably remove the (?:) though
+				')\\d*\\s*' +
+				// template parameters
+				'(\\|(?:\\{\\{[^{}]*\\}\\}|[^{}])*)?' +
+				// end template format
+				'\\}\\})+' +
+				// end capture
+				'(?:\\s*\\n)?)' +
+				// trailing whitespace
+				'\\s*)?',
+				flags), '$1' + tag
+		);
+		return this;
+	},
+
 	/** @returns {string} */
 	getText: function() {
 		return this.text;
--- a/morebits.js
+++ b/morebits.js
@@ -3894,6 +3894,8 @@ Morebits.wikitext.page.prototype = {
 	/**
 	 * Removes links to `link_target` from the page text.
 	 * @param {string} link_target
+	 *
+	 * @returns {Morebits.wikitext.page}
 	 */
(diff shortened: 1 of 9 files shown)


twinkle.js
```diff
@@ -126,6 +126,9 @@ Twinkle.defaultConfig = {
 	// Hidden preferences
 	autolevelStaleDays: 3, // Huggle is 3, CBNG is 2
 	revertMaxRevisions: 50, // intentionally limited
+	// various hatnote templates, used when tagging
+	// (csd/xfd/tag/prod/protect) to ensure MOS:ORDER
+	hatnoteRegex: 'short description|hatnote|main|correct title|dablink|distinguish|for|further|selfref|year dab|similar names|highway detail hatnote|broader|about(?:-distinguish| other people)?|other\\s?(?:hurricane(?: use)?s|people|persons|places|ships|uses(?: of)?)|redirect(?:-(?:distinguish|synonym|multi))?|see\\s?(?:wiktionary|also(?: if exists)?)',
```
siddharthvp, line 131: This is not a user preference at all. What we're missing is a way to share data/code between modules. It would be better to define this as `Twinkle.hatnoteRegex` in `twinkle.js` file at the bottom under a header called something like: 

```js
/**
 * Data and code shared between multiple modules,
 * but not generic enough to be in morebits.
 */
```

Such a section will also come handy for a lot of other stuff too.

morebits.js
```diff
@@ -4001,6 +4001,71 @@ Morebits.wikitext.page.prototype = {
...
+		if (!preRegex || !preRegex.length) {
+			preRegex = '';
+		} else if (Array.isArray(preRegex)) {
+			preRegex = preRegex.join('|');
+		}
+
+
+		// Regex is extra complicated to allow for templates with
+		// parameters and to handle whitespace properly
+		return this.text.replace(
```
siddharthvp, line 4042:
```suggestion
		return this.text = this.text.replace(
```
This should at least modify `this.text` at the least. 

I'm not sure whether it'd better to return the modified text or just `this` (which would make the method chainable). Presently, the other Morebits.wikitext methods don't return anything so changing them all to  return `this` is what would make sense IMO, so that it can be used like `text = wikipage.insertAfterTemplates(...).getText()`.

morebits.js
```diff
@@ -4001,6 +4001,71 @@ Morebits.wikitext.page.prototype = {
...
+	 */
+	insertAfterTemplates: function(tag, regex, flags, preRegex) {
+		if (!tag) {
+			return this.text;
+		}
+
+		// .length is only a property of strings and arrays so we
+		// shouldn't need to check type
+		if (!regex || !regex.length) {
+			return tag + this.text;
```
siddharthvp, line 4026: It's better to throw when the input is clearly invalid, so that the client can know they're doing something wrong.

morebits.js
```diff
@@ -4001,6 +4001,71 @@ Morebits.wikitext.page.prototype = {
...
 	},
 
+	/**
+	 * Smartly insert a tag atop page text but after specified templates,
+	 * such as hatnotes, short description, or deletion and protection templates.
+	 * Notably, does *not* insert a newline after the tag
+	 *
+	 * @param {string} tag - The tag to be inserted
+	 * @param {string|array} regex - Templates after which to insert tag,
+	 * given as either as a string or an array to be joined by pipes
```
siddharthvp, line 4011:
```suggestion
	 * @param {string|string[]} regex - Templates after which to insert tag,
	 * given as either as a string or an array to be joined by pipes
```
Also, this should clarify that the strings can be regexes too.

twinkle.js
Amorymeltzer: Exactly right.  I wrote it that way at first, but shied away from it because there simply wasn't anything else there.  I thought having it in `defaultConfig` might make it more obvious that "this is a thing to change," but, yes, it'd be better to just do it normally.

morebits.js
Amorymeltzer: No, that makes sense, and chaining would be ideal.  I can toss in some `return this`-es to the others, IIRC most of the uses are simple.

morebits.js
Amorymeltzer: Didn't know about `string[]`!

morebits.js
Amorymeltzer: Unused on enwiki, so added here.

morebits.js
Amorymeltzer: Fair, although I'm mixed on how much of an error/invalid we should actually consider these.  I guess it can just be handled on the "client" side, but this seemed nicer.  You're probably right about giving them the option of wanting to know.

Amorymeltzer:
>It would be great if you can include some unit tests too

I'll get around to #946/#944/#911 soon, hopefully.

twinkle.js
Amorymeltzer: Also a good place for #969

morebits.js
Amorymeltzer: Went with both

Amorymeltzer: I suppose technically we could check the flags, something like `/^[gimsuy]$/`

PR #1028, opened 2020-07-10
warn: Fix bug in autolevel not elevating level for singlewarn

### LINKED ISSUE #1027: warn: Autolevel didn't set to 4 if previous warning is signlewarn
Related code:
https://github.com/azatoth/twinkle/blob/07fffdf04bf8adf008862eb2ac2e5cc52c28b9d9/modules/twinklewarn.js#L1499-L1516
Level increased by 1 according to previous warning. But didn't do if previous warning is signlewarn.
If I understand correctly, ~L1508 should check `latest.type` instead of `template` (`params.sub_group`)~.

Amorymeltzer:
Was using the default selected template rather than the latest warning found
in the page object, meaning the "found" level would be set to 1, so even if a
singlewarn template was found, it wouldn't be elevated.  If you selected a
singlewarn template, then went to the autolevel menu, the default would *be* a
singlewarn, so it would appear to be processed properly.  Fixes #1027.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1497,15 +1497,13 @@ Twinkle.warn.callbacks = {
 	* @returns {Array} - Array that contains the full template and just the warning level
 	*/
 	autolevelParseWikitext: function(wikitext, params, latest, date, statelem) {
-		var template = params.sub_group.replace(/(.*)\d$/, '$1');
-
 		var level; // undefined rather than '' means the isNaN below will return true
 		if (/\d(?:im)?$/.test(latest.type)) { // level1-4im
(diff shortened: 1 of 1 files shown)


Xi-Plus:
I think it is better to change this line.
https://github.com/azatoth/twinkle/blob/07fffdf04bf8adf008862eb2ac2e5cc52c28b9d9/modules/twinklewarn.js#L1500

Amorymeltzer:
That could/should be moved, but it's for the final return, ~L1570

Amorymeltzer:
Moved down, although truth be told there should probably be a check early on to ensure it's actually there.

PR #1032, opened 2020-07-11
morebits: mb.w.p: Don't allow null append or prepend text

Amorymeltzer:


siddharthvp:
Should it really throw? The prependtext or appendtext could be determined from code and may turn out to be an empty string in some applications. In such a case, it would make more sense to silently abort rather than throw.

Amorymeltzer:
Fair point, and that mirrors regular `save` when the text is the same.  I'd
say instead this should be checked against `null`, as that's the default
context value, and only throw then.

morebits.js
```diff
@@ -2236,9 +2236,19 @@ Morebits.wiki.page = function(pageName, currentAction) {
 
 		switch (ctx.editMode) {
 			case 'append':
+				if (ctx.appendText === 'null') {
```
siddharthvp, line 2239:
```suggestion
				if (ctx.appendText === null) {
```
comparing with string null?

Amorymeltzer: Brainfart 🧠 💨

Issue #379, opened 2017-04-01
Nominating redirects for AfD puts tag on the wrong page


lee533:
See [https://en.wikipedia.org/wiki/Wikipedia:Articles_for_deletion/P!nk_Live_2017](url). The AfD nominator had redirected the page, then( wrongly of course) nominated the redirect for deletion. [https://en.wikipedia.org/wiki/Pink_(singer)](url), the redirect target, ended up receiving the AfD tag. Probably because when Twinkle loads the page to add the tag it will load the target instead of the redirect. Simple solution would be to disable nominating redirects for AfD, since I don't think there's a valid reason to do that (at best, people should revert a redirect before nominating). Could be a better way to do this, though. 

(ansh666 en-wiki)

Amorymeltzer:
This is intentional (see https://github.com/azatoth/twinkle/blob/master/modules/twinklexfd.js#L553), at least it was eight years ago.  The idea is that if an article/template/page is moved (or redirected, in this case) the intent of the nomination is to nominate the page, not the redirect, so Twinkle finds the target page in question.  I think that's helpful in case something happens, especially since the xfd module automatically suggests RfD on redirects; essentially, I think it better to trust the user selected the right option rather than trust that nothing changed in the background.

It *could* be worth adding a confirmation, though, as it might be surprising to the tagger.

siddharthvp:
It will indeed be surprising to the user, as the AfD for the article `Bar` which was redirected from `Foo` will still be called `Wikipedia:Articles for deletion/Foo` (as the determination of discussion page name takes place before tagging)

Something ought to be done about this, even if it was intentional 8 years ago.

1. If the page is a redirect as seen by the user (`mw.config.get('wgIsRedirect')`) then, Twinkle should right away stop when user submits, giving back the reason text, and telling them to use RfD instead.
2. If the page was not a redirect as seen by user (`!mw.config.get('wgIsRedirect')`) but is a redirect when loaded internally (`pageobj.getPageName() !== Morebits.pageNameNorm`, I think, as `ctx.pageName` would have been updated to become the redirect target name), then a warning should be displayed with the text `Page is now a redirect. <a>Click to check history</a> (opens in a new tab)` and be given the option to continue nominating the redirect target (via a button in the status text).

The same is applicable for PROD too, as reported in #172.

PR #1033, opened 2020-07-11
xfd: undisable all options, use warning texts instead

### LINKED ISSUE #753: XfD: Only show valid options

siddharthvp:
In my opinion, the new system of having some of the invalid options disabled but not the others is just ugly and plain unprofessional. It isn't feasible to disable all invalid options, so I think the better solution is to allow anything to be selected, but use warning texts to tell the user in the interface that the option may not be appropriate.

At the same time, with the warning showed on selecting AfD on redirects takes care of a [recent WT:TW request](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#AfDing_non-Article_page) and alleviates the issue #379. 

Closes #753.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -56,7 +56,7 @@ Twinkle.xfd.callback = function twinklexfdCallback() {
 		type: 'select',
 		name: 'category',
 		label: 'Deletion discussion venue:',
-		tooltip: 'When activated, a default choice is made, based on what namespace you are in. This default should be the most appropriate; some inappropriate options may be disabled.',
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I dunno about unprofessional, but I think I generally like this sort of thing.

PR #1034, opened 2020-07-11
Bump font size to 12px

siddharthvp:
Twinkle has been using a font-size of 11.44px (11px on monobook) on all UI dialog texts all these years. I think this is way too small that it causes accessibility issues for at least some users and is generally hard to read. 

I propose bumping the morebits font-size to 12px. This change is subtle enough that it doesn't disturb the relative positioning of any UI elements, and would likely go unnoticed, but at the same time it does make things easier to read.

### DIFF
--- a/morebits.css
+++ b/morebits.css
@@ -40,6 +40,7 @@ form.quickform
 form.quickform *
 {
 	font-family: sans-serif;
+	font-size: 12px;
 }
 
 form.quickform fieldset
@@ -186,10 +187,6 @@ div.morebits-previewbox .mw-editsection
 	background-image: none;
 }
 
-body.skin-monobook .morebits-dialog {
-	font-size: 125%;
-}
-
 body .ui-dialog.morebits-dialog .ui-dialog-titlebar {
 	height: 1em;
 	background-color: #BCCADF !important;
@@ -221,7 +218,6 @@ body.skin-monobook .morebits-dialog .ui-dialog-titlebar {
 }
 
 .ui-dialog.morebits-dialog .morebits-dialog-content {
-	font-size: 88%;  /* this just seems to be traditional - I don't quite see the point, in this day and age of huge monitors, etc. */
 	padding: 0;
 }
 


morebits.css
```diff
@@ -40,6 +40,7 @@ form.quickform
 form.quickform *
 {
 	font-family: sans-serif;
+	font-size: 12px;
```
Amorymeltzer, line 43: Why this and not removing `font-size` from `.ui-dialog.morebits-dialog .morebits-dialog-content`? https://github.com/azatoth/twinkle/blob/07fffdf04bf8adf008862eb2ac2e5cc52c28b9d9/morebits.css#L224

Amorymeltzer:
Somewhat related, but I've been meaning on looking in on bumping some things up in size after looking at some of it for the jquery tooltip menus and earlier cleanup.  Warranted IMO.

morebits.css
siddharthvp: I was thinking of this more as a proposal so wasn't paying attention to the code. Let's see. Good old days of playing with the CSS...

morebits.css
```diff
@@ -186,10 +187,6 @@ div.morebits-previewbox .mw-editsection
 	background-image: none;
 }
 
-body.skin-monobook .morebits-dialog {
-	font-size: 125%;
-}
```
siddharthvp, line 191: This (at 0,2,1) was being over-ridden by the more specific `.ui-dialog.morebits-dialog .morebits-dialog-content` selector (0,3,0) so this wasn't doing anything at all.

siddharthvp:
`.morebits-dialog-content` is almost same as `.quickform` (for scripts that use quickform) so I don't know which one is better off having the css rule. 

Should we use percentages instead of font sizes? It looks like 100% is 13px, so for 12px we'll need around 92.3%.

morebits.css
Amorymeltzer: Well, it was doing *something*.  I get 11.44px with, 11px without.  More generally, shifts things around a bit.  Continued below...

Amorymeltzer:
So, as far as relevant items, we've got (top to bottom):

- `.morebits-dialog`: `small`
- <s>`body.skin-monobook .morebits-dialog`: `125%`</s>
- `body .ui-dialog.morebits-dialog .ui-dialog-titlebar`: `1em`
- <s>`.ui-dialog.morebits-dialog .morebits-dialog-content`: `88%`</s>
- `.morebits-dialog-footerlinks`: `90%`

I haven't been able to do much testing today, but `small` on `.quickform` sets everyone to 13px afaict, would that be reasonable?  I'm inclined toward %.

PR #1035, opened 2020-07-11
docs: Put gadget definition in Gadget.md, enable checking via sync.pl

Amorymeltzer:
This should make it easier to track changes to the gadget, as well as confirm when changes are needed.

### DIFF
--- a/sync.pl
+++ b/sync.pl
@@ -63,80 +63,103 @@ my $mw = MediaWiki::API->new({
 $mw->{ua}->agent('Twinkle/sync.pl ('.$mw->{ua}->agent.')');
 $mw->login({lgname => $conf{username}, lgpassword => $conf{password}});
 
+### Main function
+if (@files) {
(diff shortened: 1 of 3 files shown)


sync.pl
```diff
@@ -63,82 +63,105 @@
...
-  my $cmd = $repo->command(diff => '--stat', '--color');
-  my $s = $cmd->stdout;
-  while (<$s>) {
-    print;
+
+  my $localGadgetDef = read_text('Gadget.md');
+  if ($wikiGadgetDef eq $localGadgetDef) {
+    print "Gadget up-to-date\n";
+  } else {
+    print colored ['red'], "MediaWiki:Gadgets-definition needs udpating!\n";
```
Xi-Plus, line 160: udpating -> updating

Amorymeltzer: Fixed, thanks!

Issue #1029, opened 2020-07-10
Deprecate Morebits.wiki.* and create a new library


siddharthvp:
Morebits.wiki.* has descended into a spaghetti mess over the years. IMO, what we should do is to deprecate the whole thing and start afresh. I would propose the creation of a new library called `<insert cool name here>` (codenamed senora for the sake of talking about it below) with some much-needed enhancements.

Problems with morebits:
- The user agent string is a nasty global. This makes it unsuitable for use by multiple scripts simultaneously. The approach I used in my [DYK-helper](https://en.wikipedia.org/wiki/User:SD0001/DYK-helper.js) and other morebits-based scripts is to avoid over-writing the global user-agent until we're certain the user is about to make an edit. Hardly nice.
- No ability to add a default edit summary or default change tags. 
- Has used the old-fashioned callbacks all this while. #911 is an attempt to fix but it would be good to get rid of callback support as a whole.
- The Morebits.wiki.actionCompleted.* system is hacky and confusing and unreliable. 
- Error callbacks generally take the page object as argument, which is an anti-pattern. Generally, most libraries would use an error object or error code.
- Morebits.wiki.api:
  - a pathetic constructor
  - uses POST requests all the time even when GET is suitable
    - the function to make the requests is called `post()`! So it would be silly to rework it use GET. 
  - doesn't retry on badtoken errors

The solution:
- senora would be a class that needs to be instantiated to access all the functionality.
  - created like `var sen = new Senora('twinkle ([[w:en:TW]]');` (takes the user agent as constructor argument).
  - This would enable each script to create a senora instance with a user agent string, an edit summary advert, and change tags used in every write API call. Presently Twinkle source code is littered with a gazillion instance of ` + Twinkle.getPref('summaryAd')` which is unnecessary.
  - This also eliminates the need of having `Morebits.wiki.actionCompleted.*` stuff around as globals (shared between scripts). They would be class attributes instead. 
- senora's api class would be an extension of `mw.Api` integrated into the status system.
  - mw.Api is better at pre-processing parameters, eg. it scrubs out any `false` or `undefined` params.
  - mw.Api's postWithToken automatically handles badtoken retries.
  - mw.Api has an `abort` method to cancel all in-progress requests made using the object. This provides a way to solve #124. 
  - mw.Api's behaviour of getting a new token for every edit is a no-no for us that we'll probably want to customise. 
  - cleanly integrating with the status system would be a challenge.
[...]

PR #1038, opened 2020-07-12
morebits: api improvements

siddharthvp:
#1029 seems rather far-fetched for now, so it might still be worth making some improvements to Morebits.wiki.api.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1774,7 +1774,7 @@ Morebits.wiki.api.prototype = {
 
 		var ajaxparams = $.extend({}, {
 			context: this,
-			type: 'POST',
+			type: this.query.action === 'query' ? 'GET' : 'POST',
 			url: mw.util.wikiScript('api'),
 			data: queryString,
 			dataType: this.query.format,
--- a/morebits.js
+++ b/morebits.js
@@ -4204,7 +4204,7 @@ Morebits.status.error = function(text, status) {
  */
 Morebits.status.actionCompleted = function(text) {
 	var node = document.createElement('div');
-	node.appendChild(document.createElement('span')).appendChild(document.createTextNode(text));
+	node.appendChild(document.createElement('b')).appendChild(document.createTextNode(text));
 	node.className = 'morebits_status_info';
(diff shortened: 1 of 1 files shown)


morebits.js
```diff
@@ -1774,7 +1777,7 @@ Morebits.wiki.api.prototype = {
 
 		var ajaxparams = $.extend({}, {
 			context: this,
-			type: 'POST',
+			type: this.query.action === 'query' ? 'GET' : 'POST',
```
Amorymeltzer, line 1780: Nice

morebits.js
```diff
@@ -1721,6 +1721,8 @@ Morebits.wiki.api = function(currentAction, query, onSuccess, statusElement, onE
 	}
 	if (!query.format) {
 		this.query.format = 'xml';
+	} else if (query.format === 'json' && !query.formatversion) {
+		this.query.formatversion = '2';
```
Amorymeltzer, line 1725: Good point about this being a good time.  I'm already using it locally, so 👍

morebits.js
```diff
@@ -1738,6 +1740,7 @@ Morebits.wiki.api.prototype = {
 	statusText: null, // result received from the API, normally "success" or "error"
 	errorCode: null, // short text error code, if any, as documented in the MediaWiki API
 	errorText: null, // full error description, if any
+	tokenRetry: false,
```
Amorymeltzer, line 1743: Looks like you renamed this to `badtokenRetry` below (or vice-versa)

siddharthvp: Indeed. So what messed me up here was the later realisation that all actions don't use a CSRF token. If we want to be fussy about it, we could do something like [what I do in my bot framework](https://github.com/siddharthvp/mwn/blob/master/src/bot.js#L525) (querying the type of token needed as well) but IMO not really necessary. 

Easier to pretend that everything is CSRF-based (badtoken error on anything which isn't will be failing as before).

Amorymeltzer: Indeed, that's roughly where I ended up thinking.  It's few and far between atm, so nbd.

PR #1039, opened 2020-07-12
block: Add DS topic codes for aeblock/aepblock

Amorymeltzer:
The restructuring at the end of `change_action` has means `change_preset` is additionally being called whenever blocking is happening, not just if in the context of also applying a template.  This leads to `change_template`, which toggles the visibility of the DS select.

This has the side effect of changing how the form behaves when a user is already blocked.  Right now, the form doesn't display any of the old options by default (block and issue template).  *If* the sysop unchecks the issue template option, the block form will prefill with all the information from the original block (expiry, reason, etc.).  It's ephemeral, though, as if a preset is chosen, that information is then wiped clean, only returning if the issue template checkbox is toggled.

That's pretty misleading/confusing, especially since no one should be expected to memorize the information.  The change to `change_action` means that, when the template box is unchecked, the old information is not prefilled, matching the behavior elsewhere.  If we want to *always* prefill, that's a separate discussion.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -163,6 +163,11 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 
 	partial.prop('disabled', !blockBox && !templateBox);
 
+	// Can be in preset or template field, so the old one in the template
+	// field will linger. No need to keep the old value around, so just
+	// remove it; saves trouble when hiding/evaluating
+	$form.find('[name=dstopic]').parent().remove();
+
 	Twinkle.block.callback.saveFieldset($('[name=field_block_options]'));
 	Twinkle.block.callback.saveFieldset($('[name=field_template_options]'));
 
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Had initially thought to do it with select2, so the page titles were searchable and clickable, but the tradeoff with a bulkier item that stands-out didn't seem worth it?

siddharthvp:
How often would these have to be updated? Doesn't this add a serious maintenance overhead? Would it make sense to maintain the list on-wiki as a JSON page that we can ajax look up using a select2 widget? I wonder what @MusikAnimal would say about this.

modules/twinkleblock.js
```diff
@@ -335,6 +340,20 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 		}
 	}
 
+	// DS selection visible in either the template field set or preset,
+	// joint settings saved here
+	var dsSelectSettings = {
+		type: 'select',
+		name: 'dstopic',
+		display: 'none',
```
siddharthvp, line 348:
```suggestion
		style: 'display: none',
```
right?

Amorymeltzer:
Well, the short answer is it's entirely up to ArbCom, but by my count, there's been roughly a half-dozen in the past two years, so, no, it's not changed that much; we've had more changes to CSD in that same time frame.  I don't think it'd be too much overhead.

modules/twinkleblock.js
Amorymeltzer: Hmm, yes.  `block` has quite a few of these, stemming back to the beginning.

Amorymeltzer: Did away with 'em all

Amorymeltzer:
There's another behavior, noted in the commit message now:

>The restructuring at the end of `change_action` has means `change_preset` is additionally being called whenever blocking is happening, not just if in the context of also applying a template.  This leads to `change_template`, which toggles the visibility of the DS select.

>This has the side effect of changing how the form behaves when a user is already blocked.  Right now, the form doesn't display any of the old options by default (block and issue template).  *If* the sysop unchecks the issue template option, the block form will prefill with all the information from the original block (expiry, reason, etc.).  It's ephemeral, though, as if a preset is chosen, that information is then wiped clean, only returning if the issue template checkbox is toggled.

>That's pretty misleading/confusing, especially since no one should be expected to memorize the information.  The change to `change_action` means that, when the template box is unchecked, the old information is not prefilled, matching the behavior elsewhere.

I'll try and play around with *always* prefilling when I get a chance, but that's probably a separate discussion.

PR #1041, opened 2020-07-13
morebits: integrate checkboxShiftClickSupport into quickForm

siddharthvp:
Integrate checkboxShiftClickSupport into quickForm by providing a new `shiftClickSupport` option for checkbox elements.

Was developing an unrelated script, and what, I have to [write one whole extra line](https://en.wikipedia.org/w/index.php?title=User:SD0001/massMoveRevert.js&diff=prev&oldid=967476421) to get shift click support?? NO way, I am gonna tweak the library instead.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -445,6 +445,9 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 					}
 				}
 			}
+			if (data.shiftClickSupport && data.type === 'checkbox') {
+				Morebits.checkboxShiftClickSupport(Morebits.quickForm.getElements(node, data.name));
+			}
 			break;
 		case 'input':
 			node = document.createElement('div');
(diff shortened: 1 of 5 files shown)


Amorymeltzer:
This seems rather clear-cut and useful!

morebits.js
```diff
@@ -445,6 +445,9 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 					}
 				}
 			}
+			if (data.shiftClickSupport && data.type === 'checkbox') {
+				Morebits.checkboxShiftClickSupport(Morebits.quickForm.getElements(node, data.name));
```
Amorymeltzer, line 449: Subtle, but this implies the same logic from #908 that `current.name` implies it's a single checkbox.  Neat!

PR #1043, opened 2020-07-14
speedy: Remove T2

Amorymeltzer:
To be removed at WT:CSD: https://en.wikipedia.org/wiki/Wikipedia_talk:Criteria_for_speedy_deletion#RfC:_Removing_T2

Leaves `Twinkle.speedy.templateList` looking rather silly...

cc @Trialpears I left you a note at WT:CSD, which should help quite a bit with the implementation.  I can't commit to being available tonight or tomorrow so I'm not closing it, but I can help out.


Amorymeltzer:
Manually done on-wiki, although I've just removed it from speedy, not config or twinkle (which should be sufficient for a short time).

Removal of a config option has irreversible consequences of potential reseting of preferences.  It's unlikely, but *just in case* something gets reversed over the next day or two, we couldn't go back.  It has the added benefit of making my life easier...

PR #1044, opened 2020-07-15
Fix testing, move addInitCallback below callback definitions

Amorymeltzer:
The gadget files work fine, but to see the benefits from #946 for testing purposes, `Twinkle.addInitCallback` needs to be *below* any function expressions found within the function (`func`) that `addInitCallback` purports to add, such as the callback used by `Twinkle.addPortletLink`; e.g., for XfD, `Twinkle.addInitCallback(Twinkle.xfd, 'xfd')` must be below `Twinkle.xfd.callback`.  These are all function expressions, so are only evaluated when reached.

For visual consistency's sake, I've moved them all to the very end. cc @siddharthvp

### DIFF
--- a/modules/friendlyshared.js
+++ b/modules/friendlyshared.js
@@ -20,7 +20,6 @@ Twinkle.shared = function friendlyshared() {
 		}, 'Shared IP', 'friendly-shared', 'Shared IP tagging');
 	}
 };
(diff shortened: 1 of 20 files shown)


siddharthvp:
Nice find!

PR #1052, opened 2020-07-17
xfd/prod: show Find sources links in AfD and PROD windows

### LINKED ISSUE #1047: Add {{find sources}} links for AfD and PROD
This should help with performing a WP:BEFORE. Inspired by [this discussion](https://en.wikipedia.org/wiki/Wikipedia_talk:Proposed_deletion#RfC_on_improving_PROD_instructions_in_%22Before_nomination%22_section). 

Regarding implementation, should we just API:Parse the {{find sources}} template? This will be easier but the links would take time to appear. A more elegant implementation would be translate the template to javascript. A bit more code but the IMO this would be proper way of doing it.

siddharthvp:
The code doesn't look particularly nice, but this should be useful in practice.

Closes #1047.

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -64,6 +64,12 @@ Twinkle.prod.callback = function twinkleprodCallback() {
 	});
 
 	field.append({
+		type: 'div',
+		label: Twinkle.makeFindSourcesDiv(),
+		style: 'margin-bottom: 5px;'
+	});
+
+	field.append({
 		type: 'radio',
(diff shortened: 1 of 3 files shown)


Amorymeltzer (approved):
Rebased.  LGTM, although, yes, agreed on the code!  I do think it's worth [discussing preview rendering](https://github.com/azatoth/twinkle/issues/1047#issuecomment-671911400) but that need not stop this in any meantime! 👏

PR #1053, opened 2020-07-17
warn: add uw-unattribcc as single-issue notice

mdaniels5757:
Per [Special:Diff/968171944](https://en.wikipedia.org/wiki/Special:Diff/967928507/968171944).


Amorymeltzer (approved):
👍

Amorymeltzer:
Merged as e2bc3e8 for a minor grammar tweak.

PR #1063, opened 2020-07-23
morebits: getInputData: Fix dumb IE bug

Amorymeltzer:
Apparently IE iterates over names of HTMLform stuff when doing `for (var i in form.elements)`?  We could also use `$.each`.

----

Reported at WT:TW around welcoming, this was super annoying to run down since there weren't any errors logged.

cc @siddharthvp just FYI this is dumb

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -735,7 +735,7 @@ Morebits.quickForm.element.generateTooltip = function QuickFormElementGenerateTo
 Morebits.quickForm.getInputData = function(form) {
 	var result = {};
 
-	for (var i in form.elements) { // eslint-disable-line guard-for-in
+	for (var i = 0; i < form.elements.length; i++) {
 		var field = form.elements[i];
 		if (field.disabled || !field.name || !field.type ||
 			field.type === 'submit' || field.type === 'button') {


Amorymeltzer:
As noted in #1066, this applies to more than just a weird IE behavior, it stems from the [discussion in #908](https://github.com/azatoth/twinkle/pull/908/commits/dccdacaee2536c74e6cfc94abf6b33a73075aeb3#r443258638) about `ownProperties`.  I don't see it in my Chrome (and clearly didn't when we were dealing with #908, with prior versions) so I wonder if there's a browser+OS issue going on.  Using tag, Chrome (for me) just lists each item, whereas Safari then goes back through and lists the nodes?  I think those are "inherited enumerables" rather than "enumerables" but I'm not sure why Chrome isn't listing them.  Probably related to the fact that this is an HTMLCollection.

At any rate, I'm not 100% that `for (var i = 0; i < form.elements.length; i++)` is definitely what we want. 
 We've got a few alternatives, in particular a modified version of [@siddharthvp's suggestion](https://github.com/azatoth/twinkle/pull/908/commits/dccdacaee2536c74e6cfc94abf6b33a73075aeb3#r446500712)

```js
if (!Object.prototype.hasOwnProperty.call(Object.keys(form.elements), i)) {
	continue;
}
```

The shorter-but-hackier `if (isNaN(i)) { continue; }` also appears to work.

Amorymeltzer:
For those interested in testing, I've pushed this change to enWiki (given that the tag updates have been reverted) but **not** testwiki.

siddharthvp:
If there's too much trouble with form.elements, we can just use jQuery,
$(form).find('input, select, textarea').each(...)  which is guaranteed to
behave the same way across browsers since it's jQuery.

Amorymeltzer:
@siddharthvp That's a good point!  I should clarify that I don't *know* it's an issue, just that I'm not confident it isn't.

PR #1067, opened 2020-07-25
protect: Fix invalid expiry when unprotecting

Amorymeltzer:
Another fallout from #908/6c38a5d: `getInputData` ignores disabled elements, so the expiry for each unprotection level is `undefined` rather than whatever was there before.  It doesn't matter what the expiry is, but it *needs* to be valid or else the API complains (the create error is... unique).


Amorymeltzer:
Closed in favor of #1068, which handles it on the Morebits side.

PR #1068, opened 2020-07-26
morebits: Default protection expirys to infinity

Amorymeltzer:
Various protection expirys only take valid time parameters, and even if not needed (removal of protection), they need to be valid.  The API defaults to `infinity` if missing, but they can't be `undefined`.  Better way of solving #1067; see also #1057.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2501,16 +2501,21 @@ Morebits.wiki.page = function(pageName, currentAction) {
 	};
 
 	// Protect-related setter functions
+	/**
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
As noted, this is a better way of fixing unprotections in twinkleprotect.  `getInputData` (#908/ 6c38a5d) ignores disabled elements, so the expiry for each unprotection level was getting passed as `undefined` rather than whatever was there before.  It doesn't matter what the expiry is, but it *needs* to be valid or else the API complains.

PR #1071, opened 2020-07-26
protect: Ensure a valid type of move protection

Amorymeltzer:
Since move tries to match edit if present, but autoconfirmed is not an option (see discussion in #451), it's possible for someone to attempt a submission without any level (sysop, template, etc.) selected for move protection if they select move protection while the default edit level of "autoconfirmed" is selected, and then uncheck edit protection.  The query would just fail, but better not to reset the form.

Also moved the empty default selection for filing a request to the top, as might be expected.

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -623,9 +623,9 @@ Twinkle.protect.callback.changeAction = function twinkleprotectCallbackChangeAct
 				name: 'expiry',
 				label: 'Duration: ',
 				list: [
+					{ label: '', selected: true, value: '' },
 					{ label: 'Temporary', value: 'temporary' },
-					{ label: 'Indefinite', value: 'infinity' },
-					{ label: '', selected: true, value: '' }
+					{ label: 'Indefinite', value: 'infinity' }
 				]
 			});
 			field1.append({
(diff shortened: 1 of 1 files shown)


modules/twinkleprotect.js
```diff
@@ -1160,7 +1160,13 @@ Twinkle.protect.callback.evaluate = function twinkleprotectCallbackEvaluate(e) {
 						thispage.setEditProtection(input.editlevel, input.editexpiry);
 					}
 					if (input.movemodify) {
-						thispage.setMoveProtection(input.movelevel, input.moveexpiry);
+						// Ensure a level has actually been chosen
+						if (input.movelevel) {
```
Xi-Plus, line 1164: I think it should be
```javascript
if (input.movelevel !== 'all') {
```
Because of
https://github.com/azatoth/twinkle/blob/3c65fa5bf409143b99f0af3f4bddbb32a65b45ab/modules/twinkleprotect.js#L388-L392

Amorymeltzer: Sorry @Xi-Plus; I've (clearly) already proven to be beyond tired today, so I must be missing what you're saying, but what's the issue with `all`?  It's a valid level for unprotection, becoming `move=all`

Xi-Plus: I wanna know how to trigger the alert (You must chose a move protection level).

Amorymeltzer: 1. Start with move protection disabled and edit protection selected at the autoconfirmed level
2. Select move protection.  The expiry would be set, but no value for move protection would be selected.
3. Uncheck edit protection.

The form would let you submit and then promptly fail.

Xi-Plus: Thanks a lot!

PR #1072, opened 2020-07-27
fluff: Fix username for IP ranges on Special:Contributions

Amorymeltzer:
First reported at [WT:TW](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=969770397#%22Unknown_user%22_and_rollback_errors), but with the restructuring and unification of rollback link creation done in #947 and the handling of revdel'd names in #993, the username for IP ranges becomes important.  This was blindly passing `null` (from the lack of `wgRelevantUserName`), so all IPs were treated as revdel'd when viewing a range's contributions.


Amorymeltzer:
Pushed live.

PR #1076, opened 2020-07-31
protect: Allow presets to override protection length

Amorymeltzer:
We naively set the protection length (aka expiry) based on the type of protection, but there's no good reason not allow the presets to set their own.  In particular, items that should be indefinite really shouldn't default to 2 days or whatever.  To begin, I've set the following defaults:

- `pp-usertalk`: infinity
- `pp-template`: infinity
- `pp-30-500-arb`: infinity
- `pp-semi-usertalk`: infinity
- `pp-move-indef`: infinity
- <s>`pp-30-500-blp`: 1 week</s>
- <s>`pp-semi-blp`: 1 week</s>
- <s>`pp-pc-blp`: 2 months</s>

What others should be covered?  I'll ask on WT:TW/WP:AN.

----

Will conflict with #1074, whose commits are included here for completion's/diffs sake.

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -711,6 +711,7 @@ Twinkle.protect.protectionWeight = {
 
 // NOTICE: keep this synched with [[MediaWiki:Protect-dropdown]]
 // Also note: stabilize = Pending Changes level
+// expiry will override any defaults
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Sticking with just the indefs per some feedback at WP:AN, but the ability for anything remains.

Issue #910, opened 2020-03-31
morebits: get rid of modifications to native prototypes


siddharthvp:
Modifications to prototypes of native classes is a bad idea in modern-day javascript.

- [x] `Date.prototype.getMonthName` and `.getUTCMonthName` (removed in #814) 
- [ ] `HTMLFormElement.prototype.getChecked` and `.getUnchecked` (most uses removed in #908) 
- [ ] `RegExp.escape` - probably replace with `Morebits.escapeRegExp`

Amorymeltzer:
After #908:

- [ ] `getChecked`:
    - [ ] two in `arv`
    - [ ] one in `speedy`
    - [ ] one in `tag`
- [ ] `getUnchecked`: two in `tag`

PR #1077, opened 2020-07-31
morebits: Deprecate RegExp.escape in favor of Morebits.string.escapeRegExp

Amorymeltzer:
Part of #910, cc @siddharthvp

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -520,7 +520,7 @@ Twinkle.arv.callback.evaluate = function(e) {
 				var $aivLink = '<a target="_blank" href="/wiki/WP:AIV">WP:AIV</a>';
 
 				// check if user has already been reported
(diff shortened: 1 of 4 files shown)


siddharthvp:
Since this is anyway a breaking change, I was thinking it would be worth improving things by enforcing `space_fix` in Morebits.string.escapeRegExp (making it the default and removing the option to turn it off). This does away with the awkward use of `true` as second param. In places where space_fix is not desired, mw.util.escapeRegExp can be used instead, which is anyway shorter.

Amorymeltzer:
That's a good point and one that I only briefly considered, if I'm being honest.  I think I was hoping to make anyone's life easier, but I do think it'd be worthwhile.  This of course predates the mw tool, so it's about damn time.

Amorymeltzer:
{{done}}.  Drew a blank on a different name, still think it'd be clearer.

mdaniels5757:
> {{done}}.

You know templates don't work here, right? :)

Best, ~~~~

Issue #1050, opened 2020-07-17
morebits: add live preview 


siddharthvp:
It would be good to have the infrastructure for live previews in morebits. I'd imagine this would be useful primarily for previewing comments (edit/log summaries) where we usually offer only a text field rather than a textarea, in which it's difficult to correctly write something long with links etc. Mentioned by @roysmith on  [WT:TW](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_43#U/I_problems_for_CSD_with_multiple_reasons.).

Relevant modules which use summaries that ought to be previewed: speedy, unlink, batchdelete, batchundelete, batchprotect. 

It doesn't look as if we can cleanly implement this as an additional option in `Morebits.wiki.preview`. So what I would do is to create a new `Morebits.livePreview` (or `Morebits.wiki.livePreview` if you're a pedant).

Amorymeltzer:
Sorry, I'm not entirely following; I think you're suggesting a live-updating preview of edit (or log) summary fields.  Is that right?

roysmith:
Yes. the coolest thing would be a live-updating preview.  But, even just having a "preview" button that you had to click to get a preview would a big improvement, and from my point of view, good enough.

siddharthvp:
> Sorry, I'm not entirely following; I think you're suggesting a live-updating preview of edit (or log) summary fields. Is that right?

Yes. I think live-updating previews are better for summary fields to avoid the UI clutter of a preview button. Do you think it would be too jarring? Notably, I think the new Reply tool uses live preview.

PR #1082, opened 2020-08-02
batchprotect/batchdelete: Add dropdowns with common protection/CSD reasons

Amorymeltzer:
Only short lists of generic, common-ish reasons.  Ideally users should continue to provide a custom reason, in which case the dropdown will provide a prefix to that reason.

----

Somewhat on the fence as to whether this is a good idea or not.  I lean yes, if only because it might reduce the chance of mis-typed links, and will hopefully lead to more thorough log messages overall, but I am wary of normalizing this sort of thing.

Open to any items that are missing, but for example I left off the A-criteria, since they're unlikely.

### DIFF
--- a/modules/twinklebatchdelete.js
+++ b/modules/twinklebatchdelete.js
@@ -104,10 +104,40 @@ Twinkle.batchdelete.callback = function twinklebatchdeleteCallback() {
 		]
 	});
 	form.append({
+		type: 'select',
+		name: 'speedy',
+		label: 'Common CSD reasons: ',
+		list: [
+			{ label: '', value: '' },
+			{ label: 'G1: Patent nonsense, meaningless, or incomprehensible', value: '[[WP:CSD#G1|G1]]: [[WP:PN|Patent nonsense]], meaningless, or incomprehensible' },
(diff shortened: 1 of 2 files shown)


siddharthvp:
Dunno about batchprotect but for batchdelete, I don't think this is a good idea at all, as this adds a number of selectables that almost will never be used (or at least till now have never been used). 

>  I lean yes, if only because it might reduce the chance of mis-typed links, and will hopefully lead to more thorough log messages overall

I think #1050 is a much better solution to this problem.

Amorymeltzer:
>or at least till now have never been used

I think you mean "never been requested" or "desired"?  While I don't disagree, I can think of a *very good reason* why something that didn't exist hasn't been used... 😆

siddharthvp:
Poor wording, but what I meant was the default CSD reasons have almost never been used for twinkle batch deletions (from some imperfect database scans that i ran, yes I'm supposed to be busy atm but somehow i still have time for this). 

Anyway if this is being done, I'd suggest something like a datalist HTML field or a nicer select2-based alternative if there is one (which basically combine the select element+ input element into one). The pattern of having one field to select some text which can't be edited and another field to supplement that text looks pretty outdated.

PR #1084, opened 2020-08-04
warn: Append warning to an existing section, even if it's not the last

Amorymeltzer:
There's been some discussion on this front recently (see [here](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_42#Warn_template_bug?_creating_unneeded_new_section_headers) and [here](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_43#Multiple_sections_for_one_month)) and I think it's reasonable to do.  The *correct* way would be to set up an additional query to `action=parse`, then go through `prop=sections`, but that seems overkill if we can just check `/\n(==+).+\1/g`.  This changes the single `pageobj.save` to either `pageobj.append` or `pageobj.newSection` as appropriate.

Note that I'm not proposing the same for the block, as I think it's less obviously useful to place a block notice anywhere but the bottom.  I did, however, restructure the block evaluation somewhat, so that we don't bother with the page text if the user is just gonna blank it.

Tested but not extensively with weird talk pages; relies upon `newSection` from #1083.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -1892,32 +1892,36 @@ Twinkle.block.callback.getBlockNoticeWikitext = function(params) {
 };
 
 Twinkle.block.callback.main = function twinkleblockcallbackMain(pageobj) {
-	var text = pageobj.getPageText(),
-		params = pageobj.getCallbackParameters(),
+	var params = pageobj.getCallbackParameters(),
+		date = new Morebits.date(pageobj.getLoadTime()),
 		messageData = params.messageData,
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
[Fixed](https://github.com/azatoth/twinkle/compare/2a7b58c1e2f5fb86a830df0220d012a65b05581c..434a068de780c1cd478d126696d578075e504337) regex and traverse sections bottom up to prefer the bottom-most section made if there are multiple.

PR #1088, opened 2020-08-05
morebits: Use common checks for wiki.page methods

Amorymeltzer:
Similar in concept to `fnCheckPageName`; just useful for move, undelete, delete, protect, stabilize

Not entirely convinced `fnProcessChecks` is worth the complexity.  It's tucked away and greatly reduces repetition, but that `if` is ugly.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2745,11 +2745,10 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		ctx.onMoveSuccess = onSuccess;
 		ctx.onMoveFailure = onFailure || emptyFunction;
 
-		if (!ctx.editSummary) {
-			ctx.statusElement.error('Internal error: move reason not set before move (use setEditSummary function)!');
-			ctx.onMoveFailure(this);
-			return;
+		if (!fnPreflightChecks.call(this, 'move', ctx.onMoveFailure)) {
+			return; // abort
 		}
+
 		if (!ctx.moveDestination) {
 			ctx.statusElement.error('Internal error: destination page name was not set before move!');
 			ctx.onMoveFailure(this);
@@ -2853,16 +2852,8 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		ctx.onDeleteSuccess = onSuccess;
 		ctx.onDeleteFailure = onFailure || emptyFunction;
 
-		// if a non-admin tries to do this, don't bother
-		if (!Morebits.userIsSysop) {
-			ctx.statusElement.error('Cannot delete page: only admins can do that');
-			ctx.onDeleteFailure(this);
-			return;
-		}
-		if (!ctx.editSummary) {
-			ctx.statusElement.error('Internal error: delete reason not set before delete (use setEditSummary function)!');
-			ctx.onDeleteFailure(this);
-			return;
+		if (!fnPreflightChecks.call(this, 'delete', ctx.onDeleteFailure)) {
+			return; // abort
 		}
 
 		if (fnCanUseMwUserToken('delete')) {
@@ -2885,16 +2876,8 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		ctx.onUndeleteSuccess = onSuccess;
 		ctx.onUndeleteFailure = onFailure || emptyFunction;
 
-		// if a non-admin tries to do this, don't bother
-		if (!Morebits.userIsSysop) {
-			ctx.statusElement.error('Cannot undelete page: only admins can do that');
-			ctx.onUndeleteFailure(this);
-			return;
-		}
-		if (!ctx.editSummary) {
-			ctx.statusElement.error('Internal error: undelete reason not set before undelete (use setEditSummary function)!');
-			ctx.onUndeleteFailure(this);
-			return;
+		if (!fnPreflightChecks.call(this, 'undelete', ctx.onUndeleteFailure)) {
+			return; // abort
 		}
 
 		if (fnCanUseMwUserToken('undelete')) {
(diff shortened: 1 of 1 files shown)


siddharthvp:
I'm opposed to the separate function for pre-flight checks: #911 (I'll revisit that to resolve the conflicts some time) reduces the code needed to handle these checks to one line ([example for deletePage](https://github.com/azatoth/twinkle/pull/911/commits/5a86c126bb297a864456afe4509ed3fd665a8097)). So we aren't saving much of code, and this will only create even more conflicts. Also the check for sysop right is something that I don't think is even needed since it's obviously checked by the API itself.

Amorymeltzer:
>#911 (I'll revisit that to resolve the conflicts some time)

[Done](https://github.com/azatoth/twinkle/compare/80d18907245f370a2816832878e17bfaf27c17e9..cb9c0ffc42b8ceecaf2bb214fb50f08d20094840)! (I handled conflicts a while ago, but haven't had a ton of time to actually look at/understand it)

> reduces the code needed to handle these checks to one line ([example for deletePage](https://github.com/azatoth/twinkle/pull/911/commits/5a86c126bb297a864456afe4509ed3fd665a8097)). So we aren't saving much of code, and this will only create even more conflicts.

I don't mind the conflicts, but regardless I'd deal with #911 first.  Admittedly, that does reduce the length quite a bit, but it's still repetitive, especially the `if`s.  I was converting wiki.page to json, and it just jumped out at me that I had to do the same thing a half-dozen times.

>Also the check for sysop right is something that I don't think is even needed since it's obviously checked by the API itself.

I presume the idea is that it's an easy check to preempt the need to actually go through the motions; no need to submit to the API, etc.  I don't think that's such a bad idea, as long as we're not too excessive about it.

morebits.js
```diff
@@ -2702,11 +2702,10 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		ctx.onMoveSuccess = onSuccess;
 		ctx.onMoveFailure = onFailure || emptyFunction;
 
-		if (!ctx.editSummary) {
-			ctx.statusElement.error('Internal error: move reason not set before move (use setEditSummary function)!');
-			ctx.onMoveFailure(this);
-			return;
+		if (!fnPreflightChecks('move', ctx.onMoveFailure)) {
```
siddharthvp, line 2705:
```suggestion
		if (!fnPreflightChecks.call(this, 'move', ctx.onMoveFailure)) {
```
because fnPreflightChecks uses `this`.

Following 911, this would need to be either:
```
var error = fnPreflightChecks.call(this, 'move', ctx.onMoveFailure);
if (error) {
    return error;
}
```
with fnPreflightChecks returning a rejected promise (in case of error) or null (in case of no error)

Or alternatively 
```
var error = fnPreflightChecks.call(this, 'move');
if (error) {
    return returnError.call(this, error.code, error.text);
}
```
with fnPreflightChecks returning an object containing the error code and text.

morebits.js
```diff
@@ -3298,6 +3267,63 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+			return false;
+		}
+		return true; // all OK
+	};
+
+	// Common checks for fnProcess functions (fnProcessDelete, fnProcessMove, etc.)
+	// Used for move, undelete, delete, protect, stabilize
+	var fnProcessChecks = function(action, onFailure, xml) {
+		var missing = $(xml).find('page').attr('missing') === '';
+		if ((action === 'undelete' && !missing) ||
```
siddharthvp, line 3292: This removes the ability to undelete revisions of existing pages. Desirable?

morebits.js
```diff
@@ -3298,6 +3267,63 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+			(action === 'protect' && (
+				(!missing && ctx.protectCreate) ||
+					(missing && (ctx.protectEdit || ctx.protectMove)))) ||
+			missing) {
+			ctx.statusElement.error('Cannot ' + action + ' the page because it ' + (missing ? 'no longer' : 'already') + ' exists');
+			onFailure(this);
+			return false;
+		}
```
siddharthvp, line 3300: This can be structured better by using named booleans. 
```js
var deletingNonexistingPage = action === 'delete' && missing;
var saltingExistingPage = ...
var protectingNonexisitingPage = ...
if (deletingNonexistingPage || saltingExistingPage || protectingNonexisitingPage ) {
    ....
}
```
```

File: morebits.js [hunk 4]
```diff
@@ -3298,6 +3267,63 @@ Morebits.wiki.page = function(pageName, currentAction) {
 
 	};
 
+	// Common checks for action methods
+	// Used for move, undelete, delete, protect, stabilize
+	var fnPreflightChecks = function(action, onFailure) {
+		// if a non-admin tries to do this, don't bother
+		if (!Morebits.userIsSysop && action !== 'move') {
+			ctx.statusElement.error('Cannot ' + action + 'page : only admins can do that');
+			onFailure(this);
```
siddharthvp, line 3276: pretty sure `this` here would be `window`. Can be fixed by changing the caller per prev cmt.

morebits.js
```diff
@@ -3298,6 +3267,63 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
+		if (action === 'undelete') {
+			editprot = $(xml).find('pr[type="create"]');
+		} else if (action === 'delete' || action === 'move') {
+			editprot = $(xml).find('pr[type="edit"]');
+		}
+		if (editprot.length > 0 && editprot.attr('level') === 'sysop' && !ctx.suppressProtectWarning &&
+			!confirm('You are about to ' + action + ' the fully protected page "' + ctx.pageName +
+			(editprot.attr('expiry') === 'infinity' ? '" (protected indefinitely)' : '" (protection expiring ' + new Morebits.date(editprot.attr('expiry')).calendar('utc') + ' (UTC))') +
+			'.  \n\nClick OK to proceed with ' + action + ', or Cancel to skip.')) {
+			ctx.statusElement.error('Aborted ' + action + ' on fully protected page.');
```
siddharthvp, line 3314: Bad wording on stabilize: `action` here is `apply FlaggedRevs to` which doesn't fit.

morebits.js
Amorymeltzer: Probably not — we had a whole colloquy about that back when I added `undelete`

morebits.js
Amorymeltzer: Constant struggle with FlaggedRevs, Pending Changes, and Stabilize all being awful, inscrutable, or not verbs!

Amorymeltzer:
>I think the real reason I am annoyed is the insulting comment that comes before it!

I *think* you might be teasing @siddharthvp, but I just want to make sure — did I insult you somewhere?  I really hope I didn't, it certainly was not my intention.  Nothing but appreciation on my end!

morebits.js
Amorymeltzer: Jah.  I'm about to be AFK for 2 weeks (and as stated would come after #911) but wanted to get it up for "yea or nay"

siddharthvp:
> 
> 
> > I think the real reason I am annoyed is the insulting comment that comes before it!
> 
> I _think_ you might be teasing @siddharthvp, but I just want to make sure — did I insult you somewhere? I really hope I didn't, it certainly was not my intention. Nothing but appreciation on my end!

Oh lol no! I was talking about the comment in the code `// if a non-admin tries to do this, don't bother` which doesn't exactly sound nice.

PR #1090, opened 2020-08-07
Support for watchlist expiry

Amorymeltzer:
This is still pretty rough — I whipped it up some time last month — but wanted to put this up as a starting point, in case there's any discussion.  Watchlist expiries in core can be tested at testWiki, but aren't enabled on enWiki (I'm not sure there's been any discussion about that process).

Basically, this:

- Adds a few default options to `watchlistEnums` in twinkleconfig
- Tweaks a few config options, even adding a few as necessary
- Overhauls how morebits handles watchlists, namely by expanding `setWatchlist` and doing away with `setWatchlistFromPreferences`.

The major sticking points are:
1. Does it work?
42. Is any of this a good idea?
42. Handling bespoke watchlist API actions (AFAICT `action=watch` has no way to default to prefs, so can either easily be *mostly* correct or rather convlutedly be correct (see `twinklexfd`)).

@MusikAnimal This is your baby as far as I see, so I'd love your thoughts/feedback on any/all of this at your leisure.  I only played with it minimally after it first went live, so I may have missed some specifics in the implementation...

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -870,17 +870,7 @@ Twinkle.arv.processSock = function(params) {
 	spiPage.setEditSummary('Adding new report for [[Special:Contributions/' + params.uid + '|' + params.uid + ']].');
 	spiPage.setChangeTags(Twinkle.changeTags);
 	spiPage.setAppendText(text);
-	switch (Twinkle.getPref('spiWatchReport')) {
-		case 'yes':
-			spiPage.setWatchlist(true);
-			break;
-		case 'no':
-			spiPage.setWatchlistFromPreferences(false);
-			break;
-		default:
-			spiPage.setWatchlistFromPreferences(true);
-			break;
-	}
+	spiPage.setWatchlist(Twinkle.getPref('spiWatchReport'));
 	spiPage.append();
 
 	Morebits.wiki.removeCheckpoint();  // all page updates have been started
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -21,7 +21,15 @@
 
 Twinkle.config = {};
 
-Twinkle.config.watchlistEnums = { yes: 'Add to watchlist', no: "Don't add to watchlist", 'default': 'Follow your site preferences' };
+Twinkle.config.watchlistEnums = {
+	yes: 'Add to watchlist (indefinitely)',
+	no: "Don't add to watchlist",
+	'default': 'Follow your site preferences',
+	'1 week': 'Watch for 1 week',
+	'1 month': 'Watch for 1 month',
+	'3 months': 'Watch for 3 months',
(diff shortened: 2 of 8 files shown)


MusikAnimal:
I think it's a great idea! Gadgets like Twinkle is what we suspected would be a major consumer. I certainly have no need to watch someone's user talk indefinitely after issuing them a single warning, same with an XfD I created, but I would for a temporary period if I could. I think people will appreciate this functionality in Twinkle.

> AFAICT action=watch has no way to default to prefs

The preferences are tied to specific actions such as delete/rollback/etc., so for action=watch there's no preference to fallback to.

I have not done much for code review, but as long as the "watchlist" and "watchlistexpiry" parameters are set it the APIs should accept it. Note also you can supply any arbitrary value accepted by [PHP's `strtotime`](https://www.php.net/manual/en/function.strtotime.php), just like you can for block durations. E.g. "1 day", "next Thursday", or an exact timestamp. It may be worth it to provide a way to enter custom expiries, but I assume the pre-supplied options are enough for most people. The only caveat is [`$wgWatchlistExpiryMaxDuration`](https://www.mediawiki.org/wiki/Manual:$wgWatchlistExpiryMaxDuration), currently set to 6 months. This limit may be removed in the future.

Thanks for doing this work!

Amorymeltzer:
>The preferences are tied to specific actions such as delete/rollback/etc., so for action=watch there's no preference to fallback to.

It'd be nice I suppose to supply an action to the watch API and have it use the default pref for that action.  Pie in the sky!

@MusikAnimal Do you know if there's a timeline or plan for rollout (to enwiki or elsewhere)?  Just wondering if/when I'll need to start debugging, etc.

MusikAnimal:
> Do you know if there's a timeline or plan for rollout (to enwiki or elsewhere)?

I don't know that we have an exact timeline yet, but I expect it wouldn't get to any group 2 wikis for another month or so. There'll be a post at WP:VPT a week or two in advance.

Amorymeltzer:
Rebased, etc.  Moved 3 commits into 2, so that each is logically and functionally sound.  Only major change from above is to support `unwatch`, although we don't use it.

Amorymeltzer:
The [latest release schedule](https://meta.wikimedia.org/wiki/Community_Tech/Watchlist_Expiry/Release_Schedule) has this slated to head out everywhere December 1.  I wonder if there's any value in pushing out most of this beforehand, hopefully flushing out any bugs in the background reworking.  Refamiliarizing myself with it, I *think* what'd be needed is mainly just removing the expiry options from `Twinkle.config.watchlistEnums`.  Thoughts?  I'll try and test it out beforehand, but I think it'll make things simpler.

Amorymeltzer:
Deployed!  I think having the expiry options available and used would just lead to a swallowed API warning, but I've disabled them for the nonce.

PR #1030, opened 2020-07-10
xfd: The Great Cleanup

siddharthvp:
Cleanup of truckloads of messy code in xfd module. As discussed in #908.

PR #1091, opened 2020-08-07
Add intestactions to loadQuery, auto-add XfD edit request 

Amorymeltzer:
There are two separate issues here, but the main question is whether adding `intestactions` to the page load query in `Morebits.wiki.page` is a good idea.  I think it adds value being able to confirm early whether an action can take place, and should avoid running into a hard error if handled elegantly.  While this is a simple proof of concept, there's a lot we can think about making use of it.  Right now, I've just added `edit`, but an expansive view could imagine checks for `move` or `delete`, whether in morebits itself or available for modules to use.  Even without additional checks, it'd be a helpful pre-edit early failure check.  I've not tested or thought through the implications for non-load things like `append`.

I've used it here to enable twinklexfd to auto-add edit protected requests when a user can't edit the page in question.  With the exception of TfM other template tagging, `wgIsProbablyEditable` would have sufficed for this implementation. (technically both are making assumptions about "protection" but in practice, that's it).  I'm not certain that it's easier than using promises (cc @siddharthvp) but it's probably more efficient.

At any rate, I'd welcome thoughts on either side of things.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -714,6 +714,24 @@ Twinkle.xfd.setWatchPref = function twinklexfdsetWatchPref(pageobj, pref) {
 };
 
 Twinkle.xfd.callbacks = {
+	// Requires having the tag text (params.tagText) set ahead of time
+	autoEditRequest: function(pageobj, params) {
+		// Update the form
+		pageobj.getStatusElement().warn('Page protected, requesting edit');
+
+		var editRequest = '{{subst:Xfd edit protected|page=' + pageobj.getPageName() +
+			'|discussion=' + params.discussionpage + '|tag=<nowiki>' + params.tagText + '\u003C/nowiki>}}'; // U+003C: <
+
+		var talk_page = new Morebits.wiki.page(new mw.Title(pageobj.getPageName()).getTalkPage().toText(), 'Automatically posting edit request on talk page');
+		talk_page.setNewSectionTitle('Edit request to complete ' + utils.toTLACase(params.venue) + ' nomination');
+		talk_page.setNewSectionText(editRequest);
+		talk_page.setCreateOption('recreate');
+		Twinkle.xfd.setWatchPref(talk_page, Twinkle.getPref('xfdWatchPage'));
+		talk_page.setFollowRedirect(true);  // should never be needed, but if the article is moved, we would want to follow the redirect
+		talk_page.setChangeTags(Twinkle.changeTags);
+		talk_page.setCallbackParameters(params);
+		talk_page.newSection();
+	},
 	getDiscussionWikitext: function(venue, params) {
 		if (venue === 'cfds') { // CfD/S takes a completely different style
 			return '* [[:' + Morebits.pageNameNorm + ']] to [[:' + params.cfdstarget + ']]\u00A0\u2013 ' +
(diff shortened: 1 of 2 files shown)


modules/twinklexfd.js
```diff
@@ -606,6 +606,29 @@ Twinkle.xfd.setWatchPref = function twinklexfdsetWatchPref(pageobj, pref) {
...
+		editRequest += 'to the top of the page to complete the nomination. Thank you. ~~~~';
+
+		var talk_page = new Morebits.wiki.page(talkTitle, 'Automatically posting edit request on talk page');
+		talk_page.setAppendText(editRequest);
+		talk_page.setEditSummary('Adding edit request to complete ' + venue + ' nomination' + Twinkle.getPref('summaryAd'));
+		talk_page.setCreateOption('recreate');
+		Twinkle.xfd.setWatchPref(talk_page, Twinkle.getPref('xfdWatchPage'));
+		talk_page.setFollowRedirect(true);  // should never be needed, but if the article is moved, we would want to follow the redirect
+		talk_page.setCallbackParameters(params);
+		talk_page.append();
```
Amorymeltzer, line 630: Would be better to do `newSection` after #1083

modules/twinklexfd.js
```diff
@@ -606,6 +606,29 @@ Twinkle.xfd.setWatchPref = function twinklexfdsetWatchPref(pageobj, pref) {
 };
 
 Twinkle.xfd.callbacks = {
+	// Requires having the tag text (params.tagText) set ahead of time
+	autoEditRequest: function(pageobj, params) {
+		// Update the form
+		pageobj.getStatusElement().warn('Page protected, requesting edit');
+
+		var venue = params.venue.toUpperCase().replace(/\BF/, 'f');
```
Amorymeltzer, line 614: Could fold `toTLACase` into `utils` after #1030

morebits.js
```diff
@@ -2056,6 +2056,7 @@ Morebits.wiki.page = function(pageName, currentAction) {
 		watchlistOption: 'nochange',
 		creator: null,
 		timestamp: null,
+		canEdit: null,  // true or false after load
```
Amorymeltzer, line 2059: It might be better for this to be a `testactions` array or something, then just assign all found actions to that array.  Then, separate getters can be used (`return ctx.testactions.indexOf('edit') !== -1`) or a bulk test (`function(action) { return ctx.testactions.indexOf(action) !== -1 }`)

Amorymeltzer: There are also concerns about initial value here.  Distinguishing `null` from `false` or `true` could be useful if there's a chance a page object hasn't been `load`ed yet...

modules/twinklexfd.js
Amorymeltzer: Done

modules/twinklexfd.js
```diff
@@ -714,6 +714,29 @@ Twinkle.xfd.setWatchPref = function twinklexfdsetWatchPref(pageobj, pref) {
...
+		pageobj.getStatusElement().warn('Page protected, requesting edit');
+
+		var venue = utils.toTLACase(params.venue);
+
+		var talkTitle = new mw.Title(pageobj.getPageName()).getTalkPage().toText();
+		var editRequest = '\n\n== Edit request to complete ' + venue + ' nomination ==\n';
+		editRequest += '{{Edit protected|' + talkTitle + '|answered=no}}\n';
+		editRequest += 'I nominated ' + pageobj.getPageName() + ' at ' + venue + ' ([[' + params.discussionpage + '|nomination]]), but it is protected so I cannot tag the page. Please add:';
+		editRequest += '\n\n<nowiki>' + params.tagText + '</nowiki>\n\n';
+		editRequest += 'to the top of the page to complete the nomination. Thank you. ~~~~';
```
siddharthvp, line 729: This should be a (substituted) template on-wiki.

Amorymeltzer: I guess?  It seems pretty one-Twinkle-task specific, but I suppose it helps with for portability.  safesubsts are top five least-favorite things.

modules/twinklexfd.js
Amorymeltzer: Done

morebits.js
Amorymeltzer: I've gone and done this, still just with `edit` for now.

modules/twinklexfd.js
siddharthvp: Also, someone creating an XfD without using Twinkle could use it. 

> safesubsts are top five least-favorite things.

Ikr

modules/twinklexfd.js
```diff
@@ -714,6 +714,28 @@ Twinkle.xfd.setWatchPref = function twinklexfdsetWatchPref(pageobj, pref) {
...
+	autoEditRequest: function(pageobj, params) {
+		// Update the form
+		pageobj.getStatusElement().warn('Page protected, requesting edit');
+
+		var venue = utils.toTLACase(params.venue);
+
+		var talkTitle = new mw.Title(pageobj.getPageName()).getTalkPage().toText();
+		var editRequest = '{{Edit protected|' + talkTitle + '|answered=no}}\n' +
+			pageobj.getPageName() + ' has been listed at ' + venue + ' ([[' + params.discussionpage + '|nomination]]), but is protected so could not be tagged. Please add:' +
+			'\n\n<nowiki>' + params.tagText + '</nowiki>\n\nto the top of the page to complete the nomination. Thank you. ~~~~';
```
siddharthvp, line 727:
```suggestion
		var editRequest = '{{subst:xfd edit protected|1=' + pageobj.getPageName() + '|discussion=' + params.discussionpage + '|tag=<nowiki>' + params.tagText + ''\u003C/nowiki>}}'; // U+003C: <
```

I created [the template](https://en.wikipedia.org/wiki/Template:Xfd_edit_protected).

And we can't really use </nowiki> here as that would break the script page on-wiki.

Amorymeltzer: Nice; I had created an earlier [draft in my userspace](https://en.wikipedia.org/wiki/User:Amorymeltzer/sandbox/temp), actually lined up pretty closely.  I made a couple of minor tweaks, I like your method of venue determination, much simpler!

> And we can't really use `</nowiki>` here as that would break the script page on-wiki.

Oof, good point!

siddharthvp: Wow, really close! I thought I was crazy to do things like `~~<noinclude></noinclude>~~` and `<noinclude>|demo=y</noinclude>`. Good to know it's the standard practise.

PR #1092, opened 2020-08-07
xfd/prod: Replace newlines in logged reason with {{pb}}

Amorymeltzer:
Otherwise, new lines entered into the reason box will mess up the numbering

I don't think we want to force `getInputData` to do this on its own, but is there a better way?  There's a reasonable argument this could/should be in `Morebits.string`, alongside something like `formatReasonText`.  I might even prefer that, especially if we want to sidestep the template.

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -386,7 +386,7 @@ Twinkle.prod.callbacks = {
 		if (params.logEndorsing) {
 			logText += 'endorsed ' + (params.blp ? 'BLP ' : params.book ? 'BOOK' : '') + 'PROD. ~~~~~';
 			if (params.reason) {
-				logText += "\n#* '''Reason''': " + params.reason + '\n';
+				logText += "\n#* '''Reason''': " + params.reason.replace(/\n+/g, '{{pb}}') + '\n';
 			}
 			summaryText = 'Logging endorsement of PROD nomination of [[:' + Morebits.pageNameNorm + ']].';
 		} else {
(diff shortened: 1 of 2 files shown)


modules/twinkleprod.js
```diff
@@ -386,7 +386,7 @@ Twinkle.prod.callbacks = {
 		if (params.logEndorsing) {
 			logText += 'endorsed ' + (params.blp ? 'BLP ' : params.book ? 'BOOK' : '') + 'PROD. ~~~~~';
 			if (params.reason) {
-				logText += "\n#* '''Reason''': " + params.reason + '\n';
+				logText += "\n#* '''Reason''': " + params.reason.replace(/\n/g, '{{pb}}') + '\n';
```
siddharthvp, line 389:
```suggestion
				logText += "\n#* '''Reason''': " + params.reason.replace(/\n+/g, '{{pb}}') + '\n';
```
Or `\n{2,}` to be more accurate and in line with normal wikitext representation.


para breaks in wikitext are `\n\n`.

Amorymeltzer: `\n{2,}` wouldn't work as we're talking about within a list, and even a single newline breaks out of that.

`+` is good!

I'm realizing now though that since successive para breaks don't replicate successive newlines — correct or not, hitting return 3+ times gives you more space — if we want truer-ish fidelity, we'd have to use {{break}}.

Amorymeltzer: For clarity, is that a +1 to `\n+` or `{{break}}`?

siddharthvp: Whichever you prefer. \n+ seems good enough. Haven't heard about {{break}}. I think it might be overkill.

Amorymeltzer: `\n+` done, as yes, I agree that {{break}} is a bit precious (and less easy to replicate elsewhere, although `br` would be simple I suppose).

I'm still debating whether it'd be better to put into `Morebits.string`, leaning yes since it's somewhat broadly useful, but then again still fairly limited in scope.

siddharthvp: > I'm still debating whether it'd be better to put into Morebits.string, leaning yes since it's somewhat broadly useful, but then again still fairly limited in scope.

Done in #1110, which added more of this sort of replacements.

Amorymeltzer:
Entirely pre-empted (and improved upon) by #1110, so closing in favor of that.

PR #1094, opened 2020-08-07
xfd: various cleanups after #1030 / 81bb669

Amorymeltzer:
Nothing major, but some minor cleanups initially proposed at https://github.com/azatoth/twinkle/pull/1030#pullrequestreview-446909818

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -62,6 +62,20 @@ var utils = {
 			return title;  // user entered invalid input; do nothing
 		}
 		return title_obj.toText();
+	},
+
+	/**
+	 * Provide Wikipedian TLA style: AfD, RfD, CfDS, RM, SfD, etc.
+	 * @param {string} venue
+	 * @returns {string}
+	 */
+	toTLACase: function(venue) {
+		return venue
+			.toString()
+			// Everybody up, inclduing rm and the terminal s in cfds
+			.toUpperCase()
+			// Lowercase the central f in a given TLA and normalize sfd-t and sfr-t
+			.replace(/(.)F(.)(?:-.)?/, '$1f$2');
 	}
 };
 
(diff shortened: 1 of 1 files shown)


modules/twinklexfd.js
```diff
@@ -51,13 +51,12 @@ var utils = {
...
 	 * Add namespace name to page title if not already given
-	 * CAUTION: namespace name won't be added if a namespace (*not* necessarily
-	 * the same as the one given) already is there in the title
+	 * If a namespace is already in the title, it will be stripped
 	 * @param {string} title
 	 * @param {number} namespaceNumber
 	 */
 	addNs: function(title, namespaceNumber) {
-		var title_obj = mw.Title.newFromUserInput(title, namespaceNumber);
+		var title_obj = mw.Title.newFromUserInput(utils.stripNs(title), namespaceNumber);
```
siddharthvp, line 59: The original was a bit intentional.

`addNs('foobar', 10)` --> `Template:Foobar`
`addNs('Module:foobar', 10)` -->`Module:Foobar`
(where 10 is template NS)

though addNs isn't actually being used for templates/modules (but i think this would be useful for twinkleimage?)

Amorymeltzer: Word.  That was what I was getting at in my original comment on this:

>Makes this less useful and is heavy-handed, but then it would do what it says on the tin. Only used for cfds right now, so it's shouldn't hurt.

modules/twinklexfd.js
```diff
@@ -51,17 +51,30 @@ var utils = {
...
 		}
 		return title_obj.toText();
+	},
+
+	/**
+	 * Provide Wikipedian TLA style: AfD, RfD, CfDS, RM, SfD, etc.
+	 * @param {string} venue
+	 * @returns {string}
+	 */
+	toTLACase: function(venue) {
```
Amorymeltzer, line 71: Moved `toTLACase` here, no need to wait on #1091

modules/twinklexfd.js
Amorymeltzer: Removed it in the interest of going forward.

PR #1096, opened 2020-08-08
morebits: exclude null values as well as undefined in api querystring

Amorymeltzer:
Stems back to 2007 (https://en.wikipedia.org/w/index.php?title=User%3AAzaToth%2Fmorebits.js&diff=prev&oldid=137920973) but the initial intent was to exclude `null` values.  There are a number of places now where we set `null` to `''` so that it's correctly parsed: `{ key: '' }` becoming `key=` rather than `{ key: null }` becoming `key=null`.  This would just leave it off entirely.

----

Is this a good idea?  I haven't tested it at all, so I'm not sure of any negative side effects.  Philosophically, it's mixed; query strings are not a monolith.  `$.param` produces `param=` for `null` values, whereas we're sending `param=null`.  `JSON.stringify` keeps `null` values around, fwiw.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1762,7 +1762,7 @@ Morebits.wiki.api.prototype = {
 		var queryString = $.map(this.query, function(val, i) {
 			if (Array.isArray(val)) {
 				return encodeURIComponent(i) + '=' + val.map(encodeURIComponent).join('|');
-			} else if (val !== undefined) {
+			} else if (val !== undefined && val !== null) {
 				return encodeURIComponent(i) + '=' + encodeURIComponent(val);
 			}
 		}).join('&').replace(/^(.*?)(\btoken=[^&]*)&(.*)/, '$1$3&$2');


siddharthvp:
mw.Api scrubs out `undefined` and `false`. I think we should do the same with `false`. I recall at least one place in twinkleblock which would be simplified by choosing to ignore `false`.

----

I am not sure whether we we need to remove `null`, though, but it seems to be good idea. 

> There are a number of places now where we set null to '' so that it's correctly parsed: { key: '' } becoming key= rather than { key: null } becoming key=null. This would just leave it off entirely.

Which are those places? Note that `{ key: ''}` can be legitimate. For example if `rvprop` is not specified, it takes the default value that's something like `ids|title|comment|user`. But if we don't want any of that in the result, its correct to put rvprop as an empty string. If someone puts `null` when they actually mean the empty string, well, that's wrong, and leads to unexpected behaviour both with this patch and without this patch. 

Ideally, we wouldn't be using null anywhere. You might want to read about the [API data types](https://www.mediawiki.org/wiki/API:Data_formats). The API doesn't have a null data type. Everything is either a string or a boolean. If null is sent for a string field, it gets treated as a string "null" obviously, and for a boolean field, "null" is treated as true. Counter-intuitive in both cases. 

With this patch, where null was being used, the API's interpreted value changes for a string field from being "null" to unspecified (would either throw a missingparams error or silently go with the default value). For a boolean field, the interpreted value changes from true to false. 

TL;DR is that it's a fairly moot point since null shouldn't be used anywhere in correctly written API calls.

Amorymeltzer:
>Ideally, we wouldn't be using null anywhere.

>Note that `{ key: ''}` can be legitimate.

Exactly.  `null` is useful as a background default for `ctx` stuff, but
that's my point — `''` is valid for *some* things, but `null` never is.
This was just to catch any forgetfulness or missed checking.

I'd agree that `false` would be good to do (as well or instead).

siddharthvp:
Hmm I didn't think about the use in morebits itself. So I take it that this will help remove a lot of if conditions in mb.w.p? If so, that's a good thing.

morebits.js
```diff
@@ -1762,7 +1762,7 @@ Morebits.wiki.api.prototype = {
 		var queryString = $.map(this.query, function(val, i) {
 			if (Array.isArray(val)) {
 				return encodeURIComponent(i) + '=' + val.map(encodeURIComponent).join('|');
-			} else if (val !== undefined) {
+			} else if (val !== undefined && val !== null) {
```
Amorymeltzer, line 1765: So if we say `undefined`, `null`, and `false` are out, that becomes, what


```suggestion
			} else if (val !== undefined && val !== null && val !== false) {
```

It'd be nicer visually to do:

```suggestion
			} else if (val || val === '') {
```

But that would remove `0` and `NaN` — I can't imagine ever wanting to pass `NaN` but it's certainly possible to want `0` (number).

PR #1103, opened 2020-08-14
xfd: Fix bugs in RfD target notifications

Amorymeltzer:
Quite a lineage here!  When first added in #527 (a5ae221f), that left a `mw.Title` object being passed to `targetNotification`, rather than the string that `Morebits.wiki.page` expects.  This appears not to have actually been problem, however, until the ability to avoid following cross-namespace redirects was added in #915 (95445604).  This meant that the `mw.Title` object was being passed into `new mw.Title`, which would quietly complain.

It turns out that bug was epistatic to another misnaming from the addition of `getInputData` in #908 (df2aa163): a single `params.usertalk` remained, when the other had all been converted to `params.notifycreator`.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -1705,7 +1705,7 @@ Twinkle.xfd.callbacks = {
 						statelem.warn('You (' + mw.config.get('wgUserName') + ') are the target; skipping target page notification');
 					}
 				} else {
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Hmm, some of that might be wrong: it seems to be working fine on enWiki based on the latest results at RfD, but if the no xns bit was the issue I would expect it to still be silently not working.

PR #1106, opened 2020-08-15
warn: Add uw-displaytitle

jackmcbarn:


modules/twinklewarn.js
```diff
@@ -926,6 +926,10 @@ Twinkle.warn.messages = {
 			label: 'Removing proper sources containing dead links',
 			summary: 'Notice: Removing proper sources containing dead links'
 		},
+		'uw-displaytitle': {
+			label: 'Incorrect use of DISPLAYTITLE',
+			summary: 'Notice: Incorrect use of DISPLAYTITLE'
```
Amorymeltzer, line 931: Not sure this conveys the intent of the template fully.  It looks like you wrote it to be used for a new account using `DISPLAYTITLE` instead of moving a page; these don't seem to fully capture that, they seem broader to me.  And might be a bit BITEy?

Amorymeltzer (commented):
Seems rather niche, is this a common issue/problem such that a template would have wide usage?

modules/twinklewarn.js
jackmcbarn: I see your point, but I'm not sure how to make it better without making it really really long. Any ideas?

jackmcbarn:
> Seems rather niche, is this a common issue/problem such that a template would have wide usage?

https://en.wikipedia.org/wiki/Category:Pages_with_disallowed_DISPLAYTITLE_modifications has almost 6000 pages in it, and seems to have a net upward trend of 5-10 per day.

modules/twinklewarn.js
Amorymeltzer: Yeah I was struggling too. Maybe something more straightforward, like "DISPLAYTITLE doesn't modify the page name"?  Out of style but as far I've gotten atm.

Amorymeltzer:
To be sure, but how many of those are new users thinking they're renaming a page as opposed to some other issue?  That's the niche I meant.

Amorymeltzer (approved):

modules/twinklewarn.js
Amorymeltzer: Fine enough I suppose!  Merged.

PR #1118, opened 2020-09-02
Don't follow user talk page redirects when issuing block templates.

wiki-ST47:
This patch fixes security/harassment issue #912 by preventing following of redirects when issuing a block.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -1559,7 +1559,6 @@ Twinkle.block.callback.issue_template = function twinkleblockCallbackIssueTempla
 
 	var wikipedia_page = new Morebits.wiki.page(userTalkPage, 'User talk page modification');
 	wikipedia_page.setCallbackParameters(params);
-	wikipedia_page.setFollowRedirect(false);
 	wikipedia_page.load(Twinkle.block.callback.main);
 };
 
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -1559,7 +1559,7 @@ Twinkle.block.callback.issue_template = function twinkleblockCallbackIssueTempla
 
 	var wikipedia_page = new Morebits.wiki.page(userTalkPage, 'User talk page modification');
 	wikipedia_page.setCallbackParameters(params);
-	wikipedia_page.setFollowRedirect(true, false);
+	wikipedia_page.setFollowRedirect(false);
 	wikipedia_page.load(Twinkle.block.callback.main);
 };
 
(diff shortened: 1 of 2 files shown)


modules/twinkleblock.js
```diff
@@ -1559,7 +1559,7 @@ Twinkle.block.callback.issue_template = function twinkleblockCallbackIssueTempla
 
 	var wikipedia_page = new Morebits.wiki.page(userTalkPage, 'User talk page modification');
 	wikipedia_page.setCallbackParameters(params);
-	wikipedia_page.setFollowRedirect(true, false);
+	wikipedia_page.setFollowRedirect(false);
```
Amorymeltzer, line 1562:
```suggestion
```
`false` is the default, so you can actually just remove the line entirely!

Amorymeltzer (dismissed):
LGTM!  As I stated back in April, I've come across to this, especially if there's an LTA abusing it.  I still don't see how it's a security issue, though?

wiki-ST47:
How is it *not* a security issue? Twinkle users' accounts are being hijacked to place warning and block templates on innocent users' talk pages, and this is being actively abused as part of a harassment attack.

This is apparently also a problem with speedy deletion notices, I'm looking to see if that is also easy to fix.

siddharthvp:
I don't think this patch is a good idea as it means that the redirect page would be edited so that there is a template under the redirect. https://github.com/azatoth/twinkle/issues/912#issuecomment-685731173 is likely the way to go. 

(But as an urgent temporary fix, I guess it's ok to directly apply this change to the on-wiki gadget).

Amorymeltzer:
> I don't think this patch is a good idea as it means that the redirect page would be edited so that there is a template under the redirect. [#912 (comment)](https://github.com/azatoth/twinkle/issues/912#issuecomment-685731173) is likely the way to go.
> 
> (But as an urgent temporary fix, I guess it's ok to directly apply this change to the on-wiki gadget).

Basically what I'm thinking, but it warrants further discussion (always provide, don't always provide)

> How is it _not_ a security issue? Twinkle users' accounts are being hijacked to place warning and block templates on innocent users' talk pages, and this is being actively abused as part of a harassment attack.

This is bikeshedding I suppose, but I would say rather that *users* are being *tricked*.  Maybe our definitions are different, but I usually take "hijacked" and "security" to be something that threatens the integrity of access to account, usually enabling generic abilities.  At any rate, agreed that it's bad and Twinkle should adapt to use, not just ideal use.

Amorymeltzer:
I've merged (as b67e2c73) fix here for `twinkleblock`, and deployed it on-wiki.

Going forward, I suppose #912 can be where we figure out exactly how to treat this sort of thing in morebits, etc.

PR #1120, opened 2020-09-06
fluff: Add MajavahBot to trustedBots

supertassu:
MajavahBot can edit directly after an edit to WP:EF/FP/R, a page with frequent vandalism edits.


mdaniels5757 (approved):
LGTM

Amorymeltzer:
Seems reasonable — this is underutilized IMO.  I looked back a few pages, didn't see a ton of places where it would have helped, but seems like it only could!

PR #1121, opened 2020-09-07
speedy: Use Module:Module wikitext to tag modules

jackmcbarn:


modules/twinklespeedy.js
```diff
@@ -1516,6 +1515,15 @@ Twinkle.speedy.callbacks = {
 				}
 			}
 
+			// Scribunto isn't parsed like wikitext, so CSD templates on modules need special handling to work
+			if (params.scribunto) {
+				var equals = '';
+				while(code.indexOf(']' + equals + ']') !== -1) {
+					equals += '=';
+				}
+				code = "require('Module:Module wikitext')._addText([" + equals + '[' + code + ']' + equals + ']);';
```
Amorymeltzer, line 1524: What are the brackets for?  This ends up with something like `require('Module:Module wikitext')._addText([[{{db-vandalism|help=off}}]]);`, which isn't what your doc suggests.

Amorymeltzer:
Neat module @jackmcbarn!  Seems pretty solid and steady on WP:CSD and [WT:CSD](https://en.wikipedia.org/wiki/Wikipedia_talk:Criteria_for_speedy_deletion#New_section_-_Modules) so would be good to add it here for sure!

I can't seem to get the module to work on test wiki, though - any idea what I'm missing?  Lua's nowhere near my strong point.

modules/twinklespeedy.js
jackmcbarn: Lua uses brackets for the same purpose that most languages use triple-quoted strings for. By putting everything inside of `[[`, `[=[`, etc., it forms a string literal where nothing needs to be escaped. I thought this was simpler/easier than using regular strings and then having to escape things, but let me know if you'd rather I do that instead.

jackmcbarn:
> I can't seem to get the module to work on test wiki, though - any idea what I'm missing? Lua's nowhere near my strong point.

You got everything with the Lua modules correct. The only missing piece was that https://test.wikipedia.org/wiki/MediaWiki:Scribunto-doc-page-show was a stub that didn't actually call any of the Lua code. I imported that from enwiki and now your tests there seem to work.

modules/twinklespeedy.js
Amorymeltzer: No, that's excellent; there's user-entered text (esp. with `{{db}}`) so that's fabulous.

Presumably it's just different in the on-wiki doc to just KISS when for manual editing?

jackmcbarn: Yep, that's exactly why. I expect that most editors following that documentation will (1) be more familiar with regular strings with escaping, and (2) will probably almost always use things that don't need any escaping.

modules/twinklespeedy.js
```diff
@@ -1516,6 +1515,15 @@ Twinkle.speedy.callbacks = {
 				}
 			}
 
+			// Scribunto isn't parsed like wikitext, so CSD templates on modules need special handling to work
+			if (params.scribunto) {
```
Amorymeltzer, line 1519: Not super critical, but there's no need for line 2175 (`params.scribunto = mw.config.get('wgPageContentModel') === 'Scribunto';`) to exist anymore, yeah?  This can just be written: 
```suggestion
			if (mw.config.get('wgPageContentModel') === 'Scribunto') {
```

modules/twinklespeedy.js
```diff
@@ -1516,6 +1515,15 @@ Twinkle.speedy.callbacks = {
...
+				var equals = '';
+				while(code.indexOf(']' + equals + ']') !== -1) {
+					equals += '=';
+				}
+				code = "require('Module:Module wikitext')._addText([" + equals + '[' + code + ']' + equals + ']);';
+			}
+
```
Amorymeltzer, line 1526: Stray whitespace; surprised linter didn't catch it...

```suggestion

```

Amorymeltzer (commented):
Works great!  I've suggested a minor cosmetic change, but nothing critical.

This will be *much* nicer, I think.  And, selfishly, for #1090

jackmcbarn:
Your suggestions seemed reasonable, so I applied them. How is this related to #1090 though?

Amorymeltzer (approved):
Fabulous

Amorymeltzer:
Merged as 9e1c1acc, thanks again!

Amorymeltzer:
>How is this related to #1090 though?

It's not strictly, but as part of that restructuring, I'm turning a couple of options for `fluff` and `speedy` modules away from booleans.  Since we weren't editing the module page directly, the watch query made to watch the module page itself was a PITA: the `watch` action is just set up to watch an item, not to do so potentially based on user preferences.  That is, if a user had the watchlist option set to follow their preferences, we had to check those preferences when deciding whether to separately post the watch query.  You've done away with the need for that.

I'm probably planning on re-introducing it separately, however, so it's all moot!

PR #1128, opened 2020-09-14
morebits.css: make font sizes consistent across skins, browser settings

siddharthvp:
A thorough version of #1034.

Fixes the issue of font size not being scaled for users with uncommon browser font settings. Also makes the sizes consistent across all skins.

The px translation of the % sizes used are: (on the default browser settings)
- 12px in most places
- 12.96px (108% of 12px) for the titlebar and h5 headers
- 11.64px (97% of 12px) for footer links
On timeless, the sizes are very slightly different (12.008px instead of 12px, etc).

This PR touches only morebits.css. Left for the future:
- select2 css tweaks invloves a hardcoded 13px, which won't scale with browser font settings
- The large text option in TWPREFS also uses an absolute 13px

This is live on testwiki.

### DIFF
--- a/morebits.css
+++ b/morebits.css
@@ -40,7 +40,6 @@ form.quickform
 form.quickform *
 {
 	font-family: sans-serif;
-	font-size: 12px;
 }
 
 form.quickform fieldset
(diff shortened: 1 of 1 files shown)


Amorymeltzer (approved):
Thanks for taking this on, all looks great AFAICS!

These are some weird-ass numbers.

PR #1133, opened 2020-09-19
protect: Add options for watching protected/requested/tagged pages

Amorymeltzer:
`watchRequestedPages`, `watchPPTaggedPages`, and `watchProtectedPages`

----

As written, dependent on #1090; it need not be.


Amorymeltzer:
Should also wait until #1136 then use that system.

PR #1134, opened 2020-09-19
fluff: Add preference to require confirmation on mobile

Amorymeltzer:


modules/twinklefluff.js
```diff
@@ -617,7 +617,10 @@ Twinkle.fluff.callbacks = {
 				break;
 		}
 
-		if (Twinkle.getPref('confirmOnFluff') && !userHasAlreadyConfirmedAction && !confirm('Reverting page: are you sure?')) {
+		if ((Twinkle.getPref('confirmOnFluff') ||
+			// Mobile user agent taken from [[en:MediaWiki:Gadget-confirmationRollback-mobile.js]]
+			(Twinkle.getPref('confirmOnMobileFluff') && /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent))) &&
```
siddharthvp, line 622:
```suggestion
			(Twinkle.getPref('confirmOnMobileFluff') && /Android|webOS|iPhone|iPad|iPod|BlackBerry|Mobile|Opera Mini/i.test(navigator.userAgent))) &&
```
This regex may not actually be that effective any longer. See https://en.wikipedia.org/wiki/Wikipedia:Village_pump_(technical)/Archive_184#Gadget%3A_%22Require_confirmation_before_performing_rollback_on_mobile_devices%22 . I would suggest using `Mobile` to catch UC browser, Edge Mobile, Samsung browser, Yandex browser etc (see https://deviceatlas.com/blog/mobile-browser-user-agent-strings).

Amorymeltzer: Those all have `Android` according to that list, but `Mobile` rather than `IEMobile` does catch the blackberry example (I guess they dropped their own moniker?), so :+1:

Truth be told, since we're largely interested in touch, not really mobile, both this and the gadget are more likely to want something with `Navigator.maxTouchPoints`, but I've no idea how well supported it is and it is probably only going to become less useful.

Issue #1131, opened 2020-09-17
Dialogs should have link to TWPREFS


siddharthvp:
To make WP:TWPREFS more discoverable, the Twinkle dialogs should have a morebits footerlink pointing to the relevant section (for that module) at TWPREFS.

Amorymeltzer:
Is that even possible at the moment?  The content is added by config, so the sections are only added after the page is loaded.  Presumably it'd involve adding a hook into config to send folks to the section once the content has been loaded; that's probably a good idea regardless!

Amorymeltzer:
See #1136, which should make this manageable.

Amorymeltzer:
I've merged bug fixes #1135 and #1136, so this should now be trivial

PR #1135, opened 2020-09-19
config: Fix jump-to-section behavior

Amorymeltzer:
Added in 46cd08038, but won't jump if there's no change, so pass a null value first.  See also #1131.

----

Feel like there must be a better way of doing this, but this works and I don't see any false jumping?


Amorymeltzer:
Pushed live

PR #1136, opened 2020-09-19
config: Use consistent heading ids for links

Amorymeltzer:
The use of `tocnumber` meant that any additional section or other rearrangement would change the location any external links would head.  That's not good!  This adds a `module` attribute, identifying each section with the specific Twinkle module being used, so any incoming section links will be consistent.  Probably overdue.

The reality is that the currently used headers have been consistent for ages, so hasn't been a huge issue; #1133 would make it one.  This doesn't make an attempt to attach the old `id`s to their long-time sections, although given the consistency/longevity of those links, that would probably be a good idea...

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -126,6 +126,7 @@ Twinkle.config.commonSets = {
  *
  * {
  *   title: <human-readable section title>,
+ *   module: <name of the associated module, used to link to sections>,
  *   adminOnly: <true for admin-only sections>,
  *   hidden: <true for advanced preferences that rarely need to be changed - they can still be modified by manually editing twinkleoptions.js>,
  *   preferences: [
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
`oldTOCNumber: 6,` etc., then:

```javascript
if (section.oldTOCNumber) {
	var oldtocdiv = document.createElement('div');
	oldtocdiv.id = 'twinkle-config-section-' + section.oldTOCNumber;
	cell.appendChild(oldtocdiv);
}
```

Amorymeltzer:
Actually this has been totally in flux since #260 when the block module was added.  Since then, the headers for sysops not only changed, but have diverged from those for non-sysops.

Amorymeltzer:
That fact about block means that nearly all the headers are inconsistent between sysops and non-sysops.  I think it'd be good to have some fix for old section links, but it's not possible to be consistently correctly.  A section link posted by a sysop will *always* be wrong for a non-sysop, and vice-versa.

Thus, while I'm open to adding something (half a loaf and all that), the reality is that, especially given #1135, there's not much value to be gained.  I'm going to merge this without the fix, I anticipate few complaints.

PR #1137, opened 2020-09-19
morebits: remove getChecked/getUnchecked from HTMLFormElement prototype

Amorymeltzer:
Replaced `getUnchecked` with xor in `getChecked`, no need to be so redundant

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -700,6 +700,74 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 	return [ node, childContainder ];
 };
 
+
+/**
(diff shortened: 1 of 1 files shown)


siddharthvp:
Sorry missed this earlier why are the functions under `Morebits.quickForm.element.prototype`? This obviously won't work. They need to be under `Morebits.quickForm`.

siddharthvp:
Besides the removed functions are still in use, [as you noted](https://github.com/azatoth/twinkle/issues/910#issuecomment-651500140).

Amorymeltzer:
Undone in #1157, no earthly idea what happened there; thanks.

Issue #1005, opened 2020-06-28
speedy / PROD? / AfD? : Please show us the page's deletion log, so we can decide if salting is appropriate or not


unforgettableid:
The problem:

Sometimes, a page may have been repeatedly recreated, but the Twinkle user doesn't know this. The user tags the page as G11, and the page gets deleted. Later on, a spammer recreates it yet again.

My proposed solution:

It would be good if Twinkle could check the logs before tagging a page for speedy deletion. If the page has been created at least three times, Twinkle could automatically request salting — even if the user hasn't ticked the 'Tag for creation protection (salting)' checkbox.

Thank you for reading this!

Dinoguy1000:
If this is added, it should be sensitive to the timing of the past deletions, e.g. a page which was last recreated and deleted a decade ago shouldn't have a request for salting made for it.

unforgettableid:
We could decide upon a threshold beyond which the previous article creations or deletions are considered "too old".  If all or almost all are considered "too old", Twinkle could refrain from requesting salting.

When considering past article creations and/or deletions, how many years back should Twinkle look?

Dinoguy1000:
Honestly, I'd be surprised if anything older than around a year would be worth considering. Though that being said, it's been quite some time since I was active on Wikipedia, and even when I was, page deletion/salting wasn't an area of focus for me.

Amorymeltzer:
I don't know that a recommendation is quantifiable.  As a sysop, I've salted pages that have been recreated only once and not salted pages that have done so a half-dozen times: it's all very subjective and dependent on the situations around the individual case.

That being said, I think we could lazily include a little text insert showing how many prior deletions took place, yeah?  That might be useful information that taggers might be able to use to make the decision themselves.  I wouldn't want to be seen as unduly influencing folks, though.

unforgettableid:
@Amorymeltzer wrote:

> I think we could lazily include a little text insert showing how many prior deletions took place, yeah? That might be useful information that taggers might be able to use to make the decision themselves.

This would be a good idea.

CSDHelper already does something vaguely like this in certain cases. It loads the deletion log and displays it in a box which is collapsed by default. If you click the "Show deletion log" link, it expands the box, revealing the list of deletion-log entries.

It might be excellent if Twinkle could do something similar.

PR #1138, opened 2020-09-20
speedy: Note number of previous deletions

Amorymeltzer:
Little black message, red if 3+, max 5.  Implements #1005.

cc @unforgettableid @Dinoguy1000

Right now, it doesn't take into account how old an entry is; that is, if a page was deleted three times in 2006, it'll be noted here.  If we want to add a time limit, we'd add something like:

```javascript
leend: new Morebits.date().subtract(2, 'years').toISOString(),
```

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -258,6 +258,12 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
 
 	form.append({
 		type: 'div',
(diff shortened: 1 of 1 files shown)


unforgettableid:
Excellent; much appreciated!

Amorymeltzer:
Maybe: don't limit the timeframe, but add if folks request it?  Could ask when put live.

Also maybe: If we remove the limit on searches, we could give an exact number and say "42 previous deletions since 2004-01-20" or something

PR #1139, opened 2020-09-27
morebits: fix error handling on no connectivity

siddharthvp:
commit 1: fix error handling on no internet 
errorCode is `null` on no internet connectivity, not `undefined` let alone the string `'undefined'` which the code was looking for.
This bug was also preventing any onFailure handler from triggering because of the TypeError on line 3260 (null doesn't have `indexOf` method).

commit 2: wait for 2 seconds after a connection error
Retrying immediately after encountering a network error is pointless, the most likely outcome is that the same error would recur. (Ever since I added a 5-sec wait in SDZeroBot, it has never ever encountered failures due to lack of connectivity. 5 may be excessive for a user-facing script though.)

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -3243,7 +3243,7 @@ Morebits.wiki.page = function(pageName, currentAction) {
 			purgeApi.post();
 
 		// check for network or server error
-		} else if (errorCode === 'undefined' && ctx.retries++ < ctx.maxRetries) {
+		} else if ((errorCode === null || errorCode === undefined) && ctx.retries++ < ctx.maxRetries) {
 
 			// the error might be transient, so try again
 			ctx.statusElement.info('Save failed, retrying');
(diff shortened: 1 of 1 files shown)


morebits.js
```diff
@@ -3243,12 +3243,16 @@ Morebits.wiki.page = function(pageName, currentAction) {
...
-		} else if (errorCode === 'undefined' && ctx.retries++ < ctx.maxRetries) {
+		} else if ((errorCode === null || errorCode === undefined) && ctx.retries++ < ctx.maxRetries) {
 
 			// the error might be transient, so try again
-			ctx.statusElement.info('Save failed, retrying');
+			ctx.statusElement.info('Save failed, retrying in 2 seconds ...');
 			--Morebits.wiki.numberOfActionsLeft;  // allow for normal completion if retry succeeds
-			ctx.saveApi.post(); // give it another go!
+
+			// wait for sometime for client to regain connnectivity
```
Amorymeltzer, line 3252: I've no idea how to trigger this to test it!

Amorymeltzer: Ah, I was dumb, needed to have `load`ed the page.  All checks out!

Amorymeltzer (changes requested):
>This bug was also preventing any onFailure handler from triggering because of the TypeError on line 3260 (null doesn't have indexOf method).

Commit 1 is actually insufficient to prevent this, as once we've exhausted replies, the `errorCode` enters the "hard error" section.  Might as well wrap `protectedpage`, `abusefilter`, and `spamblacklist` in an `if (errorCode) {` block, yeah?  Other than that, this all looks 👍

siddharthvp:
> Commit 1 is actually insufficient to prevent this, as once we've exhausted replies, the errorCode enters the "hard error" section. Might as well wrap protectedpage, abusefilter, and spamblacklist in an if (errorCode) { block, yeah? 

Oh yeah. Put it into a switch instead as that seems to makes the whole thing tidier.

Amorymeltzer (approved):

Amorymeltzer:
>Put it into a switch instead as that seems to makes the whole thing tidier.

Nice! LGTM 👍 

(linking #707 as the source of this)

Issue #1119, opened 2020-09-05
Missing token errors


siddharthvp:
Needs investigation. Been reported many times of late:

* https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_43#Error_with_XfD
* https://en.wikipedia.org/w/index.php?title=Wikipedia:Articles_for_deletion/Ry%C5%8Dhei_Arai&diff=976224156&oldid=976219898&diffmode=source
* https://en.wikipedia.org/wiki/User_talk:SD0001/DYK-helper#This_script_lost_my_nomination (DYK helper uses morebits)

One would think https://github.com/azatoth/twinkle/pull/1038/commits/af13dcede0ec3452adf835c2baf597e25b4312a0 removed all possibility of token errors, but apparently not.

Amorymeltzer:
The timing of one or two of these is weird, but I think there have recently been some server-side token issues (or, at least, others).  Good to have open to track.

Amorymeltzer:
All these came after #1038 so I wonder if #1143 won't solve them all?

siddharthvp:
Indeed it would.

PR #1143, opened 2020-10-05
morebits: Fix bug in getting new token

Amorymeltzer:
Stems from af13dcede/#1038.  Looks like, since `Morebits.wiki.api.getToken` explicitly creates a new construct, `this` in `returnError`'s return ends up going global (`window`).

[Reported at WT:TW](https://en.wikipedia.org/w/index.php?oldid=982033312#%22Getting_token%22_and_what?)


Amorymeltzer:
This should at least partially explain #1119.

siddharthvp:
Ohhhhh! Can't believe I did that. I would've preferred `.bind(this)` but anyway.

Amorymeltzer:
Fair; done.

PR #1146, opened 2020-10-10
morebits: restore button font size

siddharthvp:
Restore the original font size of 13px on the submit buttons. Missed in #1128. This was particularly conspicuous in Tag module where the "Adding x tags" text had also become much smaller.


Amorymeltzer (approved):
Good catch!

PR #1147, opened 2020-10-10
xfd: tfd: auto-select type=inline for inline templates

siddharthvp:
Inline tag templates (eg. Template:Dead link, Template:More citations needed) use the class "Inline-Template". The selector `.mw-parser-output > p .Inline-Template` ensures that any inline tags used in the template documentation aren't considered.


Amorymeltzer (approved):
Nice!  :+1:

Just noting that this comes from {{Fix}}, which covers nearly all such templates, and to an approximation, 100% of the uses.

Issue #1140, opened 2020-10-02
Move twinkle to a GitHub organisation


siddharthvp:
Suggested by @atlight in https://github.com/azatoth/twinkle/issues/129#issuecomment-702759024. 

Should we move twinkle to a new organisation called "en-wikipedia-gadgets"? I think that would be great as we could also host other gadgets without a single-owner such as popups over there. I have wanted to contribute to [popups](https://en.wikipedia.org/wiki/MediaWiki:Gadget-popups.js) -- but it's so difficult to get code reviews for any big changes due to lack of a repository.

MusikAnimal:
Maybe "wikimedia-gadgets", or even "wikimedia-scripts"? Hopefully one day Twinkle will be global, and other gadgets like Popups could be too.

MusikAnimal:
I would be happy to migrate [MoreMenu](https://github.com/MusikAnimal/MoreMenu) to the org. I am the sole author, but that's not by choice. In my view, once a script becomes a gadget, especially if it's "global" (used on many wikis), it's sort of in the community's hands since any int-admin can make changes to the deployed code.

Amorymeltzer:
Would be neat, although if we’re thinking about a broader (en)wiki gadget org, it should primarily be discussed and organized on-wiki.

It’d be nice to have a centralized(ish) place for reviews ([T71445](https://phabricator.wikimedia.org/T71445) is a long way off) but I’m not certain how it’ll be perceived, what with requiring another account, etc.  Twinkle and MoreMenu would be on-board, and @enterprisey and @evad37 are obvious candidates, but I can see folks objecting to taking things like popups off-wiki (even if it’d be for the better (Also, oi does popups need some love)).

That being said, I think at the very least *Twinkle* could be made into an organization @azatoth.  I’ve thought about asking a few times, if only for some stuff around actions/security/etc.

DannyS712:
Agree that having a centralized(ish) place for reviewers would be nice, but also that another account might not be the best idea. Perhaps use gerrit?

siddharthvp:
[...]

PR #1151, opened 2020-10-11
Lint via GitHub actions rather than travis-ci

Amorymeltzer:
Travis is [still running requests](https://travis-ci.org/azatoth/twinkle/requests), but the connection to the GitHub web interface recently stopped.  It's not a bad excuse to <s>play around with</s> move to GitHub's actions!  I've been thinking about doing so for a while, so why not now?  (ping @MusikAnimal )

If so, @azatoth I think you'll need to enable some options at https://github.com/azatoth/twinkle/settings/actions

If not, I think you'll have to reauthorize the connection; this is one of those things that only the repository owner can do, since this isn't [an organization](https://github.com/azatoth/twinkle/issues/1140#issuecomment-703013774).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,16 @@
+name: Linter
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: npm ci
+
+    - name: Lint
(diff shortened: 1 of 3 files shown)


Amorymeltzer:
Thanks @azatoth, and for setting up codeql analysis; that was gonna be my next step!

azatoth:
> Thanks @azatoth, and for setting up codeql analysis; that was gonna be my next step!

No problem; I did chuckle a bit over the "DOM text reinterpreted as HTML" issues :)

PR #1152, opened 2020-10-11
morebits: Do triage properly, with extra query to determine if page is in queue

Amorymeltzer:
Original structure from #930, where we can properly asses and handle cases, rather than just swallowing all errors and avoiding stuff like #1093/#1145 and #1150.  Restore AfD's `triage` to where it (sensibly) was before #1093.

----

Per the previously convo with @siddharthvp, not sure if this is better than the current system, but figured I'd put it up for consideration.  It's an extra query if the page can be triaged, but otherwise avoids running into any errors.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -987,6 +987,11 @@ Twinkle.xfd.callbacks = {
 
 			// Now we know we want to go ahead with it, trigger the other AJAX requests
 
+			// Mark the page as curated/patrolled, if wanted
+			if (Twinkle.getPref('markXfdPagesAsPatrolled')) {
+				new Morebits.wiki.page(Morebits.pageNameNorm).triage();
+			}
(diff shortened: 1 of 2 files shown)


siddharthvp:
The real issue here appears to be: `Morebits.status.onError(Twinkle.xfd.printRationale);` -- rationale is printed along with `Your deletion rationale is provided below, which you can copy and paste into a new XFD dialog if you wish to try again` if _any_ API request fails, which is ridiculous. It only needs to be shown if the API request that creates (or appends to) the discussion page fails, or if any API call leading up to that call fails.

siddharthvp:
I guess this is reasonable enough to merge for now, but I'll note that issues like would most likely get resolved the "proper" way when I rewrite XFD as part of i18n/l10n work. We won't need an extra query at that time.

PR #1153, opened 2020-10-11
xfd: Use a unified user notification function

Amorymeltzer:
With all the XfD notices unified (29ae3373), we don't need to handle them individually.  This handles all current notifications, including without double-logging on MfDs, TfMs, and RfD, and handles non-userspace targets in RfD target notifications; the majority of the complexity is due to handling MfD and RfD.  Should also work for #1001.

Also fixes a bug in MfD notifications, where you could only notify the userspace owner if you were also notifying the creator, regardless of whether they were the same or not.

----

I think this is a benefit, as despite the annoying handlings, it's all in one spot rather than disparate.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -496,7 +496,7 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 					type: 'checkbox',
 					list: [
 						{
-							label: 'Also notify owner of userspace if they are not the page creator',
+							label: 'Notify owner of userspace (if they are not the page creator)',
 							value: 'notifyuserspace',
 							name: 'notifyuserspace',
 							tooltip: 'If the user in whose userspace this page is located is not the page creator (for example, the page is a rescued article stored as a userspace draft), notify the userspace owner as well.',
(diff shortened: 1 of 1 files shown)


modules/twinklexfd.js
```diff
@@ -818,6 +818,100 @@ Twinkle.xfd.callbacks = {
...
+				}
+				break;
+			case 'cfd':
+				notifytext += '|action=' + params.action + (mw.config.get('wgNamespaceNumber') === 10 ? '|stub=yes' : '');
+				break;
+			default: // ffd, rfd
+				break;
+		}
+		// {{Tfd notice}} can't handle the namespace prefix
+		notifytext += '|1=' + (params.venue === 'tfd' ? mw.config.get('wgTitle') : Morebits.pageNameNorm) + '}} ~~~~';
```
siddharthvp, line 877:
```suggestion
		notifytext += '|1=' + Morebits.pageNameNorm + '}} ~~~~';
```
I have fixed {{Tfd notice}} so that we don't need the special case.

modules/twinklexfd.js
```diff
@@ -818,6 +818,100 @@ Twinkle.xfd.callbacks = {
...
+			case 'mfd':
+				notifytext += params.numbering !== '' ? '|order=&#32;' + params.numbering : '';
+				break;
+			case 'tfd':
+				if (params.xfdcat === 'tfm') {
+					notifytext = '\n{{subst:Tfm notice|2=' + params.tfdtarget;
+				}
+				if (mw.config.get('wgPageContentModel') === 'Scribunto') {
+					notifytext += '|module=yes';
+				}
```
siddharthvp, line 868:
```suggestion

```
Unnecessary per [changes made](https://en.wikipedia.org/w/index.php?title=Template:Tfd_notice&diff=983106167&oldid=954464172&diffmode=source) to Template:Tfd notice.

siddharthvp (changes requested):

Amorymeltzer:
Nice, thanks @siddharthvp!

PR #1160, opened 2020-10-17
Make patch-test.js work on Windows

siddharthvp:
Windows doesn't support the $(...) syntax like unix shells, so we run the shell commands separately.

Also, `git merge-base $(git rev-parse --abbrev-ref HEAD) master` can be written as just `git merge-base HEAD master` so that's one less command to run.

### DIFF
--- a/dev/patch-test.js
+++ b/dev/patch-test.js
@@ -1,10 +1,4 @@
 /* eslint-env node, es6 */
-/* eslint-disable no-console */
-/* eslint-disable es5/no-destructuring */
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I'll have to take your word for it — I've got a VM for XP and 7, but I'll be damned if I'm gonna go through the hassle to install git on 'em!

siddharthvp:
At least you've got just one OS to worry about! Talk of the constant switching between Windows, WSL and Linux, oof.

PR #1162, opened 2020-10-17
speedy: code cleanup

siddharthvp:
Some code cleanup. Quite a lot of crappy code in this module ...

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -1994,7 +1994,10 @@ Twinkle.speedy.getUserTalkParameters = function twinklespeedyGetUserTalkParamete
 	return utparams;
 };
 
-
+/**
+ * @param {Event} e
+ * @returns {Array}
+ */
 Twinkle.speedy.resolveCsdValues = function twinklespeedyResolveCsdValues(e) {
 	var values = (e.target.form ? e.target.form : e.target).getChecked('csd');
 	if (values.length === 0) {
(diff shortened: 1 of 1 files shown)


modules/twinklespeedy.js
```diff
@@ -253,36 +216,17 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
...
-};
-
 Twinkle.speedy.callback.modeChanged = function twinklespeedyCallbackModeChanged(form) {
 	var namespace = mw.config.get('wgNamespaceNumber');
 
 	// first figure out what mode we're in
-	var mode = Twinkle.speedy.callback.getMode(form);
-	var isSysopMode = Twinkle.speedy.mode.isSysop(mode);
+	var mode = {
+		isSysop: form.tag_only && !form.tag_only.checked,
```
Amorymeltzer, line 224: Doesn't matter since we only ever treat it as a boolean, just noting that this is `undefined` for users.
```suggestion
		isSysop: !!form.tag_only && !form.tag_only.checked,
```

modules/twinklespeedy.js
```diff
@@ -2042,28 +2045,14 @@ Twinkle.speedy.callback.evaluateSysop = function twinklespeedyCallbackEvaluateSy
...
-			if (Twinkle.getPref('warnUserOnSpeedyDelete').indexOf(norm) !== -1) {
-				if (norm === 'g6' && values[index] !== 'copypaste') {
-					return true;
-				}
-				warnusertalk = true;
-				return false;  // break
-			}
-		});
-	}
+	var warnusertalk = form.warnusertalk.checked && normalizeds.some(function (norm, index) {
```
Amorymeltzer, line 1969: Nice.  I always forget that `.some` and `.every` are available, and lump them in with `.includes()`

modules/twinklespeedy.js
```diff
@@ -2101,54 +2090,26 @@ Twinkle.speedy.callback.evaluateUser = function twinklespeedyCallbackEvaluateUse
 	}
 
 	// var multiple = form.multiple.checked;
-	var normalizeds = [];
-	$.each(values, function(index, value) {
-		var norm = Twinkle.speedy.normalizeHash[value];
 
-		normalizeds.push(norm);
+	var normalizeds = values.map(function (value) {
```
Amorymeltzer, line 2015:
```suggestion
	var normalizeds = values.map(function(value) {
```

modules/twinklespeedy.js
```diff
@@ -473,12 +414,12 @@ Twinkle.speedy.generateCsdList = function twinklespeedyGenerateCsdList(list, mod
 			return null;
 		}
 
-		if (criterion.subgroup && !hasSubmitButton) {
+		if (criterion.subgroup && mode.isRadioClick) {
```
Amorymeltzer, line 417: `mode.isRadioClick` does what it says on the tin, but it isn't an accurate replacement for `!hasSubmitButton`.  I believe this should be
```suggestion
		if (criterion.subgroup && !mode.isMultiple && mode.isRadioClick) {
```
As is, selecting multiple options is broken in `radioClick`.

Amorymeltzer (changes requested):
Just needs one bugfix, but otherwise LGTM!  This stuff was hideous, much nicer now! 💯

modules/twinklespeedy.js
siddharthvp: nice catch, weirdly I thought radioclick mode === the one without submit button.

Amorymeltzer: It is!  But when there are multiple options, it obviously has to get one.

siddharthvp:
Diff for changes in force pushes: https://github.com/azatoth/twinkle/compare/00a1d4f80f2ed04cc597eefca6223ea92076cf89..3341706

Amorymeltzer:
I see some `getInputData` work slipped in there...

PR #1164, opened 2020-10-19
tag: Add an input for redirect page name for R avoided double redirect

TheTVExpert:
Added an input to allow the user to enter the redirect page name for the R avoided double redirect template.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1680,9 +1680,8 @@ Twinkle.tag.callbacks = {
 				if (params.altLangTo) {
 					tagText += '|to=' + params.altLangTo;
 				}
-				if (params.doubleRedirectTarget) {
-					tagText += '|1=' + params.doubleRedirectTarget;
-				}
+			} else if (tagName === 'R avoided double redirect' && params.doubleRedirectTarget) {
+				tagText += '|1=' + params.doubleRedirectTarget;
 			}
 			tagText += '}}';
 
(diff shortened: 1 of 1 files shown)


modules/friendlytag.js
```diff
@@ -1671,6 +1680,9 @@ Twinkle.tag.callbacks = {
 				if (params.altLangTo) {
 					tagText += '|to=' + params.altLangTo;
 				}
+				if (params.doubleRedirectTarget) {
+					tagText += '|1=' + params.doubleRedirectTarget;
+				}
 			}
```
Amorymeltzer, line 1686: This is an indent too deep; right now it's under `(tagName === 'R from alternative language')`, which obviously won't work!

What should happen is something like:

```suggestion
				if (params.altLangTo) {
					tagText += '|to=' + params.altLangTo;
				}
			} else if (tagName === 'R avoided double redirect' && params.doubleRedirectTarget) {
				tagText += '|1=' + params.doubleRedirectTarget;
			}
```

Amorymeltzer (changes requested):
Needs a quick bug fix, but otherwise nice PR @TheTVExpert!

TheTVExpert:
Fixed.

Amorymeltzer (approved):
Looks great, nice addition!  I wonder how many others there are that might benefit from this kind of thing?

Amorymeltzer:
Merged as 2c31888

PR #1169, opened 2020-10-23
tag: fix bugs in processing of existing tags

siddharthvp:
1: For tags within {{multiple issues}}, check if the class name begins with "box-", don't just assume that it does.

2: For tags already present on the page (and not unchecked for removal), `Twinkle.tag.article.flatObject[tag]` would be undefined, and when we do `Twinkle.tag.article.flatObject[tag].excludeMI`, the result is `*_*`

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1559,7 +1559,8 @@ Twinkle.tag.callbacks = {
 
 		// To-be-retained existing tags that are groupable
 		params.tagsToRemain.forEach(function(tag) {
-			if (!Twinkle.tag.article.flatObject[tag].excludeMI) {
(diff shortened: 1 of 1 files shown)


Amorymeltzer (approved):
Awful bug, good find!

Amorymeltzer:
Not important for this, but for the first commit, is there a way to just use `className` throughout, ideally without getting overly complicated?  I don't know if they're ever different, but in theory couldn't we stop relying on the `box-` tag being first by doing `var index = className.indexOf('box-')`, `if (index !== -1`, and so on?  I suppose it'd mean using regex on the string to get the name, which probably makes this all more complicated/slower.  Or, I guess, ,ale an actual array out of the classList, but that's (also?) overkill?

PR #1170, opened 2020-10-23
tag: Add support for existing rcats

TheTVExpert:
When tagging a redirect, show existing rcats at the top of the form and remove them when unchecked.


siddharthvp:
Very nice attempt, but can I suggest a different approach? [Module:Redirect template](https://en.wikipedia.org/wiki/Module:Redirect_template) supports a |class= parameter (which I implemented about two years ago with this very usecase in mind, at the same time as I added the `rcat` class which I see you've used) which can be used to cleanly make the rcats machine-readable. That would be better than the hacks you've used here for detecting the existing tags, and would greatly help with portability to other wikis.

siddharthvp:
Well now that I go back and look at it, I see that |class= of Module:Redirect_template doesn't work the same way as |name= of Module:Message_box (which we use for untagging of articles) – there's no prefix to guard from reading some other random class.

So I just went ahead and edited the module so that it now has an |id= param which would contain "rcat-{{{id}}}".

What we'd need to do is to edit every rcat template to add `|id={{subst:PAGENAME}}`, so that the templates would then carry classes like `rcat-R_with_history` for {{R with history}}, and likewise for others.

siddharthvp:
> What we'd need to do is to edit every rcat template to add |id={{subst:PAGENAME}}

Done! All ~285 rcat templates using AWB. It was a lot easier than I thought coz all those templates had identical wikitext formatting ❤. Now we can enable untagging of any rcat template, not just the ones that can be added by Twinkle.

It's also worth writing a bot to periodically check these templates (and the ones for articles) to ensure the id param still matches the template name.

TheTVExpert:
Done.

siddharthvp:
I implemented the feature on [twinkle-core ](https://github.com/wikimedia-gadgets/twinkle-core/blob/master/src/tag.ts) / [twinkle-enwiki](https://github.com/wikimedia-gadgets/twinkle-enwiki/blob/master/src/tag.ts), centralising the tagging code for both article & redirect modes (can also be easily extended to file tags). Not sure if this still needs to be kept open here.

Amorymeltzer (changes requested):
Apologies for not getting to this earlier @TheTVExpert, I thought it was being worked on.  I'm glad to see this, but it does still need some work.  The edit summary seems to be a bit messed up, but more pressingly, there's an error on at least some template, such as {{R from acronym}} (try https://en.wikipedia.org/w/index.php?title=7mm_RSAUM&redirect=no for an example).

I can take a deeper look and try and help diagnose some of the issues if you'd like, just lemme know.

siddharthvp:
The code for untagging is deeply convoluted. Even more so for the article untagging part which is really really bad (I wrote that 2 years - certainly the last thing to be proud about). It would be helpful to structure things properly and unify the codes for articles/redirects. But as you can probably see that would be equivalent to tearing it all down and rewriting it from scratch. That's what I had to do as [linked above](https://github.com/wikimedia-gadgets/twinkle/pull/1170#issuecomment-757461565). Re-doing the whole thing here seems like a waste of resources as it has been already done there. Certainly not opposed to it if @TheTVExpert wants to fix it up and have this pushed now, but I think this can wait until twinkle on enwiki is migrated to use twinkle-core/twinkle-enwiki instead.

TheTVExpert:
@Amorymeltzer the issues should now be fixed. @siddharthvp it wasn't too difficult to fix, however I do agree that the code needs improvements.

modules/friendlytag.js
```diff
@@ -192,8 +235,12 @@ Twinkle.tag.callback = function friendlytagCallback() {
...
 						name: 'tags',
 						list: subgroup.map(function (item) {
-							return { value: item.tag, label: '{{' + item.tag + '}}: ' + item.description, subgroup: item.subgroup };
-						})
+							returnValue = undefined;
+							if (redirectTagsPresent.indexOf(item.tag) === -1) {
+								returnValue = { value: item.tag, label: '{{' + item.tag + '}}: ' + item.description, subgroup: item.subgroup };
+							}
+							return returnValue;
+						}).filter(removeUndefined => removeUndefined !== undefined)
```
Amorymeltzer, line 243: No ES6 means no arrow functions allowed
```suggestion
						}).filter(function(removeUndefined) {
							return removeUndefined !== undefined;
```

modules/friendlytag.js
```diff
@@ -1711,18 +1761,32 @@ Twinkle.tag.callbacks = {
...
+		if (rcatsToAdd.length && rcatsToRemove.length) {
+			summaryText += ' & ';
+		}
+		rcatsToRemove.forEach(function removeTag(tag) {
+			var tag_re = new RegExp('\\{\\{' + Morebits.pageNameRegex(tag) + '\\s*(\\|[^}]+)?\\}\\}\\n?');
+
+			if (tag_re.test(pageText)) {
+				pageText = pageText.replace(tag_re, '');
+			}
+			summaryText += 'Removed {{[[:' + (tag.indexOf(':') !== -1 ? tag : 'Template:' + tag + '|' + tag) + ']]}}';
```
Amorymeltzer, line 1775: This should probably be sentence-like, as with article tags, with "Added ... and removed ..." rather than "Added... & Removed".  See line ~1310.  As it now stands, aside from that it adds "Removed" after every redirect, rather than listing them.

modules/friendlytag.js
```diff
@@ -1711,18 +1761,32 @@ Twinkle.tag.callbacks = {
...
 
 		// Check for all Rcat shell redirects (from #433)
 		if (pageText.match(/{{(?:redr|this is a redirect|r(?:edirect)?(?:.?cat.*)?[ _]?sh)/i)) {
 			// Regex inspired by [[User:Kephir/gadgets/sagittarius.js]] ([[Special:PermaLink/831402893]])
 			var oldTags = pageText.match(/(\s*{{[A-Za-z\s]+\|(?:\s*1=)?)((?:[^|{}]*|{{[^}]*}})+)(}})\s*/i);
-			pageText = pageText.replace(oldTags[0], oldTags[1] + tagText + oldTags[2] + oldTags[3]);
+			if (rcatsToAdd.length) {
+				pageText = pageText.replace(oldTags[0], oldTags[1] + tagText + oldTags[2] + oldTags[3]);
+			} else {
+				pageText = pageText.replace(oldTags[0], '');
```
Amorymeltzer, line 1788: This just blindly removes all tags if we're not adding any, right?  That's not what we want!  I think the goal here was to remove the rcat shell if there were no tags left (which, heh, I guess this technically does...) but that should be handled separately.

modules/friendlytag.js
```diff
@@ -1995,7 +2060,9 @@ Twinkle.tag.callback.evaluate = function friendlytagCallbackEvaluate(e) {
 
 	// File/redirect: return if no tags selected
 	// Article: return if no tag is selected and no already present tag is deselected
```
Amorymeltzer, line 2062:
```suggestion
	// File: return if no tags selected
	// Article/redirect: return if no tag is selected and no already present tag is deselected
```

Amorymeltzer (changes requested):
Some more bugs noted.

This is indeed a rather complicated scheme, but I think it's fine to do now as long as it works.  Ideally, one of two things would happen:

1. The functions from `article` mode would be pulled out and genericized, and used here.
2. The functions from `article` mode would be largely duplicated within `redirect` mode.

The latter is easier for the time being, and probably makes the former easier later.  Either way, it's a lot of work.  Then there are things like getting the `statusNode` to work (currently at the end of `Twinkle.tag.updateSortOrder`).  This looks well on its way, so @TheTVExpert if you want to keep at it, by all means go ahead.

NovemLinguae:
Mentioned today by Qwerfjkl at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Deselect_rcats

This kind of thing might be a good candidate for unit tests and a function that accepts string wikicode and array tags, and outputs string wikicode. Then we could write unit tests for each edge case mentioned in the ticket, and when all the unit tests are passing, we know the code is somewhat reliable.

PR #1174, opened 2020-10-30
morebits date: Refactor MW timestamps to ISO 8601

Amorymeltzer:
In addition to the comma and paren-wrapped UTC issues from #936, it turns out Safari's imeplementation of `Date.parse` is much more specific about what it accepts than Chrome and Firefox.  It's basically limited to the ECMAScript-262 standard of a simplified ISO-8601 standard, so even the cleaned up format won't work.  This obviously can cause issues, in particular with warn's autolevel selection, as reported at [WT:TW](https://en.wikipedia.org/w/index.php?oldid=986156173#Twinkle_auto-select_user_warning_level_not_working).

This could also be done with `.exec`:

```javascript
var result = /(\d\d:\d\d), (\d{1,2}) (.+) (\d{4}) \(UTC\)/.exec(args[0]);
args[0] = [result[4], mw.config.get('wgMonthNames').indexOf(result[3]), result[2]].join('-')+'T'+result[1]+'Z';
```

But I think the `.replace` is clearer to read.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1403,10 +1403,19 @@ Morebits.unbinder.getCallback = function UnbinderGetCallback(self) {
 Morebits.date = function() {
 	var args = Array.prototype.slice.call(arguments);
 
+	// Date.parse implementations vary too much between browsers, and
+	// MediaWiki's format is too non-standard, so we just convert MW
+	// timestamps to ISO-8601. A paren-wrapped 'UTC' messes everyone up,
+	// and the comma after the time is only okay in modern Firefox. After
+	// this first replace, Chrome and Firefox are content. The second
+	// replace is mainly for Safari, which basically *only* accepts the
+	// simplified ECMA-262 implementation of ISO-8601.
 	if (typeof args[0] === 'string') {
-		// Attempt to remove a comma and paren-wrapped timezone, to get MediaWiki timestamps to parse
-		// Firefox (at least in 75) seems to be okay with the comma, though
 		args[0] = args[0].replace(/(\d\d:\d\d),/, '$1').replace(/\(UTC\)/, 'UTC');
+		// Safari is particular about timezone offsets, so this is intentionally specific
+		args[0] = args[0].replace(/(\d\d:\d\d) (\d{1,2}) ([A-z][a-z]+) (\d{4}) UTC$/, function(match, time, date, monthname, year) {
+			return [year, mw.config.get('wgMonthNames').indexOf(monthname), date].join('-') + 'T' + time + 'Z';
+		});
 	}
 	this._d = new (Function.prototype.bind.apply(Date, [Date].concat(args)));
 };


morebits.js
```diff
@@ -1403,10 +1403,19 @@ Morebits.unbinder.getCallback = function UnbinderGetCallback(self) {
...
+	// MediaWiki's format is too non-standard, so we just convert MW
+	// timestamps to ISO-8601. A paren-wrapped 'UTC' messes everyone up,
+	// and the comma after the time is only okay in modern Firefox. After
+	// this first replace, Chrome and Firefox are content. The second
+	// replace is mainly for Safari, which basically *only* accepts the
+	// simplified ECMA-262 implementation of ISO-8601.
 	if (typeof args[0] === 'string') {
-		// Attempt to remove a comma and paren-wrapped timezone, to get MediaWiki timestamps to parse
-		// Firefox (at least in 75) seems to be okay with the comma, though
 		args[0] = args[0].replace(/(\d\d:\d\d),/, '$1').replace(/\(UTC\)/, 'UTC');
```
Amorymeltzer, line 1414: Down the line, we could attempt to account for a timezone offset here:
```javascript
args[0] = args[0].replace(/(\d\d:\d\d),/, '$1').replace(/\(UTC(?:[+-]\d+)\)/, '$1');
```

But that's overkill, since mw doesn't use those

morebits.js
```diff
@@ -1403,10 +1403,19 @@ Morebits.unbinder.getCallback = function UnbinderGetCallback(self) {
...
+		// Safari is particular about timezone offsets, so this is intentionally specific
+		args[0] = args[0].replace(/(\d\d:\d\d) (\d{1,2}) ([A-z][a-z]+) (\d{4}) UTC$/, function(match, time, date, monthname, year) {
```
Amorymeltzer, line 1416: Not really relevant here, but Safari also treats absence of trailing `Z` as UTC, which is wrong?

Amorymeltzer:
Merging/pushing now

morebits.js
siddharthvp, line 1416: @Amorymeltzer it should have been `[A-Z]` not `[A-z]`.

Amorymeltzer: Thanks!  Done in 451b991, although I really hope nobody is providing `\ctober`!

morebits.js
```diff
@@ -1403,10 +1403,19 @@ Morebits.unbinder.getCallback = function UnbinderGetCallback(self) {
...
+			return [year, mw.config.get('wgMonthNames').indexOf(monthname), date].join('-') + 'T' + time + 'Z';
```
Amorymeltzer, line 1417: [Revenge of the join](https://github.com/azatoth/twinkle/pull/814#discussion_r379869058)

PR #1176, opened 2020-10-30
block: Favor current block information if present

Amorymeltzer:
Before 91ccacf, the block form would inconsistently and rarely load the prior options; I know folks were occasionally confused as to what was happening.  That commit made it never do so, as a measure of consistency.  This commit adds the prior block conditions as a new preset, which is the default selection upon form loading or action toggling.  This should restore the behavior but make it more explicit when the information is present, and more importantly allow the sysop to reliably *go back* to the prior block settins.

I've also toggled the initial loading of the form to account for partial status, although the preset itself is dynamically adjusted to account for `{{uw-block}}`, `{{uw-pblockindex}}`, etc. as appropriate.  Additionally, I've explicitly added `bkprops` so we can preload any partial block restrictions, as those were missing.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -60,7 +60,7 @@ Twinkle.block.callback = function twinkleblockCallback() {
 				label: 'Partial block',
 				value: 'partial',
 				tooltip: 'Enable partial blocks and partial block templates.',
-				checked: Twinkle.getPref('defaultToPartialBlocks')
+				checked: Twinkle.getPref('defaultToPartialBlocks') // Overridden if already blocked
 			},
 			{
 				label: 'Add block template to user talk page',
@@ -83,6 +83,12 @@ Twinkle.block.callback = function twinkleblockCallback() {
 	result.root = result;
 
 	Twinkle.block.fetchUserInfo(function() {
+		// Toggle initial partial state depending on prior block type,
+		// will override the defaultToPartialBlocks pref
+		if (Twinkle.block.currentBlockInfo) {
+			$(result).find('[name=actiontype][value=partial]').prop('checked', Twinkle.block.currentBlockInfo.partial === '');
+		}
+
 		// clean up preset data (defaults, etc.), done exactly once, must be before Twinkle.block.callback.change_action is called
 		Twinkle.block.transformBlockPresets();
 		if (Twinkle.block.currentBlockInfo) {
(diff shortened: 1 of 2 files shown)


modules/twinkleblock.js
```diff
@@ -163,6 +170,33 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 
 	partial.prop('disabled', !blockBox && !templateBox);
 
+	// Add current block parameters as default preset
+	if (Twinkle.block.currentBlockInfo) {
```
Amorymeltzer, line 174: This could be changed if we only wanted the original settings to be available on the same type (partial or not) of block.  I don't think that's better, but something like `((partialBox && Twinkle.block.currentBlockInfo.partial === '') || (!partialBox && Twinkle.block.currentBlockInfo.partial !== ''))` would do it.

modules/twinkleblock.js
```diff
@@ -163,6 +170,33 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
...
+			label: 'Prior block',
+			// value not a valid template selection, chosen below by setting templateName
+			list: [{ label: 'Prior block settings', value: 'prior', selected: true }]
+		};
+		// Arrays of objects are annoying to check
+		if (!blockGroup.some(function(bg) {
+			return bg.label === prior.label;
+		})) {
+			blockGroup.push(prior);
+		}
```
Amorymeltzer, line 186: Could also just add the `prior` group to both right off the bat if there's a current block.  Probably a net efficiency gain if there's any toggling going on.

modules/twinkleblock.js
```diff
@@ -1494,6 +1529,40 @@ Twinkle.block.callback.update_form = function twinkleblockCallbackUpdateForm(e,
...
+		// Add any preset options; in practice, just used for prior block settings
+		if (data.restrictions) {
+			if (data.restrictions.pages && !$pageSelect.val().length) {
+				var pages = data.restrictions.pages.map(function(pr) {
+					return pr.title;
+				});
+				// since page restrictions use an ajax source, we
+				// short-circuit that and just add a new option
+				pages.forEach(function(page) {
+					if (!$pageSelect.find("option[value='" + page + "']").length) {
```
siddharthvp, line 1554:
```suggestion
					if (!$pageSelect.find("option[value='" + $.escapeSelector(page) + "']").length) {
```

Amorymeltzer: I couldn't trigger any issues with any titles here, and still can't.  Maybe because these are values already-returned by the API?  Maybe because of select2?

siddharthvp: It would fail with a page whose title contains a `'`, like `Occam's razor`.

Amorymeltzer: Gah, of course!  I don't know why, I always parse `$.escapeSelector` as just "Escape things you select that might cause issues later" not "Escape things you select that will mess up your actual selection."  Done and thanks!

PR #1177, opened 2020-10-30
Morebits date: Add timeBetween function for human-readable time difference

Amorymeltzer:
The time ranges are taken from moment.js' [fromnow](momentjs.com/docs/#/displaying/fromnow).  I've thought about this a bit, but it didn't really seem worth it without a use-case.  Still not sure it's necessary, but block can make use of it (#1172).

In theory, I think this could interplay with `calendar`, but I haven't thought of a net-gain way.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -1757,9 +1757,14 @@ Twinkle.block.callback.getBlockNoticeWikitext = function(params) {
 		if (!/te?mp|^\s*$|min/.exec(params.expiry)) {
 			if (params.indefinite) {
 				text += '|indef=yes';
-			} else if (!params.blank_duration && !new Morebits.date(params.expiry).isValid()) {
+			} else if (!params.blank_duration) {
 				// Block template wants a duration, not date
-				text += '|time=' + params.expiry;
+				var date = new Morebits.date(params.expiry);
+				if (date.isValid()) {
+					text += '|time=' + date.timeBetween(Date.now());
+				} else {
+					text += '|time=' + params.expiry;
+				}
 			}
 		}
 
--- a/morebits.js
+++ b/morebits.js
@@ -1432,6 +1432,14 @@ Morebits.date.localeData = {
 		thisWeek: 'dddd [at] h:mm A',
 		pastWeek: '[Last] dddd [at] h:mm A',
 		other: 'YYYY-MM-DD'
+	},
+	units: {
+		second: 'second',
+		minute: 'minute',
+		hour: 'hour',
+		day: 'day',
+		month: 'month',
+		year: 'year'
 	}
 };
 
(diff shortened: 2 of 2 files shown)


siddharthvp:
Why not just add a new parameter to the block notice template that takes the date string rather than the duration? That would be more "accurate" as the user is blocked _till a date_, not _for a duration_.

This seems like too much of clutter for little purpose and the i18n scheme doesn't make sense.

Amorymeltzer:
> Why not just add a new parameter to the block notice template that takes the date string rather than the duration? That would be more "accurate" as the user is blocked _till a date_, not _for a duration_.

Well, for starters, I think it's better to include a little helper function in one place rather than attempt to adjust the block meta templates just for the sake of this relatively minor Twinkle aspect.  Easier to maintain, not imposing our whim, etc.

But more than that, "the user is blocked _till a date_, not _for a duration_" isn't really right.  Sure, as far as the API is concerned they are, but that's because the backend can unwind a duration.  As far as Wiki*edia editors are concerned, we protect pages and block users for a duration, not until a date, both in common parlance and in nearly all drop-down menus, including our own.  These templates are a good example of that.

>...the i18n scheme doesn't make sense.

I tossed it in as a nod to your previous efforts in `localeData` rather than hardcoding, but I wasn't thrilled with it and I'm certainly not wed to it; what would you suggest to improve it?

siddharthvp:
> "the user is blocked till a date, not for a duration" isn't really right

As in: in the specific case of the API being passed an absolute date (via TW or otherwise), the block log shows shows the absolute date rather than a duration. No need to go into what the API does on the backed. So giving this date to the template and letting the template do what it wants is IMO better than concocting a duration of our own. We've got to use the _right_ tools to solve problems, and here the right tool is the template.

> Easier to maintain, not imposing our whim, etc.

Oh you can't be serious! It isn't imposing our whim to implement an optional enhancement to the template that doesn't affect existing usages. TW relies heavily on templates though we take it for granted on enwiki – that's what makes things easier to maintain.

Amorymeltzer:
When people on-wiki speak, they speak about being blocked for 2 weeks or a page being protected for 3 months.  The templates even moreso, since they are explicitly about communicating.  I don't think we disagree about that, that the templates should say "you have been blocked for 31 hours," I think we're just disagreeing on where/how to make the change.

If the templates already had support for it, great, but the fact on the ground is that they don't.  Adding a parameter to every block template isn't trivial: there are a couple of meta templates, but not every template uses them.  And that's ignoring any processing of the value.  I don't see how that's simpler or easier than adapting to what has been the standard/status quo for over a decade.

siddharthvp:
Sigh. What if something like "20 November" is entered – that would seem perfectly reasonable to the user and is parsed correctly by the PHP-based API, but JS Date() parses it as 20 Nov 2001(!), so the template would say "you've been blocked for 19 years". Which is why I advocate to pass the unprocessed date to the template. The template can do something with it using #time parser function which would parse them the exactly the same way as the API (they share the same backend logic).

Amorymeltzer:
That's a fair point about `#time` handling everything much easier, but I still figure it's a bunch of complication to templates that aren't screaming out for it.

At any rate, your objection is reasonable, so I'm fine with this being closed, though I continue to think there's utility, for example with something like #1183.

Issue #1144, opened 2020-10-09
i18n/l10n roadmap


siddharthvp:
### Across modules:

- Abstract the code that extracts the filled-in form data into a class or function (defined in Morebits). In some modules (like xfd), we may want to allows per-wiki configurations that add new input fields. (#908)
- Use HTML5 required field attribute for validating required fields in most places. This does away the need for the `alert`s, which would be one less thing to localise.
- Document Twinkle's dependencies on wiki templates
- Extend use of `Morebits.taskManager`, or another promise-based system, across most modules.
- Use object-oriented pattern in the more complex modules like Tag, XFD and CSD to make them extensible more easily.
- Set up a build step for transpiling ES2020 code. This makes working with classes a lot easier. 
- Explore use of unit testing in module code, probably based on SinonJS and/or web driver tools, which would significantly reduce the time and effort needed for testing. Integrate unit tests in Travis.
- Explore the use of translatewiki.net for translation of strings. 

#### Module components

What do Twinkle modules have in common? They are all tools with a GUI containing a form in which the user can enter some data. On submitting, a series of actions are executed using the API. (Fluff is an exception -- it's completely different. But fluff is not project specific, and can be readily i18n-ed. So that is the least of our concerns.)

- UI
  - `Morebits.quickForm`, `Morebits.simpleWindow` are used for creating the UI. No changes are planned here.
- Collecting form data
  - With #908, we have a structured way of extracting form data in one line of code. No further changes planned here.
- Actions
  - `Morebits.wiki.api` and `Morebits.wiki.page` are used for the bulk of actions code. To overcome the issues mentioned in #1029 and #911, I plan to rewrite these two in modern JS with unit tests and documentation. I believe having a rich and well-documented library of reusable components is one of the best ways to enable l10n of code involving processes that are wildly different across wikis.
  - Orchestrating the actions in a proper sequence, handling failures, etc is something Twinkle does the worst, leading to bugs like #726, [this](https://en.wikipedia.org/w/index.php?oldid=980448010&diff=980928378&diffmode=source#XFD) and [this](https://en.wikipedia.org/w/index.php?oldid=981569889&diff=981605445&diffmode=source#XfD_feature_issue) (also present for speedy and other modules). I think the solution is a promise-based system like `Morebits.taskManager` (already implemented PR #1036, but not merged yet), which needs to be used throughout.
- Showing status messages
  - Uses `Morebits.status`, integrated into Morebits.wiki.page and Morebits.wiki.api
[...]

PR #1179, opened 2020-10-31
morebits: Use new error formats (default to html) in api; use $.parseHTML in status messages

Amorymeltzer:
2nd commit uses/enforces the [newer API error messages](mediawiki.org/wiki/API:Errors_and_warnings) in `Morebits.wiki.api`, defaulting to `html`, but allowing `plaintext` and `wikitext`; `raw`, `none`, and the legacy `bc` will be converted to `html`.  Provides `errorlang: mw.config.get('wgUserLanguage')`.  Errors are limited to the first found, which was the case with the old syntax and is assumed by our uses of `errorCode`/`errorText`, but maybe something to refactor at some point (not a huge gain IMO).

The 1st commit enables the html showing up nicely, by using `$.parseHTML` in `Morebits.status.codify()`.  There's a whole long commit message, but the tl;dr is: it will now automatically parse html, not wikitext; and there's now a `statRaw` for error logging.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -4645,6 +4645,7 @@ Morebits.status.onError = function(handler) {
 
 Morebits.status.prototype = {
 	stat: null,
+	statRaw: null,
 	text: null,
 	textRaw: null,
 	type: 'status',
@@ -4669,7 +4670,9 @@ Morebits.status.prototype = {
 	},
 
 	/**
-	 * Create a document fragment with the status text.
+	 * Create a document fragment with the status text, parsing as HTML.
+	 * Runs upon construction for text (part before colon) and upon
+	 * render/update for status (part after colon).
 	 *
 	 * @param {(string|Element|Array)} obj
 	 * @returns {DocumentFragment}
@@ -4681,11 +4684,13 @@ Morebits.status.prototype = {
 		var result;
 		result = document.createDocumentFragment();
 		for (var i = 0; i < obj.length; ++i) {
-			if (typeof obj[i] === 'string') {
-				result.appendChild(document.createTextNode(obj[i]));
-			} else if (obj[i] instanceof Element) {
+			if (obj[i] instanceof Element) {
 				result.appendChild(obj[i]);
-			} // Else cosmic radiation made something shit
+			} else {
+				$.parseHTML(obj[i]).forEach(function(elem) {
+					result.appendChild(elem);
+				});
+			}
 		}
 		return result;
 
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
Presumably there are places that could be simplified by making use of the 1st commit, but I've not done a survey and might as well leave well-enough alone atm.  But block and protect for sure.

siddharthvp:
Well, this was part of the plan in #1144, though I was more of expecting to rewrite Morebits.status with fancy StatusLine and StatusBlock classes.

I would bat for `uselang: 'content'`, `errorlang: 'content'` (same effect as specifying `mw.config.get('wgContentLanguage'`) as everything in Twinkle is in the content language, and `errorsuselocal: true`.

siddharthvp:
> One more note: because morebits.css has
`
form.quickform *
{
	font-family: sans-serif;
}
`
any status elements that would be expected to affect `font-family` won't work properly.  In particular, things like `<pre>`, `<kbd>`, and `<code>` won't display with `font-family: monospace`, as may be desired/expected, until/unless morebits.css is changed.

That CSS should be removed. sans-serif is I think the default font on mediawiki, so there's no need to specify it. Even if it were different for someone, I don't see why it needs to be adjusted away from that choice.

Amorymeltzer:
>I would bat for `uselang: 'content'`, `errorlang: 'content'` (same effect as specifying `mw.config.get('wgContentLanguage'`) as everything in Twinkle is in the content language, and `errorsuselocal: true`.

I can never keep track of these!  But, this is for messages being passed to the Twinkle user, not ever being added to the wiki.  The content in the Twinkle interface is... well, it's all `en` now, but in theory it should show the user 'user' and edit/act with 'content', no?  `user` is the default for `uselang`, but `errorlang` needs to be set explicitly (the default to match `uselang` was... weirdly nonfunctional for me?).

>That CSS should be removed. sans-serif is I think the default font on mediawiki, so there's no need to specify it. Even if it were different for someone, I don't see why it needs to be adjusted away from that choice.

Jah, but figured OoS for this.  IIRC from minor testing, it was basically just the select menus that would/could be affected: without this, they'd inherit from the skin, so Modern/Timeless would look different (slightly bigger).  Don't think it's a big deal, but didn't look too much into it.

siddharthvp:
> but in theory it should show the user 'user' and edit/act with 'content', no?

In theory yes, but in practise, that's hopelessly complicated to work out. There could be situations where one could want to have a tooltip say "for xxx purposes use {{other template name}} instead". Here the "other template name" obviously has to be based on the content language, but the message itself should be in user language? You see, using one language for the UI and another for editing is a lost cause. My i18n plans (assuming they approve the grant) have been to use content lang for everything.

Amorymeltzer:
That's a real good point — ugh!  Done and done!

PR #1180, opened 2020-10-31
action: Add PR labeler action based on which file they touch

Amorymeltzer:
https://github.com/marketplace/actions/labeler

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -1,0 +1,49 @@
+"Module: twinkle":
+  - twinkle.js
+  - twinkle.css
+  - twinkle-pagestyles.css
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
*sigh* I had hoped this would be smooth, sorry to keep bothering you @azatoth.  I'm not entirely sure what the issue is, is there a chance it's being prevented from running by the [repo settings](https://github.com/azatoth/twinkle/settings/actions)?

Amorymeltzer:
Reverted (c695df4) as apparently it'll only work if made from a branch on the repo, not forks (see, e.g. https://github.com/actions/first-interaction/issues/10)

Issue #1065, opened 2020-07-25
Treat redirects with RfD template as redirects


Xi-Plus:
If there is RfD template on redirects, Twinkle treats it as article (Identified by `mw.config.get('wgIsRedirect')`).
It is recommended to also check RfD template or "Category:Redirects for discussion from XXX" category to identify redirects.
Sample page: https://en.wikipedia.org/w/index.php?title=Death_of_Amir_Tehkal&oldid=969331860

Amorymeltzer:
This is indeed a major complaint of the speedy module, but I'm not entirely sure what the best bet is.  The RfD nomination doesn't have any convenient classes or ids; `twinkleprod` checks the tex for `#invoke:RfD`, which is not ideal here.  Maybe we could modify the module itself?

Amorymeltzer:
Hmm, for items that have an rcat we can check for `.rcat`, so we could use that in an imperfect world.

@siddharthvp You like adding classes to templates/modules, what do you think about adding one to [[Module:RfD]] and/or {{Rcat shell}}?  I can't imagine folks would complain about an `|id` in the latter, although I think Module:RfD would be sufficient?

siddharthvp:
@Amorymeltzer {{Rcat shell}} already had `.box-Redirect_category_shell`. For Module:RfD, I just [added](https://en.wikipedia.org/w/index.php?title=Module:RfD&diff=986499456&oldid=954817046) it (`.box-RfD`)

I was initially thinking we could just use `wgCategories` as Xi-Plus implied, but I'm surprised to see that wgCategories doesn't work while editing or viewing history (why does wgCategories exist at all then? Because while viewing the page we could just scrape #catlinks in the DOM!). So, at the end of the day, detecting neither the template nor category works in edit/history mode.

Amorymeltzer:
Mmm, yes indeed.  And thanks.  And I have no earthly idea why `wgCategories` doesn't.  History is the one I was concerned about, that came up with XfD in #863 (mainly around soft redirects); not sure there's a way back from that.

Amorymeltzer:
Opened #1182; not sure that I think `.rcat` is necessary, nor parsing `#wpTextbox1`, but open to discussion.  As noted there, XfD is left behind 😢

PR #1182, opened 2020-11-01
Detect RfD nomination in Morebits.wiki.isPageRedirect, use in CSD

Amorymeltzer:
It's been long desired to show the redirect CSDs on redirects nominated at RfD.  This adds detection of that box to `Morebits.wiki.isPageRedirect` and uses it in twinklespeedy.  Should also improve things elsewhere that it's used, in particular redirect tagging in the tag module.  XfD will have problems with it, namely finding the redirect target for RfD.

----

As noted in https://github.com/azatoth/twinkle/issues/1065#issuecomment-720045907, this will only work when `view`ing the page.  If we wanted to include editing, we'd have to use the page content, doing something like `document.getElementById('wpTextbox1').value.indexOf('#invoke:RfD') !== -1` or `/#redirect/i.test(document.getElementById('wpTextbox1').value)` or something.

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -281,7 +281,7 @@ Twinkle.speedy.callback.modeChanged = function twinklespeedyCallbackModeChanged(
 		work_area.append({ type: radioOrCheckbox, name: 'csd', list: Twinkle.speedy.generateCsdList(Twinkle.speedy.talkList, mode) });
 	}
 
-	if (!mw.config.get('wgIsRedirect')) {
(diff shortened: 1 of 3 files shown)


siddharthvp:
> document.getElementById('wpTextbox1').value

Would have to be `$('#wpTextbox1').textSelection('getContents')` for compatibility with different editors including 2017 wikitext editor.

Amorymeltzer:
Do you think this would be a sufficient start, @Xi-Plus?

Xi-Plus:
Yes. It's great!

PR #1183, opened 2020-11-01
block/protect: Align current/log displays, show expiry/timestamp of last action

Amorymeltzer:
1st commit:
- block: Show block log under current block info, tweak messages
- protect: Use `Morebits.date.calendar` to format expiry

2nd commit: Show timestamp for an unblock or unprotection action, and expiry for a block or protection, but only if the user/page is not currently blocked/protected.  For protection it's not particularly smart about finding the last expiry, it just grabs the first level, which defaults to edit.  That means that if the last action was to adjust the edit protection to expire before move protection expired, this would show the wrong one.

I'm wary of getting too cluttered here.

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -282,7 +282,7 @@ Twinkle.protect.callback.showLogAndCurrentProtectInfo = function twinkleprotectC
 			if (settings.expiry === 'infinity') {
 				protectionNode.push(' (indefinite) ');
 			} else {
-				protectionNode.push(' (expires ' + new Date(settings.expiry).toUTCString() + ') ');
+				protectionNode.push(' (expires ' + new Morebits.date(settings.expiry).calendar('utc') + ') ');
 			}
 			if (settings.admin) {
 				var adminLink = '<a target="_blank" href="' + mw.util.getUrl('User talk:' + settings.admin) + '">' + settings.admin + '</a>';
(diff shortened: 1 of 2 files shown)


modules/twinkleprotect.js
```diff
@@ -258,7 +258,7 @@ Twinkle.protect.callback.showLogAndCurrentProtectInfo = function twinkleprotectC
 			if (settings.expiry === 'infinity') {
 				protectionNode.push(' (indefinite) ');
 			} else {
-				protectionNode.push(' (expires ' + new Date(settings.expiry).toUTCString() + ') ');
+				protectionNode.push(' (expires ' + new Morebits.date(settings.expiry).calendar('utc') + ') ');
```
Amorymeltzer, line 261: I've done `Morebits.date.calendar('utc')` to match what's present in the block display and morebits `confirms`s, but I actually think that's not necessarily great, mainly because it doesn't *indicate* the UTC-ness and can be confusing for upcoming expirations.

```suggestion
				protectionNode.push(' (expires ' + new Morebits.date(settings.expiry).calendar() + ') ');
```

might be better, or even something with `.format`

Amorymeltzer:
*sigh* Only just realized this, but the expiry format in `letype=stable` for `modify` or `config` actions is inscrutable to js' date.  T267922 but that'll hold up the second commit, so I'll have to split this.

Amorymeltzer:
- First commit merged as d2a8863

- Second commit in draft PR #1194

PR #1185, opened 2020-11-04
warn: Add uw-editsummary2

TheTVExpert:
Add per https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&diff=987082046&oldid=987052409

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -931,7 +931,7 @@ Twinkle.warn.messages = {
 			summary: 'Notice: Consider drafting your article in [[Help:Userspace draft|userspace]]'
 		},
 		'uw-editsummary': {
-			label: 'Not using edit summary',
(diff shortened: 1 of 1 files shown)


modules/twinklewarn.js
```diff
@@ -934,6 +934,10 @@ Twinkle.warn.messages = {
 			label: 'Not using edit summary',
```
Amorymeltzer, line 934:
```suggestion
			label: 'New user not using edit summary',
```

Maybe given the divide, wouldn't hurt to be explicit?

Amorymeltzer (approved):
One minor suggestion but I think this is a fine thing to add — thanks for the speedy response.

TheTVExpert:
Done

Amorymeltzer (approved):

PR #1188, opened 2020-11-09
morebits: zero-pad date in constructor

Amorymeltzer:
The regex added in #1174 would match the date number, but didn't provide the leading zero needed for ISO-8601 formats.  Reported at [WT:TW](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=987793556#Twinkle_auto-select_user_warning_level_not_working_(Again))


morebits.js
```diff
@@ -1414,6 +1414,10 @@ Morebits.date = function() {
 		args[0] = args[0].replace(/(\d\d:\d\d),/, '$1').replace(/\(UTC\)/, 'UTC');
 		// Safari is particular about timezone offsets, so this is intentionally specific
 		args[0] = args[0].replace(/(\d\d:\d\d) (\d{1,2}) ([A-Z][a-z]+) (\d{4}) UTC$/, function(match, time, date, monthname, year) {
+			// zero-pad date
+			if (date < 10) {
```
Amorymeltzer, line 1418: I did this to match `pad` below, but it's moderately wrong, and will break on correctly-padded dates (e.g. `07`); fixed in c0f5591.

PR #1189, opened 2020-11-10
prod: add nomreason to old prod

TheTVExpert:
Per https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&diff=988006985&oldid=987953219

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -354,7 +354,7 @@ Twinkle.prod.callbacks = {
 		}
 
 		// Add {{Old prod}} to the talk page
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Noting @siddharthvp's valid points on-wiki, this would leave BLPPRODs blank, yeah?  An empty `reason` is fine for the log since it's already noted, but for this it'd be no good and something should be filled in.

TheTVExpert:
Once there is consensus on-wiki, I'll alter this PR to reflect the result.

TheTVExpert:
@siddharthvp, since there was no opposition to your proposal, could you add your new parameters to the template

siddharthvp:
@TheTVExpert Sorry missed your message. I'll eventually look into editing the template, but would be grateful if someone else does it!

Marking this PR as a draft as it can't be merged as of now.

PR #1190, opened 2020-11-15
Expand and standardize Morebits jsdocs, export to html and markdown

Amorymeltzer:
Not perfect, but largely completes jsdoc documentation for the library, and does so in an exportable way (see first commit).  I've done that in two ways (2nd commit):

- HTML.  Ideally, the docs directory would be hosted via github pages (@azatoth); I've enabled it from my fork at https://amorymeltzer.github.io/twinkle/ if you want to see it.
- Markdown, via `jsdoc-to-markdown`.  I've already uploaded the results to https://github.com/azatoth/twinkle/wiki/morebits for perusal.

There are advantages to both methods, but I've found both to be helpful.  `npm run ghpages` and `npm run wiki`, or both with `npm run docs` (via `./dev/docs.sh` shim).

Another option is to *not* muddy the main branch with the output files, but host them in a separate, gh-pages branch, akin to the old-school style.

cc @MusikAnimal @siddharthvp

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1,62 +1,67 @@
 // <nowiki>
 /**
- * morebits.js
- * ===========
  * A library full of lots of goodness for user scripts on MediaWiki wikis, including Wikipedia.
  *
  * The highlights include:
- *   - Morebits.quickForm class - generates quick HTML forms on the fly
- *   - Morebits.wiki.api class - makes calls to the MediaWiki API
- *   - Morebits.wiki.page class - modifies pages on the wiki (edit, revert, delete, etc.)
- *   - Morebits.wikitext class - contains some utilities for dealing with wikitext
- *   - Morebits.status class - a rough-and-ready status message displayer, used by the Morebits.wiki classes
- *   - Morebits.simpleWindow class - a wrapper for jQuery UI Dialog with a custom look and extra features
+ * - {@link Morebits.wiki.api} - make calls to the MediaWiki API
+ * - {@link Morebits.wiki.page} - modify pages on the wiki (edit, revert, delete, etc.)
+ * - {@link Morebits.date} - enhanced date object processing, sort of a light moment.js
+ * - {@link Morebits.quickForm} - generate quick HTML forms on the fly
+ * - {@link Morebits.simpleWindow} - a wrapper for jQuery UI Dialog with a custom look and extra features
+ * - {@link Morebits.status} - a rough-and-ready status message displayer, used by the Morebits.wiki classes
+ * - {@link Morebits.wikitext} - utilities for dealing with wikitext
+ * - {@link Morebits.string} - utilities for manipulating strings
+ * - {@link Morebits.array} - utilities for manipulating arrays
  *
  * Dependencies:
- *   - The whole thing relies on jQuery.  But most wikis should provide this by default.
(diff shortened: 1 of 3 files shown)


morebits.js
```diff
@@ -1401,14 +1507,14 @@ Morebits.unbinder.getCallback = function UnbinderGetCallback(self) {
 
 
 
+/* **************** Morebits.date **************** */
```
Amorymeltzer, line 1510: I kept these non-standard headers for internal searching/jumping, but I don't actually think they're necessary to keep around.

MusikAnimal:
Looking good to me!

> Another option is to not muddy the main branch with the output files, but host them in a separate, gh-pages branch, akin to the old-school style.browse around in the code.

That would be my recommendation. You could even make the docs/ folder a submodule, pointing to a dedicated repo that has gh-pages. It could be updated on post-commit hook (though you'd still need to remember to push). But maybe that's over-complicating it. I just am not a fan of all that HTML showing up when I'm searching for code within GitHub, for instance. The important part, I think, is everything being documented in a consistent way _within_ the source code, such that IDEs know the data type of each function parameter, etc. I personally probably wouldn't ever refer to the generated docs.

Amorymeltzer:
@MusikAnimal That's probably my preference as well, and for exactly that reason, plus it won't muddy up the main branch with pointless diffs/commits(/conflicts even).  We can keep the infrastructure (package.json, configs, etc.) and add to .gitignore, but it'd probably best to just move everything there.  Much simpler to muck about with, too.

> I personally probably wouldn't ever refer to the generated docs.

My goal is mainly help with adoption of the library, and I thought this as well, but I've actually found myself using it this past month or two.

siddharthvp:
Looks pretty cool! Though I have thinking of transitioning to TypeScript which has more robust type annotation support. 

Regarding the hosting, can't we host it on readthedocs.io? azatoth.github.io/twinkle sounds a bit awkward.

Amorymeltzer:
>readthedocs.io

Wary of yet another account, especially before #1140 or https://github.com/azatoth/twinkle/issues/1140#issuecomment-703013774.  gh pages means anyone with push access can update.  And, meh, it's a link.

siddharthvp:
Not just that but readthedocs is purpose-built for documentation. I think we may not need to store the docs in the git repo at all – it can be built from source directly from readthedocs. 

That being said, I've no idea how to set it up for a JS project – the site is python-oriented.

Not a big deal either way.

> Wary of yet another account

You can sign in with github.

Amorymeltzer:
I meant more given the whole Amorymeltzer !== Azatoth thing...

Amorymeltzer:
Alright, I've updated this to just have the morebits stuff on the main branch, saving the rest for gh-pages.

Amorymeltzer:
@azatoth If you could configure the repo settings to serve GitHub pages from the `docs` folder on the `gh-pages` branch, that should be all that's needed here (no template required)!

azatoth:
> @azatoth If you could configure the repo settings to serve GitHub pages from the `docs` folder on the `gh-pages` branch, that should be all that's needed here (no template required)!

done

Amorymeltzer:
Fabulous, thanks all!  I'll keep 'em updated as things change.

PR #1191, opened 2020-11-15
tag: COI: Add area for automatically starting a discussion on the talkpage

Amorymeltzer:
[Old request](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle/Archive_42#COI)

----

Presumably others could warrant from this, in particular {{POV}}, but likely most of the "neutrality" section.  Should we be disfavoring both of those or some subset?  Probably better to ask on-wiki.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -716,7 +716,12 @@ Twinkle.tag.article.tagList = {
 		],
 		'Neutrality, bias, and factual accuracy': [
 			{ tag: 'Autobiography', description: 'autobiography and may not be written neutrally' },
-			{ tag: 'COI', description: 'creator or major contributor may have a conflict of interest' },
+			{ tag: 'COI', description: 'creator or major contributor may have a conflict of interest', subgroup: mw.config.get('wgNamespaceNumber') === 0 ? {
+				name: 'coiReason',
+				type: 'textarea',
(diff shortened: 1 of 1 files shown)


modules/friendlytag.js
```diff
@@ -1225,19 +1235,30 @@ Twinkle.tag.callbacks = {
 			pageobj.setMinorEdit(Twinkle.getPref('markTaggedPagesAsMinor'));
 			pageobj.setCreateOption('nocreate');
 			pageobj.save(function() {
-				// special functions for merge tags
+
+				// COI: Start the discussion on the talk page (mainspace only)
+				if (params.coiReason) {
+					var coiTalkPage = new Morebits.wiki.page('Talk:' + Morebits.pageNameNorm, 'Starting discussion on talk page');
+					coiTalkPage.setNewSectionText(params.coiReason + ' ~~~~');
+					coiTalkPage.setNewSectionTitle('COI tag {{subst:CURRENTMONTHNAME}} {{subst:CURRENTYEAR}}');
```
Amorymeltzer, line 1243: Would welcome suggestions for improvement.

siddharthvp: Put the date in brackets?

Amorymeltzer:
Now updated to take into account #1216

siddharthvp:
No need for a new function. You can just use a ternary.

PR #1192, opened 2020-11-15
warn: Don't italicize additional notes

Amorymeltzer:
This is long-standing here, and only a few folks have taken issue with it, but it *does* stand out.  Opened to start a discussion, since I'm not sure anyone feels strongly about it.  Can ask on-wiki.


siddharthvp:
I'm not sure if i can follow the logic.

> but it does stand out.

Yes, that's the whole point, right? User written text should stand out from the template boilerplate. 

Though I've always thought that a better way of making it stand out would be putting it in a new paragraph.

Amorymeltzer:
Their issue was that it's impossible *not* to add the italics.  I don't feel like digging up the link, but IIRC some were resorting to manually adding their comments later because they couldn't undo it.  I figured a preference was silly for something so minor, but hey, maybe not

PR #1193, opened 2020-11-15
xfd: Adjust portlet tooltip to be more accurate

Amorymeltzer:
This was [briefly discussed on WT:TW](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Move_requests_deeply_hidden), but the portlet tooltip of `Start a deletion discussion` is rarely, if ever, sufficiently accurate, and indeed elides and even hides functionality (namely, RM).  I've put the two ideas up here:

- 1st commit: `Start a discussion`
- 2nd commit: Be more specific for each venue (suggested by @siddharthvp, correct me if I misinterpreted)

I (clearly) favor the former as being simpler and not duplicating checks, but honestly the more detail from the second is probably nicer and likely to be more popular.  It'll be inaccurate if folks choose something else (or in the specific cases of things like userboxes, stubs, etc.) but by the time the module window is open, I don't think *anyone* cares what the tooltip was.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -21,7 +21,7 @@ Twinkle.xfd = function twinklexfd() {
 		return;
 	}
 
-	Twinkle.addPortletLink(Twinkle.xfd.callback, 'XFD', 'tw-xfd', 'Start a deletion discussion');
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
Went option 2

PR #1194, opened 2020-11-16
block/protect: Show expiry/timestamp for last action (if not currently restricted)

Amorymeltzer:
Shows timestamp for an unblock or unprotection action, and expiry for a block or protection, but only if the user/page is not currently blocked/protected.  For protection it's not particularly smart about finding the last expiry, it just grabs the first level, which defaults to edit.  That means that if the last action was to adjust the edit protection to expire before move protection expired, this would show the wrong one.

Split off from #1183 as protect expiry is held up on https://phabricator.wikimedia.org/T267922, I believe.

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -212,7 +212,9 @@ Twinkle.protect.fetchProtectionLevel = function twinkleprotectFetchProtectionLev
 
 		// show the protection level and log info
 		Twinkle.protect.hasProtectLog = !!protectData[0].query.logevents.length;
+		Twinkle.protect.protectLog = Twinkle.protect.hasProtectLog && protectData[0].query.logevents;
 		Twinkle.protect.hasStableLog = hasFlaggedRevs ? !!stableData[0].query.logevents.length : false;
+		Twinkle.protect.stableLog = Twinkle.protect.hasStableLog && stableData[0].query.logevents;
 		Twinkle.protect.currentProtectionLevels = current;
 
 		if (adminEditDeferred) {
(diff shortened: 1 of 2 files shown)


modules/twinkleprotect.js
```diff
@@ -240,13 +242,28 @@ Twinkle.protect.callback.showLogAndCurrentProtectInfo = function twinkleprotectC
...
 		}
 
 		if (Twinkle.protect.hasStableLog) {
 			$linkMarkup.append($('<a target="_blank" href="' + mw.util.getUrl('Special:Log', {action: 'view', page: mw.config.get('wgPageName'), type: 'stable'}) + '">pending changes log</a>)'));
+			if (!currentlyProtected || !Twinkle.protect.currentProtectionLevels.stabilize) {
+				var lastStabilizeAction = Twinkle.protect.stableLog[0];
+				if (lastStabilizeAction.action === 'reset') {
+					$linkMarkup.append(' (reset ' + new Morebits.date(lastStabilizeAction.timestamp).calendar('utc') + ')');
+				} else { // config or modify
+					$linkMarkup.append(' (expired ' + new Morebits.date(lastStabilizeAction.params.expiry).calendar('utc') + ')');
```
Amorymeltzer, line 264: [Right here](https://phabricator.wikimedia.org/T267922)

siddharthvp: It's a [standard MW timestamp format](https://www.mediawiki.org/wiki/Manual:Timestamp) – comes up in a lot of places. Should we [tweak our constructor to take care of it](https://github.com/siddharthvp/mwn/blob/master/src/date.ts#L21)?

Amorymeltzer: Good point, so I guess so?  I consider it a bug as far as the return is considered, but I suppose it'd only help.  It's used in URLs, at least.

Amorymeltzer: Dunno if you were gonna submit it, but #1200

PR #1196, opened 2020-11-17
warn: Add uw-notenglishedit

TheTVExpert:
Per https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&diff=989209699&oldid=988940845


Amorymeltzer:
I'm wary of #533, but I think this one is fine, given the specific use case it fills a hole for, even though it appears fairly unutilized.

PR #1198, opened 2020-11-21
xfd: Notify WT:TW/WT:AWB/WT:REDWARN automatically (rework of #1001)

Amorymeltzer:
@mdaniels5757 This is a rework of your work in #1001, basically making use of the work in #1153.  I've restructured your PR into the first commit here; there are some renamings but it should roughly match what you created.  You're still attributed so feel free to double check it all!

The second commit expands on that a bit, mainly by moving the checkbox up top and showing it for every venue, but *only* if at least one category is found.  I think it's a better look and structure, but YMMV.

It still uses `wgCategories`; @siddharthvp's [note](https://github.com/azatoth/twinkle/pull/1001#discussion_r446497108) about [T206250](https://phabricator.wikimedia.org/T206250) is of course a concern, but I largely agree.  We could, I suppose lazy-execute a query on form opening, that shouldn't be too bad, although the potential for the form jumping ain't great.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -200,6 +200,37 @@ Twinkle.xfd.callback = function twinklexfdCallback() {
 			}
 		]
 	});
+
+	// Notify devs of common tools, regardless of venue
+	var pageCategories = mw.config.get('wgCategories');
+	Twinkle.xfd.devPages = [];
+	if (namespace === 10 && pageCategories.length) {
+		var devNotificationPageMap = {
+			'Templates used by Twinkle': 'Wikipedia talk:Twinkle',
+			'Templates used by AutoWikiBrowser': 'Wikipedia talk:AutoWikiBrowser',
+			'Templates used by RedWarn': 'Wikipedia talk:RedWarn'
+		};
+		$.each(devNotificationPageMap, function(category, page) {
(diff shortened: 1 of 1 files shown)


siddharthvp:
Would this ever be applicable outside of tfd?

Amorymeltzer:
RM, although I've noticed that it didn't actually handle that case.  I've [added that](https://github.com/azatoth/twinkle/compare/ce6e8d58334300e46077048dff2e91ab7c8902cf..035f383c4bee3eb5bf7da1b58a1bb2ee08388398), but it's not great:

- RM doesn't yet have the server time so as written this reintroduces a use of the client date.  Not ideal.
- Doesn't really fit in `notifyUser`, so maybe it's not worth doing there but rather in a separate callback?  Ugh.

Amorymeltzer:
Unless either of you have a great idea, I think what I'll probably do is merge the first commit, then open a separate PR to better-deal with the fallout from RM, if it's worth it.

siddharthvp:
No rush; I'm working on an XfD rewrite and it turns that by chance the RM notification things fits in neatly into my framework!

Amorymeltzer:
Merged your work @mdaniels5757 as 3e7a8cd; as above, I'll revisit the expansion later

PR #1200, opened 2020-11-21
morebits: Add YYYYMMDDHHmmss to morebits.date constructor

Amorymeltzer:
Noted at https://github.com/azatoth/twinkle/pull/1194#discussion_r525931042, basically just copied from @siddharthvp's work at mwn: https://github.com/siddharthvp/mwn/blob/7fa804354e817ae5e1d3faad9fea64e86bc0cf90/src/date.ts#L21-L25

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1519,13 +1519,29 @@ Morebits.unbinder.getCallback = function UnbinderGetCallback(self) {
 Morebits.date = function() {
 	var args = Array.prototype.slice.call(arguments);
 
-	// Check if it's a MediaWiki signature timestamp (which the native Date cannot parse directly)
-	// Must be first since firefox erroneously accepts the format, sans timezone
-	// See also: #921, #936, #1174, #1187
-	var dateParts;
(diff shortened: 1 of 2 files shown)


siddharthvp:
Don't put it inside `signatureTimestampFormat` hook as that's for wikis to specify the local signature format. This thing is the same everywhere, so I'd either inline it in the constructor, or create a static function like newFromMWTimestamp() to call from the constructor.

Amorymeltzer:
Ah yeah, duh!  Done, but totally invalidated by #1203, so will rework after that.

Amorymeltzer:
Now redone, and includes #1213.  As noted, the bespoke checks are aimed to be first, since not doing so causes problems.

siddharthvp (commented):
> Unfortunately, this limits us in certain ways, in particular the inability to parse the UNIX epoch after late November, 2286.

Height of future planning!

Seriously though, I can't think of any situation where the 14-digit timestamp will arise as a number rather than string. That part could be omitted.

Amorymeltzer:
> Seriously though, I can't think of any situation where the 14-digit timestamp will arise as a number rather than string. That part could be omitted.

Funny, I was just looking at that and writing some tests for it.  Agreed the `typeof` is unneeded; removed.

Issue #1209, opened 2020-11-23
XfD broken on pages in the PageTriage queue


Amorymeltzer:
Bear with me here, but, as an example, try to MfD a page in the PageTriage queue.  The page is created, but adding to the daily list fails (no warning, promise just doesn't resolve): `jQuery.Deferred exception: udate.getDayNameAbbrev is not a function`

That's because:
- MfD nomination calls `date.format('===MMMM D, YYYY===\n', 'utc');`
- `.format` detects the `utc` and `.add`s `this.getTimezoneOffset()`
- `Morebits.date.prototype.add` is broken because of [T268513](https://phabricator.wikimedia.org/T268513)/[T250932](https://phabricator.wikimedia.org/T250932), where PageTriage modifies the native Date prototype, in particular with its own `add`.

As noted on phab, I consider this an upstream issue, but barring an unlikely speedy fix, we need to deal with it.  This impacts all six main XfD venues, although obviously really just AfD, RfD, and MfD are likely.  AfD is particularly bad.  cc @siddharthvp

Amorymeltzer:
Opened #1210 as a quickfix to handle it, by merely skipping the three that collide with us: `add`, `getDayName`, and `getMonthName`.  I've merged that because I want to get a fix out (besides, quiet time is almost over..) but leaving this open in case I or anyone has a more elegant solution.

I'm leaving the version on testwiki broken for testing.

PR #1210, opened 2020-11-23
morebits.date: Don't import methods that collide with PageTriage (T268513)

Amorymeltzer:
See https://phabricator.wikimedia.org/T268513 and #1209 for more details, but basically:

- PageTriage loads an outdated version of an external library (Date.js)
- It adds *a lot* of methods to the native JS Date prototype, including a few that collide with Morebits.date
- In particular, clobbering of `add` broke `format` which broke XfD listings

This just skips importing the three that collide with us: `add`, `getDayName`, and `getMonthName`


siddharthvp (approved):
LGTM. Planning to file a patch for PageTriage, but since it's PageTriage, I wouldn't expect it to be merged anytime soon.

PR #1212, opened 2020-11-24
Add footerlink to Twinkle talk page

TheTVExpert:
Add a link to Twinkle talk page, so users can easily provide feedback on the script.

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -33,6 +33,7 @@ Twinkle.talkback.callback = function() {
 	Window.addFooterLink('About {{talkback}}', 'Template:Talkback');
 	Window.addFooterLink('Talkback prefs', 'WP:TW/PREF#talkback');
 	Window.addFooterLink('Twinkle help', 'WP:TW/DOC#talkback');
+	Window.addFooterLink('Give feedback', 'WT:TW');
 
 	var form = new Morebits.quickForm(Twinkle.talkback.evaluate);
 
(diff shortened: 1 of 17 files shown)


Amorymeltzer:
Won't this leave some/all of them fairly cluttered?  I'm skeptical of the need for this, but maybe removing the doc link?  I've not checked them lately to see how useful they are, but I'd imagine more so that the talkpage (which, incidentally, the doc would lead indirectly to).

siddharthvp:
I think this is a good idea. We could always do with more user feedback. ARV I think has the max number of footerlinks and this fill up the footer. Maybe increase the ARV dialog width by a tiny bit to better accommodate them all?

modules/friendlytalkback.js
```diff
@@ -33,6 +33,7 @@ Twinkle.talkback.callback = function() {
 	Window.addFooterLink('About {{talkback}}', 'Template:Talkback');
```
Amorymeltzer, line 33: Unrelated, but this seems pointless?

Amorymeltzer:
Extending ARV just for the footer seems weird, but I'm not sure there's a better way to incorporate an extra link.  Maybe making the guides to AIV/UAA/SPI just say "AIV info", etc.?  Or something similarly shorter?

modules/twinklearv.js
```diff
@@ -35,6 +35,7 @@ Twinkle.arv.callback = function (uid, isIP) {
 	Window.addFooterLink('About SPI', 'WP:SPI');
```
Amorymeltzer, line 35: I think Wikipedia:Sockpuppet investigations/SPI/Guide to filing cases is probably a more useful link, in line with the others?

Amorymeltzer (approved):
Let's do it!  I'll make the changes I suggested locally, but otherwise all set!

Amorymeltzer:
Merged as 1626040

PR #1142, opened 2020-10-04
Create dev folder, move dev scripts and docs there

Amorymeltzer:
Allow `sync.pl` and `patch-test.js` to be run from any subdirectory, and move them into a new dev folder, along with morebits-test and DEVELOPER.md.  oldtwinkle.js has been removed as entirely unnecessary.

Also allows providing a port number to patchtest via commandline, and adds sync.pl to the package defs as `sync` and `deployall` with the `--mode=deploy --all` options (use `npm run deployall -- --lang=`, etc. to pass other flags).

Instruct `eslint` to ignore /dev folder

----

I've been sitting on this for a while, given the additional scripts in #911, but since @MusikAnimal [mentioned](https://github.com/azatoth/twinkle/issues/129#issuecomment-703293623) the testing method not being noted, I figured I'd toss it up.  It worked over the summer!

Issue #309, opened 2016-01-15
JavaScript tests!


MusikAnimal:
Twinkle is huge and complex. It does powerful things, and we rely on humans to tell us if it malfunctions.

What do we think about adding tests? In particular [Jasmine](https://github.com/jasmine/jasmine), or to do one better, [Jest](https://github.com/facebook/jest/). The latter I've never used on a module-less application (e.g. where we require dependencies on a file-by-file basis), so not sure if it will work for us. Jest is one of the new hot commodities in the JS testing world, which is why I'm bringing it up.

Thoughts?

atlight:
Does anyone realistically have the time to write, maintain and run tests? That's largely the reason we haven't had them until now...

Our code also isn't very testable, with its long functions, so there is also the fact of having to refactor big chunks of the actual modules themselves.

If anyone's willing to work on this, I won't stand in their way, but nor will I be taking part in it myself.

MusikAnimal:
I guess acceptance tests would be the best way to go, then =P But yeah, it sure is time-consuming. Perhaps we would limit testing to the more powerful features. I'll look into it!

PR #1213, opened 2020-11-25
Add some basic QUnit tests for Morebits functions

Amorymeltzer:
tl;dr: QUnit tests (#309) for some Morebits functions `test/tests/`, run via `npm run test`, part of CI.  REPL to play around with.  cc @MusikAnimal and @siddharthvp 

Makes use of jsdom (and jquery), as well as

Makes use of jsdom, as well as jQuery (which is added @3.5.1 rather than MediaWiki's @3.4.1 since the differences don't matter here (https://phabricator.wikimedia.org/T250068)).  In the future jquery-ui could be added (the MediaWiki version (1.9.2) is no longer available (https://www.npmjs.com/package/jquery-ui)).

This a little shim for mediawiki functions and some false data for quick testing essentially creates a REPL, which we load before the QUnit tests.  Each class has its own file, and are all fairly basic right now but should hopefully catch some things down the line.  Nearly all the issues with `Morebits.date` have been browser-specific (implementations of `Date.parse`, polluted prototypes, etc.), which this won't help with.  But a good start!  The tests have been added to GitHub actions.

The mediawiki functions and data are a limited set (see files for list) as needed for the unit tests, although there are extra `mw.util` functions we don't need at the moment.  In the future, it would be nice to set this up to have multiple page/action/user examples.

Per #1142, done away with the `dev/` directory; moved test files into test folder, moved `sync.pl` into `scripts/` folder.

If you want to use the testable REPL environment for morebits, just run `require('path/to/mb_repl')`.  This essentially does:

```javascript
const jsdom = require('jsdom');
const { JSDOM } = jsdom;
global.window = (new JSDOM()).window;
global.document = window.document;

global.HTMLFormElement = window.HTMLFormElement;

global.jQuery = require('jquery');
```

And then loads in `mw_shim`, `mw_config_data`, and finally `morebits.js`, e.g. (from the top of the directory):

```javascript
global.mw = require('./test/mw_shim').mw;
require('./test/mw_config_data');
require('./morebits.js');
global.Morebits = window.Morebits;
```

----

Currently has basic tests for:
- toplevel properties and functions
- `Morebits.date`
- `Morebits.string`
- `Morebits.array`
- `Morebits.wikitext` and `Morebits.wikitext.page`
- `Morebits.batchOperation`


package.json
```diff
@@ -6,12 +6,16 @@
 	"scripts": {
 		"lint": "eslint .",
 		"lint:fix": "eslint . --fix",
-		"patchtest": "node dev/patch-test.js",
-		"sync": "perl dev/sync.pl",
+		"patchtest": "node test/patch-test.js",
+		"sync": "perl scripts/sync.pl",
+		"test": "qunit --require ./test/mb_repl.js ./test/tests/*",
```
Amorymeltzer, line 11: This could arguably be `lint && test`

test/tests/morebits.wikitext.js
```diff
@@ -0,0 +1,41 @@
+QUnit.module('Morebits.wikitext');
+QUnit.test('parseTemplate', assert => {
+	var makeTemplate = function(data) {
+		var template = '{{' + data.name;
+		Object.keys(data.parameters).forEach(function(key) {
+			// Necessary until fixing a bug in parseTemplate
```
Amorymeltzer, line 6: #1214

siddharthvp:
Aren't there licensing issues with copying large chunks of MW source code into the repo? Neither CC-BY-SA nor GFDL are GPL-compatible.

Amorymeltzer:
> Aren't there licensing issues with copying large chunks of MW source code into the repo? Neither CC-BY-SA nor GFDL are GPL-compatible.

Well, none of it is going on the wiki, so that's not a concern.  We're not explicitly licensed here (#252, #253) (although obviously all on-wiki code is implicitly cc-by-sa/gfdl), but it's a good point, I'll explicitly include the header on that one file.

test/tests/morebits.array.js
```diff
@@ -0,0 +1,15 @@
+QUnit.module('Morebits.array');
+QUnit.test('chunk', assert => {
+	assert.deepEqual(Morebits.array.chunk([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 3), [[1, 2, 3], [4, 5, 6], [7, 8, 9], [10]], '10 over 3');
+	assert.deepEqual(Morebits.array.chunk([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 42), [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]], 'Size > length');
+	assert.deepEqual(Morebits.array.chunk([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]), [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]], 'No size');
+	// assert.throws(Morebits.array.chunk(42), 'throws');
```
Krinkle, line 6: In case you were wondering, using `throws()` takes the statement as a code block. Right now it would execute outside the test and not possible for QUnit to check it. (See <https://api.qunitjs.com/assert/throws/>.)

Like so:

```js
assert.throws(function() {
  Morebits.array.chunk(42);
}, 'No array');
```

Or pick one of the shorter styles that JavaScript allows:

```js
assert.throws(
  () => { Morebits.array.chunk(42); },
  'No array'
);

assert.throws(() => Morebits.array.chunk(42), 'No array');
```

test/mw_shim.js
```diff
@@ -0,0 +1,849 @@
...
+ */
+mw.log = console && Function.prototype.bind.call(console.log, console);
+mw.log.warn = console && console.warn ?
+	Function.prototype.bind.call(console.warn, console) :
+	function(){};
+
+/**
+ * mw.Map
+ * Needed for various things, in particular mw.config
+ * https://doc.wikimedia.org/mediawiki-core/master/js/source/mediawiki.html#mw-Map-method-constructor
```
Krinkle, line 36: These are not stable links for source code.

Use something like <https://gerrit.wikimedia.org/g/mediawiki/core/+/1.35.0/resources/src/startup/mediawiki.js> or <https://github.com/wikimedia/mediawiki/blob/ea586cfa2e/resources/src/startup/mediawiki.js> instead.

When browsing GitHub, you can press `y` to get a stable permalink.

.github/workflows/mb_funcs.yml
```diff
@@ -0,0 +1,16 @@
+name: Morebits function tests
+
+on: [push, pull_request]
```
siddharthvp, line 3:
```suggestion
on:
  push:
    paths:
    - 'morebits.js'
  pull_request:
    paths:
    - 'morebits.js'
```

No need to trigger on every PR, just the ones that change morebits. Per https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths

test/mb_repl.js
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+// Consistent environment stuff (mw, window/document via jsdom, jQuery)
+require(__dirname + '/mb_repl_head.js');
+
+// Load mw.config data
+require(__dirname + '/mw_config_data.js');
+
+require(__dirname + '/../morebits.js');
+global.Morebits = window.Morebits;
```
siddharthvp, line 13: Neat, I had heard of `globalThis` but didn't know `global` was a thing!

test/mw_config_data.js
```diff
@@ -0,0 +1,97 @@
+// Includes only what's needed for morebits, util, and Title
+mw.config.set({
+	// Needed by Morebits
+	pageTriageNamespaces: [0, 2, 118],
+	wgArticleId: 18866,
+	wgIsRedirect: false,
+	wgPageName: 'Macbeth',
+	wgRestrictionEdit: ['autoconfirmed'],
+	wgUserGroups: ['interface-admin', 'oversight', 'patroller', 'reviewer', 'sysop', '*', 'user', 'autoconfirmed'],
```
siddharthvp, line 9: Ooof!

test/mw_shim.js
```diff
@@ -0,0 +1,849 @@
...
+ * user.tokens (nonfunctional, only used in Morebits.wiki.page)
+ * Title (only used in Morebits.wiki.page)
+ */
+
+var mw = {};
+
+/**
+ * mw.log
+ * We just need the `warn` method, but might as well mw.log = console.log
+ * node repl doesn't have console.warn
```
siddharthvp, line 25: ? node repl does have it.

test/tests/morebits.date.js
```diff
@@ -0,0 +1,22 @@
+QUnit.module('Morebits.date');
+var now = Date.now();
+var timestamp = '09:31, 17 April 2020 (UTC)';
+QUnit.test('Construction', assert => {
+	assert.equal(new Morebits.date().getTime(), new Date().getTime(), 'Basic constructor');
```
siddharthvp, line 5: This failed for me the first time I ran it (off by one 1 millisecond), but worked fine on every later run.

test/tests/morebits.wikitext.js
siddharthvp: You can write the test for the ideal behaviour here and disable it using `QUnit.skip`, to be enabled later. Writing flawed tests just to get them to pass defeats the point of testing.

test/tests/morebits.batchOperation.js
```diff
@@ -0,0 +1,27 @@
...
+var chunkLength = 2;
+batch.setOption('chunkSize', chunkLength);
+QUnit.test('Run', assert => {
+	var all = '';
+	var runFunc = function(page) {
+		all += page;
+		batch.workerSuccess();
+	};
+	batch.run(runFunc);
+	assert.equal(all, pageList.join(''), 'batch run');
```
siddharthvp, line 17: `assert.equal`, here and everywhere, does comparisons with `==`. Should be avoided in favour of `assert.strictEqual`. This is why Qunit is annoying.

test/mw_config_data.js
```diff
@@ -0,0 +1,97 @@
...
+	wgUserName: 'Amorymeltzer',
+
+	// Needed by mw.util
+	wgScript: '/w/index.php',
+	wgArticlePath: '/wiki/$1',
+	wgScriptPath: '/w',
+
+	// Needed by mw.Title
+	wgFormattedNamespaces: {
+                '-2': 'Media',
```
siddharthvp, line 19: indentations are messed up in this file.

siddharthvp (commented):
This looks pretty good!

Regarding the directory structure, don't do `/test/tests`. Put the mb_* and mw_* files in `/test/mocking`, and the actual tests in `/test`.

.github/workflows/mb_funcs.yml
Amorymeltzer: Dope!

test/mw_config_data.js
Amorymeltzer: s/spaces/tabs/

test/tests/morebits.array.js
Amorymeltzer: Awesome, thanks!  Done!

test/tests/morebits.batchOperation.js
Amorymeltzer: Done!

test/tests/morebits.date.js
Amorymeltzer: I initially had this with ISOString for exactly this reason (some browsers fudge milliseconds for security reasons), so maybe that would be better?

test/mb_repl.js
Amorymeltzer: Like 70% of the effort here was annoying `global`/`window`/`global.window` shit.  But global is the repl's speciality since there ain't no window

test/mw_shim.js
Amorymeltzer: Dunno WTF that was, I must've been overwriting it when first testing...

.github/workflows/mb_funcs.yml
Amorymeltzer: Mmm, what about adding `- 'test/**`?  Might as well run the tests if tests change...

Amorymeltzer:
> Regarding the directory structure, don't do `/test/tests`. Put the mb_* and mw_* files in `/test/mocking`, and the actual tests in `/test`.

Fair, although I did this so as to fit all the other stuff in there (patch-test, #1215, etc.).  I don't *think* QUnit will touch anything by default, but this seemed cleaner.

.github/workflows/mb_funcs.yml
siddharthvp: Yep, was about to say that!

test/tests/morebits.date.js
Amorymeltzer: I suppose we could lop off the final digit, since we don't really care.  It's reliably under 0.1% of times for me, so I'm fine leaving as-is if you are.

Amorymeltzer:
In the interest of just moving forward, done

PR #1215, opened 2020-11-26
add dev/server.js

siddharthvp:
The server.js file from #911. No dependencies. While it's possible to set up a server using `php -S` and `python -m SimpleHTTPServer` etc, putting it in package.json means IDEs can provide a one-click button for running it.


siddharthvp:
For the record, the security issue flagged by CodeQL isn't germane here: while providing uncontrolled access to the file system is indeed a problem for a production server, here the only files an "attacker" can gain access to are the ones on their own system.

PR #1216, opened 2020-11-26
tag: initial work on l10n

siddharthvp:
Initial work on localisation of Tag module. See commits for details.

----

I have [imported](https://test.wikipedia.org/wiki/Special:Log?type=import&user=SD0001&page=&wpdate=2020-11-27&tagfilter=&subtype=) all article tag templates to testwiki for ease of testing.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -481,6 +481,7 @@ Twinkle.tag.article = {};
 var translationSubgroups = [
 	{
 		name: 'translationLanguage',
+		parameter: '1',
 		type: 'input',
 		label: 'Language of article (if known): ',
 		tooltip: 'Consider looking at [[WP:LRC]] for help. If listing the article at PNT, please try to avoid leaving this box blank, unless you are completely unsure.'
@@ -552,6 +553,7 @@ Twinkle.tag.article.tagList = {
 				tag: 'Cleanup', description: 'requires cleanup',
 				subgroup: {
 					name: 'cleanup',
+					parameter: 'reason',
 					type: 'input',
 					label: 'Specific reason why cleanup is needed: ',
 					tooltip: 'Required.',
@@ -567,6 +569,7 @@ Twinkle.tag.article.tagList = {
 				description: 'requires copy editing for grammar, style, cohesion, tone, or spelling',
 				subgroup: {
 					name: 'copyEdit',
+					parameter: 'for',
 					type: 'input',
 					label: '"This article may require copy editing for..." ',
 					tooltip: 'e.g. "consistent spelling". Optional.',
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
This all looks pretty good, but a few comments @siddharthvp:

1. What does this have to do with localization?  I understand that this is laying the ground work, and will make localization easier, but the subject of this PR is inscrutable or misleading to anyone else.
2. Since I have to review everything before pushing it up, having something live here more than a few days (during a major US holiday) would be good practice.  Especially as there may be other, earlier PRs this may ruin with conflicts.  Personally, I've tried to keep it to a couple of weeks or until someone chimes in, unless it's urgent/simple.
3. One of the reasons b443496 wasn't already done is that if the tag in question isn't visible because the user has used the quickfilter, the field validation just quietly prevents the form from proceeding.  Now it just looks broken, with no message whatsoever.

siddharthvp:
1. This is more than just the groundwork. Localisation is all about separating the to-be-configurable per-wiki data from the code. Yet, I'm not pretending this is the whole thing, which is why it says "initial work". The subject clearly says "See commits for details". I don't see how it's misleading.
2. I pushed this early I had an another large PR in the works that would have conflicted badly with this one. I hoped to put it up yesterday but then I discovered some intractable issues in it so had to abandon it. I have a tight schedule as I'm starting a new job in January by which time the grant period is supposed to end. RL things took unexpected turns in Nov because of which I have had little contiguous free time to work on twinkle, and I don't know how much free time I'll have after January. Waiting a couple of weeks on every PR won't be possible. If you want to avoid conflicts, merge in things quickly. I can only keep track of all the new work on master, not the open PRs as well. No doubt it will lead to some merged commits that may later need changes or reversion. That's to be expected of any project in very active development. 
3. I see. Seems like there's no easy solution. Do you have ideas on how to fix it?

Amorymeltzer:
Literally all I'm asking is to not make my life even harder as we all try to *collaborate*.  I have gone out of my way to try and smooth everyone's experience and reduce conflicts while encouraging participation, but a combative stance does not help anyone.  Twinkle is not any one person's project, certainly not mine, and "if you don't like it move faster" is not conducive to working together.

siddharthvp:
Only seeing this now as I didn't check email. You are making your life hard if you try to review every line before pushing it up (which you imply you're doing) – there is no obligation on you to do that. The int-admins who usually review editreqs on-wiki don't even know JS – they just rely on trusting the user. 

I merely explained my POV and at the same time took the opportunity to explain why I've only started just about now when the official start date was 10 Nov.

Amorymeltzer:
>You are making your life hard if you try to review every line before pushing it up (which you imply you're doing) – there is no obligation on you to do that. The int-admins who usually review editreqs on-wiki don't even know JS – they just rely on trusting the user.

This isn't really the place, but FWIW I entirely disagree.  Anything I push I take responsibility for, so even for my own sake, I review every edit just like any other edit.  Moreover, there are only a few of us entrusted to edit those pages, so to "just rely on trusting the user" would defeat the purpose and, at least IMO, would be preempting the community and bordering on irresponsible.  It's trivial to break things for thousands of users, which is why I make an effort not to make notable changes if I can't guarantee to be somewhat available or responsive for a while after.

PR #1218, opened 2020-11-30
tag: duflu merged to needtrans at TfD

Amorymeltzer:
https://en.wikipedia.org/wiki/Wikipedia:Templates_for_discussion/Log/2020_October_31#Template:Duflu


Amorymeltzer:
That being said, are we fully following the procedures?  Regardless, I think they'll need to change slightly....

PR #1219, opened 2020-12-01
tag: fix bug from #1216

siddharthvp:
https://github.com/azatoth/twinkle/pull/1216#issuecomment-736019277 bullet 3.

Hacky handling of form submission when quickfilter hides required fields that weren't filled in.

Suggestion for improvement welcome.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -240,6 +240,26 @@ Twinkle.tag.callback = function friendlytagCallback() {
 		}
 	});
 
+	// Hacky handling of form submission when quickfilter hides required fields that weren't filled in:
+	// Remove the form submit handler
+	/** @param {HTMLFormElement} form */
+	var submitForm = function(form) {
+		if (form.checkValidity()) {
+			Twinkle.tag.callback.evaluate({target: form});
+		} else {
+			$(form.quickfilter).val('');
+			// XXX: $(form.quickfilter).keyup() doesn't seem to work, so manually doing what the
+			// keyup handler should do
+			$allCheckboxDivs.show();
+			$allHeaders.show();
+			$(form).find('input[type=submit]').click(); // triggers browser validation which should fail
+		}
+	};
+	result.removeEventListener('submit', Twinkle.tag.callback.evaluate);
+	$('.tw-tag-submit').off('click').on('click', function() {
+		submitForm(result);
+	});
+
 	if (Twinkle.tag.mode === 'article') {
 
 		Twinkle.tag.alreadyPresentTags = [];


modules/friendlytag.js
```diff
@@ -222,7 +222,7 @@ Twinkle.tag.callback = function friendlytagCallback() {
 			]
 		});
 	}
-	form.append({ type: 'submit', className: 'tw-tag-submit' });
+	form.append({ type: 'submit', className: 'tw-tag-submit', name: 'submitButton' });
```
siddharthvp, line 225: Was necessary because by default the button has name `submit`, which causes `form.submit` to refer to this button, shadowing HTMLFormElement API's submit() method.

modules/friendlytag.js
```diff
@@ -240,6 +240,26 @@ Twinkle.tag.callback = function friendlytagCallback() {
...
 
+	// Hacky handling of form submission when quickfilter hides required fields that weren't filled in:
+	// Remove the form submit handler
+	/** @param {HTMLFormElement} form */
+	var submitForm = function(form) {
+		if (form.checkValidity()) {
+			Twinkle.tag.callback.evaluate({target: form});
+		} else {
+			$(form.quickfilter).val('');
+			// XXX: $(form.quickfilter).keyup() doesn't seem to work, so manually doing what the
```
Amorymeltzer, line 251: Not sure what you mean here?

Amorymeltzer:
So, AFAICT, this just force-clears the quickfilter, thus showing the form and highlighting the missing thing(s)?  Seems reasonable, I suppose; it's kind of what I'd expect at an airline or survey or whatever.  Visually looks fine, I think?

It does feel like this sort of thing should be better handled by morebits, though, rather than unwriting and rewriting like this.

modules/friendlytag.js
siddharthvp: The main quickfilter function is a keyup handler on the input field. So I'd have expected `$(form.quickfilter).val('').keyup()` or `$(form.quickfilter).val('').trigger('keyup')` to automatically [make all checkboxes and headers show up](https://github.com/azatoth/twinkle/blob/3813f112ff2d7475dcc2d9b43239e4b6761043e6/modules/friendlytag.js#L82), but it doesn't work for some reason.

Amorymeltzer: Ah okay I get you now.  I know `.trigger` is imperfect, looks like `.keyup` is not calling the event at all.  IIRC events added via `addEventListener` are difficult to find, so that could be the root of it.

modules/friendlytag.js
siddharthvp: Undone as no longer necessary, submit() is not being used now.

siddharthvp:
Replaced `form.submit();` with `$(form).find('input[type=submit]').click();` as the former was not working nicely in IE 11. [MDN docs on .submit()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit) say that "Constraint validation is _not_ triggered". That seems to contradict the actual behaviour (at least in Chrome). In IE 11 too, the form wasn't getting submitted on invalid input, but no popup message was showing up. This is resolved on triggering a click on the [the `input[type=submit]`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit) which is invisible and notably different from `button.tw-tag-submit` which is added by simpleWindow.

modules/friendlytag.js
```diff
@@ -240,6 +240,26 @@ Twinkle.tag.callback = function friendlytagCallback() {
...
+			// keyup handler should do
+			$allCheckboxDivs.show();
+			$allHeaders.show();
+			$(form).find('input[type=submit]').click(); // triggers browser validation which should fail
+		}
+	};
+	result.removeEventListener('submit', Twinkle.tag.callback.evaluate);
+	$('.tw-tag-submit').off('click').on('click', function() {
+		submitForm(result);
```
siddharthvp, line 260: Despite the `off` and `on`, this still is being triggered twice. 🤔

But this doesn't seem to be a problem as  the tagging and all supplementary actions (like adding the merge tag on the other page and posting to WP:PNT) are taking place once only.

Amorymeltzer: @siddharthvp I get a very laggy interface and a `too much recursion` error here/at the `.click` line.  I'm seeing it run twice, sure, but each time it looks like these are running 13 times.  FF fwiw.

PR #1220, opened 2020-12-01
Set up use of TypeScript

siddharthvp:
Set up the use of TypeScript. I suppose there's no dearth of information online on why TypeScript is so much better than untyped JavaScript. This will help us nip a lot of bugs at the source, while also allowing the use of all ES2020+ JS features.

I have converted twinkle.js to twinkle.ts. The compiled output is very similar to what we had before. Will work on Tag and XFD hopefully soon, as they anyway need to be entirely restructured (for #1144). 

I have left morebits.js as is and just created the type definition file for it as `types/morebits.d.ts`. It's probably unnecessary and quite difficult to properly migrate morebits to TS, as most of it is morebits.wiki.page whose implementation is far removed from standard TS idiom.

Type definitions for jQuery come from [@types/jquery module](https://www.npmjs.com/package/@types/jquery). Definitions for mw.* modules are in `types/mediawiki.d.ts`, self-written as I couldn't find them online despite extensive searches. I presume we're amongst the first codebases in the mediawiki universe to start using typescript.

Run `npm i typescript` first, then `npm run build` can be used for compiling TS to JS. Compilation does give some errors, but they don't matter (output is generated anyway). The errors can be avoided by adding some `// @ts-ignore`s. 

Cc @MusikAnimal and @Amorymeltzer

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,6 +60,19 @@
 				"strip-json-comments": "^3.1.1"
 			}
 		},
+		"@types/jquery": {
+			"version": "3.5.4",
+			"resolved": "https://registry.npmjs.org/@types/jquery/-/jquery-3.5.4.tgz",
+			"integrity": "sha512-//9CHhaUt/rurMJTxGI+I6DmsNHgYU6d8aSLFfO5dB7+10lwLnaWT0z5GY/yY82Q/M+B+0Qh3TixlJ8vmBeqIw==",
+			"requires": {
+				"@types/sizzle": "*"
+			}
+		},
+		"@types/sizzle": {
+			"version": "2.3.2",
+			"resolved": "https://registry.npmjs.org/@types/sizzle/-/sizzle-2.3.2.tgz",
+			"integrity": "sha512-7EJYyKTL7tFR8+gDbB6Wwz/arpGa0Mywk1TJbNzKzHtzbwVmY4HR9WqS5VV7dsBUKQmPNr192jHr/VpBluj/hg=="
+		},
 		"acorn": {
 			"version": "7.4.1",
 			"resolved": "https://registry.npmjs.org/acorn/-/acorn-7.4.1.tgz",
@@ -868,6 +881,12 @@
 			"version": "0.8.1",
 			"resolved": "https://registry.npmjs.org/type-fest/-/type-fest-0.8.1.tgz",
 			"integrity": "sha512-4dbzIzqvjtgiM5rw1k5rEHtBANKmdudhGyBEajN01fEyhaAIhsoKNy6y7+IN93IfpFtwY9iqi7kD+xwKhQsNJA==",
+			"dev": true
+		},
+		"typescript": {
+			"version": "4.1.2",
+			"resolved": "https://registry.npmjs.org/typescript/-/typescript-4.1.2.tgz",
+			"integrity": "sha512-thGloWsGH3SOxv1SoY7QojKi0tc+8FnOmiarEGMbd/lar7QOEd3hvlx3Fp5y6FlDUGl9L+pd4n2e+oToGMmhRQ==",
 			"dev": true
 		},
 		"uri-js": {
--- a/package.json
+++ b/package.json
@@ -7,12 +7,15 @@
 		"lint": "eslint .",
 		"lint:fix": "eslint . --fix",
 		"patchtest": "node dev/patch-test.js",
-	  	"server": "node dev/server.js",
+		"server": "node dev/server.js",
 		"sync": "perl dev/sync.pl",
-		"deployall": "perl dev/sync.pl --mode=deploy --all"
+		"deployall": "perl dev/sync.pl --mode=deploy --all",
+		"build": "npx tsc -p ./tsconfig.json"
 	},
 	"devDependencies": {
-		"eslint": "^7.13.0",
-		"eslint-plugin-es5": "^1.5.0"
+	  "@types/jquery": "^3.5.4",
+	  "eslint": "^7.13.0",
+	  "eslint-plugin-es5": "^1.5.0",
+	  "typescript": "^4.1.2"
 	}
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,73 @@
+{
+  "compilerOptions": {
+	/* Visit https://aka.ms/tsconfig.json to read more about this file */
+
+	/* Basic Options */
+	// "incremental": true,                   /* Enable incremental compilation */
+	"target": "ES5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
+	"module": "none",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
+	"lib": [
+	  "dom",
+	  "es5",
+	  "es2015.collection" // Only Map and Set from this are supported in IE 11.
+	],                    /* Specify library files to be included in the compilation. */
+	// "allowJs": true,                       /* Allow javascript files to be compiled. */
+	// "checkJs": true,                       /* Report errors in .js files. */
+	// "jsx": "preserve",                     /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
+	// "declaration": true,                   /* Generates corresponding '.d.ts' file. */
+	// "declarationMap": true,                /* Generates a sourcemap for each corresponding '.d.ts' file. */
+	// "sourceMap": true,                     /* Generates corresponding '.map' file. */
+	// "outFile": "./",                       /* Concatenate and emit output to single file. */
+	// "outDir": "./",                        /* Redirect output structure to the directory. */
+	// "rootDir": "./",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
+	// "composite": true,                     /* Enable project compilation */
+	// "tsBuildInfoFile": "./",               /* Specify file to store incremental compilation information */
+	// "removeComments": true,                /* Do not emit comments to output. */
+	// "noEmit": true,                        /* Do not emit outputs. */
+	// "importHelpers": true,                 /* Import emit helpers from 'tslib'. */
+	// "downlevelIteration": true,            /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
+	// "isolatedModules": true,               /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
+
+	/* Strict Type-Checking Options */
+	//	"strict": true,                       /* Enable all strict type-checking options. */
+	// "noImplicitAny": true,                 /* Raise error on expressions and declarations with an implied 'any' type. */
+	// "strictNullChecks": true,              /* Enable strict null checks. */
+	// "strictFunctionTypes": true,           /* Enable strict checking of function types. */
+	// "strictBindCallApply": true,           /* Enable strict 'bind', 'call', and 'apply' methods on functions. */
+	// "strictPropertyInitialization": true,  /* Enable strict checking of property initialization in classes. */
+	// "noImplicitThis": true,                /* Raise error on 'this' expressions with an implied 'any' type. */
+	// "alwaysStrict": true,                  /* Parse in strict mode and emit "use strict" for each source file. */
+
+	/* Additional Checks */
+	// "noUnusedLocals": true,                /* Report errors on unused locals. */
+	// "noUnusedParameters": true,            /* Report errors on unused parameters. */
+	// "noImplicitReturns": true,             /* Report error when not all code paths in function return a value. */
+	// "noFallthroughCasesInSwitch": true,    /* Report errors for fallthrough cases in switch statement. */
+
+	/* Module Resolution Options */
+	 "moduleResolution": "node",            /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
+	// "baseUrl": "./",                       /* Base directory to resolve non-absolute module names. */
+	// "paths": {},                           /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */
+	// "rootDirs": [],                        /* List of root folders whose combined content represents the structure of the project at runtime. */
+	 "typeRoots": ["./types"],                /* List of folders to include type definitions from. */
+	// "types": [],                           /* Type declaration files to be included in compilation. */
+	// "allowSyntheticDefaultImports": true,  /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */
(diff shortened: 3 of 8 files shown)


Amorymeltzer:
I'm of course curious what @MusikAnimal thinks, but I'm mixed on this WIP, likely opposed.  Beyond the fact that this wasn't part of #1144, this is, first and foremost, an on-wiki gadget, and I think things that increase the distance between on-wiki code and contributors contributing should be avoided if possible.  That being said, I see the clear benefits, especially if/when IE11 support is lifted.  Is there a reason this *has* to be a part of that project?  It seems like a lot of extra restructuring/etc. and reviewing in and of itself, especially for something that would already be a large endeavor.

siddharthvp:
Have you ever used TypeScript? You need to use it once to appreciate why it's helpful. The main benefits are saying goodbye to all sorts of issues that occur due to type mismatch which have frequently occurred in Twinkle, and much better integration with IDE code completion tools which make development easier. You can get precise code completion for browser APIs and `mw.*` stuff that an IDE would never be able to provide for pure javascript.

> first and foremost, an on-wiki gadget, and I think things that increase the distance between on-wiki code and contributors contributing should be avoided if possible. 

Why? The dev code not being the same as production code obviously doesn't mean any behaviour would be different – these tools are battle-tested in the industry for much bigger and critical applications than Twinkle. Besides, note that the transpiled code is very much readable.

If you're worried about debugging, that can be easily done using source maps – with which you can set up breakpoints right inside the TS code in an IDE (I'll try to document these things). 

> That being said, I see the clear benefits, especially if/when IE11 support is lifted. 

What does this have to do with IE support? When IE11 support is lifted, all we've to do is to flip the compilation target in `tsconfig.json` from ES5 to ES6, so that the transpiled code looks more like similar to the original.
 
> It seems like a lot of extra restructuring/etc. and reviewing in and of itself, especially for something that would already be a large endeavor.

The only restructuring that was necessary was porting `twinkle.js`, since `Twinkle` is needed in the modules and it not being in TS makes things a bit weird. I don't get the other part of the comment, this is intended to make future work easier, not harder. Granted, TS does have some pain points, but I'm pretty sure the positives outweigh it (otherwise it wouldn't have become insanely popular as it is now, would it?).

Amorymeltzer:
JS ain't my first (or favorite!) language, I'm familiar with different typing systems; hence, "I see the clear benefits"

> > first and foremost, an on-wiki gadget, and I think things that increase the distance between on-wiki code and contributors contributing should be avoided if possible.
>
> Why? The dev code not being the same as production code obviously doesn't mean any behaviour would be different – these tools are battle-tested in the industry for much bigger and critical applications than Twinkle. Besides, note that the transpiled code is very much readable.

I think you missed what I was saying here.  I mean that this is a wiki(p|m)edia gadget, for use by on-wiki users, including by review.  We develop here because MW is awful for development, but it's where the userbase and contributor base will come from, 100%.  I meant that things moves obfuscate on-wiki code are less desirable and things that make it harder for someone to look at something on-wiki and then contribute are less desirable.  This is something that can help developers develop Twinkle/Morebits, but it makes it harder for the next wiki user to join in.  That's what I meant by "distance."

> > That being said, I see the clear benefits, especially if/when IE11 support is lifted.
> 
> What does this have to do with IE support? When IE11 support is lifted, all we've to do is to flip the compilation target in `tsconfig.json` from ES5 to ES6, so that the transpiled code looks more like similar to the original.

Yes, that's exactly what I was saying.

> Granted, TS does have some pain points, but I'm pretty sure the positives outweigh it (otherwise it wouldn't have become insanely popular as it is now, would it?).

I don't disagree, but [[[Popularity fallacy]]](https://en.wikipedia.org/w/index.php?title=Popularity_fallacy).

siddharthvp:
I can tell you from personal experience that I put together the whole of #1244 in 3 days, a feat that would never have been possible had it not been for those red underlines that appeared everytime I wrote something wrong. It is not popularity "fallacy". People use it for a reason. There have been many JS extensions out there but only one of them is popular.

> I think you missed what I was saying here. I mean that this is a wiki(p|m)edia gadget, for use by on-wiki users, including by review. We develop here because MW is awful for development, but it's where the userbase and contributor base will come from, 100%. I meant that things moves obfuscate on-wiki code are less desirable and things that make it harder for someone to look at something on-wiki and then contribute are less desirable. This is something that can help developers develop Twinkle/Morebits, but it makes it harder for the next wiki user to join in. That's what I meant by "distance."

Is there any evidence for any of this? It doesn't matter whether the code is on WP or GitHub. People would want to contribute to a codebase which they can actually read and understand, not an antique dump of legacy code. See recent on-wiki discussions like [this](https://en.wikipedia.org/wiki/Wikipedia:Village_pump_(technical)#Proper_ECMAScript_support,_modernisation_and_integration_into_the_existing_ecosystem) and [this](https://en.wikipedia.org/wiki/Wikipedia_talk:WikiProject_JavaScript#JS_on_wikimedia_is_currently_too_antiquated_to_ever_bloom). Anyone who has learnt JavaScript anytime in the last 5 years (and who learn it from now onwards) would be familiar with the recent language variants, not with ES5. 

Your only argument seems to be "obfuscating on-wiki code" – which would happen even if we were using ES2020 or ES6 or any sort of transpiler. Are you opposed to that even? You mentioned [here](https://github.com/azatoth/twinkle/issues/129#issuecomment-703013724) you wanted to do it and FWIW @MusikAnimal [agreed with it](https://github.com/azatoth/twinkle/issues/129#issuecomment-703293623). The transpiled output of TypeScript and ES6+ are the very same.

There really is no excuse for sticking with ES5, and I'm certain you're in the minority here. It would make things like #1244 and WIP #1221 _much_ more difficult to do – seeing as they rely extensively on classes which ES5 doesn't really support.

siddharthvp:
Had it been any other PR, I would have stopped arguing with you long back. But this really is the base upon which so much of other work (i18n/l10n-related) depends.

DannyS712:
@siddharthvp have you considered using OOJS's object inheritance features instead? OO.inheritClass can be used for class inheritance, and a bunch more stuff - see https://www.mediawiki.org/wiki/OOjs

siddharthvp:
I haven't considered that. But right off the bat, OOjs is a heavy library and we _also_ have to load jquery.ui and select2 on top of that. Plus, I'm yet to hear a compelling reason why we shouldn't use TS/ES6, which are useful for a lot more things than just object inheritance.

DannyS712:
> I haven't considered that. But right off the bat, OOjs is a heavy library and we _also_ have to load jquery.ui and select2 on top of that. Plus, I'm yet to hear a compelling reason why we shouldn't use TS/ES6, which are useful for a lot more things than just object inheritance.

I'm not sure I understand what you mean by "heavy" - ResourceLoader minification and caching are extremely robust, and most gadgets I've seen already depend on something that uses oojs (as do many core modules). Furthermore, jquery.ui is deprecated in favor of OOUI, and all ooui modules depend on oojs, so eventually we'll have to load it anyway.

Amorymeltzer:
Apologies if that was confusing; that second paragraph was not in reference to the quoted portion, but rather the task itself.  I was working on it before putting it on hold.

siddharthvp:
I did mention ES2020:
>- Use object-oriented pattern in the more complex modules like Tag, XFD and CSD to make them extensible more easily.
>- Set up a build step for transpiling ES2020 code. This makes working with classes a lot easier.

I didn't mention TS because I didn't even know of TS at the time. I have been figuring out things along the way, rather than having everything planned advance. I have already told you in an unrelated email discussion what led me to this, reproducing for others:
> ... The key advantage of ES6 which I'm after is the class syntax which I have been using a lot in the rewrite. My idea was that since we need to use ES6+ anyhow, why not at the same time also do TypeScript which brings in some real advantages in addition to the cosmetics of ES6. TS code looks quite similar to ES6. ...

And you never replied after that. Now you pretend to be "surprised, to say the least". Cool.

siddharthvp:
@DannyS712 OOjs/OOUI are projects that started way back in 2012-13, at a time when there was no ES6. If it had been created today, they wouldn't have built up stuff like `OO.inheritClass`. OOUI's successor, which started recently, is [WVUI](https://github.com/wikimedia/wvui), which [uses TypeScript](https://github.com/wikimedia/wvui#typescript) (which shouldn't be surprising). For a project being largely restructured/rewritten in 2020, continuing to use ES5 is unheard of.

Amorymeltzer:
I am always happy to discuss the merits of a proposal, but I do not intend to spend what little personal time I have trading rude barbs or parsing my emotions according to someone else's schedule.  If you'd like to return to the former, I'm here.

MusikAnimal:
Sorry for taking so long to reply to this! First off, I ask that we respect each other's opinions. It seems the discussion has unnecessarily gotten a little heated.

As for my opinion, I do support use of ES6+. I have never used TypeScript, and while it looks really neat, the fact that it is effectively a different language (or superset) is a little concerning. Indeed, ES6+ requires a new build step to transpile to ES5, just like TypeScript, but it's still _JavaScript_. All the principles ES6+ brings is something any JS developer will face some day. Even at Wikimedia, with our infamously 10+ year long support cycle, [the future of ES5 is limited](https://phabricator.wikimedia.org/T178356).

I will say the fact that WMF is starting to use TypeScript makes me more accepting of it. Presumably more of the Wikimedia developer community will come to adopt it, too. It still seems like a commitment that is worthy of careful thought, though. Will future Twinkle devs care for TypeScript? Maybe not. Will they for ES6+? Yes, because that is the ~~future~~ current state of JavaScript.

I don't feel either TypeScript or ES6 are _required_ for the localization project, but if we have to rewrite things anyway as part of that effort, I think it makes sense to adopt newer technologies in an effort to be more future-proof. ES6+ is a very safe bet in that regard. TypeScript seems plenty worthwhile, but only if there's consensus here to use it. WMF's adoption helps but shouldn't be a deciding factor. Similarly, my opinion shouldn't bear that much weight as I'm not really that active as a Twinkle dev :)

So my overall opinion is TypeScript probably isn't that hard to grasp and seems to have good payoffs, but any attempt to rewrite Twinkle should at least be ES6+ and liberally take advantage of its features. I do not recommend using OOjs. At the same time, we don't want to end up with a freshly rewritten codebase that is harder to contribute to and/or alienates active and future contributors.

Hope this helps.

siddharthvp:
Thanks for the comment, MusikAnimal. 

I would suggest taking a look at https://github.com/wikimedia-gadgets/twinkle/blob/l10n/src/xfd.ts. While there is some infrastructure behind the scenes to ensure robust type checking, for a casual contributor wanting to improve Twinkle, the only way that the code is different from ES6 are that there are a few lines of property declarations at the top of classes, and some functions have type specifiers for return types and/or parameters (they aren't required everywhere – for methods returning literals and for inherited methods, TS is smart enough to make out the types automatically). If we were using pure JS instead, anyway those type specifiers would have had to be included in the JSDocs instead, which aren't any easier to maintain. I hope this alleviates the concerns of it being a "different language" – it isn't.

I would also suggest comparing that file with https://github.com/wikimedia-gadgets/twinkle/blob/master/modules/twinklexfd.js. Which one looks more organised and easier to maintain and expand?

A very important point which I cannot emphasise enough: TS is a very handy tool to have while working with unfamiliar libraries – which would be the case for people writing per-wiki localisations using morebits/twinkle-core as a library.

From a project management perspective as well, TS shines. It is one tool that replaces so many others. We would be using at least 5 different tools if we are not using TS:
* babel
* babel-plugin-proposal-class-properties and a few other plugins to support useful ES2021 features (TS supports them all out of the box). 
* babel-plugin-add-header-comment [which MA suggested](https://github.com/wikimedia-gadgets/twinkle/issues/1144#issuecomment-707897860) (not needed as TS compiler is pretty conservative – it doesn't even remove comments!).
* A JSDoc linter/validator to ensure that they are in sync with code
* eslint-plugin-es5 (added in #968; won't be needed as TS by itself calls out use of functions not supported in IE11, and needless to say TS can detect "hidden uses" of those functions which the eslint plugin can't).

And wow, this turned out to be a pretty big comment and I haven't yet talked about the _existing bugs present in Twinkle_ which are trivially avoided when using TS. Will bring those up if anyone needs further convincing.

sportshead:
Hello Twinkle maintainers, I saw this PR and just wanted to comment on a few things. @ChlodAlejandro and I am currently rewriting RedWarn from JS into TS, and since the old codebase isn't very ES6 friendly, rewriting the codebase basically meant rewriting the entire tool. Overall, TypeScript would be easier for users new to the codebase to contribute. From my experience, a project at the relatively small scale of Twinkle and RedWarn may still take a while for new contributors to explore the JavaScript codebase. Poor documentation (no offense to @Ed6767) of the old RedWarn code meant that while rewriting the code I had to look at what each function accepted as parameters, what they returned, and how it was used. With TypeScript, many IDEs provide auto-complete and show what types a said function accepts and returns. Now, with all good things, there must be a downside - and there are two that I can think of, which is lack of typings for the MediaWiki API (see [T263914](https://phabricator.wikimedia.org/T263914)), and learning the new syntax of TypeScript. The former is already being worked on by @siddharthvp ([wikimedia-gadgets/types-mediawiki](https://github.com/wikimedia-gadgets/types-mediawiki)), and for the latter, TypeScript syntax is relatively easy to learn if you already know other similar object-oriented programming languages such as C# and/or Java. From what I understand, Twinkle uses modular classes just like RedWarn, and RedWarn's current approach is to convert every file that extends `rw` (e.g. `rw.rollback`) into its own exported class (e.g. `export default class Rollback {...}`). I hope this comment helped you decide on whether or not to use TypeScript.

siddharthvp:
> From what I understand, Twinkle uses modular classes just like RedWarn, and RedWarn's current approach is to convert every file that extends rw (e.g. rw.rollback) into its own exported class (e.g. export default class Rollback {...})

This is not the case. In the current setup on https://github.com/wikimedia-gadgets/twinkle/tree/l10n branch, the TS source files aren't modules, they are normal script files. They are compiled to JS using tsc, and concatenation and minification is left to the MediaWiki ResourceLoader. (In the current ES5 twinkle, the Twinkle scripts are put in a folder named `modules` (done back in 2011), but of course they aren't [ECMAScript modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules)).

Using modules would be neat, I think, even though at the moment I am yet to find a pressing need to be using modules, as the different Twinkle scripts don't depend on each other. But the trouble is with deciding on how to configure a module bundler like webpack. The development mode bundle that webpack produces is quite readable if we aren't using ES6+ syntax excessively. But it leaves a  lot of `__WEBPACK_IMPORTED_MODULE_1__`-like variables in the code. Is all that acceptable? An alternative is to use webpack's production mode bundling which produces fully minified code (minified a lot better than ResourceLoader), but I _think_ there are some people who take exception at the idea of putting minified code on-wiki gadget pages. 

Open to any and all ideas on module bundling.

siddharthvp:
The l10n work continues at https://github.com/wikimedia-gadgets/twinkle-core and https://github.com/wikimedia-gadgets/twinkle-enwiki. I'm now using TypeScript with modules – clearly very useful for a number of reasons. Thanks for the suggestion @sportshead.

siddharthvp:
Accidentally closed due to branch deletion, reopening to let discussion (if any) to continue, though the contents of this branch are very outdated.

PR #1221, opened 2020-12-05
tag: WIP: rewrite for localizability 

siddharthvp:
Builds up on #1220 (as that isn't yet merged, its commits are included here for now).

This is an object-oriented rewrite of the tag module where each mode (article, redirect, file) is a class extending the abstract base class `TagMode`. Common functionality are shared in either the `Tag` class or in `TagMode`.

All functionality are methods on the class, for example `captureFormData()`, `validateInputs()`, `action()`.

This is a WIP. `TagMode` could probably be improved to _do_ something (`action()`) by default. I think I can figure that out while working on article/file modes.

Also, #1170 needs to be integrated into this.

### DIFF
--- a/modules/tag-data.js
+++ b/modules/tag-data.js
@@ -38,13 +38,15 @@ var redirectTagList = {
                         name: 'altLangFrom',
                         type: 'input',
                         label: 'From language (two-letter code): ',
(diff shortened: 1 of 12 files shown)


siddharthvp:
Development of localised Twinkle is now taking place on https://github.com/wikimedia-gadgets/twinkle-core and https://github.com/wikimedia-gadgets/twinkle-enwiki repos.

PR #1224, opened 2020-12-05
Convert all morebits/twinkle modules to use json, not xml, by default

Amorymeltzer:
Does what it says on the tin.  `xml` is still the default for `Morebits.wiki.api` for backwards compatibility, but this should convert everyone.

Not live on test.wiki since that's being actively tested, but I had it up for a month or so this fall, everything was smooth.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1416,13 +1416,16 @@ Twinkle.tag.callbacks = {
 				'redirects': 1,  // follow redirect if the class name turns out to be a redirect page
 				'lhnamespace': '10',  // template namespace only
 				'lhshow': 'redirect',
-				'lhlimit': 'max' // 500 is max for normal users, 5000 for bots and sysops
+				'lhlimit': 'max', // 500 is max for normal users, 5000 for bots and sysops
+				'format': 'json'
 			}, function removeRedirectTag(apiobj) {
-
-				$(apiobj.responseXML).find('page').each(function(idx, page) {
+				var pages = apiobj.getResponse().query.pages.filter(function(p) {
+					return !p.missing && !!p.linkshere;
+				});
+				pages.forEach(function(page) {
 					var removed = false;
-					$(page).find('lh').each(function(idx, el) {
-						var tag = $(el).attr('title').slice(9);
+					page.linkshere.forEach(function(el) {
+						var tag = el.title.slice(9);
 						var tag_re = new RegExp('\\{\\{' + Morebits.pageNameRegex(tag) + '\\s*(\\|[^}]*)?\\}\\}\\n?');
 						if (tag_re.test(pageText)) {
 							pageText = pageText.replace(tag_re, '');
@@ -1432,7 +1435,7 @@ Twinkle.tag.callbacks = {
 					});
 					if (!removed) {
 						Morebits.status.warn('Info', 'Failed to find {{' +
-						$(page).attr('title').slice(9) + '}} on the page... excluding');
+						page.title.slice(9) + '}} on the page... excluding');
 					}
 
 				});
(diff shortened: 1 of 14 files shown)


modules/twinklefluff.js
```diff
@@ -748,7 +756,7 @@ Twinkle.fluff.callbacks = {
 			if (apiobj.params.reviewRevert) {
 				var query = {
 					'action': 'review',
-					'revid': $edit.attr('newrevid'),
+					'revid': edit.attr('newrevid'),
```
siddharthvp, line 759: Aha!

Amorymeltzer: Heh, good eye — caught one!

Amorymeltzer:
BTW @siddharthvp lmk if you're not using the test.wiki files atm, I can push these up, nbd either way.

siddharthvp:
@Amorymeltzer Sure go ahead. I always forget rollbacking test edits on testwiki.

Amorymeltzer:
{{done}}

modules/twinkledeprod.js
```diff
@@ -48,38 +48,43 @@ Twinkle.deprod.callback = function() {
...
+		'inprop': [ 'protection' ],
+		'format': 'json'
 	};
 
 	var statelem = new Morebits.status('Grabbing list of pages');
 	var wikipedia_api = new Morebits.wiki.api('loading...', query, function(apiobj) {
-		var $doc = $(apiobj.responseXML);
-		var $pages = $doc.find('page[ns!="6"]');  // all non-files
+		var response = apiobj.getResponse();
+		var pages = response.query.pages.filter(function(p) {
```
siddharthvp, line 58: Breaks when invoked on an empty category

siddharthvp (commented):
Needless to say, this looks hella great! I guess this would speed up things a tiny bit too as all the jquery xml parsing is going away. 

I'm wondering do we really want to use `.getResponse()` all over the place, maybe just use `.response`, as in JS everything is anyway public. 

Also, we should note somewhere in the guidelines that morebits.wiki.api applies formatversion=2 behind the scene, as otherwise the output from the ApiSandbox (which sadly uses formatversion=1 by default) doesn't match, and it wouldn't be clear to a new contributor why.

modules/friendlytag.js
```diff
@@ -1415,13 +1415,13 @@ Twinkle.tag.callbacks = {
...
+				'lhlimit': 'max', // 500 is max for normal users, 5000 for bots and sysops
+				'format': 'json'
 			}, function removeRedirectTag(apiobj) {
-
-				$(apiobj.responseXML).find('page').each(function(idx, page) {
+				apiobj.getResponse().query.pages.forEach(function(page) {
 					var removed = false;
-					$(page).find('lh').each(function(idx, el) {
-						var tag = $(el).attr('title').slice(9);
+					page.linkshere.forEach(function(el) {
```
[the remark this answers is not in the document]
Amorymeltzer: Good catch.  `filter`ed it out, as well as items without a `linkshere`

modules/twinkledeprod.js
Amorymeltzer: Indeed!  Handled in batchdelete and batchprotect, but not here or batchundelete, so fixed x2.

Amorymeltzer:
> Also, we should note somewhere in the guidelines that morebits.wiki.api applies formatversion=2 behind the scene, as otherwise the output from the ApiSandbox (which sadly uses formatversion=1 by default) doesn't match, and it wouldn't be clear to a new contributor why.

I thought I did this when adding #1179, but I'm realizing now that I think I held off for want of conflicts with #1190, and it slipped through. Added in 31b8721, sufficient?

> I'm wondering do we really want to use `.getResponse()` all over the place, maybe just use `.response`, as in JS everything is anyway public.

¯\\\_(ツ)\_/¯  Six and one half.  Looks better alongside all the other methods?  Honestly I think I did it for diffs and regex.

siddharthvp (approved):
Cool. All LGTM.

PR #1227, opened 2020-12-07
tag: fix minor bug (from #1216)

siddharthvp:
Affects {{Expand language}} only. This tag is weird in that for some reason we want to add the parameter topic= with no value into the wikitext. #1216 has a somewhat hacky way to accommodate this situation. But it had a bug in that undefined was being put into the wikitext instead of the empty string. That is, it was adding `{{Expand language|topic=undefined}}` when `{{Expand language|topic=}}` is desired.

[Example affected edit](https://en.wikipedia.org/w/index.php?title=Byron_Hyland&diff=next&oldid=919634235&diffmode=source)

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1470,7 +1470,7 @@ Twinkle.tag.callbacks = {
 					var subgroups = Array.isArray(subgroupObj) ? subgroupObj : [ subgroupObj ];
 					subgroups.forEach(function(gr) {
 						if (gr.parameter && (params[gr.name] || gr.required)) {
-							currentTag += '|' + gr.parameter + '=' + params[gr.name];
+							currentTag += '|' + gr.parameter + '=' + (params[gr.name] || '');
 						}
 					});
 				}
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -781,6 +781,7 @@ Twinkle.tag.article.tagList = {
 					type: 'hidden',
 					name: 'expandLangTopic',
 					parameter: 'topic',
+					value: '',
 					required: true // force empty topic param in output
 				}, {
 					name: 'expandLanguageLangCode',


Amorymeltzer:
I think you mean #1216/853bedc.

modules/friendlytag.js
```diff
@@ -1470,7 +1470,7 @@ Twinkle.tag.callbacks = {
 					var subgroups = Array.isArray(subgroupObj) ? subgroupObj : [ subgroupObj ];
 					subgroups.forEach(function(gr) {
 						if (gr.parameter && (params[gr.name] || gr.required)) {
-							currentTag += '|' + gr.parameter + '=' + params[gr.name];
+							currentTag += '|' + gr.parameter + '=' + (params[gr.name] || '');
```
Amorymeltzer, line 1473: This sadly won't solve the issue, since the issue is that `params[gr.name]` actually *is* defined, as the string `undefined` i.e. `{expandLangTopic: 'undefined'}`.

Amorymeltzer (changes requested):

modules/friendlytag.js
siddharthvp: Wow, really? It works file on my end (Chrome/Windows). Hidden inputs should take the value of an empty string. But might be platform-dependent. I've manually put it in `value: ''` in there. Can you try now?

siddharthvp: Oh okay, it looks like it's a morebits bug that the value is rendered into the HTML as `undefined`, and I had some stary quickform changes locally which somehow cancelled out the bug.

siddharthvp:
> I think you mean #1216 / [853bedc](https://github.com/azatoth/twinkle/commit/853bedc99f137dc94bc9e95bb80d84b252248783).

Indeed, fixed.

modules/friendlytag.js
Amorymeltzer: Yeah, setting `value: ''` worked for me earlier.  Although, since this is the first instance, wouldn't the more proper way be to do `if (data.value)` in the case of `hidden`?

siddharthvp: morebits has
```
			node.values = data.value;
			node.setAttribute('value', data.value);
```

in the case of hidden. Are you suggesting wrapping these two lines with the `if(data.value)`?

Not sure what the `node.values` thing is.

Amorymeltzer: Yes, wasn't that what you were implying?  It's done for the `input`ish (sub)nodes in `input` and `textarea`, and sort of in `submit` and `button`, but (notably?) not `checkbox`/`radio`.

`node.values` is beyond me atm, but I do see that it's used throughout as well.

Anyway, adding `value: ''` is simple, and doesn't establish a precedent or attempt to improve anything, so probably fine here for ease of not breaking anything.  Would be good to note, though, in case anyone wants to tackle `compute` (should be possible to write tests in the cli repl...)

Amorymeltzer: [`node.values` originated] way back in 2008 with the message:

>for checkboxes, radiobuttons and selects, add the given value to an other variable as well (element.values), to allow complex values, i.e. value: {a:b,c:d}, using this if present in getChecked

siddharthvp: Yep. Let's leave quickForm changes for the later. 

> but (notably?) not checkbox/radio.

Because of this, I _think_ we've got a couple of cases where value=undefined comes up in the HTML (though harmlessly), and a [couple of cases](https://github.com/azatoth/twinkle/blob/31b8721eac3aba8a0d4e3baace2b43812203f133/modules/twinkleblock.js#L286) where we put a silly `value: '1'` just to prevent that from happening. 

Also maybe it would make sense to just put `data.value = data.value || ''` somewhere at the top of quickForm's `compute`.

Amorymeltzer: Mmm, good point, forgot about those.  Hey, precedent!

PR #1228, opened 2020-12-08
tests: Expand Morebits.date tests, begin basic unit tests for Morebits.taskManager

Amorymeltzer:
Expands the `Morebits.date` tests to include some representative methods, notably: getters, `isAfter`/`isBefore`, the month header functions, and a couple non-UTC timezone formatting.  It also fixes the issue where testing the basic constructor (`new Morebits.date()` or `new Date()`) would occasionally (~0.068% of the time on my laptop) fail as the two times would be off by ` in the millisecond portion.  This creates a new assertion that allows values of +1 or -1 to pass, used just there.

The `taskManager` tests don't make use of the `steps`/`verifySteps` assertions mainly because I didn't like the way I had to duplicate function creation.

A few minor cleanups as well.

### DIFF
--- a/tests/morebits.date.js
+++ b/tests/morebits.date.js
@@ -1,26 +1,66 @@
 QUnit.module('Morebits.date');
+// Allow off-by-one values in milliseconds for not-quite-simultaneous date contructions
+QUnit.assert.pmOne = function(actual, expected, message) {
+	this.pushResult({
+		result: actual === expected || actual === expected+1 || actual === expected-1,
+		actual: actual,
+		expected: expected,
+		message: message
+	});
+};
+
 var now = Date.now();
 var timestamp = '16:26, 7 November 2020 (UTC)';
 QUnit.test('Construction', assert => {
-	assert.strictEqual(new Morebits.date().getTime(), new Date().getTime(), 'Basic constructor');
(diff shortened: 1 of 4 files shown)


tests/morebits.date.js
```diff
@@ -1,26 +1,71 @@
...
-	assert.strictEqual(date.format('dddd D MMMM YY h:mA', 'utc'), 'Saturday 7 November 20 4:26PM', 'Some weirder stuff');
-	assert.strictEqual(date.format('MMt[h month], [d]a[y] D, h [o\'clock] A', 'utc'), '11th month, day 7, 4 o\'clock PM', 'Format escapes');
+QUnit.test('Methods', assert => {
+	assert.true(date.isValid(), 'Valid');
+	// Logs a message; not a failure, but annoying
+	assert.false(new Morebits.date('no').isValid(), 'Invalid');
+
+	// non-UTC timezone to test getUTC
+	var local = new Morebits.date('2020-11-30T18:42:00-0800');
+	assert.strictEqual(local.getDayName(), 'Monday', 'getDayName');
```
Amorymeltzer, line 30: *sigh* Works on-wiki, on my local node, etc.  I'll look into it.

PR #1232, opened 2020-12-11
xfd: rfd: unbreak target page notification for userspace

siddharthvp:
If the target's talk page is in user talk space, the notification wasn't going through, though it appears to be unintentional from the code.


Amorymeltzer (approved):
Ahh indeed!  I see it now, absolutely right.  💯

PR #1233, opened 2020-12-11
morebits: page: allow binding to existing statusElement

siddharthvp:
This makes it possible for a morebits.wiki.page object to use an existing Morebits.status object rather than always creating a new one.

This can be done either using `.setStatusElement()` or by just providing a Morebits.status object to the constructor as second argument rather than a string.


Amorymeltzer:
Mmmm, nice.  I haven't thought through the ramifications yet, but seems useful.

siddharthvp:
Thinking about this again, while the constructor improvement will definitely help improve status handling, the `setStatusElement` seems mostly unnecessary but provided anyway for max flexibility.

PR #1235, opened 2020-12-12
xfd: logging: minor tweaks

siddharthvp:
Put the "notifed {{user|1=USERNAME}}" text at the end – just before the reason text.

This means there's just one area in which the text is different for different modes. The variable text is sandwiched between two pieces of non-variable texts. In such a setup, it's a lot easier to provide the option for localisation.

The only downside I see is that for MfD userspace owner + creator notification, we can't say "; notified X and Y" but only "; notified X; notified Y". Not a big deal.

Also:
- added a colon after the [[page name]] before the [[nominated]] link (otherwise it seems as if "nominated" is part of the page name).
- Removed the colon before the name of the 2nd page name in CfD.


Amorymeltzer (approved):
All seems fine to me!

PR #1237, opened 2020-12-12
mb.wikitext.page: Throw for unprovided options, allow no regex flags on insertAfterTemplates

Amorymeltzer:
- morebits.wikitext.page: Throw for unprovided options.  Was part of #1022 for `insertAfterTemplates`, but wasn't present for `removeLink`, `commentOutImage`, `addToImageComment`, and `removeTemplate`.
- morebits.wt.p.insertAfterTemplates: Allow '' for no regex flags.  Previously, all falsey values would default to `'i'` for the regex, but that meant there was no way to *not* apply any flags.  Can't imagine anyone would want to.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -4623,7 +4623,8 @@ Morebits.wikitext.page.prototype = {
 	 * @param {string} tag - The tag to be inserted.
 	 * @param {string|string[]} regex - Templates after which to insert tag,
 	 * given as either as a (regex-valid) string or an array to be joined by pipes.
-	 * @param {string} [flags=i] - Regex flags to apply.
+	 * @param {string} [flags=i] - Regex flags to apply.  `''` to provide no flags;
+	 * other falsey values will default to `i`.
 	 * @param {string|string[]} [preRegex] - Optional regex string or array to match
 	 * before any template matches (i.e. before `{{`), such as html comments.
 	 * @returns {Morebits.wikitext.page}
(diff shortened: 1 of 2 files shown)


morebits.js
```diff
@@ -4595,7 +4596,7 @@ Morebits.wikitext.page.prototype = {
 			regex = regex.join('|');
 		}
 
-		flags = flags || 'i';
+		flags = flags || flags === '' ? '' : 'i';
```
siddharthvp, line 4599: Probably cleaer to write as `if (typeof flags !== 'string') flags = 'i';`

siddharthvp:
I don't think the rest is needed. If #1220 is merged, when developers use these functions wrongly, they will conveniently be flagged with red underlines in code editors. Enforcing the types with JS bloats up the final JS payload with unreachable code (from the end-user point of view).

morebits.js
Amorymeltzer: Much!  Done.

Amorymeltzer:
It's a library, so even if #1220 is merged, that won't do anything for anyone without that set up in their IDE, incorporating it elsewhere, or editing on-wiki.

siddharthvp:
That still misses the point of why would anyone expect these functions to work without args in the first place? It's useful to throw for functions where the args are non-intuitive, but here it's simply clutter.

Amorymeltzer:
There are plenty of things that *do* work without a parameter or with a default, so it's not crazy.  Clutter is only clutter if it offers nothing.  More to the point, I think "here's what you did wrong" is more helpful than "Cannot read property 'whatever' of undefined," but if you feel that adamant, it's a minor point. e8dd56b merged as 00b3c57.

siddharthvp:
Hmm. I had expected you to go ahead regardless. I am not adamant about this. Feel free to merge if you want to. I mostly wrote that in a "just saying" sense and because it seemed like naively trying to preempt ... another PR.

PR #1238, opened 2020-12-12
morebits.pageNameRegex: Properly escape entire string, use in Morebits.wikitext.page

Amorymeltzer:
There's more to matching a page name than just the first character; this applies `Morebits.string.escapeRegExp` (spaces/underscores to `[_ ]` and all the loveliness from `mw.util.escapeRegExp`) as well.  Underutilized since its inception, it's only used in the tag module, so I don't think there should be anything relying on it *not* matching pages it should be.  It is, however, perfect for the `Morebits.wikitext.page` methods.


siddharthvp:
👍 I think the reason I didn't do the regex escaping from the start was from a misinformed notion of performance and simplicity. Still, I don't think the default value or the error check are useful. NOt a big deal either way.

morebits.js
```diff
@@ -131,14 +131,26 @@ Morebits.pageNameNorm = mw.config.get('wgPageName').replace(/_/g, ' ');
...
 Morebits.pageNameRegex = function(pageName) {
-	return '[' + pageName[0].toUpperCase() + pageName[0].toLowerCase() + ']' + pageName.slice(1);
+	pageName = pageName || mw.config.get('wgTitle');
+	if (typeof pageName !== 'string') {
+		throw new Error('Not a valid string');
+	}
+	var firstChar = Morebits.string.escapeRegExp(pageName[0]),
+		remainder = Morebits.string.escapeRegExp(pageName.slice(1));
+	if (firstChar.toUpperCase() !== firstChar.toLowerCase()) {
+		return '[' + pageName[0].toUpperCase() + pageName[0].toLowerCase() + ']' + remainder;
```
siddharthvp, line 151: This is still incomplete. `pageName[0].toUpperCase()` should probably be written as `mw.Title.phpCharToUpper(pageName[0])` to account for caps differences between PHP and JS.

Amorymeltzer: Cool, didn't know about `phpCharToUpper`!  I almost left a note about this was guaranteed to be imperfect given utf-8, but that probably(?) reduces errors to much much rarer edge cases, that I'm not sure MediaWiki itself handles.

Amorymeltzer: Worth doing for `Morebits.string.toUpperCaseFirstChar`, presumably

Amorymeltzer: Done, now.  Incorporated the json file for our mocking.

Also, the cases where `escapeRegExp` does something are never cases where the upper and lower differ, so that was unnecessary.

Amorymeltzer:
I'm wary of introducing more stuff that assumes you're on a page, but I figured it was a smart default that could encourage more use.

Amorymeltzer:
I guess I'm wondering what to do when nothing or `''` is passed.  I kind of want to `return ''` on the latter, so maybe it'd be smarter to skip `wgTitle` and do that?

Amorymeltzer:
I've gone and done that.

Amorymeltzer:
I suppose I could argue it likewise shouldn't treat leading and trailing whitespace/`_` nicely, but I feel like if you're there, things have already gone wrong...

morebits.js
siddharthvp: Cool. There shouldn't be any edge cases that MW doesn't handle. Any char that gives different uppercases on PHP and JS is included in that file.

Amorymeltzer: Yeah, that's what I was saying — anything this would be wrong on MW itself is going to be.

PR #1242, opened 2020-12-13
morebits: Force loading of a page if a watchlist expiry is provided

Amorymeltzer:
Tried a few different things, but after thinking about it for a bit longer, I think this is just the only really viable option, at least unless/until the information is exposed in an easier way (T270057).  Would largely revert #1241 as unnecessary, opting to build on/expand #1230.  Still imperfect since there's no information about whether the page is watched indefinitely or with an expiry (T268834), but at least this means we'll never overwrite indefinite with expiry; it also means we won't extend an expiration.

To do *that*, right now I think the only way would be to rewrite everything to not use the `watchlist` option, but rather a new callback with `action=watch`, although that doesn't account for userpreferences well, and would, of course, be awful.


Amorymeltzer:
@MusikAnimal I don't mean to mix your business and pressure, but wanted to ping you in case I missed something obvious.

MusikAnimal:
No worries and thank you for highlighting this issue. I think [T268834](https://phabricator.wikimedia.org/T268834) definitely needs to be addressed, and I've already started on a patch. The iOS team was asking for the same thing, actually. I'll hopefully have that up for review soon, but in all likelihood it won't get deployed until the new year due to the usual code freeze in late December.

I left a comment at [T270057](https://phabricator.wikimedia.org/T270057).

I don't think you're missing anything obvious. I don't completely follow what this PR does, but whatever workaround is probably worth the while if you need a solution soon.

Amorymeltzer:
Appreciate that!  FWIW this basically just short-circuits any efficiencies where we didn't need to load a page, such as when appending a notice or or new section or something, and instead forces a wait on the watched status query.

PR #1243, opened 2020-12-14
morebits: use blue colour for "Redirected from" message

siddharthvp:
Morebits.status.info() creates a green-coloured message, which should ideally be reserved for success messages.

 Blue colour is suitable for other info. The "Redirected form X to Y" message being in green, it's easy to miss among the rest.


Amorymeltzer (approved):
Seems reasonable!

Issue #1049, opened 2020-07-17
Avoid hanging status messages


siddharthvp:
There are some status messages around that appear to be stuck in a 'doing' stage and never resolve(become green), even when the underlying task has finished without issues. Eg. looking up of author in speedy module.

siddharthvp:
This is happening in protect module as well.

PR #1244, opened 2020-12-14
XfD rewrite 

siddharthvp:
Rewrite of the XfD module. Each XfD process is represented as a class, all of them inheriting a common `XfDMode` class (should we call that `XfDBase` instead?). No more switch statements anywhere. Every existing feature of the previous XFD module is supported. All code needed for a venue is together, making it easy to add and remove venues. This rewrite should greatly facilitate localisation, is faster thanks to Morebits.taskManager which allows for precise scheduling, and avoids issues like #364.

`XfDMode` implements some base methods a few of which are directly used by the subclasses (eg. `fetchCreatorInfo`, `notifyCreator`, `autoEditRequest`), some implement all common functionality and provide hooks for the subclasses to interfere (eg. `addToLog` with the hook `getUserspaceLoggingExtraInfo`). Some are used as-is in most venues but overridden in specific venues (eg. `preview` and `showPreview`). `getMenuTooltip` methods elegantly handles #1193. `preprocessParams` methods process the form data - these were previously being duplicated in `evaluate` and `preview` functions. No longer the case.

Possible future improvements:
- The evaluate method of every venue looks similar. It's just the taskManager description, and code to replicate Morebits.wiki.actionCompleted redirect. Maybe this could be abstracted out into a function.
- Haven't looked into it, but some common parts of page tagging methods could also probably be abstracted out. 
- I don't like the Morebits.status.onError much. It's a lot better than what we had before (no need for hacks like #1093 and #1153 is now unnecessary), but it'd still better if somehow only the failure of tasks _in path of_  `createDiscussionPage`/`addToList` would trigger printUserText. Maybe that's something to think about adding to taskManager. 

Does not yet take into account #1198 or #1224. Not yet integrated #1208 for TfDs. 

Depends on #1220, which made this a _lot_ easier to write.


siddharthvp:
This should work exactly the same as the current XFD module. I have not made any behaviour changes here, opting to open them as separate PRs (#1232, #1235, #1236). The only behaviour changes are a few fixes of status message handling (to avoid #1049), facilitated by #1233.

modules/xfd.js
```diff
@@ -0,0 +1,2052 @@
...
+        var params = this.params;
+        var def = $.Deferred();
+        var notifyTitle = mw.Title.newFromText(notifyTarget, 3); // 3: user talk
+        var targetNS = notifyTitle.getNamespaceId();
+        var usernameOrTarget = notifyTitle.getRelativeText(3);
+        statusElement = statusElement || new Morebits.status('Notifying initial contributor (' + usernameOrTarget + ')');
+        var notifyPageTitle = notifyTitle.toText();
+        if (targetNS === 3) {
+            // Disallow warning yourself
+            if (usernameOrTarget === mw.config.get('wgUserName')) {
```
DannyS712, line 258: suggest combining this check with the targetNS check

modules/xfd.js
```diff
@@ -0,0 +1,2052 @@
...
+    XfdMode.prototype.notifyCreator = function () {
+        if (!this.params.notifycreator) {
+            this.params.intialContrib = null;
+            return $.Deferred().resolve();
+        }
+        return this.notifyTalkPage(this.params.initialContrib);
+    };
+    // Should be called after notifyTalkPage() which may unset this.params.intialContrib
+    XfdMode.prototype.addToLog = function () {
+        var params = this.params;
```
DannyS712, line 302: params is only needed once, suggest using this.params.venue instead of needing a temporary variable

modules/xfd.js
```diff
@@ -0,0 +1,2052 @@
...
+            '.select2-results .select2-results__group { padding-top: 1px; padding-bottom: 1px; } ' +
+            // Adjust font size
+            '.select2-container .select2-dropdown .select2-results { font-size: 13px; }' +
+            '.select2-container .selection .select2-selection__rendered { font-size: 13px; }' +
+            // Make the tiny cross larger
+            '.select2-selection__choice__remove { font-size: 130%; }');
+    };
+    Afd.prototype.evaluate = function () {
+        var _this = this;
+        var tm = _super.prototype.evaluate.call(this);
```
DannyS712, line 480: what does `tm` stand for? Similar to above, suggest more informative variable names

modules/xfd.js
```diff
@@ -0,0 +1,2052 @@
...
+            else {
+                var number = 0;
+                for (var i = 0; i < titles.length; ++i) {
+                    var title = titles[i].getAttribute('title');
+                    // First, simple test, is there an instance with this exact name?
+                    if (title === 'Wikipedia:Articles for deletion/' + Morebits.pageNameNorm) {
+                        number = Math.max(number, 1);
+                        continue;
+                    }
+                    var order_re = new RegExp('^' +
```
DannyS712, line 523: since the creation of a new RegExp object is likely a bit expensive, and since it doesn't change each time through the `for` loop, suggest creating once outside of the loop

modules/xfd.js
```diff
@@ -0,0 +1,2052 @@
...
+                        Morebits.string.escapeRegExp('Wikipedia:Articles for deletion/' + Morebits.pageNameNorm) +
+                        '\\s*\\(\\s*(\\d+)(?:(?:th|nd|rd|st) nom(?:ination)?)?\\s*\\)\\s*$');
+                    var match = order_re.exec(title);
+                    // No match; A non-good value
+                    if (!match) {
+                        continue;
+                    }
+                    // A match, set number to the max of current
```
DannyS712, line 531: can combine the `if (!match)` etc. into just

```
if (match) {
    number = Math.max(number, Number(match[1]));
}
```
which may be simpler

modules/xfd.js
```diff
@@ -0,0 +1,2052 @@
...
+                pageobj.save(defs[idx].resolve, defs[idx].resolve); // failures aren't important
+            });
+        });
+        return $.when.apply($, defs);
+    };
+    Afd.prototype.patrolPage = function () {
+        if (Twinkle.getPref('markXfdPagesAsPatrolled')) {
+            new Morebits.wiki.page(Morebits.pageNameNorm).triage();
+        }
+        return $.Deferred().resolve(); // XXX
```
DannyS712, line 691: should this not wait until after patrolling is done? (If this is a carry-over from the prior behavior, feel free to ignore)

modules/xfd.js
```diff
@@ -0,0 +1,2052 @@
...
+        }
+        // Modules can't be tagged, TfD instructions are to place on /doc subpage
+        this.params.scribunto = mw.config.get('wgPageContentModel') === 'Scribunto';
+        if (this.params.xfdcat === 'tfm') {
+            this.params.otherTemplateName = (this.params.scribunto ? 'Module:' : 'Template:') + this.params.tfdtarget;
+        }
+    };
+    Tfd.prototype.evaluate = function () {
+        var _this = this;
+        var tm = _super.prototype.evaluate.call(this);
```
DannyS712, line 789: same question about what `tm` means - not going to comment this every time a more informative variable name can be used, but applies elsewhere too

modules/xfd.js
```diff
@@ -0,0 +1,2052 @@
...
+    Tfd.prototype.getNotifyText = function () {
+        var text = "{{subst:tfd notice";
+        if (this.params.xfdcat === 'tfm') {
+            text = '\n{{subst:Tfm notice|2=' + this.params.tfdtarget;
+        }
+        text += "|1=" + Morebits.pageNameNorm + "}} ~~~~";
+        return text;
+    };
+    Tfd.prototype.getUserspaceLoggingExtraInfo = function () {
+        var params = this.params, text = '';
```
DannyS712, line 981: some places declare multiple variables with one variable per line, others combine them - suggest being consistent (personally, I prefer one per line)

modules/xfd.js
```diff
@@ -0,0 +1,2052 @@
...
+                _this.autoEditRequest(pageobj).then(def.resolve, def.reject);
+            }
+        });
+        return def;
+    };
+    Ffd.prototype.addToList = function () {
+        var _this = this;
+        var params = this.params;
+        var def = $.Deferred();
+        var wikipedia_page = new Morebits.wiki.page(params.logpage, "Adding discussion to today's list");
```
DannyS712, line 1072: same as before, suggest camelCase

siddharthvp:
Hi @DannyS712, sorry for not mentioning it outright here, but this code is written in TypeScript (`.ts`) and the `.js` file is just the transpilation. 

Most of your comments seem to be from the insides of functions (such as caps inconsistencies) – they're copied them from the original `twinklexfd.js` file, and I am holding off from making unnecessary changes to them as far as possible.

modules/xfd.js
siddharthvp: Yes, but Morebits.wiki.page#triage doesn't support callbacks or promises (till now), so there's no way to wait.

DannyS712:
> Hi @DannyS712, sorry for not mentioning it outright here, but this code is written in TypeScript (`.ts`) and the `.js` file is just the transpilation.
> 
> Most of your comments seem to be from the insides of functions (such as caps inconsistencies) – they're copied them from the original `twinklexfd.js` file, and I am holding off from making unnecessary changes to them as far as possible.

Oh, oops - wasn't paying attention to the fact that this now uses typescript. Feel free to ignore the prior comments on the js, let me take a look at the actual ts source

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+		// @ts-ignore
+		this.mode = new mode();
+		this.mode.result = this.result;
+		this.mode.Window = this.Window;
+
+		$('#wrong-venue-warn').text(this.mode.getVenueWarning() || '');
+		form.previewer.closePreview();
+
+		let fieldset = this.mode.generateFieldset();
+		let renderedFieldset = fieldset.render();
```
DannyS712, line 133: `fieldset` is only used once, suggest combining

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+		this.mode.postRender(renderedFieldset as HTMLFieldSetElement);
+	}
+}
+
+
+abstract class XfdMode {
+	static venueCode: string
+	static venueLabel: string
+
+	// must be overridden, unless the venue is never the default choice
```
DannyS712, line 145: "unless..." means that it isn't that this method "**must** be overridden", only that it generally should be

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+	}
+
+	autoEditRequest(pageobj: Morebits.wiki.page) {
+		let params = this.params;
+		let def = $.Deferred();
+		var talkName = new mw.Title(pageobj.getPageName()).getTalkPage().toText();
+		if (talkName === pageobj.getPageName()) {
+			pageobj.getStatusElement().error('Page protected and nowhere to add an edit request, aborting');
+			return def.reject();
+		} else {
```
DannyS712, line 246: since the `if` branch has a return, no need to wrap in an `else` block

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+			pageobj.getStatusElement().warn('Page protected, requesting edit');
+
+			var editRequest = '{{subst:Xfd edit protected|page=' + pageobj.getPageName() +
+				'|discussion=' + params.discussionpage + '|tag=<nowiki>' + params.tagText + '\u003C/nowiki>}}'; // U+003C: <
```
DannyS712, line 250: same comment as from the javascript about documenting why U+003C must be used manually

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+
+			var talk_page = new Morebits.wiki.page(talkName, 'Automatically posting edit request on talk page');
```
DannyS712, line 252: same suggestion regarding camelCase variable naming

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+		usl.changeTags = Twinkle.changeTags;
+		return usl.log(this.getUserspaceLoggingText(), this.getUserspaceLoggingEditSummary());
+	}
+
+	getUserspaceLoggingEditSummary() {
+		return 'Logging ' + utils.toTLACase(this.params.venue) + ' nomination of [[:' + Morebits.pageNameNorm + ']].';
+	}
+
+	getUserspaceLoggingText(): string {
+		let params = this.params;
```
DannyS712, line 366: definitely not too familiar with typescript, but is there a functional difference here between `let` and `var`? If not, suggest being consistent one way or the other

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+
+class Afd extends XfdMode {
+	static venueCode = 'afd';
+	static venueLabel = 'AfD (Articles for deletion)';
+
+	static isDefaultChoice() {
+		return mw.config.get('wgNamespaceNumber') === 0;
+	}
+
+	getFieldsetLabel() {
```
DannyS712, line 401: why not have the field set label as a static value, like venueCode and venueLabel?

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+
+	getVenueWarning() {
+		if (mw.config.get('wgNamespaceNumber') !== 0) {
+			return 'AfD is generally appropriate only for articles.';
+		} else if (mw.config.get('wgIsRedirect')) {
+			return 'Please use RfD for redirects.';
+		}
+	}
+
+	getMenuTooltip() {
```
DannyS712, line 413: same question about static value for menu tooltip

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+			'.select2-container .select2-dropdown .select2-results { font-size: 13px; }' +
+			'.select2-container .selection .select2-selection__rendered { font-size: 13px; }' +
+
+			// Make the tiny cross larger
+			'.select2-selection__choice__remove { font-size: 130%; }'
+		);
+	}
+
+	evaluate() {
+		let tm = super.evaluate();
```
DannyS712, line 545: please use a more informative variable name for `tm`

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+		tm.add(this.fetchCreatorInfo, []);
+		tm.add(this.notifyCreator, [this.fetchCreatorInfo]);
+		tm.add(this.notifyUserspaceOwner, [this.fetchCreatorInfo]);
+		tm.add(this.addToLog, [this.notifyCreator, this.notifyUserspaceOwner]);
+		tm.execute().then(() => this.redirectToDiscussion());
+	}
+
+	determineDiscussionPage() {
+		let params = this.params;
+		let wikipedia_api = new Morebits.wiki.api('Looking for prior nominations of this page', {
```
DannyS712, line 1647: same suggestion for camelCase

modules/xfd.ts
```diff
@@ -0,0 +1,2218 @@
...
+				for (var i = 0; i < titles.length; ++i) {
+					var title = titles[i].getAttribute('title');
+
+					// First, simple test, is there an instance with this exact name?
+					if (title === 'Wikipedia:Miscellany for deletion/' + Morebits.pageNameNorm) {
+						number = Math.max(number, 1);
+						continue;
+					}
+
+					var order_re = new RegExp('^' +
```
DannyS712, line 1673: same comment as on the javascript - suggest creating the RegExp object outside of the for loop

modules/xfd.ts
siddharthvp: I used `let` in all new code, and retained `var`s from the code copied from the old version. Maybe just do a find-and-replace of all `var`s with `let`?

modules/xfd.ts
siddharthvp: It has to do with how it's invoked – [on line 162](https://github.com/siddharthvp/twinkle/blob/8e12b1070f3a104fc37ee4e285c45e3933fb177f/modules/xfd.ts#L162). Here we call `getFieldsetLabel` on `this` without actually knowing what class `this` is an object of – it could be any of `Afd`, `Tfd`, etc.

modules/xfd.ts
siddharthvp: Same as above. It's invoked on an object reference rather than a class reference.

modules/xfd.ts
siddharthvp: You're right. But this [predates the initial github import 10 years ago](https://github.com/azatoth/twinkle/commit/71267fd4f1a0057c01ddc9ec5eb8ea40cf2a75be#diff-b0bf7dacc07995995276a293b020a474a08313a5f921eb1d7f623fdcd3b6394cR406)!

modules/xfd.ts
siddharthvp: Hmm. The `Morebits.taskManager` object has been consistently referred to as `tm`. Maybe using `taskManager` everywhere would be better.

modules/xfd.ts
siddharthvp: I take your point that per the spirit of OOP, the method should be static, as the value is the same for all objects constructed from the class. In a truly object-oriented language like Java, this could've been made static since `cls.staticMethod()` and `obj.staticMethod()` both work (in Java), but in JS/TS, static methods can unfortunately only be called on the class itself, not on the objects.

modules/xfd.ts
siddharthvp: This was added in #1091 though [on my suggestion](https://github.com/azatoth/twinkle/pull/1091#discussion_r507613528). In hindsight, yeah it was a stupid suggestion. Much better to do `'</no' + 'wiki>'`

siddharthvp:
Development of localised Twinkle is now taking place on https://github.com/wikimedia-gadgets/twinkle-core and https://github.com/wikimedia-gadgets/twinkle-enwiki repos.

PR #1250, opened 2020-12-20
batchdelete: fix type for page namespace

siddharthvp:
JSON API gets the page namespace as number, not string.

Stems from #1224.


Amorymeltzer (approved):
Nice catch!  There was another case in here, merged as bd043e0

PR #1251, opened 2020-12-20
batch/deprod/unlink: Sort lists of pages by namespace then title

Amorymeltzer:
Noted at [WT:TW](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=995362073#Dbatch) by @Siddharthvp, but apparently json `formatversion=2` preferentially sorts by `pageid`, whereas xml and `formatversion=1` sorts by namespace then title.  This manually reintroduces that sorting.

Technically, it's not identical, as xml/`formatversion=1` sorts preferentially by existence, but that isn't hugely important for our purposes.

### DIFF
--- a/modules/twinklebatchdelete.js
+++ b/modules/twinklebatchdelete.js
@@ -171,6 +171,10 @@ Twinkle.batchdelete.callback = function twinklebatchdeleteCallback() {
 		pages = pages.filter(function(page) {
 			return !page.missing && page.imagerepository !== 'shared';
 		});
+		// json formatversion=2 doesn't sort pages by namespace
+		pages.sort(function(one, two) {
+			return one.ns - two.ns || (one.title > two.title ? 1 : -1);
+		});
 		pages.forEach(function(page) {
 			var metadata = [];
 			if (page.redirect) {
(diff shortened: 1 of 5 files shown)


modules/twinklebatchdelete.js
```diff
@@ -171,6 +171,10 @@ Twinkle.batchdelete.callback = function twinklebatchdeleteCallback() {
 		pages = pages.filter(function(page) {
 			return !page.missing && page.imagerepository !== 'shared';
 		});
+		// json formatversion=2 doesn't sort pages by namespace
+		pages = pages.sort(function(one, two) {
+			return one.ns - two.ns || one.title > two.title;
```
siddharthvp, line 176:
```suggestion
			return one.ns - two.ns || (one.title > two.title ? 1 : -1);
```

sort's compareFn should return a number

Amorymeltzer: Word.  Done.

Amorymeltzer:
Added unlink, forgot that initially.

I'm thinking this should be pulled out to twinkle.js, as a `Twinkle.sortNamespaceTitle` or something, as this uses it a half-dozen times.

PR #1258, opened 2020-12-25
speedy/image: don't share addToLog

siddharthvp:
Speedy module needs to be changed a lot for localizability. It's no longer wise for image module to piggyback on speedy's `addToLog` function. This means some duplication but not much (particularly since #878). The integration was never elegant to begin with (noted in a comment as "params data structure when coming from DI is quite different"), so there was a large `if` separating the behaviour for the two modules.


Amorymeltzer (approved):
Sure!  Was nice to duplicate code, but it's hacky and imperfect.  Wonder if we should attempt splitting the preferences?

Amorymeltzer:
Merged as 312686f to remove unnecessary `params.fromDI` in image

Issue #1173, opened 2020-10-29
Option for IPv6 blocks to include the /64 range


CaptainEek:
Feature request. English admin here, was just blocking an IPv6 user, and was thinking how convenient it would be if one could simply check a box in the blocking interface to block the /64 when dealing with IPv6 users. Since IPv6 allocations are so large, I almost always block the /64 range, but find it annoying that I have to go to the non-Twinkle block interface and manually add the /64. I suspect I am not alone in this. For a more detailed reasoning on blocking the 64, see https://en.wikipedia.org/wiki/User:TonyBallioni/Just_block_the_/64

This of course would require Twinkle to contextually know when its dealing with an IPv6, and display a relevant option only in those cases.

L235:
As another enwiki admin I can say this would be a tremendous help

siddharthvp:
Related: #775, #924

ToBeFree:
This would be wonderful. Thanks Amory for working on this

Issue #924, opened 2020-04-16
Rollback should treat IPv6 users in the same /64 as one user


pppery:
> It also would be nice if Twinkle rollback reverted IPv6 edits on the basis of the /64, vandalism by a IPv6 switching withing its /64 range often gets left in. 

[Eostrix at Wikipedia:Village pump (technical)](https://en.wikipedia.org/wiki/Special:Diff/951300234)

QEDK:
Don't think Twinkle should implement this before MediaWiki actually implements this - https://phabricator.wikimedia.org/T112325

siddharthvp:
> Don't think Twinkle should implement this before MediaWiki actually implements this

lol no, MediaWiki could take a decade to implement that. We don't need to wait for that. We are already stuff in rollback that MediaWiki doesn't, like the bot whitelist thing.

QEDK:
> > Don't think Twinkle should implement this before MediaWiki actually implements this
> 
> lol no, MediaWiki could take a decade to implement that. We don't need to wait for that. We are already doing stuff in rollback that MediaWiki doesn't, like the bot whitelist thing.

Yes, but there's still the implication of policy. **Usually**, /64 are the smallest ranges assigned to end-users but it's also possible to assign it to small companies, large households and other institututions. We should not assume all /64 are the same user and this is reflected in MediaWiki docs as of now:

> Like in IPv4, where a single static IPv4 address may supply a household or institution of some kind, a static IPv6 /64 range may also supply a household or institution, so multiple users can be affected by /64 rangeblocks.

So, until there is a reasonable consensus that /64 should be treated as one user (a MediaWiki core change would signify that, for e.g.), no reason for Twinkle to change the status quo.

martijnhoekstra:
> no reason for Twinkle to change the status quo.

This would be totally true if the status quo were workable.

Eostrix77:
Twinkle rollback should be independent of the MediaWiki question.

In Combing talk pages on MediaWiki for the /64 (a good safe thing, but... forever) you want to be fairly certain that project-wide, in totally different pages, the same /64 is the same, or that the shared IP benefits (messages actually arriving) outweighs any possible drawback.

Twinkle rollback poses a much NARROWER question: whether CONSECUTIVE edits by the SAME /64 to the SAME ARTICLE are likely to be the same person. The answer to that is probably 99.999% yes, and it is a much easier question to answer yes to than the MediaWiki project-wide combining.

[...]

Issue #775, opened 2019-12-01
Better handling of IP ranges on Special:Contribs


Amorymeltzer:
Some discussion at [meta](https://meta.wikimedia.org/w/index.php?title=Talk:MoreMenu&oldid=19600805#IP_tools_option_missing_on_contribs_and_user_pages_for_ranges) with @MusikAnimal, but the underlying issue is that `wgRelevantUserName` does not exist for valid IP ranges; see [T206954](https://phabricator.wikimedia.org/T206954) and #476 (and [MusikAnimal/MoreMenu@13d46f8](https://github.com/MusikAnimal/MoreMenu/commit/13d46f8d0decd5ef5eae339d5e4ef5aca884fab4#diff-9bbfa14b377f5cad61e2b8ab064df9b9R69-R71) and [T211910](https://phabricator.wikimedia.org/T211910)).

Modules that Twinkle displays on contribs for IPs are `arv`, `block`, `talkback`, `warn`, and `welcome`, but right now none of those show up for valid ranges since they all rely on `wgRelevantUserName`.  That's good for warn and welcome — we don't *want* to warn or welcome ranges — but, at least *in theory*, some of the functions of arv and block should be allowed, if somewhat broken.  In particular:

- [x] Block should be enabled for IP ranges
    - [x] Shouldn't leave a block template
    - [x] Querying `bkusers` for previous blocks [doesn't work on IPv4 ranges](https://phabricator.wikimedia.org/T239527)
- [x] ARV should allow reporting ranges to AIV
    - [x] Everything else (except maybe SPI???) should be disabled, in particular AN3 reporting since `rvuser` doesn't accept IP ranges.
- [x] Don't show *anything* for ranges outside of the CIDR limit ([T211910](https://phabricator.wikimedia.org/T211910))

In the end, this is just a PITA that will complicate and obfuscate the code a fair bit. 
 Of course, depending on if and how [T206954](https://phabricator.wikimedia.org/T206954) is resolved, we'll want to *exclude* a number of these modules and their functions.

Amorymeltzer:
Should be doable starting in January sometime

Amorymeltzer:
In theory #1266 did all this, but since [T206954](https://phabricator.wikimedia.org/T206954) still isn't merged/live, and #1296 is outstanding, gonna keep this open to track ongoing/future work needed.

PR #1266, opened 2021-01-07
Improve and expand handling of IP ranges

Amorymeltzer:
Most of this should be straightforward, with the complicated changes mainly found in the block module.  Commits are as follows:

1. <s>fluff: Use `wgRelevantUserName` for IP ranges on contribs (#775) (Ends hacky reliance on `MediaWiki:Sp-contributions-footer-anon-range` from #476)</s> EDIT: moved to #1296
2. warn/welcome/talkback: Don't show for IP ranges (#775)
3. arv: Allow IP ranges but limit to AIV and SPI (`{{rvuser}}` doesn't support ranges) (#775)
    * Builds on #817/#1184/#1204 cc @DannyS712
4. block: Enable for IP ranges, improve handling of rangeblocks (#775)
    * Enable on ranges
        * Leaving a template is disabled, but the "templated reason" {{rangeblock}} is used as the default
    * Determine if IP is caught in a range block
        * Display the range, whether it's the same block, etc.
    * Use correct information for IPs covered by two blocks (e.g. a single IP and a range block)
    * Move/hide unblock link for IPs covered by ranges but not directly blocked
5. morebits: `Morebits.get64` to get the /64 subnet for an IPv6 address
6. fluff: Use Morebits.get64 to revert consecutive edits by same IPv6 /64 (#924)
    * cc @pppery @QEDK @martijnhoekstra @Eostrix77
7. block: Add checkbox to just block the /64 (#1173)
    * Obviously relies on `Morebits.get64`, but otherwise, a large part of this is just making increased use of `relevantUserName` (rather than `mw.config.get('wgRelevantUserName')`), and simply adjusting that when (un)checking the checkbox.  A number of things previously stable need to be changeable, but the overall structure isn't too different, since this form is used to reloading frequently.  The main things adjusted:
    * Unblock footerlink and window title toggled
    * Template option unchecked/disabled as necessary
    * Prior block preset reset
    * Data processing in `Twinkle.block.fetchUserInfo` is pulled out into `Twinkle.block.processUserInfo`
        * `Twinkle.block.processUserInfo` caches its results in `Twinkle.block.fetchedData`, thus allowing `Twinkle.block.fetchUserInfo` to only be rerun as necessary on multiple togglings
    * cc @ToBeFree @L235 @molly @MusikAnimal
8. Added `Morebits.validCIDR` to test if an IP range is within CIDR limits; use to limit block and arv
    * This is operating under the assumption that ranges that are too wide but are valid will still show the username.
    

The first four depend on [T206954](https://phabricator.wikimedia.org/T206954) being fully implemented to be of any use, but I *think* only the first one (fluff) should have any material harm if done beforehand (EDIT: This commit has been moved to #1296).  The last three (the /64 ones) should work without it, so in theory could be separate, but changes to the block module as written depend on the prior commit.  There's plenty of opinionated stuff in here, so comments/tests/etc. welcome!  I've got some thoughts for further changes, but this is what's needed for now.

Two questions off the top of my head:
- Is it worth adding a preference for fluff to disable treating the same /64 as the same user?  I'd argue no, but that might be worth a conversation (here and/or on-wiki), since it's an opinionated change.
- Is it worth adding a preference for block to default to the /64 checkbox?  If we did, should it default to on or off?  ATM I'd say no to both: opinionated change, essay not policy, probably new for many folks, etc.

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -9,13 +9,14 @@
  *** friendlytalkback.js: Talkback module
  ****************************************
  * Mode of invocation:     Tab ("TB")
- * Active on:              Any page with relevant user name (userspace, contribs, etc.)
+ * Active on:              Any page with relevant user name (userspace, contribs, etc.) except IP ranges
  * Config directives in:   FriendlyConfig
  */
 
 Twinkle.talkback = function() {
 
-	if (!mw.config.get('wgRelevantUserName')) {
+	if (!mw.config.exists('wgRelevantUserName') ||
+		mw.util.isIPAddress(mw.config.get('wgRelevantUserName')) !== mw.util.isIPAddress(mw.config.get('wgRelevantUserName'), true)) {
 		return;
 	}
 
--- a/modules/friendlyshared.js
+++ b/modules/friendlyshared.js
@@ -9,7 +9,7 @@
  *** friendlyshared.js: Shared IP tagging module
  ****************************************
  * Mode of invocation:     Tab ("Shared")
- * Active on:              Existing IP user talk pages
+ * Active on:              IP user talk pages
  */
 
 Twinkle.shared = function friendlyshared() {
--- a/modules/friendlywelcome.js
+++ b/modules/friendlywelcome.js
@@ -84,7 +84,9 @@ Twinkle.welcome.normal = function() {
 			}
 		}
 	}
-	if (mw.config.get('wgRelevantUserName')) {
+	// Users and IPs but not IP ranges
+	if (mw.config.exists('wgRelevantUserName') &&
+		mw.util.isIPAddress(mw.config.get('wgRelevantUserName')) === mw.util.isIPAddress(mw.config.get('wgRelevantUserName'), true)) {
 		Twinkle.addPortletLink(function() {
 			Twinkle.welcome.callback(mw.config.get('wgRelevantUserName'));
 		}, 'Wel', 'friendly-welcome', 'Welcome user');
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -19,6 +19,10 @@ Twinkle.arv = function twinklearv() {
 	}
 
 	var isIP = mw.util.isIPAddress(username, true);
+	// Ignore ranges wider than the CIDR limit
+	if (isIP && !mw.util.isIPAddress(username) && !Morebits.validCIDR(username)) {
+		return;
+	}
 	var userType = isIP ? 'IP' + (!mw.util.isIPAddress(username) ? ' range' : '') : 'user';
 
 	Twinkle.addPortletLink(function() {
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -18,12 +18,12 @@ Twinkle.arv = function twinklearv() {
 		return;
 	}
 
-	var isIP = mw.util.isIPAddress(username);
-	var title = isIP ? 'Report IP to administrators' : 'Report user to administrators';
+	var isIP = mw.util.isIPAddress(username, true);
+	var userType = isIP ? 'IP' + (!mw.util.isIPAddress(username) ? ' range' : '') : 'user';
 
 	Twinkle.addPortletLink(function() {
 		Twinkle.arv.callback(username, isIP);
-	}, 'ARV', 'tw-arv', title);
+	}, 'ARV', 'tw-arv', 'Report ' + userType + ' to administrators');
 };
 
 Twinkle.arv.callback = function (uid, isIP) {
@@ -36,6 +36,8 @@ Twinkle.arv.callback = function (uid, isIP) {
 	Window.addFooterLink('ARV prefs', 'WP:TW/PREF#arv');
 	Window.addFooterLink('Twinkle help', 'WP:TW/DOC#arv');
 	Window.addFooterLink('Give feedback', 'WT:TW');
+
+	var isIPRange = isIP && !mw.util.isIPAddress(uid);
 
 	var form = new Morebits.quickForm(Twinkle.arv.callback.evaluate);
 	var categories = form.append({
@@ -53,7 +55,7 @@ Twinkle.arv.callback = function (uid, isIP) {
 		type: 'option',
 		label: 'Username (WP:UAA)',
 		value: 'username',
-		disabled: mw.util.isIPAddress(uid)
+		disabled: isIP
 	});
 	categories.append({
 		type: 'option',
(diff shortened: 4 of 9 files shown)


martijnhoekstra:
Wonderful!

Amorymeltzer:
I'll try to hammer out any issues/etc. with this this week now that [T206954](https://phabricator.wikimedia.org/T206954) is live, but as noted there, there's a bug where, outside of Special:Contributions, the wrong IP is returned for IP ranges.  That might muck some things up...

Amorymeltzer:
One thing to add is that it's valid for CIDRs that are too large to act on.  Not sure of the best way to deal with that aside from hardcoding some limits and checking...

modules/friendlytalkback.js
```diff
@@ -9,13 +9,14 @@
...
- * Active on:              Any page with relevant user name (userspace, contribs, etc.)
+ * Active on:              Any page with relevant user name (userspace, contribs, etc.) except IP ranges
  * Config directives in:   FriendlyConfig
  */
 
 Twinkle.talkback = function() {
 
-	if (!mw.config.get('wgRelevantUserName')) {
+	if (!mw.config.exists('wgRelevantUserName') ||
+		mw.util.isIPAddress(mw.config.get('wgRelevantUserName')) !== mw.util.isIPAddress(mw.config.get('wgRelevantUserName'), true)) {
```
siddharthvp, line 19: This is so clever that using it directly is confusing. Also the IP-related functions in Morebits could do with some organisation. How about
```
Morebits.ip = {
	isRange: ... // single line implementation but still worth it
	sanitizeIPv6: ...
	get64Subnet: ...
}
```

?

Amorymeltzer: Funny you should mention that... I've got an additional local commit I didn't push up since I wanted to see how the feature was actually implemented.  It adds yet another IP function (we'll likely need to validate the CIDR range isn't too wide) and it proposes just that!  With *two* additional functions, yeah, I absolutely agree.

Amorymeltzer:
Tired of waiting on phab (movement, stalled, some movement, blah) when it's really just the one commit that *requires* it, so moved that to #1296.  This is all now live on testwiki.

Amorymeltzer:
Gonna merge this for now, but wait a bit before pushing live so it's easier to test/not holding up other things (like #1297).

In theory, this should all be fine when @Ammarpad's work gets merged in, but probably worth a once-over at that point.

modules/friendlytalkback.js
Amorymeltzer: #1297

molly:
Meant to comment on this earlier; sorry for the post-merge comment. I have a couple thoughts on this:
1. Does `{{rangeblock}}` make sense as a default? I never use that when I block /64s, and usually just use the standard block templates meant for a single user.
2. This may just be something I do, and may not be convention, in which case feel free to ignore. However, when I block a /64 I usually leave a templated block reason on the talk page of the most-recently used IP in the range. Would that make sense to support?

ToBeFree:
Regarding #1, the reason for blocking the /64 is the assumption that it is used by one single person only, and is extremely unlikely to be re-assigned to a different household. I never use {{rangeblock}} in these cases either.Regarding #2, especially when blocking for filter log entries, a talk page message is very helpful. There is also no reason not to send one.Thanks for doing this!

Amorymeltzer:
I Basically agree with @ToBeFree, @molly.  In theory, this was (and will) be for all ranges, so Twinkle could be used for *any* rangeblock.  In practice, we're not there yet, so maybe turn it off for now?  I suppose what would be ideal would be if checking the /64 box (or, I suppose, anytime we're blocking a /64), use the standard, but for other ranges, default to {{rangeblock}}.  Does that sound reasonable?  I can take a look.

As for the template, yeah, that'd be neat.  I was wondering how desired that would be — of course, you can always navigate to the specific IP/any IPs and leave just template(s), but that's not ideal.  It'd've been a fair bit more work I think, so didn't consider it for this (especially since this is/was already quite a lot).  But I like the idea!  Will think about how to do it.

molly:
> Good points @molly; I think I basically agree with you and @ToBeFree. In theory, this was (and will) be for all ranges, so Twinkle could be used for _any_ rangeblock. In practice, we're not there yet, so maybe turn {{rangeblock}} off for now? I suppose what would be ideal would be if checking the /64 box (or, I suppose, anytime we're blocking a /64), use the standard, but for other ranges, default to {{rangeblock}}. Does that sound reasonable? I can take a look. ETA: #1320

I agree this sounds very reasonable.

> As for the template, yeah, that'd be neat. I was wondering how desired that would be — of course, you can always navigate to the specific IP/any IPs and leave just template(s), but that's not ideal. It'd've been a fair bit more work I think (no _great_ way to query API for range contribs IDT (T177150)), so didn't consider it for this (especially since this is/was already quite a lot). But I like the idea! Will think about how to do it. Could also ask/see what'd be ideal. I don't think most folks do this regularly, but then again that's probably because it's not been easy!

My workflow with this I think will generally be to navigate to the talk page of the most-recently-used IPv6 address, check the /64 box, and then leave a template on the talk page I'm already on for the single IP. I imagine that would be easier to code than trying to query for all contribs in the /64, though it could definitely lead to people leaving the message on the wrong talk page if they get there from an older edit and the IP has since changed.

Amorymeltzer:
Word.  So that'd be, at least for now:
- Template off for ranges
- If the original user is a single IPv6, template *on*, even if "just block the 64" box is checked

As you say, I think that's roughly in-line with what folks might expect or hope to (or should) do.  That's easy enough to do, I can toss it together to look at tonight unless you wanna take a crack at it?

molly:
Happy to take a stab at it if you like, though I also won't be able to get to it until tonight. Either way!

Amorymeltzer:
Go for it!

molly:
Hmm, I'm poking at putting a PR together for this but realizing: how would one pull up the Twinkle block dialog for a range without coming through a single IP's talk or Special:Contribs? It looks like the Twinkle tab doesn't appear on contribs pages for ranges, and there's no such thing as a talk page for a full range.

Amorymeltzer:
> Hmm, I'm poking at putting a PR together for this but realizing: how would one pull up the Twinkle block dialog for a range _without_ coming through a single IP's talk or Special:Contribs? It looks like the Twinkle tab doesn't appear on contribs pages for ranges, and there's no such thing as a talk page for a full range.

(Not now, but this was all done with the idea that `wgRelevantUserName` would be defined on ranges in the near/coming future; this is [T206954](https://phabricator.wikimedia.org/T206954).  Right now, of course, it's not at all.)

You can fake it yourself if you want with `mw.config`, but I don't think you really need to for this?  We were talking about navigating to a single IPv6 and then selecting the "just block the /64" option, which this PR enables regardless of upstream MW change, and should work for you now.  The changes would be to keep the template issuance on for that case, and ensure it goes to that original, single IP (i.e. use `mw.config.get('wgRelevantUserName')` rather than the internal `relevantUserName` var).  Yeah?

molly:
Ah, okay, just wanted to make sure I wasn't missing some hidden entry point to this dialog I wasn't aware of. I shall carry on then :)

PR #1267, opened 2021-01-07
morebits: Allow html in labels for most quickForm element types

Amorymeltzer:
Already allowed in `div`, now done in `select`, `field`, `checkbox`/`radio`, `input`, `textarea`, `header`, and `dyninput`.

`option`, `optgroup`, `_dyninput_element`, `submit`, and `button` don't really make sense; neither does `type=header` in `checkbox`/`radio`.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -293,6 +293,24 @@ Morebits.quickForm.element.prototype.render = function QuickFormElementRender(in
 	return currentNode[0];
 };
 
+/**
+ * Render the label for a quickform element.
+ * @param {HTMLElement} labelElement
+ * @param {Node|string|(Node|string)[]} input
+ */
+Morebits.quickForm.element.prototype.generateLabel = function QuickFromElementGenerateLabel(labelElement, input) {
+	if (!Array.isArray(input)) {
+		input = [ input ];
+	}
+	for (var i = 0; i < input.length; ++i) {
+		if (typeof input[i] === 'string') {
+			labelElement.appendChild(document.createTextNode(input[i]));
+		} else if (input[i] instanceof Node) {
+			labelElement.appendChild(input[i]);
+		}
+	}
+};
+
 /** @memberof Morebits.quickForm.element */
 Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(data, in_id) {
 	var node;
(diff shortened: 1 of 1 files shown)


morebits.js
```diff
@@ -321,7 +325,16 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
...
+				if (!Array.isArray(data.label)) {
+					data.label = [ data.label ];
+				}
+				for (i = 0; i < data.label.length; ++i) {
+					if (typeof data.label[i] === 'string') {
+						label.appendChild(document.createTextNode(data.label[i]));
+					} else if (data.label[i] instanceof Element) {
+						label.appendChild(data.label[i]);
+					}
+				}
```
siddharthvp, line 337: I think this can be written as `$(label).append(data.label);` and there's no reason morebits should continue to pretend jquery doesn't exist. 

In any case, this is repeated 7 times in this PR itself and I think once in Morebits.status, so should be pulled out as a function.

siddharthvp: > I think this can be written as $(label).append(data.label);

Oh, this does mean that if label text is based on user input it would need to be sanitised first to avoid XSS issue - which I'd prefer be done on the client script side - which means a breaking change.

Amorymeltzer: I'm not entirely sure what you're suggesting with this comment, but

>done on the client script side

I'd suppose so, yes.

Amorymeltzer: Oh, I think I get it now.  You're saying that `.append` would allow for xss, which is a reason against doing the jQuery route, but that in theory that should be dealt with by whatever is using the library, right?  And that's a breaking change?

siddharthvp: Yes, but it (the XSS issue) may not be a worthwhile consideration as I doubt anyone's using quickform that innovatively!

morebits.js
```diff
@@ -321,7 +325,16 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
...
 				label = node.appendChild(document.createElement('label'));
 				label.setAttribute('for', id);
-				label.appendChild(document.createTextNode(data.label));
+				if (!Array.isArray(data.label)) {
+					data.label = [ data.label ];
+				}
+				for (i = 0; i < data.label.length; ++i) {
+					if (typeof data.label[i] === 'string') {
+						label.appendChild(document.createTextNode(data.label[i]));
+					} else if (data.label[i] instanceof Element) {
```
siddharthvp, line 334:
```suggestion
					} else if (data.label[i] instanceof Node) {
```

Something like this PR turns out to be necessary for i18n (since some modules like unlink are using HTML in labels using the array form - but for i18n we need a single string that can be put into the messages json file).

Changing `Element` to `Node` will take care of that by allowing use of $.parseHTML in the modules.

That is,
```js
label: $.parseHTML(msg('msg-name'))
```
here `msg-name` can be an html string. Say `hello <strong>world</strong>` which when passed through $.parseHTML gives `[ textNode, strongElement ]`. textNode is instance of `Node` but not `Element`.

I think we can avoid jquery for now, and pull this duplicated code into a function.

siddharthvp:
Made modifications per [previous comment](https://github.com/wikimedia-gadgets/twinkle/pull/1267#discussion_r593877538).

siddharthvp:
There's still more de-duplication that can be achieved (b/w quickForm and status) especially since I suppose it would make sense to have wikilink syntax support (added in #1366 for status) for quickForm labels too. Will open that in a separate PR.

PR #1269, opened 2021-01-08
welcome: Fix welcome-short parameter syntax

### LINKED ISSUE #1268: {{welcome-short}} on a new page doesn't place a heading
If you go to a user talk page that doesn't exist, and you click TW->Wel, and you select `{{welcome-short}}`, it will place the message without creating a corresponding heading. For the heading, I recommend `== Welcome! ==`.

You may also want to spot check this behavior for another random welcome message. Maybe this problem exists for other messages, too.

Thanks for your time.

TheTVExpert:
Fixes #1268


Amorymeltzer (approved):
Good catch!

PR #1270, opened 2021-01-08
tag: Update namespace redirect descriptions following RfC

Amorymeltzer:
The results of the [RfC](https://en.wikipedia.org/wiki/Template_talk:R_to_project_namespace#RfC:_Should_we_categorize_redirects_to_the_same_namespace?) are still in flux, so this should wait a bit, but I think this captures the intent?


Amorymeltzer:
RfC reopened?

Amorymeltzer:
RfC [(re)closed as no consensus](https://en.wikipedia.org/w/index.php?oldid=1008824905&diff=1009320766) so this can be closed as unneeded

PR #1277, opened 2021-01-13
morebits/fluff/xfd/protect: Update watchlist expiry handling for expiry return after T268834

Amorymeltzer:
This expands on #1230 and #1242, but with [T268834](https://phabricator.wikimedia.org/T268834) going through, we can now get an individual page's watchlist expiry upon query/`Morebits.wiki.page.load`.  As such, we can now confirm if the page is being watched temporarily or permanently, meaning we don't need to shy away from attempting to extend already-temp watched pages.

Accordingly, morebits will now take `watchlistexpiry` from page load and use it to avoid overwriting permanent watching while applying/extending temporary watching if the page is unwatched or temporarily watched.  `ctx.watched` will now either either be a boolean (as before) or, if the page is watched temporarily, a string.

The similar stuff in fluff and the `action=watch` calls in xfd and protect are likewise updated (there was a bug in protect, to boot).

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -454,7 +454,7 @@ Twinkle.fluff.callbacks = {
 			} else {
 				query.watchlist = 'watch';
 				// number allowed but not used in Twinkle.config.watchlistEnums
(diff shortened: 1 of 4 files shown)


Amorymeltzer:
Meant to ping you @MusikAnimal fwiw

PR #1280, opened 2021-01-17
Return from Morebits.wiki.preview, use to generate {{Find sources}}

Amorymeltzer:
Follows up on comments at https://github.com/wikimedia-gadgets/twinkle/issues/1047#issuecomment-671911400 cc @siddharthvp 

<s>Adds a callback option to `Morebits.wiki.preview`, then uses that to fill in xfd and prod.</s>  See below https://github.com/wikimedia-gadgets/twinkle/pull/1280#issuecomment-779032190

----

Could skip the Morebits edit and just use the api directly, but feels like we might as well use what we got.

{{Find sources}} is TE protected, but, uh, html being executed, etc.

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -65,8 +65,9 @@ Twinkle.prod.callback = function twinkleprodCallback() {
 
 	field.append({
 		type: 'div',
-		label: Twinkle.makeFindSourcesDiv(),
-		style: 'margin-bottom: 5px;'
+		label: '', // Added later by Twinkle.makeFindSourcesDiv()
+		id: 'twinkle-prod-findsources',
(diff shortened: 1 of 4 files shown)


siddharthvp:
Why callback? You just have to put a `return` on the [renderApi.post() line](https://github.com/wikimedia-gadgets/twinkle/pull/1280/commits/fc35bbc3bcf2bb77122495b616667268c9a5fa1b#diff-20b1ff1a748addb118304a1b428701029480dc64e319f65a82cd213ea3dc8083R4383). 

Then the subsequent commit can do `parser.beginRender(...).then(function() {...})`.

Amorymeltzer:
Hmm, I would've guessed that'd cause issues elsewhere, but you're quite right.  Much nicer, and done!

PR #1282, opened 2021-01-17
morebits.date: Make unitMap available through Morebits.date.unitMap

Amorymeltzer:
<s>`localeData` needs work, but</s> Doing this makes the available units for `add` and `subtract` accessible elsewhere; since using wrong units returns an error, we had no way of checking whether we had a correct unit without hardcoding the values.

In a similar vein, also makes `Morebits.date.add` and `Morebits.date.subtract` ignore unit case.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -1667,6 +1667,22 @@ Morebits.date.localeData = {
 	}
 };
 
+/**
+ * Map units with getter/setter function names, for `add` and `subtract`
+ * methods.
+ *
+ * @memberof Morebits.date
+ * @type {object.<string, string>}
+ */
+Morebits.date.unitMap = {
+	seconds: 'Seconds',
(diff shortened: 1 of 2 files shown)


siddharthvp:
This isn't locale data as it's used only in the code and so doesn't need to be customised per-language. Not sure where the exposed unitMap is going to be used, but it should be called `Morebits.date.unitMap`.

Amorymeltzer:
It's the strings provided to `add` and `subtract`, so it's used by anyone using those methods, not *just* in code.  My thinking was that folks might want to allow for localized units (`segundos`, etc.) but that'll get rapidly complicated with plurals, gender, etc., so `Morebits.date.unitMap` is fine by me.

Amorymeltzer:
{{done}}

Issue #980, opened 2020-06-17
Do not use native prompts


Huji:
Twinkle uses `prompt()` in a few places. The problem with `prompt()` is that is it not RTL-friendly, that is, the textbox it provides to the user to enter a response in is always left-to-right.

Instead of relying on native `prompt()` create a prompt dialog that is internationalizeable.

Amorymeltzer:
I wonder if something like [jQuery UI's dialog](https://api.jqueryui.com/dialog/) would be helpful.

siddharthvp:
Isn't this something to be fixed in the browsers? 

> I wonder if something like [jQuery UI's dialog](https://api.jqueryui.com/dialog/) would be helpful.

I don't think so, as jqueryUI doesn't blcok the event loop. OOUI has prompts that uses promises (resolved when the user submits), but then doing something like that would be a pretty big change on the codebase.

Amorymeltzer:
Ahh, I thought some of the `stopPropagation` stuff would work, I'm not very familiar with it.  I honestly think the OOUI stuff might be obsoleted by the vue stuff beforehand...

ed6767:
The issue with replacing prompt() and alert() is that they both functions that suspend until user interaction, so they may be implemented with if statements and other inline things. Here, I would think OOUI is ideal, then partially restructuring every instance of alert, prompt, ext. using a callback instead, which could be fed into the same if statements - so instead of:
```
if (prompt("What's your name?") == "dave") {
    alert("Ayup Dave!");
} else {
    alert("Hello");
}
// And things to go after
```
It should be:
```
promptFunction("What's your name?", function(name){
    if (name== "dave") {
       alert("Ayup Dave!");
   } else {
       alert("Hello");
  }
});
```
This shouldn't require too much work past some copy-pasting and a rough restructuring.

I'm very happy to aid slowly transitioning to OOUI rather than jQueryUI and outdated Javascript functions such as prompt ext.

Huji:
I believe there is a strong desire against using OOUI in Twinkle for two reasons; one is that OOUI doesn't allow its modal dialogs to be moved; the other is that OOUI uses lots of white space in its elements and with something like twinkle's speedy module that has tens of options, that may result in lots of scrolling.

Amorymeltzer:
[...]

PR #1283, opened 2021-01-18
fluff: Allow for a list of custom edit summaries

TheTVExpert:
Per https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&diff=997640771&oldid=997285448. Also partially helps #980.


siddharthvp:
Haven't tested, but this sounds pretty good to me. I had the same idea in mind to resolve #980. Use of a [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) would reduce the number of form fields from 2 to 1, but here I suppose it's better not to use datalist as the form otherwise would look too empty. Also datalist isn't supported in Safari <12, though not sure if we care much about those browsers.

It would be good to provide some some defaults in the `customReversionSummaryList`. Try copying from RedWarn.

TheTVExpert:
@siddharthvp I've added some defaults for the `customReversionSummaryList`.

modules/twinklefluff.js
```diff
@@ -395,13 +395,61 @@ Twinkle.fluff.revertToRevision = function revertToRevision(oldrev) {
...
+Twinkle.fluff.getSummary = function getSummary(apiobj) {
+	Twinkle.fluff.apiobj = apiobj;
+	var params = apiobj.params;
+	var type = params.type;
+	if (type === 'vand' || (type === 'norm' && !Twinkle.getPref('offerReasonOnNormalRevert'))) {
+		Twinkle.fluff.callbacks.main('noSummary');
+	} else {
+		Twinkle.fluff.simpleWindow = new Morebits.simpleWindow(600, 350);
+		Twinkle.fluff.simpleWindow.setTitle('Enter reversion edit summary');
+		Twinkle.fluff.simpleWindow.setScriptName('Twinkle');
```
Amorymeltzer, line 412: Should this be modal?

```suggestion
		Twinkle.fluff.simpleWindow.setScriptName('Twinkle');
		Twinkle.fluff.simpleWindow.setModality('true');
```

Amorymeltzer (changes requested):
Nice!  I've a few comments with a quick first pass.  It's a new use of `Morebits.simpleWindow` for us, but it seems to work pretty nicely!

Regarding #980, we had a discussion there @siddharthvp about not stopping the event loop.  I never played around too deeply, is that a concern here?

twinkle.js
```diff
@@ -69,6 +69,7 @@ Twinkle.defaultConfig = {
 	markRevertedPagesAsMinor: [ 'vand' ],
 	watchRevertedPages: [ 'agf', 'norm', 'vand', 'torev' ],
 	watchRevertedExpiry: 'yes',
+	customReversionSummaryList: [{value: 'Unsourced addition', label: 'Unsourced'}, {value: 'Unexplained removal of content', label: 'Removal of content'}, {value: 'Potential copyright violation', label: 'Copyright violation'}, {value: 'Disruptive editing', label: 'Disruptive editing'}],
```
Amorymeltzer, line 72: Not sure about this.  Would it make more sense to make this list be empty and allow users to provide it themselves, then offer a few options (such as these?) in addition to that list?  Kind of like the other customizations (warn/tag/welcome templates).  It'd be annoying if you want to remove one, but I think a dropdown would help with that (though see other comment).

modules/twinklefluff.js
```diff
@@ -395,13 +395,61 @@ Twinkle.fluff.revertToRevision = function revertToRevision(oldrev) {
...
+				event: function(event) {
+					var esForm = event.target.form;
+					esForm.summary.value = this.value;
+					event.stopPropagation();
+				}
+			});
+		}
+		form.append({
+			id: 'editSummaryInput',
+			type: 'input',
```
Amorymeltzer, line 430: What about a `select` option?  Less space, simpler interface as the list gets long?  An extra click, but maybe that's not a big deal, especially when compared to the current manual entry.

siddharthvp:
> Regarding #980, we had a discussion there @siddharthvp about not stopping the event loop. I never played around too deeply, is that a concern here?

No I just meant that a prompt can't be _easily_ swapped for a simpleWindow; it requires quite a bit of restructuring – which has been done here.

TheTVExpert:
@Amorymeltzer Done.

twinkle.js
siddharthvp: Agreed this looks a bit silly. Not so much agreed on the solution. I see that twinkleconfig is missing a "plainlist" data type that would have been the natural solution here. It might be worth adding that just for this usecase.

Amorymeltzer: We'd have to prepopulate it with something, no?

Amorymeltzer: There are two things to discuss here:

One is how to store the data.  A `plainlist` would be nice, but I don't think `customlist` is a huge hindrance.  The latter might make the display easier to parse for folks if there are multiple options.  So maybe "sources: removed reliable sources" and "sources: needs reliable sources" or something.  A plainlist does seem simpler given we just need one string of text.

The other is how to incorporate defaults, which I'm more concerned about.  With this setup, if someone customizes the list, they are forever left out of any changes we make unless they hit "reset, view the default options, then re-add from memory what they wanted; that's not good.  Instead we can do what is done elsewhere, namely include the basic uses in the module, and then add any items from `customReversionSummaryList` on top of that.  The advantage of the former is that users could cull the defaults, though.

NovemLinguae:

PR #1286, opened 2021-01-22
speedy: Add preference to watchlist user page on notification

Amorymeltzer:
This wasn't asked for - and I don't even recommend it - but it's a hole in our preferences, especially stark since we have `deliWatchUser`.  <s>Unlike others, defaulting to `nochange` since that's the current practice.  That being said, we could default to `1 month` to match the new temporary settings in #1249?  I might lean that way, minimal impact in the end.</s>  I've done this, defaulting to `1 month` to match `deliWatchUser` (and `xfdWatchUser`) after #1249.

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -489,6 +489,15 @@ Twinkle.config.sections = [
 				type: 'boolean'
 			},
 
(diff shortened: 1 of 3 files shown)


Amorymeltzer:
Maybe 1 week would be better?

PR #1287, opened 2021-01-22
fluff: Add good faith links everywhere

Amorymeltzer:
This makes inline (contributions, history, recent changes; aka not diffs) that used the two link (`rollback` & `vandalism`) use the three link format, using the inline format to toggle the length and capitalization of the links.

Meant to propose this once the dust settled after #947 and #942 but slipped my mind.

----

Uses just `good faith` for the link text, but could do `agf` or `AGF` to be even shorter, although that's maybe not too helpful.  `good faith rollback` is too long, IMO.  Current and possible text, in order:

```
agf
rollback
vandalism
good faith
rollback (agf)
good faith rollback
restore this revision
```

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -85,10 +85,10 @@ Twinkle.fluff.linkBuilder = {
 	/**
 	 * @param {string} [vandal=null] - Username of the editor being reverted
 	 * Provide a falsey value if the username is hidden, defaults to null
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
There was exactly no feedback on the issue: https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Add_AGF_links_everywhere?

PR #1288, opened 2021-01-25
tag: Hide subheadings on quick filter

TheTVExpert:
Extends hiding subheadings to redirect tagging, as they are already hidden on article tagging, and file tagging doesn't include subheadings. Partially per https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&diff=1002467677&oldid=1002159993


Amorymeltzer (approved):
LGTM! 👍

PR #1289, opened 2021-01-27
twinkle: Remove unused preferences from default config

TheTVExpert:
Remove preferences from the default config that are not used in any of the modules.


DannyS712 (approved):
Github search shows no other uses in the repository

Amorymeltzer (approved):
Looks like these were all removed in 12046df (April 2012)

PR #1295, opened 2021-02-08
twinkle: Add mw-portlet class to Vector outerdiv, removes newly-(re)established bullets

Amorymeltzer:
Not sure when this changed (somewhere in 1.36-wmf28 or 29, presumably), but looks like the top-level of the menu gained a `.mw-portlet`, which apaprently means no bullets.  See also #713 and #763.

@MusikAnimal [MoreMenu](https://github.com/wikimedia-gadgets/MoreMenu) will need an update as well, I believe.  The train has been messed up lately, but I *think* things are going forward this week so it might need to be?  29 is on testwiki, where I've just pushed this, so good for testing I guess.

(I should note that in the intervening few hours, the hover state of the vector menus at testwiki have become unrelatedly borked)

### DIFF
--- a/twinkle.js
+++ b/twinkle.js
@@ -277,7 +277,7 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
 			if (navigation !== 'portal' && navigation !== 'left-navigation' && navigation !== 'right-navigation') {
 				navigation = 'mw-panel';
 			}
-			outerNavClass = 'vector-menu vector-menu-' + (navigation === 'mw-panel' ? 'portal' : type === 'menu' ? 'dropdown' : 'tabs');
+			outerNavClass = 'mw-portlet vector-menu vector-menu-' + (navigation === 'mw-panel' ? 'portal' : type === 'menu' ? 'dropdown' : 'tabs');
 			innerDivClass = 'vector-menu-content';
 			break;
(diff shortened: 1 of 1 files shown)


MusikAnimal:
@Amorymeltzer Shall we deploy? I see some have complained over at WP:VPT

Amorymeltzer:
@MusikAnimal Didn't realize the train was leaving early!  Thanks for the ping, now deployed.

MusikAnimal:
> @MusikAnimal Didn't realize the train was leaving early! Thanks for the ping, now deployed.

It left late, actually. The train blocker for wmf.29 was finally fixed. Sadly they did not consider the hover states a train blocker, so that won't get fixed until wmf.30 which arrives to enwiki on Thursday, if all goes according to plan.

I say "if all goes according to plan" because Release Engineering (understandably) adopted a new policy to block the train whenever any new errors surface. So moving forward we might see more erratic train deployments like this, with changes arriving on Monday / Tuesday, etc., instead of the usual Thursday. You can always check the current state at https://versions.toolforge.org/

Amorymeltzer:
The emails have been helpful, but yeah, I didn't realize that when blockers get fixed, they'd get pushed out earlier.  Good to be aware of!  Bummer about the menus though

PR #1301, opened 2021-02-12
morebits: Add/subtract weeks in Morebits.date and create fnApplyWatchlistExpiry to decide whether/when to apply a watchlist expiry

Amorymeltzer:
We only want to attempt a watchlist expiry if (1) the page isn't watched, or (2) the page is already watched temporarily; in the latter case, we probably only want extend an existing temporary watching, not shrink it.  To do that last bit properly, we needed a function to do some date calculations.  This tries not to be too over-enthusiastic, as there are plenty of valid MW expiry strings that `Morebits.date` doesn't recognize or can't `add` with.  Depends on #1298 rather than doing it itself.

One thing that needed adding, though, was parsing "weeks."  I first started writing a whole `getWeek`/`setWeek` set for ISO weeks, which largely worked, but it was really pointless since we don't really care about them per se, just as a means to an end.  For that, `*= 7` seems fine.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2689,7 +2689,7 @@ Morebits.wiki.page = function(pageName, status) {
 			query.tags = ctx.changeTags;
 		}
 
-		if (ctx.watchlistExpiry && ctx.watched !== true) {
+		if (fnApplyWatchlistExpiry()) {
 			query.watchlistexpiry = ctx.watchlistExpiry;
 		}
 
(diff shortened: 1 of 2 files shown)


morebits.js
```diff
@@ -3647,6 +3664,53 @@ Morebits.wiki.page = function(pageName, status) {
...
+					(Morebits.date.unitMap[rel[1]] || Morebits.date.unitMap[rel[1] + 's'])) {
+					// Expiry looks like a relative time (e.g. `1 week`), so treat it like one
+					newExp = new Morebits.date().add(rel[0], rel[1]);
+					if (!newExp.isValid()) {
+						// Oh well!
+						return false;
+					}
+				} else {
+					newExp = new Morebits.date(ctx.watchlistExpiry);
+				}
```
siddharthvp, line 3696:
```suggestion
				var rel = ctx.watchlistExpiry.toLowerCase().split(' ');
				try {
					newExp = new Morebits.date().add(rel[0], rel[1]);
				} catch() {
					newExp = new Morebits.date(ctx.watchlistExpiry);
				}
```

No need to rely on the implementation details of Morebits.date or duplicate its error handling clauses here. The date#add() will always either return a valid date or throw.

morebits.js
```diff
@@ -3647,6 +3664,53 @@ Morebits.wiki.page = function(pageName, status) {
...
+	 * ensured of knowing the watch status by the use of this.
+	 *
+	 * @returns {boolean}
+	 */
+	var fnApplyWatchlistExpiry = function() {
+		if (ctx.watchlistExpiry) {
+			if (!ctx.watched) {
+				return true;
+			} else if (typeof ctx.watched === 'string') {
+				var newExp;
```
siddharthvp, line 3683:
```suggestion
				var newExpiry;
```

I first thought this stood for new expression or something.

morebits.js
Amorymeltzer: I was hoping to avoid unnecessary console messages but yeah, this would be nicer to look at!

morebits.js
```diff
@@ -3647,6 +3664,48 @@ Morebits.wiki.page = function(pageName, status) {
...
+	 * requires page loading if a watchlistexpiry is provided, so we are
+	 * ensured of knowing the watch status by the use of this.
+	 *
+	 * @returns {boolean}
+	 */
+	var fnApplyWatchlistExpiry = function() {
+		if (ctx.watchlistExpiry) {
+			if (!ctx.watched) {
+				return true;
+			} else if (typeof ctx.watched === 'string') {
```
Amorymeltzer, line 3682: Could short-circuit indef options

```js
if (Morebits.string.isInfinity(ctx.watchlistExpiry)) {
	return true;
}
```

PR #1304, opened 2021-02-13
morebits: Add number input type to quickform, use in tag

Amorymeltzer:
Suggested at https://github.com/wikimedia-gadgets/twinkle/pull/1293#issuecomment-778195064; the `{{Do not move to Commons}}` expiry is the only current case where we require/validate a number, I believe, although we have been using the number input type in config for a while.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1074,9 +1074,10 @@ Twinkle.tag.fileList = {
 					required: true
 				},
 				{
-					type: 'input',
+					type: 'number',
 					name: 'DoNotMoveToCommons_expiry',
 					label: 'Expiration year: ',
+					min: new Morebits.date().getFullYear(),
 					tooltip: 'If this file can be moved to Commons beginning in a certain year, you can enter it here (optional).'
 				}
 			]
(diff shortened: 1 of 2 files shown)


modules/friendlytag.js
```diff
@@ -1074,9 +1074,10 @@ Twinkle.tag.fileList = {
 					required: true
 				},
 				{
-					type: 'input',
+					type: 'number',
 					name: 'DoNotMoveToCommons_expiry',
 					label: 'Expiration year: ',
+					min: new Morebits.date().getYear() + 1900,
```
siddharthvp, line 1080:
```suggestion
					min: new Morebits.date().getFullYear(),
```
.getYear() is deprecated.

morebits.js
```diff
@@ -569,7 +573,18 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 				subnode.setAttribute('value', data.value);
 			}
 			subnode.setAttribute('name', data.name);
-			subnode.setAttribute('type', 'text');
+
+			if (data.type === 'input') {
+				subnode.setAttribute('type', 'text');
+			} else {
+				subnode.setAttribute('type', 'number');
+				['min', 'max', 'step', 'placeholder', 'list'].forEach(function(att) {
```
siddharthvp, line 581: Not strictly related to this PR, but placeholder is valid (and quite useful) for input type=text too so can be added there.

Amorymeltzer: Fair point, although I think in general it's to be avoided (though I personally like 'em). Might as well now.  Done, with some other reorganizing.

Issue #1303, opened 2021-02-12
Bad edit request for RfD nominations


pppery:
See https://en.wikipedia.org/wiki/Wikipedia_talk:RFP#Edit_request_to_complete_RfD_nomination. The request says to add unclosed template syntax.

Amorymeltzer:
Yeah... this is because we use [{{Xfd edit protected}}](https://en.wikipedia.org/wiki/Template:Xfd_edit_protected), and RfD is the only one to need something else, so it was left off.  I suppose the thing to do would be to add a `rfd=yes` parameter or something to the template which rewords it to something to the effect of

>[top of the page] and `}}` to the bottom to [complete the nomination].

We can then use that parameter here.  It's a bit ugly IMO, but I'm not sure there's another way without making the template itself less useful on-wiki for non-Twinkle.  I tossed it together at [Template:Xfd_edit_protected/sandbox](https://en.wikipedia.org/wiki/Template:Xfd_edit_protected/sandbox), thoughts?  cc @siddharthvp

siddharthvp:
@Amorymeltzer Sounds right to me.

Amorymeltzer:
Template now updated (thanks @siddharthvp) so gonna go do #1305

PR #1305, opened 2021-02-13
xfd: Use new rfd=yes parameter for {{Xfd edit protected}}

Amorymeltzer:
RfD nominations wrap the entire page content in a template, but the requesting template was just noting the top bit.  Everything should have worked fine, but it's Wrong.  This makes use of the new `rfd=yes` to do it Right.

----

Placeholder pending #1303

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -750,7 +750,8 @@ Twinkle.xfd.callbacks = {
 			pageobj.getStatusElement().warn('Page protected, requesting edit');
 
 			var editRequest = '{{subst:Xfd edit protected|page=' + pageobj.getPageName() +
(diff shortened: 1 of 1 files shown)


modules/twinklexfd.js
```diff
@@ -750,7 +750,8 @@ Twinkle.xfd.callbacks = {
 			pageobj.getStatusElement().warn('Page protected, requesting edit');
 
 			var editRequest = '{{subst:Xfd edit protected|page=' + pageobj.getPageName() +
-				'|discussion=' + params.discussionpage + '|tag=<nowiki>' + params.tagText + '\u003C/nowiki>}}'; // U+003C: <
+				'|discussion=' + params.discussionpage + (params.venue === 'rfd' ? 'rfd=yes' : '') +
```
JJMC89, line 753: '|rfd=yes'

Amorymeltzer: Ugh yes, thanks!

PR #1308, opened 2021-02-13
warn: Fix bug in section insertion when non-L2 headings present

### LINKED ISSUE #1307: warn: Inserting templates to wrong section if there are level 3 sections
If there are level 3 sections, Twinkle insert warning templates to wrong section.
See https://en.wikipedia.org/w/index.php?title=User_talk%3AA2093064-test&oldid=1006550366 and previous edits.

https://github.com/wikimedia-gadgets/twinkle/blob/dc05676310d4f7a4376bd546274d33817ba99b41/modules/twinklewarn.js#L1768-L1769
I think it's required to get all sections instead of level 2 sections.

Related: #1084

Amorymeltzer:
We were only counting L2 sections, so when there were non-L2 sections present, the section index would be wrong.  Stems from introduction in  #1084, closes #1307.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1765,15 +1765,15 @@ Twinkle.warn.callbacks = {
 		var sectionExists = false, sectionNumber = 0;
 		// Only check sections if there are sections or there's a chance we won't create our own
 		if (!messageData.heading && text.length) {
-			// Get all L2 sections
-			var sections = text.match(/^(==)[^=].+\1/gm);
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
@Xi-plus This is untested since I'm not really around atm, but I think this should do it?

Amorymeltzer:
Seems to work, although there's still an issue should there be a subsection within the dated L2; not sure what to do about that.  For now, though, I'll go ahead with this.

PR #1309, opened 2021-02-14
morebits: Correct jquery dialog event name

Xi-Plus:
See https://api.jqueryui.com/dialog/#event-resizeStop

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -5460,7 +5460,7 @@ Morebits.simpleWindow = function SimpleWindow(width, height) {
 				this.scrollbox.style.maxHeight = 'none';
 			}
 		},
-		resizeEnd: function() {
+		resizeStop: function() {
 			this.scrollbox = null;
 		},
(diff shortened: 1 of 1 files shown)


Amorymeltzer (approved):
Weird!  Curious how you found it, @Xi-Plus.  AFAICT, it doesn't actually make a difference given how we use it, is that right?

Xi-Plus:
Yes. The typo didn't cause any bug.

siddharthvp:
The linked docs are for a different version, but this nevertheless applies to our jQuery.ui version: https://github.com/wikimedia/mediawiki/blob/b1c3685a1a4c15e3431ceb2fec59c94f4984e4ad/resources/lib/jquery.ui/jquery.ui.dialog.js#L472

Amorymeltzer:
https://api.jqueryui.com/1.9/dialog/#event-resizeStop

PR #1311, opened 2021-02-14
server.js: Set encoding in content-type header

Xi-Plus:
Setting encoding in `response.end` is not enough. I saw mojibake when I visited the script file in browser. Set encoding in Content-Type header to fix it.


Amorymeltzer (approved):
Must've been a pain!

PR #1312, opened 2021-02-14
tag: Avoid regex inefficiency

Amorymeltzer:
CodeQL brought this up (https://github.com/wikimedia-gadgets/twinkle/security/code-scanning/17) and yes, this does have the potential for a catastrophic backtracking reDoS.  In practice, we're unlikely to run into it, but still.


siddharthvp:
Well, like all of TW, it's not running on a server so there's no DoS that can affect us!

Amorymeltzer:
Self-redos then, or reverse.

PR #1313, opened 2021-02-14
remove type=error in mw.notify on TWPREF load failures

siddharthvp:
This occurs frequently on slow internet connections. Twinkle is very much usable without the preferences so there's no need to show a jarring red message.

### DIFF
--- a/twinkle.js
+++ b/twinkle.js
@@ -397,7 +397,7 @@ $.ajax({
 	dataType: 'text'
 })
 	.fail(function () {
-		mw.notify('Could not load your Twinkle preferences', {type: 'error'});
+		mw.notify('Could not load your Twinkle preferences, resorting to default preferences');
 	})
 	.done(function (optionsText) {
 


twinkle.js
```diff
@@ -397,7 +397,7 @@ $.ajax({
 	dataType: 'text'
 })
 	.fail(function () {
-		mw.notify('Could not load your Twinkle preferences', {type: 'error'});
+		mw.notify('Could not load your Twinkle preferences');
```
Amorymeltzer, line 400: That makes sense, I suppose, but given that it'll be working *differently* than folks might suspect, I think it should at least include a note to that effect, e.g.
```suggestion
		mw.notify('Could not load your Twinkle preferences, resorting to default Twinkle preferences');
```
Is it worth noting that a reload might (likely?) help?  We could also extend the length it appears, or even require clicking to dismiss; not sure those are necessary.

twinkle.js
```diff
@@ -426,7 +426,7 @@ $.ajax({
 				Twinkle.prefs.optionsVersion = Twinkle.prefs.optionsVersion || 1;
 			}
 		} catch (e) {
-			mw.notify('Could not parse your Twinkle preferences', {type: 'error'});
+			mw.notify('Could not parse your Twinkle preferences');
```
Amorymeltzer, line 429: Isn't this actually an error, though, as it suggests the prefs are corrupted somehow?  We could clarify this note in that case, but I'd be inclined to keep it as an error since it's likely either their doing or a busted Twinkle, yeah?

siddharthvp: Yes indeed!

twinkle.js
siddharthvp: Done. Requiring click to dismiss would be quite annoying on mobile (where this seems to occur more often).

Amorymeltzer (approved):
Sounds good!

PR #1316, opened 2021-02-15
protect: Fetch watched status of the page to properly watch when requesting

### LINKED ISSUE #1315: protect: Watching expiry for rpp
https://github.com/wikimedia-gadgets/twinkle/blob/2e5954a40eab5a58f6c139d95f574c13f95c7cf7/modules/twinkleprotect.js#L1516-L1535
`pageobj` is `Wikipedia:Requests for page protection`.
So you decide whether to set the expiry for **article** by checking **rppPage**'s watching status?

Amorymeltzer:
I was using the `Morebits.wiki.page` object returned from the RfPP page itself to determine watchlist expiry status of the tagged page, which is clearly wrong.  Since we're not loading the page directly in that case, we can just tack on to the protection fetching and store it for later.  Fixes #1315.

cc @Xi-Plus

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -153,7 +153,7 @@ Twinkle.protect.fetchProtectionLevel = function twinkleprotectFetchProtectionLev
 		letype: 'protect',
 		letitle: mw.config.get('wgPageName'),
 		prop: hasFlaggedRevs ? 'info|flagged' : 'info',
-		inprop: 'protection',
+		inprop: 'protection|watched',
 		titles: mw.config.get('wgPageName')
 	});
 	var stableDeferred = api.get({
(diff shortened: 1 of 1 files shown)


modules/twinkleprotect.js
```diff
@@ -181,6 +181,9 @@ Twinkle.protect.fetchProtectionLevel = function twinkleprotectFetchProtectionLev
 		var page = protectData[0].query.pages[pageid];
 		var current = {}, adminEditDeferred;
 
+		// Save requested page's watched status for later in case needed when filing request
+		Twinkle.protect.watched = page.watchlistexpiry || page.watched === '';
```
Xi-Plus, line 185: `page.watched` will be true or false when API formatversion=2.
```suggestion
		Twinkle.protect.watched = page.watchlistexpiry || page.watched;
```

Xi-Plus (changes requested):

modules/twinkleprotect.js
Amorymeltzer: @Xi-Plus We're not actually using `formatversion=2` here; `Twinkle.protect.fetchProtectionLevel` uses `mw.Api`, not `Morebits.wiki.api`.  I'd planned on converting it to `formatversion=2` but didn't want to do it here when this should be a quick bugfix (I don't think there's much work needed, though).

Xi-Plus: Oh!

Xi-Plus:
LGTM now. I have no further questions.

PR #1317, opened 2021-02-15
morebits: lookupCreation: add failure callback

siddharthvp:
This is unlikely to fail much due to API errors being just a read operation, but things like connection drops can always happen.

There ought to be _some_ way in which a failure can be detected.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2522,6 +2522,7 @@ Morebits.wiki.page = function(pageName, status) {
 		onSaveSuccess: null,
 		onSaveFailure: null,
 		onLookupCreationSuccess: null,
+		onLookupCreationFailure: null,
 		onMoveSuccess: null,
 		onMoveFailure: null,
 		onDeleteSuccess: null,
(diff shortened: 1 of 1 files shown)


Amorymeltzer (commented):
Seems reasonable and good!  Untested atm but 👍

morebits.js
```diff
@@ -3221,12 +3222,13 @@ Morebits.wiki.page = function(pageName, status) {
 	 * @param {Function} onSuccess - Callback function to be called when
 	 * the username and timestamp are found within the callback.
 	 */
```
Amorymeltzer, line 3224:
```suggestion
	* @param {Function} onSuccess - Callback function to be called when the lookup fails
	*/
```

morebits.js
```diff
@@ -3221,12 +3222,13 @@ Morebits.wiki.page = function(pageName, status) {
...
-	this.lookupCreation = function(onSuccess) {
+	this.lookupCreation = function(onSuccess, onFailure) {
 		if (!onSuccess) {
 			ctx.statusElement.error('Internal error: no onSuccess callback provided to lookupCreation()!');
 			return;
```
Amorymeltzer, line 3228:
```suggestion
			ctx.onLookupCreationFailure(this);
			return;
```

morebits.js
```diff
@@ -3221,12 +3222,13 @@ Morebits.wiki.page = function(pageName, status) {
...
 		}
 		ctx.onLookupCreationSuccess = onSuccess;
+		ctx.onLookupCreationFailure = onFailure || emptyFunction;
```
Amorymeltzer, line 3231: Should probably move these up to accomodate returning the failure if `onSuccess` not provided.

morebits.js
```diff
@@ -3848,6 +3850,7 @@ Morebits.wiki.page = function(pageName, status) {
 		var rev = response.pages[0].revisions && response.pages[0].revisions[0];
```
Amorymeltzer, line 3850: Shouldn't we provide the failure function to `fnCheckPageName` on line 3846?

Amorymeltzer (changes requested):
Couple more places for the failure callback, nothing major

morebits.js
siddharthvp: things get hazy when there are that many lines ....

Amorymeltzer (approved):

PR #1318, opened 2021-02-15
morebits: page: remove revert()

siddharthvp:
Morebits has now grown into a whooping 5800 lines and counting. Morebits.wiki.page is alone about 2000 lines I guess. Time to start removing cruft before the situation gets totally out of hand!

----

Unused ever since this was first added. Revert functionality in twinklefluff uses its own implementation, and there is no advantage in porting that to use this. There is no conceivable use of this anywhere else.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -4,7 +4,7 @@
  *
  * The highlights include:
  * - {@link Morebits.wiki.api} - make calls to the MediaWiki API
- * - {@link Morebits.wiki.page} - modify pages on the wiki (edit, revert, delete, etc.)
+ * - {@link Morebits.wiki.page} - modify pages on the wiki (edit, move, delete, etc.)
  * - {@link Morebits.date} - enhanced date object processing, sort of a light moment.js
  * - {@link Morebits.quickForm} - generate quick HTML forms on the fly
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
>Revert functionality in twinklefluff uses its own implementation, and there is no advantage in porting that to use this

Why do you say that?  I've not looked, but the PITA that is fluff seems like it would make not just a good use case but also an example of why this is a useful feature in a library.

siddharthvp:
Reverting edits doesn't seem like something that userscripts/gadgets commonly do, except the ones _specifically_ designed for that (twinklefluff/redwarn/...) – which all end up doing their own implementation anyway.

fluff doesn't seem to use any core feature of mb.w.p, except probably the new watchlist expiry thing. I too didn't look much into this. It seemed at a glance that not using morebits in fluff was an advantage as it allowed for greater flexibility.

Amorymeltzer:
I gave it a go at #1324, it wasn't too bad: the fluff stuff was written way before `revert, which turns out to have been busted from inception.  Fluff can actually make use of quite a bit of it, it's rollbacks finding the last good revision where we need lots of processing.  `revert` could theoretically be expanded to help with that, I suppose, but it's bespoke for a reason (hence why things tend to roll their own).  For the actual reversion, though, as well as everything about revert-to-revision, it works rather neatly.

PR #1319, opened 2021-02-15
morebits: taskManager: enable failure handling and setting context

siddharthvp:
1. Enables the value of `this` in the task functions to be set via the constructor. A minor limitation here is that all tasks get the same value for `this`.

2. Enable handling of failures. For each task a failure function can be set. This will be invoked if the task errors out or returns a rejected promise. It will also be invoked if the task was never started due to a failure in any of its dependencies.

---- 

Written for twinkle-core, but at the moment, I'm completely lost on how to manage morebits in twinkle-core. 

The failure handling is used in XfdCore as a replacement for the `Morebits.status.onError` which twinklexfd uses (which has resulted in many issues). A lot cleaner and allows for much better programmatic control.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -5454,7 +5454,7 @@ Morebits.batchOperation = function(currentAction) {
 
 /**
  * Given a set of asynchronous functions to run along with their dependencies,
- * figure out an efficient sequence of running them so that multiple functions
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
>2nd commit

Eh?

PR #1320, opened 2021-02-16
block: Only select {{rangeblock}} for non-/64 ranges

Amorymeltzer:
As mentioned in https://github.com/wikimedia-gadgets/twinkle/pull/1266#issuecomment-779521407

----

@Molly @ToBeFree This does what I suggested as a resolution for item 1, namely uses {{rangeblock}} only for non-/64 ranges (I'm heading to bed/The Expanse but figured I could whip it up quickly for your perusal).

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -1541,7 +1541,7 @@ Twinkle.block.blockGroups = [
 			{ label: 'checkuserblock-wide', value: 'checkuserblock-wide' },
 			{ label: 'colocationwebhost', value: 'colocationwebhost' },
 			{ label: 'oversightblock', value: 'oversightblock' },
-			{ label: 'rangeblock', value: 'rangeblock', selected: true }, // Only for IP ranges
+			{ label: 'rangeblock', value: 'rangeblock' }, // Only for IP ranges, selected for non-/64 ranges in filtered_block_groups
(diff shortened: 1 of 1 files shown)


modules/twinkleblock.js
```diff
@@ -1578,6 +1578,11 @@ Twinkle.block.callback.filtered_block_groups = function twinkleblockCallbackFilt
 				return;
 			}
 
+			// Select rangeblock for non-/64 ranges only
+			if (blockPreset.value === 'rangeblock') {
+				blockPreset.selected = Morebits.ip.get64(relevantUserName) !== relevantUserName;
```
Amorymeltzer, line 1583: Wanted to note this separately, but I think we should do this instead, thus also disabling {{rangeblock}} for anything smaller than a /64 as well (/65, etc.)  It also has the greatly added benefit of *not* selecting {{rangeblock}} for functionally-equivalent /64 ranges (e.g., 2409:4051:4E93:5F27:0:0:0:**1**/64).

```suggestion
				blockPreset.selected = !Morebits.ip.get64(relevantUserName);
```

Amorymeltzer: Even easier if doing #1314 first

Amorymeltzer: Done and done.

PR #1321, opened 2021-02-16
config: partially port to use jQuery

siddharthvp:
Ported most of Twinkle.config.listDialog (this is the code that creates the popup dialog for editing customList type configs - such as the ones for adding custom templates) to use jQuery for UI construction rather than vanilla javascript – which is hellishly inscrutable. The jQuery rewrite is a quite more easily digestible. 

No functional changes.

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -1350,10 +1350,10 @@ Twinkle.config.init = function twinkleconfigInit() {
 
 Twinkle.config.listDialog = {};
 
-Twinkle.config.listDialog.addRow = function twinkleconfigListDialogAddRow(dlgtable, value, label) {
+Twinkle.config.listDialog.addRow = function twinkleconfigListDialogAddRow($dlgtable, value, label) {
 	var $contenttr, $valueInput, $labelInput;
 
-	$(dlgtable).append(
+	$dlgtable.append(
 		$contenttr = $('<tr>').append(
 			$('<td>').append(
 				$('<button>')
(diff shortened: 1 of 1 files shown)


modules/twinkleconfig.js
```diff
@@ -1396,88 +1396,79 @@ Twinkle.config.listDialog.display = function twinkleconfigListDialogDisplay(e) {
...
+										})
+								)
+						)
+					)
+				),
+			$('<button>')
+				.text('Save changes')
+				.attr('type', 'submit') // so Morebits.simpleWindow puts the button in the button pane
+				.on('click', function () {
+					Twinkle.config.listDialog.save($prefbutton, $dlgtbody);
```
Amorymeltzer, line 1443: Minor, but as long as we're updating one place `Twinkle.config.listDialog.save` and `Twinkle.config.listDialog.reset` are called, mightn't we likewise change their parameters to `$button, $tbody`?  Could be some minor cleanup in each.

modules/twinkleconfig.js
```diff
@@ -1399,88 +1396,79 @@ Twinkle.config.listDialog.display = function twinkleconfigListDialogDisplay(e) {
...
-	dlgtr.appendChild(dlgth);
-	// label column header
-	dlgth = document.createElement('th');
-	dlgth.style.width = '60%';
-	dlgth.textContent = curpref.customListLabelTitle ? curpref.customListLabelTitle : 'Label';
-	dlgtr.appendChild(dlgth);
-	dlgtbody.appendChild(dlgtr);
+	var $dlgtbody;
+
+	dialog.setContent(
```
Amorymeltzer, line 1401: Dunno if you tested this, but actually running it will fail here: `Uncaught TypeError: Node.appendChild: Argument 1 does not implement interface Node.`

Pretty sure this needs to be `$('<div>').append(...snip...)[0]` to properly `setContent`

Amorymeltzer (changes requested):
Definitely easier to follow a linear thread.  Two minor (well, minor to fix!) bugs

modules/twinkleconfig.js
Amorymeltzer: Doing this worked in my testing, so I can just do this if you'd like.

modules/twinkleconfig.js
siddharthvp: done

modules/twinkleconfig.js
```diff
@@ -1396,88 +1396,79 @@ Twinkle.config.listDialog.display = function twinkleconfigListDialogDisplay(e) {
...
+				.css({
+					margin: '1.4em 1em',
+					width: 'auto'
+				})
+				.append(
+					$dlgtbody = $('<tbody>').append(
+						// header row
+						$('<tr>').append(
+							$('<th>') // top-left cell
+								.css('width', '35%'),
```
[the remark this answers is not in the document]
siddharthvp: oof, serves me right for copy-pasting!

modules/twinkleconfig.js
siddharthvp: I though I tested this, but now that I think of it, patch-test.js wouldn’t really work for config.

modules/twinkleconfig.js
```diff
@@ -1396,88 +1396,79 @@ Twinkle.config.listDialog.display = function twinkleconfigListDialogDisplay(e) {
...
+						)
+					),
+					$('<tfoot>').append(
+						$('<tr>').append(
+							$('<td>')
+								.attr('colspan', '3')
+								.append(
+									$('<button>')
+										.text('Add')
+										.css('minWidth', '8em')
```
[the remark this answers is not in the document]
Amorymeltzer: fwiw, I meant in general; jQuery handles both just fine.

modules/twinkleconfig.js
Amorymeltzer: Works fine for me?

Amorymeltzer (approved):
👍

PR #1322, opened 2021-02-16
morebits: Expand setWatchlist handling of expiry, default to infinity for watch/preferences

Amorymeltzer:
`setWatchlist` now takes a second parameter, allowing you to explicitly set the watchlist expiry if so desired.

To break it down, the MediaWiki API allows four watching options:

- `watch`: Explicitly watch the page, using the expiry if provided
- `preferences`: Watch (or not) the page according to preferences, using the expiry if appropriate and provided
- `nochange` (default): Don't change the binary watch status, but if an expiry is provided and appropriate, use it
- `unwatch`: Expiry doesn't matter

We've made two small editorial decisions here:
- If no expiry is provided for a `watch` or `preferences` option, default to indefinite watching.  This mirrors prior (and expected) behavior.
- If `nochange` is provided, explicitly set the expiry to `null`, so `nochange` really means "no change."

The former is a bugfix for us: if a user wanted to indefinitely watch a temporarily watched page, we weren't handling that possibility just by using `setWatchlist` in Twinkle.  For the latter, `setWatchlistExpiry` continues to exist and remains useful.  This all - including `setWatchlist` falling back to treating a string as `watch` + expiry - is a bit kludgy in order to make Twinkle prefs work smoothly, but continues to make things easier.  Passing a falsey value for `watchlistExpiry` will get the expiry treated as nonexistent (with `fnApplyWatchlistExpiry` from #1301), using whatever exists on-wiki.

cc @MusikAnimal in case you want to peruse

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -2930,37 +2930,84 @@ Morebits.wiki.page = function(pageName, status) {
 	};
 
 	/**
-	 * @param {boolean|string} [watchlistOption=false] -
+	 * Set whether and how to watch the page, including setting an expiry.
+	 *
+	 * @param {boolean|string|Morebits.date|Date} [watchlistOption=false] -
 	 * Basically a mix of MW API and Twinkley options available pre-expiry:
-	 * - `true`|`'yes'`: page will be added to the user's watchlist when the action is called
-	 * - `false`|`'no'`|`'nochange'`: watchlist status of the page will not be changed.
-	 * - `'default'`|`'preferences'`: watchlist status of the page will
-	 * be set based on the user's preference settings when the action is
-	 * called.  Ignores ability of default + expiry.
-	 * - `'unwatch'`: explicitly unwatch the page
-	 * - {string|number}: watch page until the specified time (relative or absolute datestring)
+	 * - `true`|`'yes'`|`'watch'`: page will be added to the user's
+	 * watchlist when the action is called. Defaults to an indefinite
+	 * watch unless `watchlistExpiry` is provided.
+	 * - `false`|`'no'`|`'nochange'`: watchlist status of the page (including expiry) will not be changed.
+	 * - `'default'`|`'preferences'`: watchlist status of the page will be
+	 * set based on the user's preference settings when the action is
+	 * called. Defaults to an indefinite watch unless `watchlistExpiry` is
+	 * provided.
+	 * - `'unwatch'`: explicitly unwatch the page.
+	 * - Any other `string` or `number`, or a `Morebits.date` or `Date`
+	 * object: watch page until the specified time, deferring to
+	 * `watchlistExpiry` if provided.
+	 * @param {string|number|Morebits.date|Date} [watchlistExpiry=infinity] -
+	 * A date-like string or number, or a date object.  If a string or number,
+	 * can be relative (2 weeks) or other similarly date-like (i.e. NOT "potato"):
+	 * ISO 8601: 2038-01-09T03:14:07Z
+	 * MediaWiki: 20380109031407
+	 * UNIX: 2147483647
+	 * SQL: 2038-01-09 03:14:07
+	 * Can also be `infinity` or infinity-like (`infinite`, `indefinite`, and `never`).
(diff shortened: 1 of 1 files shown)


morebits.js
```diff
@@ -2930,36 +2930,72 @@ Morebits.wiki.page = function(pageName, status) {
...
 
 	/**
-	 * Set an expiry. setWatchlist can handle this by itself if passed a
-	 * string, so this is here largely for completeness and compatibility.
+	 * Set a watchlist expiry. setWatchlist can mostly handle this by
+	 * itself, so this is here largely for completeness and compatibility
+	 * with the full suite of options.
 	 *
-	 * @param {string} watchlistExpiry - A date-like string or array of strings
+	 * @param {string|number} watchlistExpiry=infinity - A date-like string or number.
```
MusikAnimal, line 2998: `[watchlistExpiry=infinity]` I think is the valid JSDoc syntax

MusikAnimal (approved):
One tiny nitpick, otherwise looks good to me! It doesn't look like it enforces user input to not exceed 6 months from the present, but in doing so the API will simply watch for 6 months instead and return a warning, so it's not critical.

Thanks for making use of watchlist expiry :)

Amorymeltzer:
> One tiny nitpick, otherwise looks good to me! It doesn't look like it enforces user input to not exceed 6 months from the present, but in doing so the API will simply watch for 6 months instead and return a warning, so it's not critical.

Given the API is (moderately) forgiving here, I figured better to just leave it.  Could check that in `fnApplyWatchlistExpiry` though

> Thanks for making use of watchlist expiry :)
It rules!

siddharthvp:
> > Thanks for making use of watchlist expiry :)
> 
> It rules!

Meanwhile, people on-wiki [are mystified](https://en.wikipedia.org/wiki/Wikipedia:Village_pump_(technical)#Temporary_watching_preference).

morebits.js
```diff
@@ -2930,36 +2930,72 @@ Morebits.wiki.page = function(pageName, status) {
...
+	 * watch unless `watchlistExpiry` is provided.
+	 * - `false`|`'no'`|`'nochange'`: watchlist status of the page (including expiry) will not be changed.
+	 * - `'default'`|`'preferences'`: watchlist status of the page will be
+	 * set based on the user's preference settings when the action is
+	 * called. Defaults to an indefinite watch unless `watchlistExpiry` is
+	 * provided.
+	 * - `'unwatch'`: explicitly unwatch the page.
+	 * - Any other `string` or `number`: watch page until the specified
+	 * time, deferring to `watchlistExpiry` if provided.
+	 * @param {string|number} [watchlistExpiry=infinity] - A date-like string or number.
```
siddharthvp, line 2948: What number? The switch doesn't seem to cover the case of a number.

siddharthvp: Actually wouldn't it make sense to allow a Date or Morebits.date object here (to be passed to API after `.toISOString`ing it)?

morebits.js
```diff
@@ -2930,36 +2930,72 @@ Morebits.wiki.page = function(pageName, status) {
...
+				break;
+			case 'watch':
+			case 'yes':
+			case true:
+				ctx.watchlistOption = 'watch';
+				ctx.watchlistExpiry = watchlistExpiry;
+				break;
+			default: // Not really a "default" per se but catches "any other string"
+				ctx.watchlistOption = 'watch';
+				ctx.watchlistExpiry = watchlistExpiry || watchlistOption;
```
siddharthvp, line 2988: `watchlistExpiry || watchlistOption`  ?

morebits.js
Amorymeltzer: > Actually wouldn't it make sense to allow a Date or Morebits.date object here (to be passed to API after `.toISOString`ing it)?

Oh that'd be neat, for all involved!

morebits.js
Amorymeltzer: Mmm, yeah, good call.  This was a holdover from a previous structuring, but I was trying to be nice by defaulting to the explicit `watchlistExpiry` if both were provided.  With the overwriting to `infinity`, though, this clearly stopped working!  I've removed it and just used `watchlistOption`.

I suppose another idea (more in line with what I had initially) would be to remove the `typeof watchlistExpiry === 'undefined'` line and use that in just the `watch` and `preferences` options, then this could be a `typeof` ternary.  Probably excessive.

PR #1323, opened 2021-02-17
block: Leave block template on the page of the current IP even when blocking a /64

molly:
Following from the conversation beginning at https://github.com/wikimedia-gadgets/twinkle/pull/1266#issuecomment-779498685, this changes the default behavior to leave a block template on the talk page for the IP whose talk page or contribs the block is originating from, even when the block is being applied to the /64.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -73,9 +73,7 @@ Twinkle.block.callback = function twinkleblockCallback() {
 				label: 'Add block template to user talk page',
 				value: 'template',
 				tooltip: 'If the blocking admin forgot to issue a block template, or you have just blocked the user without templating them, you can use this to issue the appropriate template. Check the partial block box for partial block templates.',
-				// Disallow for IP ranges
-				checked: !Morebits.ip.isRange(relevantUserName),
-				disabled: Morebits.ip.isRange(relevantUserName)
+				checked: true
 			}
 		]
 	});
@@ -112,7 +110,6 @@ Twinkle.block.callback = function twinkleblockCallback() {
 			list: [{
 				checked: relevantUserName !== mw.config.get('wgRelevantUserName'), // In case the user closes and reopens the form
 				label: 'Block the /64 instead',
-				tooltip: 'Will eschew leaving a template.',
 				value: 'block64'
 			}]
 		});
@@ -251,10 +248,8 @@ Twinkle.block.callback.change_block64 = function twinkleblockCallbackChangeBlock
 	var priorName = relevantUserName;
 	if ($block64.is(':checked')) {
 		relevantUserName = Morebits.ip.get64(mw.config.get('wgRelevantUserName'));
-		$form.find('[name=actiontype][value=template]').prop('disabled', true).prop('checked', false);
 	} else {
 		relevantUserName = mw.config.get('wgRelevantUserName');
-		$form.find('[name=actiontype][value=template]').prop('disabled', Morebits.ip.isRange(relevantUserName)).prop('checked', !Morebits.ip.isRange(relevantUserName));
 	}
 
 	// Refetch/reprocess user info then regenerate the main content
@@ -1992,7 +1987,9 @@ Twinkle.block.callback.evaluate = function twinkleblockCallbackEvaluate(e) {
 };
 
 Twinkle.block.callback.issue_template = function twinkleblockCallbackIssueTemplate(formData) {
-	var userTalkPage = 'User_talk:' + relevantUserName;
+	// Use wgRelevantUserName to ensure the block template goes to a single IP and not to the
+	// "talk page" of an IP range (which does not exist)
+	var userTalkPage = 'User_talk:' + mw.config.get('wgRelevantUserName');
 
 	var params = $.extend(formData, {
 		messageData: Twinkle.block.blockPresetsInfo[formData.template],
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -73,7 +73,9 @@ Twinkle.block.callback = function twinkleblockCallback() {
 				label: 'Add block template to user talk page',
 				value: 'template',
 				tooltip: 'If the blocking admin forgot to issue a block template, or you have just blocked the user without templating them, you can use this to issue the appropriate template. Check the partial block box for partial block templates.',
-				checked: true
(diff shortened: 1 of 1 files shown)


modules/twinkleblock.js
```diff
@@ -73,9 +73,7 @@ Twinkle.block.callback = function twinkleblockCallback() {
 				label: 'Add block template to user talk page',
 				value: 'template',
 				tooltip: 'If the blocking admin forgot to issue a block template, or you have just blocked the user without templating them, you can use this to issue the appropriate template. Check the partial block box for partial block templates.',
-				// Disallow for IP ranges
-				checked: !Morebits.ip.isRange(relevantUserName),
-				disabled: Morebits.ip.isRange(relevantUserName)
+				checked: true
```
Amorymeltzer, line 76:
```suggestion
				// Disallow for (initial) IP ranges
				checked: !Morebits.ip.isRange(relevantUserName),
				disabled: Morebits.ip.isRange(relevantUserName)
```

I think actually we want to keep this.  Although not (yet) supported, when eventually we have `wgRelevantUserName` for IP ranges, this will set the default for those cases.  That is, it won't make a difference now, since this will never get triggered atm, but will be necessary to be restored when it does come.

modules/twinkleblock.js
```diff
@@ -251,10 +248,8 @@ Twinkle.block.callback.change_block64 = function twinkleblockCallbackChangeBlock
 	var priorName = relevantUserName;
 	if ($block64.is(':checked')) {
 		relevantUserName = Morebits.ip.get64(mw.config.get('wgRelevantUserName'));
-		$form.find('[name=actiontype][value=template]').prop('disabled', true).prop('checked', false);
 	} else {
 		relevantUserName = mw.config.get('wgRelevantUserName');
-		$form.find('[name=actiontype][value=template]').prop('disabled', Morebits.ip.isRange(relevantUserName)).prop('checked', !Morebits.ip.isRange(relevantUserName));
```
Amorymeltzer, line 257: Again, not critical now, but we will need something once `wgRelevantUserName` is defined for IP ranges.  At that point, if blocking a range smaller than a /64 (/65, etc.), the `block64` will be offered, but we don't want to offer a template in that case.

Like below with the template, though, it'd awkwardly depend on the original IP, so maybe something like:

```js
// No templates for ranges, but if the original user is a single IP, offer the option
// (done separately in Twinkle.block.callback.issue_template)
var originalIsRange = Morebits.ip.isRange(mw.config.get('wgRelevantUserName'));
$form.find('[name=actiontype][value=template]').prop('disabled', originalIsRange).prop('checked', !originalIsRange);
```

modules/twinkleblock.js
```diff
@@ -112,7 +110,6 @@ Twinkle.block.callback = function twinkleblockCallback() {
 			list: [{
 				checked: relevantUserName !== mw.config.get('wgRelevantUserName'), // In case the user closes and reopens the form
 				label: 'Block the /64 instead',
-				tooltip: 'Will eschew leaving a template.',
```
Amorymeltzer, line 115: Since this is new/weird, I think it'd be nice to note exactly what will happen.  It'll also be necessary when ranges are enabled, since in those cases we *will* skip the template.  Maybe something like:

```js
tooltip: Morebits.ip.isRange(mw.config.get('wgRelevantUserName')) ? 'Will eschew leaving a template.' : 'Any template issued will go to the original IP: ' + mw.config.get('wgRelevantUserName'),
```

modules/twinkleblock.js
```diff
@@ -1992,7 +1987,9 @@ Twinkle.block.callback.evaluate = function twinkleblockCallbackEvaluate(e) {
 };
 
 Twinkle.block.callback.issue_template = function twinkleblockCallbackIssueTemplate(formData) {
-	var userTalkPage = 'User_talk:' + relevantUserName;
+	// Use wgRelevantUserName to ensure the block template goes to a single IP and not to the
+	// "talk page" of an IP range (which does not exist)
```
Amorymeltzer, line 1991: Indeed!

Amorymeltzer (changes requested):
Looks good!  I've made a couple of suggestions to ensure everything works smoothly once/if blocking ranges gets actually supported, but nothing too major.

modules/twinkleblock.js
```diff
@@ -251,11 +251,13 @@ Twinkle.block.callback.change_block64 = function twinkleblockCallbackChangeBlock
...
 		relevantUserName = Morebits.ip.get64(mw.config.get('wgRelevantUserName'));
-		$form.find('[name=actiontype][value=template]').prop('disabled', true).prop('checked', false);
 	} else {
 		relevantUserName = mw.config.get('wgRelevantUserName');
-		$form.find('[name=actiontype][value=template]').prop('disabled', Morebits.ip.isRange(relevantUserName)).prop('checked', !Morebits.ip.isRange(relevantUserName));
 	}
+  // No templates for ranges, but if the original user is a single IP, offer the option
+  // (done separately in Twinkle.block.callback.issue_template)
+  var originalIsRange = Morebits.ip.isRange(mw.config.get('wgRelevantUserName'));
+  $form.find('[name=actiontype][value=template]').prop('disabled', originalIsRange).prop('checked', !originalIsRange);
```
molly, line 260: Am I correct that this should happen for both cases?

modules/twinkleblock.js
```diff
@@ -73,7 +73,7 @@ Twinkle.block.callback = function twinkleblockCallback() {
 				label: 'Add block template to user talk page',
 				value: 'template',
 				tooltip: 'If the blocking admin forgot to issue a block template, or you have just blocked the user without templating them, you can use this to issue the appropriate template. Check the partial block box for partial block templates.',
-				// Disallow for IP ranges
+				// Disallow when viewing the block dialog on an IP range
```
Amorymeltzer, line 76: 👍

modules/twinkleblock.js
Amorymeltzer: Yeah — both to offer the template for the single IPv6 if the box is checked (as per this PR), and as normal if the box is unchecked.

Amorymeltzer (approved):
Awesome — all checks out to me.  Nice catch on the utility of this, glad you saw it! 🎖️

PR #1324, opened 2021-02-19
Fix Morebits.wiki.page.revert, use in fluff

Amorymeltzer:
`Morebits.wiki.page.revert` has apparently been largely crippled since its inception in 2011 (3617c19) due to overriding the revid in question, and further since 3caaede (late 2013), which unintentionally disabled the (required) user.  It should now finally work!

I've enabled `load`ing the page before reverting (in partcular by making `setOldID` set `editMode === 'revert'`), in case you want to process the user; ths also opens the door to including additional extra information in the future (content?  multiple revisions?).  It now roughly mirrors `setAppendText` or `setPrependText`, which makes sense, although will always require loading, so is perhaps a bit of a hybrid.  I also changed the default revert edit summary to be set on processing, not beforehand, and then only if one is not explicitly provided, as well as fixed handling of hidden users.

I've also set up `baserevid` for both `revert` and `all` modes, in addition to the already-present `basetimestamp` and `starttimestamp`, which is a fairly new (T34037) addition.  I also transitioned to using `page.touched` timestamp rather than `rev.timestamp`, since that's always available, means we don't need to query `rvprop=timestamp`, and should be more useful.  `getLastEditTime` will now return that.  I also added `getRevisionID` to get the fetched revision (mirrors `getRevisionUser`).

I've also done various cleanups and renames of internal context variables for clarity.  There should be no outward-facing changes, but I renamed:
- `timestamp` to `creationTimestamp`
- `revertCurID` to `latestRevID`
- `lastEditTime` to `lastTouchedTime`

With that done, we can (finally) use it in `fluff` to simplify things.  `revertToRevision` needs a `load` to snag the user for `Twinkle.fluff.formatSummary`, but otherwise the whole process is straightforward what `Morebits.wiki.page.revert` was designed for.  The rollback links of course still need the direct API query in order to get and process the slew of revisions, but we can simplify proccesing quite a bit, and finally remove the duplication in the completion callback.

Two prior commits set the groundwork:
- Don't treat a nochange response as a success
- Add getter for saveApi response to allow post-save processing


morebits.js
```diff
@@ -3614,10 +3654,36 @@ Morebits.wiki.page = function(pageName, status) {
...
+			ctx.latestRevID = page.lastrevid;
+
+			rev = page.revisions[0];
+			if (ctx.editMode === 'revert') {
+				// Is this ever even possible?
+				if (rev.revid !== ctx.revertOldID) {
+					ctx.statusElement.error('The retrieved revision does not match the requested revision.');
+					ctx.onLoadFailure(this);
+					return;
+				}
```
Amorymeltzer, line 3670: Is it?  There was a check for this in fluff, but I can't say I've seen or heard of it happening.  At the moment, this is the only reason we [still need to query](https://github.com/wikimedia-gadgets/twinkle/pull/1324/files#diff-20b1ff1a748addb118304a1b428701029480dc64e319f65a82cd213ea3dc8083R2602) for `rvprop=ids` when `revert`ing.

morebits.js
```diff
@@ -3614,10 +3654,36 @@ Morebits.wiki.page = function(pageName, status) {
 		var page = response.pages[0], rev;
 		ctx.pageExists = !page.missing;
 		if (ctx.pageExists) {
-			rev = page.revisions[0];
-			ctx.lastEditTime = rev.timestamp;
-			ctx.pageText = rev.content;
 			ctx.pageID = page.pageid;
+			ctx.lastTouchedTime = page.touched; // Used as basetimestamp when saving
```
Amorymeltzer, line 3658: Noted in the commit summary, but I've changed to using `touched` rather than `rev.timestamp`: `revert` will still need to use `touched`, and it's probably better anyway.  It also means we don't need to query for `rvprop=timestamp` ever.

That being said, we *could* keep `timestamp` for `revert`, and include a separate getter for that, to go alongside e.g. `getRevisionUser`.

It does make `getLastEditTime` slightly incorrect or misleasing, but does that matter?  I suppose we could default to `rev.timestamp` for non-`revert` modes, then use `touched` just for revert`, but that seems worse.

morebits.js
```diff
@@ -2724,17 +2732,28 @@ Morebits.wiki.page = function(pageName, status) {
...
 				query.starttimestamp = ctx.loadTime; // check that page hasn't been deleted since it was loaded (don't recreate bad stuff)
 				break;
 			default: // 'all'
 				query.text = ctx.pageText; // replace entire contents of the page
-				if (ctx.lastEditTime) {
-					query.basetimestamp = ctx.lastEditTime; // check that page hasn't been edited since it was loaded
+				// check that page hasn't been edited since it was loaded
+				if (ctx.lastTouchedTime) {
+					query.basetimestamp = ctx.lastTouchedTime;
+				}
```
Amorymeltzer, line 2754: With `baserevid` is it worth holding on to `basetimestamp`?  Is there synergy?

modules/twinklefluff.js
```diff
@@ -672,112 +614,82 @@ Twinkle.fluff.callbacks = {
...
-			undoafter: params.goodid,
-			basetimestamp: touched,
-			starttimestamp: loadtimestamp,
-			minor: Twinkle.getPref('markRevertedPagesAsMinor').indexOf(params.type) !== -1 ? true : undefined,
-			format: 'json'
-		};
-		// Handle watching, possible expiry
-		if (Twinkle.getPref('watchRevertedPages').indexOf(params.type) !== -1) {
-			var watchOrExpiry = Twinkle.getPref('watchRevertedExpiry');
+			params.csrftoken = apiobj.getResponse().query.tokens.csrftoken;
```
Amorymeltzer, line 617: We have to query so I suppose we might as well, but seems silly.

modules/twinklefluff.js
```diff
@@ -360,6 +360,8 @@ Twinkle.fluff.revert = function revertPage(type, vandal, rev, page) {
 		revid: revid
 	};
 
+	// Largely recapitulates Morebits.wiki.page.load, but we want to
+	// process multiple revisions as well as discover flagged status
 	var query = {
```
Amorymeltzer, line 365: I think everything here is still used/needed?

modules/twinklefluff.js
```diff
@@ -379,121 +381,61 @@ Twinkle.fluff.revert = function revertPage(type, vandal, rev, page) {
...
+		var revs = page.revisions;
 		if (revs.length < 1) {
 			statelem.error('We have less than one additional revision, thus impossible to revert.');
 			return;
 		}
 		var top = revs[0];
 		var lastuser = top.user;
 
+		// Should be handled by the API, but nice to quit early
 		if (lastrevid < params.revid) {
```
Amorymeltzer, line 439: The API handles this, but we could do a similar early-fail in Morebits if we wanted.

modules/twinklefluff.js
```diff
@@ -672,112 +614,82 @@ Twinkle.fluff.callbacks = {
...
+			};
+			var wikipedia_api = new Morebits.wiki.api('Automatically accepting your changes', query);
+			wikipedia_api.post();
 		}
 	}
 };
 
+// Format a nicer edit summary than the default Morebits revert one, mainly by
+// including user contribs and talk links and appending a custom reason.
 // If builtInString contains the string "$USER", it will be replaced
```
Amorymeltzer, line 693: I wouldn't be opposed to putting this in Morebits itself...

Amorymeltzer:
Currently live on testwiki

Amorymeltzer:
@siddharthvp I take it from your close of #1318 you no longer oppose this?

siddharthvp:
> Don't treat a nochange response as a success

This can be problematic in other contexts. Both the API itself and `mw.Api` treat nochange as success. I don't think this is a positive apart from the case of reverting, so maybe this behaviour should be restricted for type=revert mode only. (Which is one of the reasons why I don't think things like reverting should be part of morebits, but not that I want to get into that now...)

PR #1329, opened 2021-02-21
talkback: Remove cleanup for talkbackHeading prefs with erroneous equals signs

Amorymeltzer:
This was added in 2010 ([sysop-only link](https://en.wikipedia.org/w/index.php?title=Special:Undelete&target=User:Ioeth/friendlytalkback.js&timestamp=20100623142531&diff=prev)) to fix an issue with the [old Friendly documentation](https://en.wikipedia.org/w/index.php?title=Wikipedia:Friendly&oldid=410035412#Parameter_Descriptions).  In an ideal world, we'd have ways to clean up erroneous entries in preferences, but for this, I think we can remove it.  There are, on enWiki, on [five instances](https://en.wikipedia.org/w/index.php?title=Special:Search&limit=250&offset=0&ns2=1&sort=last_edit_desc&search=insource:/talkbackHeading/+intitle:twinkleoptions&advancedSearc$), with only three of those editors active.  I can ping them or clean those up directly, but I don't think we need to hold on to this.

I added tooltip instructions in the prefs for this, and for the mail heading, which we don't have a check for.

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -392,14 +392,12 @@ Twinkle.talkback.callbacks = {
 				}
 				break;
 			case 'see':
-				// clean talkback heading: strip section header markers that were erroneously suggested in the documentation
-				var heading = Twinkle.getPref('talkbackHeading').replace(/^\s*=+\s*(.*?)\s*=+$\s*/, '$1');
+				var heading = Twinkle.getPref('talkbackHeading');
 				text = '{{subst:Please see|location=' + input.page + (input.section ? '#' + input.section : '') +
 				'|more=' + input.message + '|heading=' + heading + '}}';
 				break;
 			default:  // talkback
-				// clean talkback heading: strip section header markers that were erroneously suggested in the documentation
-				text = '==' + Twinkle.getPref('talkbackHeading').replace(/^\s*=+\s*(.*?)\s*=+$\s*/, '$1') + '==\n' +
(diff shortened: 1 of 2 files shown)


Amorymeltzer:
> In an ideal world, we'd have ways to clean up erroneous entries in preferences

On this point, it's not particularly complicated and we already do it a bit for integers, but it's something to discuss first since I can imagine it ballooning.

siddharthvp:
[Link to the 5 instances](https://en.wikipedia.org/w/index.php?sort=last_edit_desc&search=insource%3A%2FtalkbackHeading%5C%22%3A+%5C%22%5Cs*%3D.*%3F%5C%22%2F+intitle%3Atwinkleoptions&title=Special%3ASearch&profile=advanced&fulltext=1&advancedSearch-current=%7B%7D&ns2=1). I see there are also [15 instances](https://en.wikipedia.org/w/index.php?sort=last_edit_desc&search=insource%3A%2FwelcomeHeading%5C%22%3A+%5C%22%5Cs*%3D.*%3F%5C%22%2F+intitle%3Atwinkleoptions&title=Special%3ASearch&profile=advanced&fulltext=1&advancedSearch-current=%7B%7D&ns2=1) of equals signs in welcomeHeading.

siddharthvp:
> I see there are also 15 instances of equals signs in welcomeHeading.

Looks like welcomeHeading config is long deprecated.

siddharthvp:
Notified 4 of the 5 affected users on their talk page (the other user was indef-blocked).

PR #1331, opened 2021-02-22
morebits.css: Add 2px margin-right to checkboxes, clearer spacing between elements

Amorymeltzer:
The label for a checkbox or radio button was abutting just too close to the box/button on Vector/Timeless/Monobook (not Modern?), this should space them out a little more clearly.

### DIFF
--- a/morebits.css
+++ b/morebits.css
@@ -68,8 +68,9 @@ form.quickform select
 form.quickform input[type=checkbox],
 form.quickform input[type=radio] {
 	height: 13px;
-	margin-bottom: 2px;
 	margin-top: 2px;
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I think there's possibly some value in just doing `margin: 2px;`.  Basically, without the left margin, the boxes are aligned with text (e.g. XfD) but *not* H5 headers (e.g. CSD).  You can see this in the tag module, where the main headers are H5s, so indented a bit (`margin: .5em 0 0;` and `padding: .3em .2em .2em;`), but the subsections are not.  Not sure how to split that baby, or if it's worth unindenting the H5s.

PR #1335, opened 2021-02-25
protect: Use input type=number for revid, enforce minimum

Amorymeltzer:
As suggested at https://github.com/wikimedia-gadgets/twinkle/pull/1293#issuecomment-778195064.

The minimum is just to catch any silly mistakes; it could be regularly updated, but the length and leading digit probably matter the most.

----

I don't really think the minimum is worth much.  It's not a great thing to attempt to maintain, but enWiki happens to have recently hit a major milestone, so using `1000000000` is semi convenient.  Maybe it's a net-positive on-wiki, but I don't really think it's a benefit here.

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -92,9 +92,15 @@ Twinkle.protect.callback = function twinkleprotectCallback() {
 };
 
 
+/** Installation-specific information, likely needing customization */
 // A list of bots who may be the protecting sysop, for whom we shouldn't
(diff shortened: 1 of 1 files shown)


siddharthvp:
I'm not sure about keeping the min check, as that's very awkward to localise ... Can we do something smart to avoid the hardcoding, maybe check if the input is of the same order of magnitude as `mw.config.get('wgRevisionId')` or something?

PR #1339, opened 2021-02-26
morebits: Update statusElement on successful lookupCreation

### LINKED ISSUE #1049: Avoid hanging status messages

Amorymeltzer:
`lookupCreation` would often leave a hanging status message when used on a new/separate instance of `Morebits.wiki.page` since we were never actually registering a success.  When using the original page object, things would be fine, since the corresponding method would typically pick up the slack.  Fixes #1049.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -3883,6 +3883,8 @@ Morebits.wiki.page = function(pageName, status) {
 				ctx.onLookupCreationFailure(this);
 				return;
 			}
+
+			ctx.statusElement.info('retrieved page creation information');
 			ctx.onLookupCreationSuccess(this);
 
 		} else {
@@ -3925,6 +3927,7 @@ Morebits.wiki.page = function(pageName, status) {
 			return;
 		}
 
+		ctx.statusElement.info('retrieved page creation information');
 		ctx.onLookupCreationSuccess(this);
 
 	};


siddharthvp:
Does this fully fix #1049? I thought there were many more of these.

Amorymeltzer:
Maybe yes?  Maybe no?  I haven't done an exhaustive survey, but the easy ones to find were all this case.

The other culprit would be similar behavior or unresolved issues after a `.load`.  That's possible of course, but it's less likely to happen since we're almost always trying to use that same object — especially given that the success callback is required — whereas `lookupCreation` is often orthogonal/in-addition to the main page object.  It *does* happen in a few places I know of offhand (a few preview functions, warn's autoLevel processing), but they aren't actually displaying the status object so doesn't matter.

morebits.js
```diff
@@ -3883,6 +3883,8 @@ Morebits.wiki.page = function(pageName, status) {
 				ctx.onLookupCreationFailure(this);
 				return;
 			}
+
+			ctx.statusElement.info('found page creation information');
```
siddharthvp, line 3887:
```suggestion
			ctx.statusElement.info('retrieved page creation information');
```
?

Amorymeltzer: Better!  Done.

PR #1345, opened 2021-03-02
morebits: quickform: fix duplication of IDs in select/optgroup

### LINKED ISSUE #1344: quickForm: select or optgroup with list give invalid HTML
If quickForm types `select` or `optgroup` are used with list (the list of `option`s), the select element, optgroup element and all option elements inside get the same ID, which is invalid HTML. Browsers still seem able to render them correctly but external tools like select2 freak out on seeing this. 

Steps to replicate:
```js
new Morebits.quickForm.element({
  type: 'optgroup',
  label: 'group label',
  list: [
    { label: 'option 1', value: 'value1' },
    { label: 'option 2', value: 'value2' },
  ]
}).render()
```

or 

```js
new Morebits.quickForm.element({
  type: 'select',
  label: 'group label',
  list: [
    { label: 'option 1', value: 'value1' },
    { label: 'option 2', value: 'value2' },
    { label: 'group', list: [
		{ label: 'option 3', value: 'value3' },
    	{ label: 'option 4', value: 'value4' }
	  ]  
	} 
  ]
}).render()
```

siddharthvp:
Fixes #1344.

Reminiscent of f92f0233b, although unrelated.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -252,7 +252,6 @@ Morebits.quickForm.prototype.append = function QuickFormAppend(data) {
 Morebits.quickForm.element = function QuickFormElement(data) {
 	this.data = data;
 	this.childs = [];
-	this.id = Morebits.quickForm.element.id++;
 };
 
 /**
@@ -302,7 +301,7 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 	var node;
 	var childContainder = null;
 	var label;
-	var id = (in_id ? in_id + '_' : '') + 'node_' + this.id;
+	var id = (in_id ? in_id + '_' : '') + 'node_' + Morebits.quickForm.element.id++;
 	if (data.adminonly && !Morebits.userIsSysop) {
 		// hell hack alpha
 		data.type = 'hidden';


Amorymeltzer:
So, to explain: The sub-`compute`s would have been stuck using the wrong info since they weren't being supplied a `Morebits.quickForm.element` but rather a `data` object, yeah?  Does the same thing happen for `dyninput`?

morebits.js
```diff
@@ -302,7 +301,7 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 	var node;
 	var childContainder = null;
 	var label;
-	var id = (in_id ? in_id + '_' : '') + 'node_' + this.id;
+	var id = (in_id ? in_id + '_' : '') + 'node_' + Morebits.quickForm.element.id++;
```
Amorymeltzer, line 304: So, this means that simply doing `new Morebits.quickForm.element({data})` no longer results in an `id`, since that's not created until here.  I'm not sure that matters? Maybe in theory for `getElements`?

I wonder if it wouldn't be better to make use of the second parameter to `compute` in both these cases?  It would mean not incrementing the value numerically, but is this not what `in_id` is for?  Alternatively, what about explicitly supplying the new `data` with `Morebits.quickForm.element.id++`?  I've not tested either yet.

siddharthvp: This approach feels better in theory - every time we enter compute we are going to have an HTMLElement by the end of it with a unique id. The 2nd param of compute() is only used in radio/checkbox subgroups. Using that seems like a more complicated approach, as we don't care about what the IDs are - just that they be distinct. 

>So, this means that simply doing new Morebits.quickForm.element({data}) no longer results in an id, since that's not created until here. I'm not sure that matters? Maybe in theory for getElements?
 
it shouldn't matter at all.

siddharthvp:
Live on testwiki.

morebits.js
Amorymeltzer: Agree that it really doesn't matter since there's no real meaning to any of it, but semantically, I would think it'd make sense for the optgroup and options to appear with some connection to the parent.  At any rate, a quick test with `in_id` didn't work, so I'm not sure it's actually doing what I think it should.  Maybe down the road but clearly no need to deal with it now.  I did have luck with just setting the id explicitly, but I suppose that's no real difference!

Issue #593, opened 2019-04-24
What bots should we avoid notifying?


Amorymeltzer:
People should generally avoid notifying bots of CSD/PROD/XFD nominations.  People do it anyway.  Absent an improved attention to detail when using the tools, we might consider blacklisting some commons targets.  `twinklespeedy.js` already won't notify SoxBot and Cyberbot I on F2 taggings.  Relatedly, `fluff` skips over AnomieBOT and SineBot, reverting the user beforehand.

Not a lot of bots create many pages; excluding archives and spam subpages, these seem to be the main ones:

- Cydebot
- Eubot
- AnomieBOT
- MenoBot
- DumbBOT

Some redirect to the botop and some don't, but only Cydebot and Eubot seem to receive any appreciable amount of notifications.  Cydebot has the usertalk page full protected, which suggests it actively doesn't want any.  We could skip Cydebot for C1 nominations and Eubot for RfD noms, but are there any others?

Amorymeltzer:
Although new, @JJMC89 bot III is likely to end up in the same situation as Cydebot, I think.

JJMC89:
718 Bot has a bunch of file creations from 2008–2009 and a fully protected talk page to prevent notices.

If it were up to me, bots would never get Twinkle notices.

I don't want deletion notices regarding any of my bots' creations.

siddharthvp:
> If it were up to me, bots would never get Twinkle notices.

+1. Let's disable all deletion notifications to bots for all non-mainspace pages. When an article is nominated, I think we should notify since the content-creator bots (like Rambot and Qbugbot) may want to be notified.

Amorymeltzer:
Given that some folks redirect the talk page and some don't I imagine there's likely to be variability.  I think maybe I'll start a convo at BOTN.

PR #1349, opened 2021-03-03
Add Morebits.wiki.user class for user info and processing

Amorymeltzer:
We have `Morebits.wiki.page`, but a user-specific class was missing from the library, meaning common actions like (un)blocking would be left to bespoke queries with `Morebits.wiki.api`.  Even more commonly, finding out information about a given user is often desired and those (equally-bespoke) queries get complicated depending on what's desired.  The ability *not* to notify bots or indef-blocked users is oft-requested, as is the ability to opt-out of Twinkle notifications.  To do that we need to actually know something about the user, for which `Morebits.wiki.page` is insufficient.

Key features of the `Morebits.wiki.user` class include:
- Simple loading of user info, allowing tools to easily access useful information (blocked status, user groups, registration time, etc.)
- Methods to block and unblock a user, as well as the ability to change user groups
- A `notify` method which checks users in order to skip bots or sitewide indef-blocked users, as well as usertalks with given templates or an opt-out link.

Its structure is very similar to `Morebits.wiki.page` but there are a few subtle differences and improvements.  The most notable difference is that here the user object is required to be `load`ed, so the methods, rather than checking if we need to query for protection/token/etc., will instead autoload the user if need be.  The only exception is when `notify`ing and you aren't checking for bot or indef-block status or optout templates/links; there's no need to `load` the user in that case.

Other things to note:
- The `load` success callback is optional, which while unlikely to make a difference writing tools should make single uses or testing easier
- The `notify` method uses `Morebits.wiki.page` (`append` or `newSection`) behind the scenes
- (un)block and user groups methods share success and error callbacks
- Added `Morebits.botUsernameRegex` to standardize a regex to match a "bot" username, set to `/bot\b` (see also #858 and #876)
- Answers/solves #593 (What bots should we avoid notifying?) as "all" (by default)
- Class is supported in `Morebits.batchOperation`
- Usable optout template example at [[User:Amorymeltzer/sandbox/el]]

With this PR, I've made use of it to:
- block: Get info, issue block
- xfd/csd/image/prod/tag/arv: Send notifications
- xfd/csd/image/prod/tag: Enable optout of notifications, don't notify {{deceased}} or {{retired}}
- talkback: Load {{no talkback}} optout preference
- arv: Load block status
- warn: Add note if blocked

talkback and shared need the `setMinorEdit` option, which could be done right now but hardly seems worth it, especially since we don't care about optouts for either.  Likewise for welcome, which also has the `topWelcome` pref, although it's possible the `auto` option might benefit.

(cc @MusikAnimal and @siddharthvp)

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -85,27 +85,24 @@ Twinkle.talkback.callback = function() {
 	result.tbtarget[0].dispatchEvent(evt);
 
 	// Check whether the user has opted out from talkback
-	var query = {
-		action: 'query',
-		prop: 'extlinks',
-		titles: 'User talk:' + mw.config.get('wgRelevantUserName'),
-		elquery: 'userjs.invalid/noTalkback',
-		ellimit: '1',
-		format: 'json'
-	};
-	var wpapi = new Morebits.wiki.api('Fetching talkback opt-out status', query, Twinkle.talkback.callback.optoutStatus);
-	wpapi.post();
+	var user = new Morebits.wiki.user(mw.config.get('wgRelevantUserName'), 'Fetching talkback opt-out status');
+	user.setNotifySkips('userjs.invalid/noTalkback', Twinkle.optoutTemplates);
+	user.load(Twinkle.talkback.callback.optoutStatus);
 };
 
 Twinkle.talkback.optout = '';
-
-Twinkle.talkback.callback.optoutStatus = function(apiobj) {
-	var el = apiobj.getResponse().query.pages[0].extlinks;
-	if (el && el.length) {
-		Twinkle.talkback.optout = mw.config.get('wgRelevantUserName') + ' prefers not to receive talkbacks';
-		var url = el[0].url;
-		var reason = mw.util.getParamValue('reason', url);
-		Twinkle.talkback.optout += reason ? ': ' + reason : '.';
+Twinkle.talkback.callback.optoutStatus = function(userobj) {
+	var tl = userobj.getTalkTemplates();
+	if (tl && tl.length) {
+		Twinkle.talkback.optout = mw.config.get('wgRelevantUserName') + ' transcludes {{' + tl[0] + '}}, so a talkack might not be helpful';
+	} else {
+		var el = userobj.getTalkLinks();
+		if (el && el.length) {
+			Twinkle.talkback.optout = mw.config.get('wgRelevantUserName') + ' prefers not to receive talkbacks';
+			var url = el[0].url;
+			var reason = mw.util.getParamValue('reason', url);
+			Twinkle.talkback.optout += reason ? ': ' + reason : '.';
+		}
 	}
 	$('#twinkle-talkback-optout-message').text(Twinkle.talkback.optout);
 };
--- a/morebits.js
+++ b/morebits.js
@@ -4693,7 +4693,7 @@ Morebits.wiki.user = function(userName, currentAction) {
 	userName = Morebits.ip.sanitizeIPv6(userTitle.getMainText());
 
 	if (!currentAction) {
-		currentAction = 'Querying user "' + userName + '"';
+		currentAction = msg('querying-user', userName, 'Querying user "' + userName + '"');
 	}
 
 	/**
@@ -4881,7 +4881,7 @@ Morebits.wiki.user = function(userName, currentAction) {
 			ctx.loadQuery.ellimit = 42;
 		}
 
-		ctx.userApi = new Morebits.wiki.api('Retrieving user information...', ctx.loadQuery, fnLoadSuccess, ctx.statusElement, ctx.onLoadFailure);
+		ctx.userApi = new Morebits.wiki.api(msg('fetching-userinfo', 'Retrieving user information...'), ctx.loadQuery, fnLoadSuccess, ctx.statusElement, ctx.onLoadFailure);
 		ctx.userApi.setParent(this);
 		ctx.userApi.post();
 	};
@@ -4904,7 +4904,7 @@ Morebits.wiki.user = function(userName, currentAction) {
 		// clients can do as they please.  This tells when the user
 		// was loaded, which is useful for time-based functions.
 		if (!ctx.loadTime) {
-			ctx.statusElement.error('Failed to retrieve current timestamp.');
+			ctx.statusElement.error(msg('failed-timestamp', 'Failed to retrieve current timestamp.'));
 			ctx.onLoadFailure(this);
 			return;
 		}
@@ -4913,10 +4913,10 @@ Morebits.wiki.user = function(userName, currentAction) {
 		response = response.query;
 
 		// Even if this is unnecessary (notification), an issue here
-		// is a likely indicates *something* went wrong.  The same
+		// likely indicates *something* went wrong.  The same
 		// as Morebits.wiki.page, though it's more necessary there.
 		if (!response.tokens.csrftoken || !response.tokens.userrightstoken) {
-			ctx.statusElement.error('Failed to retrieve tokens.');
+			ctx.statusElement.error(msg('failed-token', 'Failed to retrieve token.'));
 			ctx.onLoadFailure(this);
 			return;
 		}
@@ -4926,7 +4926,7 @@ Morebits.wiki.user = function(userName, currentAction) {
 		var user = response.users && response.users[0];
 		// Not sure scenario could lead to this, but might as well be safe
 		if (!user) {
-			ctx.statusElement.error('Failed to retrieve user ' + ctx.userName);
+			ctx.statusElement.error(msg('failed-userinfo', ctx.userName, 'Failed to retrieve user info for ' + ctx.userName));
 			ctx.onLoadFailure(this);
 			// force error to stay on the screen
 			++Morebits.wiki.numberOfActionsLeft;
@@ -5135,11 +5135,11 @@ Morebits.wiki.user = function(userName, currentAction) {
 		// If blocked and reblock is missing, assume we didn't know
 		// the user was already blocked, so ask to toggle
 		if (directBlock && !ctx.reblock) {
-			var message = ctx.userName + ' is already blocked (';
-			message += Morebits.string.isInfinity(this.getBlockExpiry()) ? 'indefinitely' : 'until ' + new Morebits.date(this.getBlockExpiry()).calendar();
-			message += '; by ' + this.getBlockingSysop() + '), would you like to override the block?';
+			var message = Morebits.string.isInfinity(this.getBlockExpiry())
+				? msg('already-blocked-indef', ctx.userName, this.getBlockingSysop(), ctx.userName + ' is already blocked (indefinitely; by ' + this.getBlockingSysop() + '), would you like to override the block?')
+				: msg('already-blocked', ctx.userName, this.getBlockExpiry(), this.getBlockingSysop(), ctx.userName + ' is already blocked (until ' +  new Morebits.date(this.getBlockExpiry()).calendar() + '; by ' + this.getBlockingSysop() + '), would you like to override the block?')
 			if (!confirm(message)) {
-				ctx.statusElement.error('Reblock aborted.');
+				ctx.statusElement.error(msg('reblock-aborted', 'Reblock aborted.'));
 				ctx.onBlockFailure(this);
 				return;
 			}
@@ -5149,7 +5149,7 @@ Morebits.wiki.user = function(userName, currentAction) {
 		// setExpiry allows arrays because userrights accepts it, but block doesn't
 		if (Array.isArray(ctx.expiry)) {
 			if (ctx.expiry.length !== 1) {
-				ctx.statusElement.error('You must provide a valid block expiration.');
+				ctx.statusElement.error(msg('invalid-block-expiry', 'You must provide a valid block expiration.'));
 				ctx.onBlockFailure(this);
 				return;
 			}
@@ -5159,16 +5159,14 @@ Morebits.wiki.user = function(userName, currentAction) {
 
 		// Check before indefing IPs or blocking sysops
 		if (ctx.isIP && Morebits.string.isInfinity(ctx.expiry) &&
-			!confirm(ctx.userName + ' is an IP address, do you really want to block it indefinitely?' +
-			'\n\nClick OK to proceed with the block, or Cancel to skip this block.')) {
-			ctx.statusElement.error('Infinite block of IP addressed was aborted.');
+			!confirm(msg('ip-indef-confirm', ctx.userName, ctx.userName + ' is an IP address, do you really want to block it indefinitely?' +
+			'\n\nClick OK to proceed with the block, or Cancel to abort.'))) {
+			ctx.statusElement.error(msg('ip-indef-aborted', 'Indefinite block of IP address was aborted.'));
 			ctx.onBlockFailure(this);
 			return;
 		} else if (this.isSysop() &&
-			!confirm(ctx.userName + ' is an administrator' +
-			(Morebits.string.isInfinity(ctx.grantedGroups.sysop) ? '' : ' (expires ' + new Morebits.date(ctx.grantedGroups.sysop).calendar('utc') + ' (UTC))') +
-			', are you sure you want to block them?  \n\nClick OK to proceed with the block, or Cancel to skip this block.')) {
-			ctx.statusElement.error('Block of administrator was aborted.');
+			!confirm(msg('admin-block-confirm', ctx.userName, ctx.userName + ' is an administrator, are you sure you want to block them?  \n\nClick OK to proceed with the block, or Cancel to abort.'))) {
+			ctx.statusElement.error(msg('admin-block-aborted', 'Block of administrator was aborted.'));
 			ctx.onBlockFailure(this);
 			return;
 		}
@@ -5194,7 +5192,7 @@ Morebits.wiki.user = function(userName, currentAction) {
 					(Array.isArray(ctx.namespacerestrictions) && ctx.namespacerestrictions.indexOf(3) === -1 && ctx.namespacerestrictions.indexOf('3') === -1) ||
 					(typeof ctx.namespacerestrictions === 'string' && ctx.namespacerestrictions.split('|').indexOf('3') === -1) ||
 					(typeof ctx.namespacerestrictions === 'number' && ctx.namespacerestrictions !== 3))) {
-					ctx.statusElement.error('Partial blocks cannot prevent talk page access unless also restricting User talk space.');
+					ctx.statusElement.error(msg('partial-usertalk', 'Partial blocks cannot prevent talk page access unless also restricting User talk namespace.'));
 					ctx.onBlockFailure(this);
 					return;
 				}
@@ -5229,7 +5227,7 @@ Morebits.wiki.user = function(userName, currentAction) {
 			}
 		}
 
-		ctx.blockApi = new Morebits.wiki.api('blocking user...', query, fnBlockSuccess, ctx.statusElement, fnBlockError);
+		ctx.blockApi = new Morebits.wiki.api(msg('blocking', 'blocking user...'), query, fnBlockSuccess, ctx.statusElement, fnBlockError);
 		ctx.blockApi.setParent(this);
 		ctx.blockApi.post();
 	};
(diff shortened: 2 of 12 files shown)


morebits.js
```diff
@@ -4566,215 +4577,1725 @@ Morebits.wiki.page = function(pageName, status) {
...
+	 * question.  Will be performed last, so is useful for applying
+	 * additional bespoke parameters, such as `setMinorEdit` or more complex
+	 * watch options to `setWatchlist`.
+	 *
+	 * @param {object} pageobjectFunctions - An object with `{function:
+	 * functionValue}` parameters.  Each key is the name of a
+	 * {@link Morebits.wiki.page} function, and its value is what will be
+	 * provided to that function.
+	 */
+	this.setPageobjectFunctions = function(pageobjectFunctions) {
```
Amorymeltzer, line 5849: This is hacky.  Basically, we want to be able to affect the `Morebits.wiki.page` object used to send the notification, but with minimal duplication of efforts.  What happens if `setPageobjFunctions` provides something like `append`?  It allows for additional things like `setMinor`, etc., but is weird.


It's required for full handling of the suite of `watchlistEnums`, which at the moment is xfd, speedy, and image.  That's maybe enough to justify doing something more involved within `Morebits.wiki.user` but that has its own pitfalls, since `watchuser` is binary.

morebits.js
```diff
@@ -4566,215 +4577,1725 @@ Morebits.wiki.page = function(pageName, status) {
...
+		if (ctx.notifySkipTemplates && ctx.notifySkipTemplates.length && ctx.talkTemplates && ctx.talkTemplates.length) {
+			// More efficient to do a for loop, but this is prettier?
+			var tlDups = Morebits.array.dups(ctx.talkTemplates.concat(ctx.notifySkipTemplates));
+			if (tlDups.length) {
+				ctx.statusElement.error('User talk page transcludes ' + tlDups[0] + ', aborting notification');
+				ctx.onNotifyFailure(this);
+				return;
+			}
+		} else if (ctx.notifySkipLink && ctx.talkLinks && ctx.talkLinks.length) {
+			// Should be without leading protocol; relying on mw.Uri could help
```
Amorymeltzer, line 5313: Is it worth leaning on `mw.Uri` just for this ?  I'm not sold, but it could simplify some possible user errors.

modules/friendlytalkback.js
```diff
@@ -85,27 +85,24 @@ Twinkle.talkback.callback = function() {
...
-		prop: 'extlinks',
-		titles: 'User talk:' + mw.config.get('wgRelevantUserName'),
-		elquery: 'userjs.invalid/noTalkback',
-		ellimit: '1',
-		format: 'json'
-	};
-	var wpapi = new Morebits.wiki.api('Fetching talkback opt-out status', query, Twinkle.talkback.callback.optoutStatus);
-	wpapi.post();
+	var user = new Morebits.wiki.user(mw.config.get('wgRelevantUserName'), 'Fetching talkback opt-out status');
+	user.setNotifySkips('userjs.invalid/noTalkback', Twinkle.optoutTemplates);
```
Amorymeltzer, line 89: Is it worth restructuring talkback/{{notalkback}} to use the new optout method/link?  It's not quite correct since we're not using it to optout of receiving a message, but then again we're not using this to send a message anyway.  In theory, others could use {{notalkback}}, so no, but then again I don't think anybody actually does.

morebits.js
```diff
@@ -4566,215 +4577,1725 @@ Morebits.wiki.page = function(pageName, status) {
...
+			var elDups = Morebits.array.dups(ctx.talkLinks.concat(ctx.notifySkipLink));
+			if (elDups.length) {
+				ctx.statusElement.error('User has opted out of this notification, aborting');
```
Amorymeltzer, line 5316: Should these messages on optouts be warnings rather than errors?  Likewise for indef or bot skipping?

morebits.js
```diff
@@ -4566,215 +4577,1725 @@ Morebits.wiki.page = function(pageName, status) {
...
+			if (ctx.namespacerestrictions) {
+				// This awfulness is to ensure other namespaces (e.g. 13) don't get caught up in here
+				if (!ctx.allowusertalk && (
+					(Array.isArray(ctx.namespacerestrictions) && ctx.namespacerestrictions.indexOf(3) === -1 && ctx.namespacerestrictions.indexOf('3') === -1) ||
+					(typeof ctx.namespacerestrictions === 'string' && ctx.namespacerestrictions.split('|').indexOf('3') === -1) ||
+					(typeof ctx.namespacerestrictions === 'number' && ctx.namespacerestrictions !== 3))) {
+					ctx.statusElement.error('Partial blocks cannot prevent talk page access unless also restricting User talk space.');
+					ctx.onBlockFailure(this);
+					return;
+				}
```
Amorymeltzer, line 5113: Is this full checking for partial namespace 3 not 13 stupid or what?  Should we just be less accommodating?

morebits.js
```diff
@@ -4566,215 +4577,1725 @@ Morebits.wiki.page = function(pageName, status) {
...
+		talkExists: null,
+		talkTimestamp: null,
+		talkLastEditor: null,
+		talkTemplates: null,
+		talkLinks: null,
+		message: null,
+		sectionTitle: null,
+		notifyBots: false,
+		notifyIndef: false,
+		notifySelf: false,
```
Amorymeltzer, line 4686: Pointless?  Unused atm because it's nicer to quit earlier but should this even be an option, or just always a given?

morebits.js
```diff
@@ -4566,215 +4577,1725 @@ Morebits.wiki.page = function(pageName, status) {
...
+	var fnGroupsSuccess = function() {
+		fnSuccess('groups');
+	};
+	var fnSuccess = function(action) {
+		ctx.actionResponse = ctx[action + 'Api'].response;
+
+		// `block: block` and `unblock: unblock`, but `groups: userrights`
+		var exactName = action === 'groups' ? 'userrights' : action;
+		// The API thinks we're successful if there's a response for the action,
+		// i.e. there isn't `result: 'Success'` like action=edit
```
Amorymeltzer, line 5526: Not necessarily relevant here, but the variability in successful API response among various actions is *maddening*, even down to different capitalization among parameters.

morebits.js
```diff
@@ -4566,215 +4577,1725 @@ Morebits.wiki.page = function(pageName, status) {
...
+			// as equivalent, meaning functionally-equivalent IP ranges may be
+			// misinterpreted.  Without logevents redirecting (like Special:Block does)
+			// we would need a function to parse ranges, which is a pain.
+			// IPUtils has code, but it'd be a lot of cruft for one purpose.
+			letype: 'block',
+			letitle: userTitle.toText(),
+			lelimit: 1,
+
+			// Get the talk page content, categories, etc.
+			titles: ctx.talkTitle,
```
Amorymeltzer, line 4761: Originally I had a `setLoadTalk` method to optionally load the talk, but since we rely on it for optouts, it was just getting in the way and more trouble than it was worth.

morebits.js
```diff
@@ -4566,215 +4577,1725 @@ Morebits.wiki.page = function(pageName, status) {
...
+			}
+
+		}
+
+		if (!ctx.notifyBots && this.isBot()) {
+			ctx.statusElement.error('User is a bot, aborting notification');
+			ctx.onNotifyFailure(this);
+			return;
+		}
+		// Clients may find this most useful iff notalk or the block isn't brand new
```
Amorymeltzer, line 5328: >block isn't brand new

Would be a good feature to add, IMO; maybe only for blocks older than two weeks or something.

morebits.js
```diff
@@ -4566,215 +4577,1725 @@ Morebits.wiki.page = function(pageName, status) {
...
-		parameters: {}
+		// Talk page stuff
+		// Ignore unresolved, invalid page titles (e.g. circular redirects)
+		var page = response.pages && response.pages[0];
+		if (page && !page.invalid) {
+			ctx.talkExists = !page.missing;
+			if (ctx.talkExists) {
+				// Update to redirect target or normalized name;
+				// no status message so as to avoid duplication when notifying
+				ctx.talkTitle = page.title;
```
Amorymeltzer, line 4964: Originally attempted to notify on redirection here, like in mb.w.page, but would mean duplication of that message when notifying.  Maybe not a big deal to leave off here?

siddharthvp:
This is craziness. Where was this discussed?

Amorymeltzer:
>Where was this discussed?

Here, where I'm proposing it, although I'd prefer to do so with civil discourse and constructive criticism rather than "[t]his is craziness."

MusikAnimal:
I can't say I have any strong opinions, but if the end result makes the module code more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), it seems like a good idea. Also other scripts might make use of the new class.

The opting out of Twinkle notifications seems a bit odd. You're saying this has been requested before? If we go the route of using a template, I worry it could be misused, especially if it's invisible. That bit might deserve broader discussion. A similar concept is with {{[unlocked userpage](https://en.wikipedia.org/wiki/Template:Unlocked_userpage)}}; however there only the owner of the userpage can add and remove the template (enforced by edit filter), and it also doesn't work if used in a transcluded template. Similar safeguards might be warranted here, but maybe I'm overthinking it.

siddharthvp:
> Here, where I'm proposing it, although I'd prefer to do so with civil discourse and constructive criticism rather than "[t]his is craziness."

How is it any different from your vitriolic comments at [other](https://github.com/wikimedia-gadgets/twinkle/pull/1220#issuecomment-747656779) [places](https://github.com/wikimedia-gadgets/twinkle/issues/1226#issuecomment-757500787)? I'd prefer to keep the discussion on facts rather than on civility lawyering. 

Do we write 2000 lines of code before discussing the approach in the first place? (All I know is that if someone does that at a company - they get fired.) Which is for good reason as it not easy to give constructive criticism when there are this many lines of code - because asking to rewrite the whole thing won't be constructive criticism.

Before that, one question: Do you expect me to port all of the changes to csd, xfd, warn and tag to twinkle-core/twinkle-enwiki? You never gave a hint that such a thing was in progress. If I had known that before, I would've prioritised other mdoules and/or structured my changes keeping it in mind. I'm going out of my way in localising twinkle while preserving every single twinkle feature that enwiki has, and I'm astonished to see that you've been making things more difficult starting from day 1.

siddharthvp:
- Aping the structure of Morebits.wiki.page is (forgive me I don't know a more civil word for this) craziness. Mb.w.p's design pattern may have made sense in 2011 when it was written, but it makes no sense now.
  - The methods are defined in the constructor rather than on the prototype which is an anti-pattern that among other things makes extending (sub-classing) very awkward (see the [Page class](https://github.com/wikimedia-gadgets/twinkle-core/blob/master/src/Page.ts) in twinkle-core). 
  - Use promises, and only promises. #911 was complicated because it tried to support both promises and callbacks. Using just promises makes for a very clean API which is more flexible, and also requires less code in morebits. It is also facilitates better error handling (say if a TypeError occurs in the middle of morebits - failure callback won't get triggered, but a promise automatically gets rejected). It is also more easily testable as in most test runners, just returning the promise in the test method will cause the runner to wait till it's resolved - it's less intuitive in case of callbacks).
- All of the oversighter-specific things and the function to change user rights appear out of scope. These are things that make sense in classes that extend Mb.w.u defined in admin-specific scripts. 95% of the users aren't even admins to use them. It is wasteful to load all this for everyone. Morebits isn’t even used much outside twinkle so I’m not sure who you’re expecting would use these features.
- Even the block/unblock things seem bit out of scope as they are used in just one module – no DRYness is being achieved here - that too an admin-only module again.
- On a minor note, please keep the comments in the code concise.

Amorymeltzer:
> The opting out of Twinkle notifications seems a bit odd. You're saying this has been requested before? If we go the route of using a template, I worry it could be misused, especially if it's invisible. That bit might deserve broader discussion. A similar concept is with {{[unlocked userpage](https://en.wikipedia.org/wiki/Template:Unlocked_userpage)}}; however there only the owner of the userpage can add and remove the template (enforced by edit filter), and it also doesn't work if used in a transcluded template. Similar safeguards might be warranted here, but maybe I'm overthinking it.

@MusikAnimal Yeah, there've been quite a few different requests over the years for basically every iteration; there's a recent [thread at the top of WT:TW](https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=1010170462#Notifying_deceased_users) about deceased users which lead to some talk on this, after which I rewrote some of this to agree with @siddharthvp's idea.  Some folks don't want certain notifications (IIRC AfC editors were a big one), and being able to choose is nice.  There's no real need for the template to be invisible, it's just that the text isn't particularly important.

siddharthvp:
Hey @Amorymeltzer Sorry about https://github.com/wikimedia-gadgets/twinkle/pull/1349#issuecomment-790431744 - it looks like I may have overreacted a bit as the changes in modules aren't really that extensive to cause problems with the l10n project. 

Back to the morebits changes:
- Use of string concatenation to make messages in many places, like the passing around of "change groups", creates a big challenge to i18n. These can't be translated. This means that there won't be a point in having functions like `fnProcessChecks` as every action needs a separate message. That means hell lot of messages and a heck lot of code for generating them - this can be mitigated by avoiding unnecessary error handling like the [isSysop check](https://github.com/Amorymeltzer/twinkle/blob/0a793d64e28bced6ba10c2c8c7c6a8ce0f091b8a/morebits.js#L5419) (API does it anyway with the localised messages); the [check for missing tokens](https://github.com/Amorymeltzer/twinkle/blob/0a793d64e28bced6ba10c2c8c7c6a8ce0f091b8a/morebits.js#L5431) - if they were requested with the API call they can't be missing in the response. Plus, even if token-less requests get sent, Mb.w.api would automatically retry.
- I'm not sure if using private context variables is a good idea when most of them are being exposed anyway via getters/setters. It would be simpler to just make the variables public (along with defining methods on the prototype). We can try avoiding being like Java (in which apparently everything needs to have setters for the spring framework to wrok.)
- get/setCallbackParameters are unlikely to be even necessary in properly written code and IMO ought to be deprecated even in mb.w.p. They stem from a time when accessing variables via closures were considered inefficient. For example, [twinkleprod.js](https://github.com/wikimedia-gadgets/twinkle/blob/master/modules/twinkleprod.js) doesn't use callbackParameters at all and if you ask me it's really good in terms of readability. Same case with [xfdCore.ts](https://github.com/wikimedia-gadgets/twinkle-core/blob/master/src/xfdCore.ts)/[xfd.ts](https://github.com/wikimedia-gadgets/twinkle-enwiki/blob/master/src/xfd.ts).  https://github.com/wikimedia-gadgets/twinkle/pull/1349/files#diff
- Things like https://github.com/wikimedia-gadgets/twinkle/pull/1349/files#diff-7fca53e39d0d4e891c6292a03ff9de5ef80eed3c0bf2dac6432ec0d9988bdf42R143-R144 may look ok, but note that these are really 5-6 different messages.
- This should probably be put in a separate file in the morebits gadget since at 7500 lines, the main file is too big for on-wiki syntax highlightening to work, navigating github diffs becomes tough, etc. (But moving older stuff from morebits to separate files would break non-gadget importScript loads – so may not want to rush into doing that.) Also separating unit-testable code from the rest would enable setting up code coverage tooling. (On that note I did write some E2E tests for morebits a while ago but didn't upload).

Amorymeltzer:
@siddharthvp

>How is it any different from your vitriolic comments at other places? I'd prefer to keep the discussion on facts rather than on civility lawyering.

Asking you not to treat me poorly is not vitriol.  Civility is a founding principle and one of the five pillars, not something to be dismissed.  If you want to discuss facts, let's.

>Do we write 2000 lines of code before discussing the approach in the first place? (All I know is that if someone does that at a company - they get fired.) Which is for good reason as it not easy to give constructive criticism when there are this many lines of code - because asking to rewrite the whole thing won't be constructive criticism.

This ain't a company, but I take your point.  Unfortunately, there isn't exactly a thriving team here, but I'm not wed to anything.

>Before that, one question: Do you expect me to port all of the changes to csd, xfd, warn and tag to twinkle-core/twinkle-enwiki? You never gave a hint that such a thing was in progress. If I had known that before, I would've prioritised other mdoules and/or structured my changes keeping it in mind. I'm going out of my way in localising twinkle while preserving every single twinkle feature that enwiki has, and I'm astonished to see that you've been making things more difficult starting from day 1.

You can see that the original work was all from last June, so well before you began that project.  Once you began, I shelved this in deference.  When you missed your initial deadline, I picked it back up, as I had some time at the beginning of this year here and there.  I've no earthly idea what is happening there, especially after the aforementioned incivility and your months-long disengagement and silence.

>Aping the structure of Morebits.wiki.page... The methods are defined in the constructor rather than on the prototype

I did it thinking it would make it easier for others to review, given past familiarity.  I agree the prototype would be better; I'd like to convert mb.w.page to it.

>Use promises, and only promises. #911 was complicated because it tried to support both promises and callbacks.

I actually did so at first, having seen the excellent work in #911.  However, as you [reminded us](https://github.com/wikimedia-gadgets/twinkle/pull/911#issuecomment-777516960) recently, that simply [doesn't work](https://github.com/wikimedia-gadgets/twinkle/pull/911#pullrequestreview-459305758).  I'd like to fix `actionCompleted`, as doing so is apparently a blocker.

>All of the oversighter-specific things and the function to change user rights appear out of scope. These are things that make sense in classes that extend Mb.w.u defined in admin-specific scripts. 95% of the users aren't even admins to use them. It is wasteful to load all this for everyone. Morebits isn’t even used much outside twinkle so I’m not sure who you’re expecting would use these features... Even the block/unblock things seem bit out of scope as they are used in just one module – no DRYness is being achieved here - that too an admin-only module again

For starters, the purpose of a library is to provide a resource.  I don't know what people would use if they could.  Most of those same arguments apply to `deletePage`, `protect`, `stabilize`, which I don't think you'd suggest removing.  Even the PageTriage stuff is fairly niche.  Blocking in particular is incredibly tricky to manage right, and an easy to use library would be helpful.

Secondly, the heart of this is really the user processing.  I don't think you're against any of that, just the structure of it?  Unblocking and changing user groups are incredibly trivial to implement.

Amorymeltzer:
>Use of string concatenation to make messages in many places, like the passing around of "change groups", creates a big challenge to i18n.

This isn't your main point, but "userrights" is like "stabilize" or "pagetriageaction" in that the MW action name is problematic when trying to convey something to end users (see also `deletePage` (and thus `undeletePage`) as opposed to `delete`).  I would think it unavoidable that there will have to be an internal and external moniker.

>These can't be translated. This means that there won't be a point in having functions like fnProcessChecks as every action needs a separate message. That means hell lot of messages and a heck lot of code for generating them - this can be mitigated by avoiding unnecessary error handling like the isSysop check (API does it anyway with the localised messages);

I don't know how you've been dealing with it, but it sounds like you're suggesting removing (have already removed?) pretty much all the error checking/handling across the board, is that right?

>the check for missing tokens - if they were requested with the API call they can't be missing in the response. Plus, even if token-less requests get sent, Mb.w.api would automatically retry.

Indeed!  One advantage here is that (tokenwise) we're guaranteed to have 'em, so that's very true.  Though I'll note that mb.w.api only grabs the csrf (although that's easy to change).

>I'm not sure if using private context variables is a good idea when most of them are being exposed anyway via getters/setters. It would be simpler to just make the variables public (along with defining methods on the prototype). We can try avoiding being like Java (in which apparently everything needs to have setters for the spring framework to wrok.)

I've long thought about exposing 'em for mb.w.p, especially re: prototypes above, although truth be told that's mostly because as someone who's debugging a lot I'm tired of applying my `this.getctx` stash.  It'd also make testing vastly more convenient.  Semantically, though, I think it's a mixed bag to do away with the functions.  The getters and setters can smooth some stuff over and ensure things get handled properly (watchlist is a good example), while still being functional for use.


>get/setCallbackParameters are unlikely to be even necessary in properly written code and IMO ought to be deprecated even in mb.w.p...

Currently unused, added for parity completion/parallelism.  No real comment on them in general except to say that it doesn't seem to be a huge payoff to remove 'em from mb.w.p.

>Things like #1349 (files) may look ok, but note that these are really 5-6 different messages.

I think you mean this in connection to the concatenating note above, right?  Or is there a separate issue?

>This should probably be put in a separate file in the morebits gadget since at 7500 lines, the main file is too big for on-wiki syntax highlightening to work, navigating github diffs becomes tough, etc.

I think this is agreeing about the importance of on-wiki readability?  I'm less concerned about the on-wiki editor since no morebits development is taking place there.  What are the issues you're having with diffs?

At any rate, I do think it'd be nice to separate the files, both for testing reasons and because things like the utility classes (wikitext, date, string, ip, etc.) are quite useful broadly, but lost in the weighty morass of quickForm, simpleWindow, etc.  But as you say, that's a separate conversation.

morebits.js
siddharthvp: There is no need to allow namespacerestrictions to be pipe-delimited string. Either a single number or array of numbers or a single namespace as string or array of namespace strings is enough flexibility.

morebits.js
siddharthvp: '.warn` would be better here, I think.

morebits.js
siddharthvp: mw.Uri is loaded by default on all pages, so there's no harm in using it. (Though I'm not sure why/how it's helpful here.)

morebits.js
siddharthvp: Is there a reason not to just expose the underlying Morebits.wiki.page object?

twinkle.js
```diff
@@ -494,6 +494,16 @@ Twinkle.changeTags = 'twinkle';
...
+// Function to generate an appropriate module-specific optout link i.e. notwinkle.test/?module=xfd
+// Missing protocol defaults to http
+Twinkle.makeOptoutLink = function makeOptoutLink(module) {
+	if (!module) {
+		return '';
+	}
+	return 'notwinkle.test/?module=' + module;
+};
+// Templates which opt a user talk page out of certain notifications
+Twinkle.optoutTemplates = ['Template:Retired', 'Template:Deceased Wikipedian'];
```
siddharthvp, line 506: This can also be handled by editing these two templates to include the {{no talkback}} hidden template - so that they will be skipped per the external link-based skip logic.

NovemLinguae:

PR #1352, opened 2021-03-09
block: Add option to default to blocking the /64, off by default

Amorymeltzer:
Originally suggested at https://github.com/wikimedia-gadgets/twinkle/pull/1266#issue-551310264

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -213,6 +213,14 @@ Twinkle.config.sections = [
 		module: 'block',
 		adminOnly: true,
 		preferences: [
+			// TwinkleConfig.defaultToBlock64 (boolean)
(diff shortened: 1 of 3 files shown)


modules/twinkleconfig.js
```diff
@@ -213,6 +213,14 @@ Twinkle.config.sections = [
 		module: 'block',
 		adminOnly: true,
 		preferences: [
+			// TwinkleConfig.defaultToBlock64 (boolean)
+			// Whether to default to just blocking the /64 on or off
+			{
+				name: 'defaultToBlock64',
+				label: 'For IPv6 addressed, select the option to Just Block the /64 by default',
```
siddharthvp, line 220:
```suggestion
				label: 'For IPv6 addresses, select the option to block the /64 range by default',
```

Wording

PR #1354, opened 2021-03-13
add i18n support for morebits

siddharthvp:
Adds `Morebits.i18n` which allows plugging in an i18n library through which localised messages can be provided.

The English fallbacks continue to be part of the source code so for using morebits in English no i18n library is required. That is, the changes here are backward-compatible.

Messages in code now become:
```js
msg('msg-name', ...messageParams, fallbackEnglishMessage)
```

where the `fallbackEnglishMessage` is a string that directly uses the message params. That is, it is
```js
msg('deleting-pages', numPages, 'Deleting ' + numPages + ' page' + (numPages > 1 ? 's ' : ' '))
```
instead of
```js
msg('deleting-pages', numPages, 'Deleting $1 {{plural:$1|page|pages}}'
```
The latter would have been nicer but for that I'd have to rewrite all the messages and also then an i18n library would be required to use morebits. Maybe something to think about for the future – but not required for the twinkle-core project.

In case of some messages where a MediaWiki message exists - such as the month names - our message names match the ones in mediawiki.

I've skipped the messages that only the programmers should see -- such as parameter validations for functions, isSysop check for admin-only operations, etc.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -5462,7 +5462,7 @@ Morebits.batchOperation = function(currentAction) {
 		},
 
 		// internal counters, etc.
-		statusElement: new Morebits.status(currentAction || 'Performing batch operation'),
+		statusElement: new Morebits.status(currentAction || msg('batch-starting', 'Performing batch operation')),
 		worker: null, // function that executes for each item in pageList
 		postFinish: null, // function that executes when the whole batch has been processed
 		countStarted: 0,
@@ -5529,7 +5529,7 @@ Morebits.batchOperation = function(currentAction) {
 
 		var total = ctx.pageList.length;
 		if (!total) {
-			ctx.statusElement.info('no pages specified');
+			ctx.statusElement.info(msg('batch-no-pages', 'no pages specified'));
 			ctx.running = false;
 			if (ctx.postFinish) {
 				ctx.postFinish();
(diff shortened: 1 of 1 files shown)


DannyS712:
Why not just use mw.msg since this is intended for use with MediaWiki?
Have messages in json objects, with keys as the message names and values as the message text, and mw.messages.set() and then mw.msg() or similar to fetch them

siddharthvp:
This implementation is agnostic to the choice of i18n library which may well be mw.msg:
```js
Morebits.i18n.setParser({ get: mw.msg });
```
Then load messages into it using mw.messages.set().

However, twinkle-core can't rely on mw.msg as that's intended for use with the user language rather than content language. It instead uses https://github.com/wikimedia-gadgets/orange-i18n (some cursory notes regarding this are at wikimedia-gadgets/twinkle-core#1).

> Have messages in json objects, with keys as the message names and values as the message text

I agree in general (and [twinkle-core does this](https://github.com/wikimedia-gadgets/twinkle-core/blob/master/src/messages.json)) but here it creates more work for this PR as all existing messages need to be reworked into the message string format. Also, for the message strings to be parsed it would become essential to use an i18n library for using morebits even in english. The default-loaded mw.msg doesn't support operations like plural – either mediawiki.jquerymsg or jquery.i18n RL dependency would have to be added to the morebits gadget (and clients directly importing morebits as script would break). So I can hardly can see this as a net positive for now.

Amorymeltzer:
I don't really understand the why behind this.  Your work in the other repo(s) involved a full rewrite/restructure of Morebits, so if that's to be used, why this?  If that is to be scrapped, this is a (somewhat?) opinionated implementation, no?

siddharthvp:
> Your work in the other repo(s) involved a full rewrite/restructure of Morebits

Nope. No such rewrite was done anywhere. I initially planned to, but it's beyond my bandwidth to keep up with all the changes being made to it. Twinkle-core uses [Api](https://github.com/wikimedia-gadgets/twinkle-core/blob/master/src/Api.ts), [Page](https://github.com/wikimedia-gadgets/twinkle-core/blob/master/src/Page.ts) and [User](https://github.com/wikimedia-gadgets/twinkle-core/blob/master/src/User.ts) classes which access Morebits.wiki.api/page/user under the hood. 

> this is a (somewhat?) opinionated implementation, no?

Huh? There's nothing opinionated here. Something needs to be done to get those messages to show in languages other than English. This is the least drastic way to get there - without any breaking changes (last para of https://github.com/wikimedia-gadgets/twinkle/pull/1354#issuecomment-799577111).

PR #1361, opened 2021-03-19
Revert "Fixes after renaming of {{Rfd}} and Module:RfD"

pppery:
The module rename was reverted due to causing issues with XfDcloser.

This reverts commit 61cace2eb53da954806f75161eccb53842c3a48c.

### DIFF
--- a/modules/twinkleprod.js
+++ b/modules/twinkleprod.js
@@ -257,7 +257,7 @@ Twinkle.prod.callbacks = {
 			var text = pageobj.getPageText();
 
 			// Check for already existing deletion tags
-			var tag_re = /{{(?:db-?|delete|article for deletion\/dated|AfDM|ffd\b)|#invoke:Redirect for discussion/i;
+			var tag_re = /{{(?:db-?|delete|article for deletion\/dated|AfDM|ffd\b)|#invoke:RfD/i;
 			if (tag_re.test(text)) {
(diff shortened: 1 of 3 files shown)


pppery:
Note that at the time I write this PR, things are in a weird half-done state in which the module code still has the new name, but all uses call a redirect from the old name. I've submitted a RMTR request to fix that and fully move the module back to the old name. 

In any case, what Twinkle cares about is what wikitext redirects use, and what HTML the template uses, and both have used the "RfD" name since February 28.

JJMC89:
> causing issues with XfDcloser

The correct solution to that is to update XfDcloser as was done for Twinkle.

pppery:
I agree that the correct eventual result is for the module to be renamed and all tools to be updated, and all I am trying to accomplish here is to get things out of the half-renamed state, which is clearly worse than either the module being fully renamed or not being renamed at all. I figured the best way to do that was to return to the status quo, and then do a more coordinated move in the future.

The module move also broke article alerts. The technical problems this caused were discusses at https://en.wikipedia.org/wiki/Template_talk:Redirect_for_discussion#Requested_move_4_February_2021 (after the formal requested move)

Amorymeltzer:
Christ that's annoying.  Entirely agree with both of you!

Amorymeltzer (approved):

PR #1370, opened 2021-03-29
morebits: render wikilink syntax and sanitize HTML in status and quickForm labels

siddharthvp:
Makes it much easier to include wikilinks in status messages and quickForm labels. Also de-duplicates some code in the two clases.

If literal `[[...]]` are to be included in status for whatever reason, that can still be done using the HTML entities `&#91;&#91;...&#93;&#93;`

$.parseHTML() removes <script> tags in the parsed output, so there should be no security issues.

A `Morebits.createHtml` function is added. This obsoletes `Morebits.status.prototype.codify` which is hence removed.

----

2nd commit: morebits: createHtml: sanitize html attributes 

Removes attributes that are used for executing scripts (onclick, onerror, etc). Also disallows anything that isn't a text node or Element node.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -131,7 +131,17 @@ Morebits.createHtml = function(input) {
 			fragment.appendChild(input[i]);
 		} else {
 			$.parseHTML(Morebits.createHtml.renderWikilinks(input[i])).forEach(function(node) {
-				fragment.appendChild(node);
+				if (node.nodeType === 3) { // text node
+					fragment.appendChild(node);
+				} else if (node.nodeType === 1) { // Element node, strip dangerous attributes
+					Array.prototype.slice.call(node.attributes).forEach(function(attr) {
+						// onclick, onerror, onload, etc
+						if (attr.name.indexOf('on') === 0) {
+							node.removeAttribute(attr.name);
+						}
+					});
+					fragment.appendChild(node);
+				} // any other node type is suspicious
 			});
 		}
 	}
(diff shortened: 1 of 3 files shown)


siddharthvp:
I guess it's still possible to cause XSS issues using things href="javascript:()" and the like. I'm not sure how translatewiki.net works, but I hope we can trust translators not to put evil stuff in messages, and if it happens there is someone who checks messages before the translatewiki bot pushes them into the repo?

MusikAnimal:
> I hope we can trust translators not to put evil stuff in messages, and if it happens there is someone who checks messages before the translatewiki bot pushes them into the repo?

As far as I know translatewiki doesn't do these checks. The volume of translations would make manual review infeasible. There was a major security incident involving malicious translations that made its way to WMF production a year or so ago, so while it is rare, these things do happen. I think they solved this issue by introducing a new validator job that runs in CI, but as far as I know that's specific to Gerrit/Jenkins and not external repos like this one.

I think we're probably safe, it was just something I wanted to make you aware of. You generally can decipher malicious translations pretty easily (look for extraneous markup, etc.). Just give it a quick look-see whenever you (or I guess @Amorymeltzer) do deploys, and you should be fine :)

siddharthvp:
>  I think they solved this issue by introducing a new validator job that runs in CI, but as far as I know that's specific to Gerrit/Jenkins and not external repos like this one.

We do have a Gerrit mirror now (https://gerrit.wikimedia.org/r/admin/repos/mediawiki%2Fgadgets%2FTwinkleCore) so maybe that can be looked into some time.

PR #1371, opened 2021-04-01
move some css to morebits.css, rather than embed in js

siddharthvp:
This applies to `<code>` elements within quickform divs, which should be shown in monospace font without italicisation.


siddharthvp:
As before the code text looks quite small (at least in Vector), but I suppose that's for another PR to fix.

PR #1373, opened 2021-04-03
morebits: i18n: add qqx support

siddharthvp:
Using uselang=qqx in the URL will make morebits show its message names, just like in MediaWiki.


siddharthvp:
Abandoned in favour of implementing it at twinkle-core side (https://github.com/wikimedia-gadgets/twinkle-core/commit/a7ee7462d87e97e64cfde881ea8c0cc7f61e13d1).

PR #1388, opened 2021-04-18
fluff: wrap rollback link separators in span tags

### LINKED ISSUE #1386: Make [vandalism] rollback links properly hideable via CSS
Instead of an external stylesheet, Twinkle uses inline CSS for the rollback links. This can be fine, but leads to issues when attempting to hide specific elements using CSS selectors in one's Special:MyPage/common.css.

A practical example that affects me: I currently (attempt to) use the following code to hide the red `[vandalism]` rollback link in favor of the MediaWiki core rollback link (which has a convenient JavaScript-based confirmation feature and is still faster at pushing the change to the database on large pages).

My custom CSS:

```
/* Hide Twinkle's "rollback" and "vandalism" links in favor of core rollback link in Special:Contributions */
.mw-special-Contributions .mw-contributions-current strong>a {
    display: none;
}

/* Hide Twinkle's (non-warning, instant-revert, default-summary) "vandalism" link in favor of the core rollback link above diffs */
.diff #tw-revert strong:nth-child(3) {
	display: none;
}
```

These CSS selectors are pretty far-fetched. MediaWiki is full of classes and IDs, Twinkle lacks them. However, I'd even use my code if it worked. It doesn't:

twinkle-selectors-issue-20210417

The very least I need is a workaround to remove the two pipes between the blue "rollback" and the red "vandalism" link above diffs.

ToBeFree:
fix #1386

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -123,6 +123,13 @@ Twinkle.fluff.linkBuilder = {
 
 		var separator = inline ? ' ' : ' || ';
 
+		var sepNode1 = document.createElement('span');
+		sepNode1.setAttribute('class', 'tw-rollback-link-separator');
+		var sepText = document.createTextNode(separator);
+		sepNode1.appendChild(sepText);
+
+		var sepNode2 = sepNode1.cloneNode(true);
+
 		if (!inline) {
 			var agfNode = document.createElement('strong');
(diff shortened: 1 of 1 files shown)


ToBeFree:
While I'm at it, this can be improved still

PR #1391, opened 2021-04-20
morebits: put 3px margin after input labels

siddharthvp:
For almost every input field label, such a margin was being artificially added till now by ending the label with a trailing space.

However, trailing spaces are not allowed on translatewiki.net (technical limitation) and so needs to be written as `&#32;` (https://phabricator.wikimedia.org/T278627#7019324) which unnecessarily obfuscates so many messages.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -47,7 +47,7 @@ Twinkle.tag.callback = function friendlytagCallback() {
 
 	form.append({
 		type: 'input',
-		label: 'Filter tag list: ',
+		label: 'Filter tag list:',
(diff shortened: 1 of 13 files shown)


siddharthvp:
need to check if a css adjustment is needed for select & dyninput labels as well. Don't think it's needed for textarea.

siddharthvp:
Added it for `select` and `_dyninput_element` labels as well.

siddharthvp:
Ideally this would have been better applied via CSS but there doesn't seem to be a way to select only these labels via CSS.

PR #1393, opened 2021-04-25
fluff: make rollback links and separators selectable via CSS

### LINKED ISSUE #1386: Make [vandalism] rollback links properly hideable via CSS

ToBeFree:
fix #1386

twinklefluff-20210425-span
twinklefluff-20210425-sep
twinklefluff-20210425-a

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -103,11 +103,22 @@ Twinkle.fluff.linkBuilder = {
 			revNode.setAttribute('id', 'tw-revert');
 		}
 
-		var normNode = document.createElement('strong');
-		var vandNode = document.createElement('strong');
+		var separator = inline ? ' ' : ' || ';
+		var sepNode1 = document.createElement('span');
+		var sepText = document.createTextNode(separator);
+		sepNode1.setAttribute('class', 'tw-rollback-link-separator');
+		sepNode1.appendChild(sepText);
+
+		var sepNode2 = sepNode1.cloneNode(true);
+
+		var normNode = document.createElement('span');
+		var vandNode = document.createElement('span');
 
 		var normLink = Twinkle.fluff.linkBuilder.buildLink('SteelBlue', 'rollback');
 		var vandLink = Twinkle.fluff.linkBuilder.buildLink('Red', 'vandalism');
+
+		normLink.style.fontWeight = 'bold';
+		vandLink.style.fontWeight = 'bold';
 
 		$(normLink).click(function() {
 			Twinkle.fluff.revert('norm', vandal, rev, page);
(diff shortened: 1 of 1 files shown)


modules/twinklefluff.js
```diff
@@ -118,10 +129,14 @@ Twinkle.fluff.linkBuilder = {
...
 
-		var separator = inline ? ' ' : ' || ';
+		normNode.appendChild(sepNode1);
+		vandNode.appendChild(sepNode2);
+
+		normNode.appendChild(normLink);
+		vandNode.appendChild(vandLink);
 
 		if (!inline) {
 			var agfNode = document.createElement('strong');
```
Xi-Plus, line 142: You have changed norm and vand buttons from `<strong>` to `<span>`, but you forget it for agf button.

ToBeFree: That's correct, sorry. Thanks for catching this.

ToBeFree: Fixed in 9992f5d1a10835b636a853bd5692b94f713a72f5:

twinklefluff-20210430-patchfix1

twinklefluff-20210430-patchfix2

twinklefluff-20210430-patchfix3

ToBeFree:
@Xi-Plus thanks for the review, I think I have now fixed the issue. The "request re-review" button doesn't seem to work for me :)

siddharthvp:
@ToBeFree Sorry for the long delay, and thanks for the patch. Is it intentional that `.tw-rollback-link-separator` is inside `.tw-rollback-link-normal` or `. tw-rollback-link-vandalism`. This makes it possible to hide the normal and vandalism links, but not the AGF link.

ToBeFree:
No worries, thank you very much for reviewing. :) That was in response to https://github.com/wikimedia-gadgets/twinkle/issues/1386#issuecomment-822183967 ; I'll have a look again but think it should be fine

ToBeFree:
I can't see a different way to implement this, and it would work for the use-case described in #1386 that led to the whole discussion.

ToBeFree:
@siddharthvp, hiding the AGF link can be done by targeting both the AGF link and the normal rollback's separator. Should be fine.

PR #1413, opened 2021-06-07
Problem with {{og}} and redirects

Daniuu-Wikipedia:
Fixing an issue when a block was used with the 'og'-template.
Adding support for talk pages that are redirects to a page corresponding to some larger range.
At various places, adding comments to explain the code more clearly (and why certain choices are made).


Daniuu-Wikipedia:
Apologies, this was intended for the Dutch Wikipedia only

PR #1430, opened 2021-07-16
protect: Implement #1429

### LINKED ISSUE #1429: twinkleprotect: Use /Increase and /Decrease instead of the main WP:RFPP page
per https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Requests_for_page_protection&oldid=1033930148#Technical_roadmap , this is now going live

https://github.com/wikimedia-gadgets/twinkle/blob/master/modules/twinkleprotect.js#L1374 needs an update to 'Wikipedia:Requests for page protection/Increase', but that will break the 'Decrease'/'Unprotect' functionality in https://github.com/wikimedia-gadgets/twinkle/blob/master/modules/twinkleprotect.js#L1564-L1569

TheTVExpert:
Closes #1429.

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -1601,13 +1601,13 @@ Twinkle.protect.callbacks = {
 					}
 				});
 			} else {
-				var originalTextLength = text2.length;
+				var originalTextLength2 = text2.length;
 				text2 += '\n' + newtag;
-				if (text2.length === originalTextLength) {
-					var linknode = document.createElement('a');
-					linknode.setAttribute('href', mw.util.getUrl('Wikipedia:Twinkle/Fixing RPP'));
-					linknode.appendChild(document.createTextNode('How to fix RPP'));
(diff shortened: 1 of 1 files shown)


ToBeFree:
Thank you very much! 😊

siddharthvp:
Thanks for the PR.

This appears to work and I was about to merge before I saw an uncaught TypeError when clicking the "Request page protection" radio on a page that has semi-protection. Doesn't affect functionality though.
```
Uncaught TypeError: Cannot read property '0' of undefined
    at Function.twinkleprotectCallbackShowLogAndCurrentProtectInfo [as showLogAndCurrentProtectInfo] ((index):18481)
    at HTMLInputElement.twinkleprotectCallbackChangeAction ((index):18832)
```
@TheTVExpert can you take a look?

TheTVExpert:
@siddharthvp I am not seeing this error when I test the code.

siddharthvp:
I see that the error no longer occurs. There's quite a bit of duplicated code here that definitely can be improved. But merging this for now as this has remained open for quite some time.

PR #1436, opened 2021-08-15
Changes to CONTRIBUTING.md about testing

siddharthvp:
CONTRIBUTING.md changes:

- mention the new dev-server.js script as the tool for testing. Remove mention of the patch-test.js script which is harder to use and more prone to issues.
- Suggest using testwiki more clearly.
- Omit "a good policy is that every commit should be capable of replacing the live on-wiki Gadget file for all users" -- while that likely is a good policy, that exceeds expectations and IMO unnecessary.

Cc: @TheTVExpert @ToBeFree @Trialpears let's discuss if there's anything else in the way of scripts/documentation that can make twinkle development easier.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,9 +28,13 @@ If you believe you have found a security issue, follow the guidelines in [SECURI
 
 If you'd like to help with Twinkle's development, wonderful!  Anyone can contribute, and it's easy to get set up to do so.
 
+- Download and install Node.js if you don't have it already (https://nodejs.org/en/download/).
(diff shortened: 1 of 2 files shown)


ToBeFree:
Thank you very much, @siddharthvp – I wasn't aware of the testwiki option at all, and the local testing method was indeed a bit tricky to use by itself. I'm happy to see this new alternative and am sure that the new process will encourage more contributions.

siddharthvp:
That reminds me – we should also make note of the various specialised sandboxes (Draft:Sandbox for testing tagging, 192.0.2.16 for testing IP blocks, etc)

PR #1450, opened 2021-10-08
morebits: load RL modules required for previews

siddharthvp:
Some additional ResourceLoader modules may be required for parsed wikitext to be rendered correctly. See matmarex's comment in https://phabricator.wikimedia.org/T292345#7399140.

This fixes the stray `[]` appearing at the end of comments (they now get hidden with CSS loaded from DiscussionTools). Reported at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Bug:_preview_of_CfD/S_has_trailing_brackets. 

Also added `preview` and `disableeditsection` flags to preview API.

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -4729,11 +4729,13 @@ Morebits.wiki.preview = function(previewbox) {
 
 		var query = {
 			action: 'parse',
-			prop: 'text',
-			pst: 'true',  // PST = pre-save transform; this makes substitution work properly
+			prop: ['text', 'modules'],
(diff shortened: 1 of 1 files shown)


siddharthvp:
The specific issue with DiscussionTools has since been fixed upstream.

nardog:
I was going to file a bug and pull request to eliminate the glaring subscribe button on RM preview by adding `preview: true` and/or `disableeditsection: true`, but this PR does just that. What's holding it up?

PR #1460, opened 2021-11-17
speedy: fix bugs involving watchlist preferences

### LINKED ISSUE #1427: All CSD tagged articles getting added to watchlist, despite selecting "don't add to watchlist"
Pretty much the title. Here's a screenshot of my settings.

image

NovemLinguae:
Fixes #1427 All CSD tagged articles getting added to watchlist, despite selecting "don't add to watchlist"

Also fixes a bug where all CSD watchlisting was permanent, instead of taking into account the duration set in the user's preferences.

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -2027,9 +2027,19 @@ Twinkle.speedy.callback.evaluateUser = function twinklespeedyCallbackEvaluateUse
 	});
 
 	// analyse each criterion to determine whether to watch the page/notify the creator
-	var watchPage = normalizeds.some(function(norm) {
-		return Twinkle.getPref('watchSpeedyPages').indexOf(norm) !== -1 && Twinkle.getPref('watchSpeedyExpiry');
+	// analyse each criterion to determine whether to watch the page/notify the creator
+	var watchPage = normalizeds.some(function(csdCriteria) {
+		let isWatchingThisCriteria = ( Twinkle.getPref('watchSpeedyPages').indexOf(csdCriteria) !== -1 );
+		let isWatchingAllCSDs = ( Twinkle.getPref('watchSpeedyExpiry') !== "no" );
+		let shouldWatchThis = ( isWatchingThisCriteria && isWatchingAllCSDs );
+		return shouldWatchThis;
 	});
+
+	// if we should watch the page, replace "true" with the duration we should watch for
+	if ( watchPage ) {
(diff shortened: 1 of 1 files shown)


siddharthvp (changes requested):
`let` is ES6, so can't be used here. Please use `var` instead and make sure that `npm run lint:fix` is passing.

modules/twinklespeedy.js
```diff
@@ -2027,9 +2027,19 @@ Twinkle.speedy.callback.evaluateUser = function twinklespeedyCallbackEvaluateUse
 	});
 
 	// analyse each criterion to determine whether to watch the page/notify the creator
-	var watchPage = normalizeds.some(function(norm) {
-		return Twinkle.getPref('watchSpeedyPages').indexOf(norm) !== -1 && Twinkle.getPref('watchSpeedyExpiry');
+	// analyse each criterion to determine whether to watch the page/notify the creator
```
siddharthvp, line 2030:
```suggestion
```
this comment has been duplicated

NovemLinguae:
This one is ready for re-review :)

modules/twinklespeedy.js
```diff
@@ -2027,9 +2027,18 @@ Twinkle.speedy.callback.evaluateUser = function twinklespeedyCallbackEvaluateUse
...
+		var isWatchingAllCSDs = Twinkle.getPref('watchSpeedyExpiry') !== 'no';
+		var shouldWatchThis = isWatchingThisCriteria && isWatchingAllCSDs;
+		return shouldWatchThis;
 	});
+
+	// if we should watch the page, replace "true" with the duration we should watch for
+	if (watchPage) {
+		watchPage = Twinkle.getPref('watchSpeedyExpiry');
+	}
+
```
siddharthvp, line 2041: I think this is equivalent to just 
```js
	var watchPage = normalizeds.some(function(csdCriteria) {
		return Twinkle.getPref('watchSpeedyPages').indexOf(csdCriteria) !== -1;
	}) && Twinkle.getPref('watchSpeedyExpiry');
```
?
Twinkle.getPref('watchSpeedyExpiry') doesn't need to be checked every time in the loop, and if normalizeds.some returns true, watchSpeedyExpiry being `'no'` gives the desired effect

siddharthvp:
From the issue, I can't reproduce the "unchecking all the boxes still [adds pages to watchlist]" bug. Also don't how this patch would fix that issue assuming it's there.

PR #1466, opened 2021-11-18
tag: allow wrapping existing tags in {{multiple issues}}

### LINKED ISSUE #1334: Article tagging - when tags already present and ungrouped, not able to group them
When tagging articles with maintenance tags, if tags already exist and they are not grouped, and the user has the "group inside" checkbox checked, consider not throwing an error message. This would allow the user to quickly group them.

To reproduce:

- Go to https://en.wikipedia.org/w/index.php?title=Dennis_Inyang&oldid=1008637658
- TW -> Tag
- (there's already 3 tags, but they are not grouped)
- make sure "Group inside" is checked
- hit "Submit Query"
- You get a JS browser alert saying "You must select at least one tag!"

NovemLinguae:
Fixes #1334 Article tagging - when tags already present and ungrouped, not able to group them

~~edit: Converted to draft. I found a bug involving Twinkle hanging when an article has no tags. I will debug and then resubmit.~~

edit2: Bug fixed. Resubmitting.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -2012,14 +2012,6 @@ Twinkle.tag.callback.evaluate = function friendlytagCallbackEvaluate(e) {
 			break;
 	}
 
-	/*
-	// return if no tags selected
-	if (params.tags.length === 0 && params.existingTags.length === 0) {
-		alert('You must select at least one tag!');
-		return;
-	}
-	*/
-
 	Morebits.simpleWindow.setButtonsEnabled(false);
 	Morebits.status.init(form);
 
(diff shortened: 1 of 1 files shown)


siddharthvp:
This seems to give edit summary as ["Removed tag"](https://test.wikipedia.org/w/index.php?title=SD0001test&diff=prev&oldid=494696).

NovemLinguae:
This code path is complicated. Would require a lot of testing to ensure it's bug free. Article with multiple issues ticked, article without multiple issues ticked, redirect. Add tag, no change, remove tag. etc.

I think I'd rather make a ticket to remove the "Group inside {{multiple issues}} if possible" check box, and just do that always by default. That would simplify the code instead of making it more complex.

PR #1467, opened 2021-11-20
tag: add namespace rules for redirect tags

### LINKED ISSUE #1459: Fix redirect category namespace detection
Twinkle recently let me categorize a template redirect page with {{t|R from other capitalisation}}. It was only after saving that I noticed the warning that that template shouldn't be used outside of mainspace, and the documentation states that {{t|R from modification}} should be used in those circumstances instead. Could the Twinkle menu options please be adjusted so that only tags that belong on a page in a given namespace are offered?

NovemLinguae:
Fixes #1459 Fix redirect category namespace detection

I found 184 {{R to/from X}} style redirect tags that are namespace-specific. This PR codes Twinkle to not display these unless it is the correct namespace.

I copy/pasted a utility function from Stack Overflow, and used a comment with a URL as attribution.

Here's a couple test cases to check:

````
			var insideMainspaceOnly = [
				'R from acronym',
			];
			var insideTalkNamespaceOnly = [
				'R from remote talk page'
			];
			var disambiguationPagesOnly = [
				'R to disambiguation page'
			];
			var outsideUserspaceOnly = [
				'R to user namespace'
			];
````

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -219,12 +219,6 @@ Twinkle.tag.callback = function friendlytagCallback() {
 		case 'redirect':
 			Window.setTitle('Redirect tagging');
 
-			// https://stackoverflow.com/questions/280634/endswith-in-javascript/2548133#2548133
-			// Author chakrit, CC BY-SA 4.0
-			var endsWith = function(str, suffix) {
-				return str.indexOf(suffix, str.length - suffix.length) !== -1;
-			};
-
 			// If a tag has a restriction for this namespace or title, return true, so that we know not to display it in the list of check boxes.
 			var isRestricted = function(item) {
 				if (typeof item.restriction === 'undefined') {
@@ -248,7 +242,7 @@ Twinkle.tag.callback = function friendlytagCallback() {
 						}
 						break;
 					case 'disambiguationPagesOnly':
-						if (!endsWith(mw.config.get('wgPageName'), '_(disambiguation)')) {
+						if (!mw.config.get('wgPageName').endsWith('_(disambiguation)')) {
 							return true;
 						}
 						break;
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -219,6 +219,247 @@ Twinkle.tag.callback = function friendlytagCallback() {
 		case 'redirect':
 			Window.setTitle('Redirect tagging');
 
+			// Certain redirect tags shouldn't be used in certain namespaces
+			// I generated the below lists using this search, and also added some manually.
+			// https://en.wikipedia.org/w/index.php?title=Special:Search&limit=500&offset=0&ns0=1&ns1=1&ns3=1&ns4=1&ns5=1&ns6=1&ns7=1&ns8=1&ns9=1&ns10=1&ns11=1&ns12=1&ns13=1&ns14=1&ns15=1&ns100=1&ns101=1&ns118=1&ns119=1&ns710=1&ns711=1&ns828=1&ns829=1&ns2300=1&ns2301=1&ns2302=1&ns2303=1&search=insource%3A%22Use+this+rcat+template+on%22
+			var mainspaceOnly = [
+				'R animal with possibilities',
+				'R comics from alternative name',
+				'R comics from merge',
+				'R comics from related word',
+				'R comics naming convention',
+				'R comics to list entry',
+				'R comics to section',
+				'R comics with possibilities',
+				'R fictional work',
+				'R from abbreviation',
+				'R from acronym',
+				'R from adjective',
+				'R from adverb',
+				'R from airline code',
+				'R from airport code',
+				'R from album',
+				'R from alternative military designation',
+				'R from alternative scientific name',
+				'R from alternative translation',
+				'R from alternative transliteration',
+				'R from animal',
+				'R from ATC code',
+				'R from ATCvet code',
+				'R from birth name',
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/friendlytag.js [hunk 1]
```diff
@@ -183,6 +183,248 @@ Twinkle.tag.callback = function friendlytagCallback() {
 		case 'redirect':
 			Window.setTitle('Redirect tagging');
 
+			// Certain redirect tags shouldn't be used in certain namespaces
+			// I generated the below lists using this search, and also added some manually.
+			// https://en.wikipedia.org/w/index.php?title=Special:Search&limit=500&offset=0&ns0=1&ns1=1&ns3=1&ns4=1&ns5=1&ns6=1&ns7=1&ns8=1&ns9=1&ns10=1&ns11=1&ns12=1&ns13=1&ns14=1&ns15=1&ns100=1&ns101=1&ns118=1&ns119=1&ns710=1&ns711=1&ns828=1&ns829=1&ns2300=1&ns2301=1&ns2302=1&ns2303=1&search=insource%3A%22Use+this+rcat+template+on%22
+			var mainspaceOnly = [
```
siddharthvp, line 189: All configurational data should be in`Twinkle.tag.redirectList`. You can add namespace filtering there, such as with `namespaces: [0, 1]` within the objects (with unspecified indicating all namespaces).

This avoids having to change things in multiple places while adding/removing tags to twinkle.

modules/friendlytag.js
```diff
@@ -183,6 +183,248 @@ Twinkle.tag.callback = function friendlytagCallback() {
...
+					deleteTag(tag, redirectList);
+				});
+			};
+			// https://stackoverflow.com/a/2548133/3480193
+			var endsWith = function(str, suffix) {
+				return str.indexOf(suffix, str.length - suffix.length) !== -1;
+			};
+
+			var namespace = mw.config.get('wgNamespaceNumber');
+			if (namespace !== 0 && namespace !== 1) { // not mainspace
```
siddharthvp, line 412: These templates are applicable for talk namepace as well? Doesn't look like it: [R from pseudonym](https://en.wikipedia.org/w/index.php?title=Special%3AWhatLinksHere&hidelinks=1&target=Template%3A+R+from+pseudonym&namespace=1), [R from gene symbol](https://en.wikipedia.org/w/index.php?title=Special%3AWhatLinksHere&hidelinks=1&target=Template%3AR+from+gene+symbol&namespace=1)

siddharthvp (changes requested):

modules/friendlytag.js
NovemLinguae: Instead of doing `disallowedNamespaces: [1,2]` or something similar, I decided to do human readable statements, since one of the restrictions is not namespace related, and so that we wouldn't have to list out a bunch of numbers. The human readable restrictions are `restriction: insideMainspaceOnly, insideTalkNamespaceOnly, disambiguationPagesOnly, outsideUserspaceOnly`.

modules/friendlytag.js
```diff
@@ -183,6 +183,47 @@ Twinkle.tag.callback = function friendlytagCallback() {
...
+							return true;
+						}
+						break;
+					case 'insideTalkNamespaceOnly':
+						if (namespace % 2 !== 1 || namespace < 0) {
+							return true;
+						}
+						break;
+					case 'disambiguationPagesOnly':
+						if (!endsWith(mw.config.get('wgPageName'), '_(disambiguation)')) {
```
siddharthvp, line 215: This wouldn't cover all disambig pages (eg. [Android](https://en.wikipedia.org/wiki/Android)). Checking wgCategories for "All disambiguation pages" should be exhaustive, I think. Another option would have been checking for `$('.dmbox').length` – but that will only work while viewing a page (not in edit/history mode).

NovemLinguae: I think it'd be OK to leave as-is. It's counter intuitive, but the [documentation for {{R to disambiguation page}}](https://en.wikipedia.org/wiki/Template:R_to_disambiguation_page) says "this template should only appear on a redirect page that has "(disambiguation)" at the end of its title".

NovemLinguae:
@siddharthvp Is this patch OK to merge despite not agreeing with the 2 code review comments? Are your concerns sufficiently addressed?

NovemLinguae:
@siddharthvp After I fix the merge conflict, is this patch OK to merge despite not agreeing with the 2 code review comments? Are your concerns sufficiently addressed?

siddharthvp:
@NovemLinguae Yeah go ahead. Sorry about the delayed response. Didn't realise this has been sitting for more than a year. How time flies.

modules/friendlytag.js
```diff
@@ -183,6 +183,47 @@ Twinkle.tag.callback = function friendlytagCallback() {
 		case 'redirect':
 			Window.setTitle('Redirect tagging');
 
+			// https://stackoverflow.com/questions/280634/endswith-in-javascript/2548133#2548133
+			// Author chakrit, CC BY-SA 4.0
+			var endsWith = function(str, suffix) {
+				return str.indexOf(suffix, str.length - suffix.length) !== -1;
```
siddharthvp, line 189: Not necessary to implement this anymore - we can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith which is possible now as IE 11 is no longer supported.

PR #1468, opened 2021-11-22
xfd: add the option of discuss=no to Template:RMassist

### LINKED ISSUE #1458: Add the option of discuss=no to Template:RMassist
Add the option of `discuss=no` to [Template:RMassist](https://en.wikipedia.org/wiki/Template:RMassist) which is supported by the template.

NovemLinguae:
Fixes #1458 Add the option of discuss=no to Template:RMassist

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -715,6 +715,21 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 				]
 			});
 			work_area.append({
+				type: 'checkbox',
+				list: [
+					{
+						label: 'Opt out of discussion if the request is contested',
+						value: 'rmtr-discuss',
+						name: 'rmtr-discuss',
+						tooltip: 'Use this option when when you are filing an uncontroversial technical request AND you prefer to withdraw the request if it is contested, rather than discuss it. This will suppress the "discuss" link which may be used to convert your request to a discussion on the talk page of the page which is requested to be moved.',
+						checked: false,
+						event: function() {
+							form.newname.required = this.checked;
+						}
+					}
+				]
+			});
+			work_area.append({
 				type: 'input',
 				name: 'newname',
 				label: 'New title:',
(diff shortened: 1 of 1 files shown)


siddharthvp:
Since this is only applicable for RMTR, the new input field would be better as a `subgroup` of the rmtr checkbox, so that it's only visible if rmtr is checked.

NovemLinguae:
> Since this is only applicable for RMTR, the new input field would be better as a `subgroup` of the rmtr checkbox, so that it's only visible if rmtr is checked.

Requested change complete.

modules/twinklexfd.js
```diff
@@ -710,6 +710,21 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
 							form.newname.required = this.checked;
+						},
+						subgroup: {
+							type: 'checkbox',
+							list: [
+								{
+									label: 'Opt out of discussion if the request is contested',
+									value: 'rmtr-discuss',
+									name: 'rmtr-discuss',
+									tooltip: 'Use this option when when you are filing an uncontroversial technical request AND you prefer to withdraw the request if it is contested, rather than discuss it. This will suppress the "discuss" link which may be used to convert your request to a discussion on the talk page of the page which is requested to be moved.',
```
siddharthvp, line 721: Can you update the tooltip as well? It no longer needs to say " when you are filing an uncontroversial technical request AND"

modules/twinklexfd.js
```diff
@@ -710,6 +710,21 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
+									checked: false,
+									event: function() {
+										form.newname.required = this.checked;
```
siddharthvp, line 724: This looks like a remnant from copy-pasting? The `newname` is required if rmtr is checked, it doesn't depend on rmtr-discuss param, so the event should be removed.

modules/twinklexfd.js
```diff
@@ -718,11 +718,8 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 									label: 'Opt out of discussion if the request is contested',
 									value: 'rmtr-discuss',
 									name: 'rmtr-discuss',
-									tooltip: 'Use this option when when you are filing an uncontroversial technical request AND you prefer to withdraw the request if it is contested, rather than discuss it. This will suppress the "discuss" link which may be used to convert your request to a discussion on the talk page of the page which is requested to be moved.',
-									checked: false,
-									event: function() {
-										form.newname.required = this.checked;
-									}
+									tooltip: 'Use this option when you prefer to withdraw the request if it is contested, rather than discuss it. This will suppress the "discuss" link, which may be used to convert your request to a discussion on the talk page of the page which is requested to be moved.',
```
siddharthvp, line 721:
```suggestion
									tooltip: 'Use this option if you prefer to withdraw the request if contested, rather than discuss it. This suppresses the "discuss" link, which may be used to convert your request to a discussion on the talk page.',
```

One last suggested tweak to reduce the verbosity here, after which this should be ready to merge.

siddharthvp:
Looks good. Thanks.

PR #1471, opened 2021-11-22
xfd: read delsort topics from wiki page

### LINKED ISSUE #1428: Read delsort topics from wiki page
[Wikipedia:WikiProject Deletion sorting/Computer-readable.json](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Deletion_sorting/Computer-readable.json) was recently created and is being used by User:Enterprisey/delsort.js. Twinkle also should read delsort topics from there to avoid having to make source code changes whenever delsort pages have to be added/removed. 

The JSON page should be read via the API only when needed (i.e, when user opens the XFD module) and cached for a suitable period (say 1 day), using the API's maxage/smaxage parameters.

NovemLinguae:
Fixes #1428 Read delsort topics from wiki page

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -364,21 +364,17 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 				maxage: '86400' // cache for 1 day
 			};
 			new Morebits.wiki.api('Get JSON list of deletion sorting categories', query, function(apiobj) {
-				try {
-					var wikitext = apiobj.getResponse().parse.wikitext;
-					var delsortCategories = JSON.parse(wikitext);
-					var select = $('[name="delsortCats"]');
-					$.each(delsortCategories, function(groupname, list) {
-						var optgroup = $('<optgroup>').attr('label', groupname);
-						var delsortCat = $(select).append(optgroup);
-						list.forEach(function(item) {
-							var option = $('<option>').val(item).text(item);
-							$(delsortCat).append(option);
-						});
+				var wikitext = apiobj.getResponse().parse.wikitext;
+				var delsortCategories = JSON.parse(wikitext);
+				var select = $('[name="delsortCats"]');
+				$.each(delsortCategories, function(groupname, list) {
+					var optgroup = $('<optgroup>').attr('label', groupname);
+					var delsortCat = $(select).append(optgroup);
+					list.forEach(function(item) {
+						var option = $('<option>').val(item).text(item);
+						$(delsortCat).append(option);
 					});
-				} catch (err) {
-					// If unable to pull usable Deletion Sort categories, just give up and exit this function. The user won't be able to pick any categories, but that won't stop the AFD nomination.
-				}
+				});
 			}).post();
 
 			appendReasonBox();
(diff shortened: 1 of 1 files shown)


modules/twinklexfd.js
```diff
@@ -346,53 +346,43 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
+				page: 'Wikipedia:WikiProject_Deletion_sorting/Computer-readable.json',
+				prop: 'wikitext',
+				smaxage: '86400', // cache for 1 day
+				maxage: '86400' // cache for 1 day
+			};
+			new Morebits.wiki.api('Get JSON list of deletion sorting categories', query, function(apiobj) {
+				try {
+					var wikitext = apiobj.getResponse().parse.wikitext;
+					var delsortCategories = JSON.parse(wikitext);
+					var select = $('[name="delsortCats"]');
```
siddharthvp, line 370:
```suggestion
					var $select = $('[name="delsortCats"]');
```
Per convention, jQuery vars have a $ prefix.

modules/twinklexfd.js
```diff
@@ -346,53 +346,43 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
+					$.each(delsortCategories, function(groupname, list) {
+						groupname = escapeDoubleQuotes(groupname);
+						var delsortCat = $(select).append('<optgroup label="' + groupname + '">');
+						list.forEach(function(item) {
+							item = escapeDoubleQuotes(item);
+							$(delsortCat).append('<option value="' + item + '" label="' + item + '">' + item + '</option>');
+						});
+					});
+				} catch (err) {
+					// If unable to pull usable Deletion Sort categories, just give up and exit this function. The user won't be able to pick any categories, but that won't stop the AFD nomination.
```
siddharthvp, line 383: The behaviour you describe in the comment is ideal, but it doesn't actually require the try–catch to be present. 

It can be removed since it only prevents the warning with the error trace from being logged to the console, which would have been be helpful for debugging.

modules/twinklexfd.js
```diff
@@ -346,53 +346,36 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
-			$.each(delsortCategories, function(groupname, list) {
-				var group = delsort.append({ type: 'optgroup', label: groupname });
-				list.forEach(function(item) {
-					group.append({ type: 'option', label: item, value: item });
+			// delsort categories taken from [[Wikipedia:WikiProject Deletion sorting/Computer-readable.json]]
+			var query = {
+				action: 'parse',
+				format: 'json',
+				page: 'Wikipedia:WikiProject_Deletion_sorting/Computer-readable.json',
+				prop: 'wikitext',
```
siddharthvp, line 362: The caching isn't working with action=parse. From the network tab of devtools, I am seeing the page being being fetched every time. Use action=query&prop=revisions instead.

siddharthvp:
Thanks for this, this is a good step forward in improving maintainability of the gadget. Something like this could be done in other places like #1477.

PR #1472, opened 2021-11-23
tag: add validations to redirect tags

### LINKED ISSUE #1037: tag: Add validations to redirect tags
From https://github.com/azatoth/twinkle/pull/1002#issuecomment-651451863:

>One thing to consider is whether there is/are any validations that should be done.

>The short answer is yes: there are plenty that conflict (e.g. `{{R printworthy}}` and `{{R unprintworthy}}`) but the long answer is it's complex and we should only try to handle the obvious cases.

NovemLinguae:
Fixes #1037 tag: Add validations to redirect tags

image

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -2007,6 +2007,22 @@ Twinkle.tag.callback.evaluate = function friendlytagCallbackEvaluate(e) {
 			break;
 
 		case 'redirect':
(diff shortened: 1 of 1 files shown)


siddharthvp:
I don't see much benefit as it seems unlikely someone would make the mistake of selecting such contradictory tags together, but at the same there seems to be no downside.

PR #1478, opened 2021-11-28
dev-server: modify wgUserGroups to add sysop

siddharthvp:
As many features are not applicable to non-sysops, this aids testing of those features by non-sysops. This is done by default but can be turned off with `--no-sysop` flag. (Use as `node server.js --no-sysop` or `npm start -- --no-sysop`)


NovemLinguae:
As someone who just typed `Twinkle.block.callback()` 10 times or so on test wiki, I like this idea :)

PR #1479, opened 2021-11-28
block: read discretionary sanctions information from a wiki page

### LINKED ISSUE #1477: Read discretionary sanctions information from a wiki page
`Twinkle.block.dsinfo` added in #1039 is currently hard-coded which necessitates Twinkle maintainers to update add/remove DS sanctions as and when updated by ArbCom. I wonder if it isn't already outdated. This could instead be read from a JSON page on-wiki so that Twinkle maintainers don't have to follow Arbcom proceedings. Then it could even be updated by ArbCom clerks. 

As for implementation, the JSON fetch logic used for delsort could be abstracted out into morebits.js or twinkle.js files, and reused here.

NovemLinguae:
- Fixes #1477 Read discretionary sanctions information from a wiki page
- I created the function `Morebits.wiki.getCachedWikitext(title, description, parent, callback)`, and I refactored the code to use the function in 3 places.
- I created https://en.wikipedia.org/wiki/Template:Ds/topics.json
- I added COVID-19 and Longevity to the list of discretionary sanctions

image

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -517,8 +517,7 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 	}
 
 	// grab discretionary sanctions list from en-wiki
-	var parent = this;
-	new Morebits.wiki.getCachedJson('Template:Ds/topics.json', parent).then(function(dsinfo) {
+	new Morebits.wiki.getCachedJson('Template:Ds/topics.json').then(function(dsinfo) {
 		var $select = $('[name="dstopic"]');
 		var $options = $.map(dsinfo, function (value, key) {
 			return $('<option>').val(value.code).text(key).prop('label', key);
@@ -1560,17 +1559,16 @@ Twinkle.block.callback.toggle_ds_reason = function twinkleblockCallbackToggleDSR
 	);
 
 	// grab discretionary sanctions list from en-wiki
-	var parent = this;
-	new Morebits.wiki.getCachedJson('Template:Ds/topics.json', parent).then(function(dsinfo) {
-		var sanctionCode = parent.selectedIndex;
-		var sanctionName = parent.options[sanctionCode].label;
+	new Morebits.wiki.getCachedJson('Template:Ds/topics.json').then(function(dsinfo) {
+		var sanctionCode = this.selectedIndex;
+		var sanctionName = this.options[sanctionCode].label;
 		Twinkle.block.dsReason = dsinfo[sanctionName].page;
-		if (!parent.value) {
-			parent.form.reason.value = reason;
+		if (!this.value) {
+			this.form.reason.value = reason;
 		} else {
-			parent.form.reason.value = reason + ' ([[' + Twinkle.block.dsReason + ']])';
+			this.form.reason.value = reason + ' ([[' + Twinkle.block.dsReason + ']])';
 		}
-	});
+	}.bind(this));
 };
 
 Twinkle.block.callback.update_form = function twinkleblockCallbackUpdateForm(e, data) {
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -521,10 +521,10 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 	new Morebits.wiki.getCachedWikitext('Template:Ds/topics.json', 'Get JSON list of discretionary sanctions', parent, function(wikitext) {
 		var dsinfo = JSON.parse(wikitext);
 		var $select = $('[name="dstopic"]');
-		$.each(dsinfo, function(key, value) {
-			var $option = $('<option>').val(value.code).text(key).prop('label', key);
-			$select.append($option);
+		var $options = $.map(dsinfo, function (value, key) {
+			return $('<option>').val(value.code).text(key).prop('label', key);
 		});
+		$select.append($options);
 	});
 
 	// DS selection visible in either the template field set or preset,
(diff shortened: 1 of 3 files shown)


morebits.js
```diff
@@ -2471,6 +2471,25 @@ Morebits.wiki.api.prototype = {
 
 };
 
+/** Retrieves wikitext from a page. Caching enabled, duration 1 day. */
+Morebits.wiki.getCachedWikitext = function(title, description, parent, callback) {
```
siddharthvp, line 2475: Can return a promise here instead of using callbacks (Mb.w.api#post returns a promise). No need to follow the design of Morebits.wiki.* much of which is really old-fashioned. 

Also, do we need the `description`? I could be mistaken, but I think Morebits.status.init() isn't done until the form is submitted so the `description` is never actually displayed on the UI.

modules/twinkleblock.js
```diff
@@ -516,6 +516,17 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
...
 
+	// grab discretionary sanctions list from en-wiki
+	var parent = this;
+	new Morebits.wiki.getCachedWikitext('Template:Ds/topics.json', 'Get JSON list of discretionary sanctions', parent, function(wikitext) {
+		var dsinfo = JSON.parse(wikitext);
+		var $select = $('[name="dstopic"]');
+		$.each(dsinfo, function(key, value) {
+			var $option = $('<option>').val(value.code).text(key).prop('label', key);
+			$select.append($option);
+		});
```
siddharthvp, line 527: A more efficient way to update the DOM is `$options = $.map(dsinfo, (k, v) => $('<option>').val(...)...)` then `$select.append($options)` (so that DOM is manipulated just once instead of `$options.length` number of times)

modules/twinkleblock.js
```diff
@@ -525,9 +536,7 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 		value: '',
 		tooltip: 'If selected, it will inform the template and may be added to the blocking message',
 		event: Twinkle.block.callback.toggle_ds_reason,
-		list: $.map(Twinkle.block.dsinfo, function(info, label) {
-			return {label: label, value: info.code};
-		})
+		list: ''
```
siddharthvp, line 539:
```suggestion
```

morebits.js
siddharthvp: Also, maybe it's better to call this getCachedJson and do the JSON.parse as well in this.

NovemLinguae: As requested, I eliminated the description, and I renamed the function to `getCachedJson()`.

Regarding your feedback to use promises, are promises ES5? I thought they were ES6.

I began refactoring to use promises but I am having trouble passing the JSON variable from the `getCachedJson()` function to the `then()` function. I submitted a commit below with my attempt so far. Any tips?

siddharthvp: The native `Promise` is ES6, but jQuery has its own promises (which existed before promises were standardized as part of ECMAScript). Most MediaWiki code use jQuery promises (for instance, all the mw.Api methods return jQuery promises).

As to why it's not working, see the [jsdoc for Morebits.wiki.api](https://github.com/wikimedia-gadgets/twinkle/blob/277d5b6b52c/morebits.js#L2265). It'll work on changing 
```js
return new Morebits.wiki.api(description, query, function() {...}).post();
```
to
```js
return new Morebits.wiki.api(description, query).post().then(function() {...})

File: modules/twinkleblock.js [hunk 4]
```diff
@@ -516,6 +516,16 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 		}
 	}
 
+	// grab discretionary sanctions list from en-wiki
+	var parent = this;
+	new Morebits.wiki.getCachedJson('Template:Ds/topics.json', parent).then(function(dsinfo) {
```
siddharthvp, line 521: `new` is not needed as getCachedJson is a simple function, not a class. The `parent` parameter can also be removed as it's not used anywhere. 

Sidenote: If we wanted to access `this`, it could be accessed anyway using `function(dsinfo) { ... }.bind(this)` so that `this` inside the function refers to the same thing as `this` outside the function.

modules/twinkleblock.js
```diff
@@ -1692,12 +1558,17 @@ Twinkle.block.callback.toggle_ds_reason = function twinkleblockCallbackToggleDSR
...
 	);
 
-	Twinkle.block.dsReason = Twinkle.block.dsinfo[this.options[this.selectedIndex].label].page;
-	if (!this.value) {
-		this.form.reason.value = reason;
-	} else {
-		this.form.reason.value = reason + ' ([[' + Twinkle.block.dsReason + ']])';
-	}
+	// grab discretionary sanctions list from en-wiki
+	Morebits.wiki.getCachedJson('Template:Ds/topics.json').then(function(dsinfo) {
```
siddharthvp, line 1562: This is probably a nitpick as it's anyway cached, ... but still fetching it a second time seems bad. 

I'd suggest using this approach instead:

Twinkle.block.dsinfo = Morebits.wiki.getCachedJson('Template:Ds/topics.json'); // this is a promise

// 1st use 
Twinkle.block.dsinfo.then(...)

// 2nd use 
Twinkle.block.dsinfo.then(...)

siddharthvp:
In block, there was a stray status element ": done" showing up under block options. So I was wrong about https://github.com/wikimedia-gadgets/twinkle/pull/1479#discussion_r759486911 -- the description _is_ displayed in case of block module. But we can just hide as fetching of the json page is not something the user would be interested in being notified about. I've fixed this in the new commit.

Issue #1382, opened 2021-04-17
speedy/batchdelete: delete redirect talk pages as well


siddharthvp:
Both speedy and batchdelete modules allow deletion of talk pages and redirects, but not talk pages _of_ redirects. I can't think of a reason for this not to be supported.

NovemLinguae:
Also requested by Liz on 12/12/21 at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Twinkle's_delete_function

siddharthvp:
Also originally requested by Liz at https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=1021836648#Twinkle_glitch :)

NovemLinguae:
Requested by Hey man im josh on Discord, Jan 2024

NovemLinguae:
"this is pretty easy to do now as the API natively supports a deletetalk option" -SD0001

PR #1488, opened 2021-12-13
batchdelete: also delete redirect talk pages

NovemLinguae:
Related #1382 speedy/batchdelete: delete redirect talk pages as well

This PR provides code for the "batchdelete" part of the above ticket. I will code the speedy part in a future PR.

I chose to code this so that when someone selects the "delete redirects" check box, it will automatically delete the redirect talk pages.

The CSD log message for both redirects and redirect talk pages is now: "G8: Redirect or redirect talk page to deleted page "NovemTest1""

### DIFF
--- a/modules/twinklebatchdelete.js
+++ b/modules/twinklebatchdelete.js
@@ -55,7 +55,7 @@ Twinkle.batchdelete.callback = function twinklebatchdeleteCallback() {
 							checked: true
 						},
 						{
-							label: 'Delete redirects to deleted pages',
+							label: 'Delete redirects to deleted pages (and delete the redirect talk pages)',
 							name: 'delete_redirects',
 							value: 'delete_redirects',
 							checked: true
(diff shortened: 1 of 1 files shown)


siddharthvp:
There's actually a more efficient way to use the API here. We can replace the current query for listing redirects with generator=redirects&prop=info&inprop=talkid – which returns essentially the same info but with `talkid` for the pages that have talk pages – hence eliminating the separate query to check if the talk pages exist.

Also, let's use mw.Title to get the talk page name (`= new mw.Title(pageName).getTalkPage().toText()`) instead of regex replaces.

NovemLinguae:
Refactor complete. Ready for review.

modules/twinklebatchdelete.js
```diff
@@ -55,7 +55,7 @@ Twinkle.batchdelete.callback = function twinklebatchdeleteCallback() {
 							checked: true
 						},
 						{
-							label: 'Delete redirects to deleted pages',
+							label: 'Delete redirects to deleted pages (and delete the redirect talk pages)',
```
siddharthvp, line 58: While I agree a separate checkbox for redirect talk pages is unnecessary, maybe it would be better to delete them only if delete_talk above is checked?

modules/twinklebatchdelete.js
```diff
@@ -571,20 +573,34 @@ Twinkle.batchdelete.callbacks = {
...
-		var redirectDeleter = new Morebits.batchOperation('Deleting redirects to ' + apiobj.params.page);
+		// delete redirects and redirect talk pages
+		var redirectDeleter = new Morebits.batchOperation('Deleting redirects to ' + apiobj.params.page + ' and their associated talk pages');
 		redirectDeleter.setOption('chunkSize', Twinkle.getPref('batchChunks'));
-		redirectDeleter.setPageList(pages);
+		redirectDeleter.setPageList(pagesToDelete);
 		redirectDeleter.run(function(pageName) {
 			var wikipedia_page = new Morebits.wiki.page(pageName, 'Deleting ' + pageName);
-			wikipedia_page.setEditSummary('[[WP:CSD#G8|G8]]: Redirect to deleted page "' + apiobj.params.page + '"');
+			wikipedia_page.setEditSummary('[[WP:CSD#G8|G8]]: Redirect or redirect talk page to deleted page "' + apiobj.params.page + '"');
```
siddharthvp, line 603: I think it would be better to do two batch operations as you had in an earlier iteration so distinct log summaries are used for the two cases – "Redirect to deleted page" or "Talk page of redirect".

NovemLinguae:
This patch is too much work to change further. However the code here could be a nice base for a patch to the CSD module that does the same thing.

PR #1490, opened 2021-12-15
speedy: remove CSD U3

NovemLinguae:
Per request at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#CSD_U3

CSD U3 was removed from the WP:CSD policy in July 2021. This patch puts Twinkle back into alignment with the policy.

Did a quick CSD test on a userspace page. Worked.

Did a quick test loading a user's local config with "u3" in its preferences. Not seeing console errors.


siddharthvp:
Looks like a clean removal.

PR #1491, opened 2021-12-17
tb: fix WP:NORN description

NovemLinguae:
Fixes this incorrect description. WP:NORN is not the reliable sources noticeboard. It is the no original research noticeboard.

### DIFF
--- a/modules/friendlytalkback.js
+++ b/modules/friendlytalkback.js
@@ -280,7 +280,7 @@ Twinkle.talkback.noticeboards = {
 		editSummary: 'You have replies at the [[Wikipedia:Help desk|Wikipedia help desk]]'
 	},
 	norn: {
-		label: 'WP:NORN (Reliable sources noticeboard)',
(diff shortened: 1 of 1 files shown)


modules/friendlytalkback.js
```diff
@@ -280,7 +280,7 @@ Twinkle.talkback.noticeboards = {
 		editSummary: 'You have replies at the [[Wikipedia:Help desk|Wikipedia help desk]]'
 	},
 	norn: {
-		label: 'WP:NORN (Reliable sources noticeboard)',
+		label: 'WP:NORN (No original research noticeboard)',
 		text: '{{subst:Norn-notice|thread=$SECTION}} ~~~~',
 		editSummary: 'Notice of discussion at [[Wikipedia:Reliable sources/Noticeboard]]'
```
siddharthvp, line 285: Edit summary can be fixed as well.

siddharthvp:
wow, how did this end up being overlooked all this while

PR #1493, opened 2021-12-17
xfd: fix "2nd AFD of article ending in number" bug

NovemLinguae:
Twinkle bug. Reported here: https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#2015th_nomination

Steps to reproduce:
- Create article Test (2000)
- Create article Wikipedia:Articles for deletion/Test (2000)
- Use Twinkle->XFD->AFD on article Test (2000). It will create an AFD called Wikipedia:Articles for deletion/Test (2001st nomination)

There's a complex RegEx that accepts (2000) and (2000th nom) and (2000th nomination). I hesitate to change it because if someone bothered to code it that way, these are probably AFD formats that were used in the past.

Instead, I opted to institute a maximum AFD number. I set 100 as the max. This should eliminate most year false positives.

image

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -1080,7 +1080,8 @@ Twinkle.xfd.callbacks = {
 					var match = order_re.exec(title);
 
 					// No match; A non-good value
-					if (!match) {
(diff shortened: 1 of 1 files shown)


siddharthvp:
> I hesitate to change it because if someone bothered to code it that way, these are probably AFD formats that were used in the past.

Other formats seem unlikely, but anyway this seems like a reasonable approach too.

PR #1494, opened 2021-12-18
welcome: fix topWelcomes placement bug

NovemLinguae:
Reported at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Welcome_messages_added_above_%22Talk_header%22

If you have the Twinkle preference topWelcomes set, and you try to place a welcome message on a page with {{Talk header}}, the order ends up being `welcomeText + {{Talk header}} + text`.

This patch fixes the order for simple cases.

image

### DIFF
--- a/modules/friendlywelcome.js
+++ b/modules/friendlywelcome.js
@@ -630,7 +630,14 @@ Twinkle.welcome.callbacks = {
 		var welcomeText = Twinkle.welcome.getTemplateWikitext(params.type, params.template, params.article);
 
 		if (Twinkle.getPref('topWelcomes')) {
-			text = welcomeText + '\n\n' + text;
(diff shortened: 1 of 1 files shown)


modules/friendlywelcome.js
```diff
@@ -630,7 +630,14 @@ Twinkle.welcome.callbacks = {
 		var welcomeText = Twinkle.welcome.getTemplateWikitext(params.type, params.template, params.article);
 
 		if (Twinkle.getPref('topWelcomes')) {
-			text = welcomeText + '\n\n' + text;
+			var hasTalkHeader = /^\{\{Talk header\}\}/i.test(text);
```
siddharthvp, line 633: Let's also check for the common redirects such as `talkheader`

PR #1497, opened 2021-12-24
tag: increase whitespace in redirects

NovemLinguae:
https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Redirect_format

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1755,7 +1755,7 @@ Twinkle.tag.callbacks = {
 					oldPageTags += '\n' + pageTag;
 				});
 			}
-			pageText += '\n\n{{Redirect category shell|' + tagText + oldPageTags + '\n}}';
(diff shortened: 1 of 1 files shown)


modules/friendlytag.js
```diff
@@ -1755,7 +1755,7 @@ Twinkle.tag.callbacks = {
 					oldPageTags += '\n' + pageTag;
 				});
 			}
-			pageText += '\n{{Redirect category shell|' + tagText + oldPageTags + '\n}}';
+			pageText += '\n\n{{Redirect category shell|' + tagText + oldPageTags + '\n}}';
```
siddharthvp, line 1758:
```suggestion
			pageText = pageText.trim() + '\n\n{{Redirect category shell|' + tagText + oldPageTags + '\n}}';
```
not sure if it's possible for pageText to already have some newlines, if it does this change would fix it

PR #1504, opened 2022-01-25
unlink: fix colon in title bug

### LINKED ISSUE #1503: unlink: mainspace pages with colon are parsed incorrectly
Example: https://test.wikipedia.org/wiki/Cyber_Seduction:_His_Secret_Life

Running TW->Unlink on this page will not work because it has a colon in the title. It will throw this error.

Reported by Liz at WT:TWINKLE - https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Problem_with_unlinking_backlinks

Line of code where error is occurring: https://github.com/wikimedia-gadgets/twinkle/blob/10021f12f3f3389d86a50120eb73a85f1423aac1/morebits.js#L4929-L4932

NovemLinguae:
Fixes #1503 unlink: mainspace pages with colon are parsed incorrectly

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -4927,7 +4927,7 @@ Morebits.wikitext.page.prototype = {
 
 		var mwTitle = mw.Title.newFromText(link_target);
 		var ns = mwTitle.getNamespacePrefix().slice(0, -1); // remove trailing colon
-		var title = mwTitle.title;
+		var title = mwTitle.getMainText();
 
 		var link_regex_string = '';
 		if (ns) {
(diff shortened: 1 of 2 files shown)


morebits.js
```diff
@@ -4924,13 +4924,13 @@ Morebits.wikitext.page.prototype = {
 		if (link_target.indexOf(':') === 0) {
 			link_target = link_target.slice(1);
 		}
```
siddharthvp, line 4926: not needed anymore I think

PR #1508, opened 2022-02-04
twinkle,twinkleconfig: add support for vector-2022

MusikAnimal:


MusikAnimal:
Trivial change and I've tested it. Self-merging seeing as we need to get this out ASAP

PR #1509, opened 2022-02-05
Switch to {{socklist}} in SPI report

TamzinHadasa:
I've proposed this as the new default for {{SPI report}}, and think Twinkle should use it as well. It would eliminate the longstanding problem that Twinkle reports, because they use `{{{socksraw}}}`, do not generate editor interaction links. This way, {{SPI report}} need only pass usernames as numbered parameters, and {{sock list}} does all the heavy lifting. Also makes it easier for people to add more sox after the report is filed. See https://en.wikipedia.org/wiki/Template_talk:SPI_report#Proposed_change_to_sock_list.

Resolves #1505 by superseding, just because it coincidentally affects the same lines of code. I can remove the part that copies 1505 if y'all would prefer they stay two separate PRs.

Blocked until the relevant change is implemented to {{SPI report}} (pending consensus).

Obligatory disclaimer that I am not very good at JavaScript, and *think* this change should work but am not 100% sure.

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -810,10 +810,10 @@ Twinkle.arv.processSock = function(params) {
 	Morebits.wiki.addCheckpoint(); // prevent notification events from causing an erronous "action completed"
 
 	// prepare the SPI report
-	var text = '\n\n{{subst:SPI report|socksraw=' +
-		params.sockpuppets.map(function(v) {
-			return '* {{' + (mw.util.isIPAddress(v, true) ? 'checkip' : 'checkuser') + '|1=' + v + '}}';
-		}).join('\n') + '\n|evidence=' + params.evidence + ' \n';
+	var text = '\n{{subst:SPI report|' +
+            params.sockpuppets.map(function(sock, index) {
+                return (index + 1) + '=' + sock;
+            }).join('|') + '\n|evidence=' + params.evidence + ' \n';
 
 	if (params.checkuser) {
 		text += '|checkuser=yes';


TamzinHadasa:
@Dreamy-Jazz FYI

Dreamy-Jazz:
Looks good to me. I would note that consensus about this template is likely not to take too long, so this pull request may be ready to go soon.

Dreamy-Jazz (approved):
Verified this change works using Firefox inspect element's console. No changes needed to the code, and as long as this doesn't have major opposition on-wiki I think this would be fine to merge.

NovemLinguae:
Didn't work on testwiki. May be a missing template on testwiki, or may be a bug in the code. Can you investigate?

https://test.wikipedia.org/w/index.php?title=Wikipedia%3ASockpuppet_investigations%2FNovem_Linguae&type=revision&diff=499719&oldid=499714

image

TamzinHadasa:
@NovemLinguae Yeah, there was a lot that had to be imported first. Think I've got it sorted. Try again?

NovemLinguae:
Works now, thanks for updating those templates. The side by side looks good, the old and new code render exactly the same as far as I can tell.

https://test.wikipedia.org/w/index.php?title=Wikipedia%3ASockpuppet_investigations%2FNovemBot&type=revision&diff=500322&oldid=500321

Detailed discussion onwiki, for the record:

https://en.wikipedia.org/wiki/Template_talk:SPI_report#Proposed_change_to_sock_list

I'll go ahead and make a post on the above wiki page, and pause for responses, and also to give @siddharthvp a chance to weigh in. If everything looks good in a couple of days, I'll merge it in.

PR #1463, opened 2021-11-17
welcome: remove 3, add 1, split a group into two groups

NovemLinguae:
Fixes #1406 Remove welcome-graphical, welcome-menu, and welcome-screen, and add welcoming

I also split "General welcome templates" into two groups: "General welcome templates" and "Specific welcome templates".

PR #1513, opened 2022-02-18
add back 2 welcome templates

NovemLinguae:
Related #1463. Partial revert.

Tested. Besides the revert, changed code to use welcome templates directly instead of redirects.

Requested onwiki: https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Welcome-menu_template_missing_from_the_latest_iteration_of_the_drop-down_welcome_menu

### DIFF
--- a/modules/friendlywelcome.js
+++ b/modules/friendlywelcome.js
@@ -248,6 +248,14 @@ Twinkle.welcome.templates = {
 				description: 'a welcome message with some helpful links and a plate of cookies',
 				syntax: '{{subst:welcome cookie}} ~~~~'
 			},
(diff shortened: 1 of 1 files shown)


NovemLinguae:
This should fix Bauerjack's issue too: https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Double_signature

sdkb:
I have requested [here](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Restoration_of_poorly_designed_welcome_templates) that this patch be reverted.

PR #1524, opened 2022-02-25
xfd: Use Deletion sorting/multi

### LINKED ISSUE #1519: Use Template:Deletion sorting/multi instead of multiple copies of Template:Deletion sorting
Enterprisey's delsort tool has been recently upgraded to use {{[Deletion sorting/multi](https://en.wikipedia.org/wiki/Template:Deletion_sorting/multi)}} instead of multiple copies of {{[Deletion sorting](https://en.wikipedia.org/wiki/Template:Deletion_sorting)}} when more than one sorting category is specified. See [Wikipedia:Articles for deletion/Index of Abkhazia-related articles](https://en.wikipedia.org/wiki/Wikipedia:Articles_for_deletion/Index_of_Abkhazia-related_articles) for an example of how this can save space on nominations with lots of categories.

Twinkle should be updated to use this new template as well.

TheTVExpert:
Have not tested, but should resolve #1519.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -806,10 +806,8 @@ Twinkle.xfd.callbacks = {
 
 		text += '}}';
 
-		if (params.delsortCats) { // Only for AFDs
-			params.delsortCats.forEach(function (cat) {
-				text += '\n{{subst:delsort|' + cat + '|~~~~}}';
-			});
+		if (params.delsortCats.length) { // Only for AFDs
+			text += '\n{{subst:Deletion sorting/multi|' + params.delsortCats.join('|') + '|sig=~~~~}}';
 		}
 
 		return text;


NovemLinguae:
Looks like Template:Deletion sorting/multi is more compact, which has advantages when listing a large number of delsort categories. So this looks like a good feature.

Old:
image

New:
image

I found an issue while testing. If there are no delsort cats, it will try to write the template anyway, throwing an error. This is probably because empty arrays are truthy in JavaScript. Would you mind fixing when you get a chance? Thanks.

modules/twinklexfd.js
```diff
@@ -807,9 +807,11 @@ Twinkle.xfd.callbacks = {
 		text += '}}';
 
 		if (params.delsortCats) { // Only for AFDs
+			text += '\n{{subst:Deletion sorting/multi';
 			params.delsortCats.forEach(function (cat) {
```
siddharthvp, line 811: this can be refactored to use `.join('|')` instead

TheTVExpert: Done

PR #1528, opened 2022-02-27
Add {{uw-islamhon}} series of warnings

christophorum:


NovemLinguae:
Does this have enough consensus? I note that this template series was [recently TFD'd, and levels 2 and 3 were redirected](https://en.wikipedia.org/wiki/Wikipedia:Templates_for_discussion/Log/2021_July_11#Template:Uw-islamhon1).

It might be better to add this as a single issue template. I notice in the TFD discussion that you say that Twinkle does not have good support for single issue templates, but I am not sure this is correct. We have single-issue notices and single-issue warning categories where this template could be placed. Maybe you could update your pull request to insert into one of those categories?

Personally I would also recommend swapping Template:Uw-islamhon and Template:Uw-islamhon1 on Wikipedia, and make Template:Uw-islamhon1 redirect to Template:Uw-islamhon. But that's up to you of course. If you decide to do that, you can request a round robin swap at [[WP:RM/TR]].

image

christophorum:
Updated to use it as a single-issue notice.

PR #1544, opened 2022-03-18
protect: Bug fix from #1433 and fixes #1543

### LINKED ISSUE #1543: protect: wrap protection template with noinclude label not fully visible
Screenshot 2022-03-18 at 9 38 25 PM

Probably needs some html escaping.

TheTVExpert:
Fixes bug from #1433 when selecting a preset to protect an AfD and fixes #1543.


NovemLinguae:
I refactored the code to be more self-documenting. Let me know if the refactor looks OK.

TheTVExpert:
@NovemLinguae Looks good to me. Thanks.

NovemLinguae:
Related #1550. Not sure if we should tackle that issue in this PR, or do a separate one for that. It's something I noticed while looking over this code though.

PR #1556, opened 2022-04-04
Improve wording for merges

fredster33:


NovemLinguae:
"Other template destination template for merge" sounds a bit redundant, says template twice.

"Other module destination template for merge" sounds incorrect, we don't usually merge modules into templates.

Old wordings might be better.

fredster33:
Pushed an edit, hopefully this is better.

NovemLinguae:
Thanks for submitting. But I don't see anything to indicate that there is a "destination for merge". The documentation (TFD notices, TFD links) seem to not presume a destination, but rather let the folks at TFD hammer out the final destination.

Closing for now.

PR #1562, opened 2022-04-11
diff: Use mw.config to get revision id

### LINKED ISSUE #1561: diff: Uncaught exception on deleted revisions diff
View https://en.wikipedia.org/w/index.php?diff=2
```
Uncaught TypeError: Cannot read properties of null (reading '1')
```
at https://github.com/wikimedia-gadgets/twinkle/blob/8e7fd292af0f59074422db00631c28602f88faf2/modules/twinklediff.js#L30

This function should be disabled on deleted diff page.

Xi-Plus:
Use robuster way to get revision id. And wgDiffNewId is missing on deleted diff, fix #1561


NovemLinguae:
This would probably fix bug #1582 too

PR #1579, opened 2022-05-07
config: handle ombox deprecation

NovemLinguae:
Requested by Izno at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#ombox_use_in_MediaWiki:Gadget-twinkleconfig.js

There is a message box in Twinkle that uses MediaWiki's ombox CSS class, but this will be deprecated soon. This patch copies the styling for this to Twinkle's morebits CSS stylesheet, eliminating a dependency.

Could potentially add all the ombox styles, but I decided to just add the ones we use.


NovemLinguae:
image

Todo: Need to make tickets for or fix these items too:

image

NovemLinguae:
image

PR #1580, opened 2022-05-07
dev-server: fix single apostrophes in CSS bug

NovemLinguae:
There is a SyntaxError in the dev version of Twinkle if you add any apostrophe in morebits.css. For example, if you want to add a comment such as `/* Don't do this. */`

This patch escapes single apostrophes, fixing the issue.

### DIFF
--- a/scripts/dev-server.js
+++ b/scripts/dev-server.js
@@ -29,6 +29,7 @@ const server = http.createServer(async (request, response) => {
 	}
 	for (let file of cssFiles) {
 		let css = (await readFile(file)).replace(/\s+/g, ' ');
+		css = css.replace(/'/g, "\\'"); // escape apostrophes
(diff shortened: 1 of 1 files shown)


NovemLinguae:
@siddharthvp I notice this is failing CI. Any suggestions on the fix?

siddharthvp:
@NovemLinguae The [expanded guideline](https://github.com/wikimedia-gadgets/twinkle/security/code-scanning/18) appears to suggest escaping backslashes first before escaping single quote. If that doesn't work, we could also just dismiss the alert as it's more relevant to untrusted inputs - the dev-server only runs on one's own system so there's no security issue.

siddharthvp:
JSON.stringify sounds like the perfect and fool-proof choice!

PR #1586, opened 2022-05-21
tag: add part and reason params for {{update}}

loganrosen:
Adding support for the suggested `part` and `reason` parameters of `{{update}}`. Tested on [the test wiki](https://test.wikipedia.org/w/index.php?title=Watty62_test&action=history).

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -714,7 +714,24 @@ Twinkle.tag.article.tagList = {
 		],
 		'Timeliness': [
 			{ tag: 'Current', description: 'documents a current event', excludeMI: true }, // Works but not intended for use in MI
-			{ tag: 'Update', description: 'needs additional up-to-date information added' }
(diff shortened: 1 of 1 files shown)


NovemLinguae:
Hey there. I removed the "type" parameter since the options are article (default) and section, and Twinkle can only place tags at the top of the article at the moment, so would basically never need to use section. You OK with this change? If so I think this is ready to merge. Thanks.

NovemLinguae:
Note to self: Look into widening the reason box.

loganrosen:
The `part` parameter can be more descriptive than just "section" (see the first example in [the usage section](https://en.wikipedia.org/wiki/Template:Update#Usage)). I'm okay removing it regardless if you think it won't add much value.

NovemLinguae:
Ah, I see, thank you for the explanation. I added back `updatePart`. How's it look now?

loganrosen:
Looks good to me, thanks!

Issue #1537, opened 2022-03-11
warn: Template:Botun doesn't allow linked article


TheTVExpert:
Requested by InterstateFive at https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle#Slight_error_when_warning_a_user_for_having_a_%22bot_username%22

Add to list at https://github.com/wikimedia-gadgets/twinkle/blob/f8e5769e0aea9e2bad6894f9faa62d7d41b1a494/modules/twinklewarn.js#L1402-L1409

NovemLinguae:
So this template for its 1st parameter doesn't seem to want a username, it wants some kind of message. The exact code in the template is `{{<includeonly>safesubst:</includeonly>#if:{{{1|}}}|&nbsp;In addition, it {{{1}}}.}}`

The proposed algorithm above I think would just change "linked page" to the above text. But then would create a situation where a user supplies a username, and it just prints `In addition, it USERNAME.`

I think #1538 might be a better approach. Thoughts?

TheTVExpert:
>The proposed algorithm above I think would just change "linked page" to some custom text that asks for a username.

I think this object just changes the label for the linked article input, so for this template we would ask for an additional reason, not a username.

I do agree that #1538 might be better though, since the reason parameter for this template probably doesn't get much use.

PR #1604, opened 2022-06-15
warn: ability to hide linked article and reason

### LINKED ISSUE #1408: Small bug with {{Summary2}}
When warning a user with [{{Summary2}}](https://en.wikipedia.org/wiki/Template:Uw-editsummary2), there's a box for putting in a linked article. The template does not support linking any specific article (nor do I think it should, since it's designed to warn about a broad behavioral pattern), so this should be removed.

### LINKED ISSUE #1538: warn: option to suppress "linked page"
We have an option to suppress the edit summary...

````
		'uw-bite': {
			label: '"Biting" newcomers',
			summary: 'Notice: "Biting" newcomers',
			suppressArticleInSummary: true
		},
````

But no option to suppress the Linked Page text input box

````
		'uw-botun': {
			label: 'Bot username',
			summary: 'Warning: Bot username',
			suppressLinkedPage: true
		},
````

This ticket proposes adding code to handle `suppressedLinkedPage: true`

Would fix the issue reported by InterstateFive at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Slight_error_when_warning_a_user_for_having_a_%22bot_username%22, where typing a linked article for Template:Uw-botun displays non-grammatical text.

image

NovemLinguae:
Fixes #1408
Fixes #1538
Related #1537

When warning a user with `uw-editsummary2`, hides the "linked page" and "reason" text boxes.

I also created infrastructure to easily mark other templates with `hideLinkedPage` and `hideReason`, if we need to in the future.

I also added unit tests. I changed some unit test configuration files so that in the future, if we want to unit test modules, it will be easier.

Also renamed a couple variables.

image

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -1,4 +1,4 @@
-name: Morebits function tests
+name: Unit tests
 
 on:
   push:
@@ -24,5 +24,5 @@ jobs:
     - name: Install dependencies
       run: npm ci
 
-    - name: Morebits tests
+    - name: Unit tests
       run: npm run test
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,7 +1,7 @@
 /* eslint-env node */
 
 module.exports = {
-	testMatch: ['**/tests/morebits*.js'],
+	testMatch: ['**/tests/morebits*.js', '**/tests/twinkle*.js'],
 	testEnvironment: 'jsdom',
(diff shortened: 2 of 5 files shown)


NovemLinguae:
@siddharthvp. Hey there. I wrote some unit tests for what is basically a private method. Think it's OK to keep them in or should I remove them? I know unit tests for private methods are sometimes considered a code smell.

siddharthvp:
@NovemLinguae I think it's perfectly ok. Modules don't really have any "public" methods so this is the only way we can unit-test anything.

PR #1611, opened 2022-06-22
twinkle: add preferences link to dropdown menu

### LINKED ISSUE #1326: Add preferences to dropdown menu?
Just a suggestion. Consider adding "Prefs" (preferences) to the Twinkle dropdown menu, at the bottom. That seems like an intuitive spot for it. It would point to the page Wikipedia:Twinkle/Preferences, so you wouldn't have to type WP:TWINKLE and click the link at the top, which is more steps than ideal.

NovemLinguae:
Fixes #1326

I think myself and Qwerfjkl are in favor (see onwiki discussion at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Add_%22Prefs%22_to_Twinkle_menu%3F), and @Amorymeltzer had some reservations. Open to feedback.

### DIFF
--- a/twinkle.js
+++ b/twinkle.js
@@ -486,6 +486,8 @@ Twinkle.load = function () {
 	if (isVector && Twinkle.getPref('portletType') === 'menu' && $('#p-twinkle').length === 0) {
 		$('#p-cactions').css('margin-right', 'initial');
 	}
+
+	Twinkle.addPortletLink(mw.util.getUrl('Wikipedia:Twinkle/Preferences'), 'Prefs', 'tw-prefs', 'Open Twinkle preferences page');
 };
 
 
--- a/twinkle.js
+++ b/twinkle.js
@@ -487,7 +487,9 @@ Twinkle.load = function () {
 		$('#p-cactions').css('margin-right', 'initial');
 	}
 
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
I'll reiterate what I said earlier, in particular that adding another tab for non-vector/non-timeless users is a big deal, especially as the name "prefs" is vague and prone to confusion/misleading.

NovemLinguae:
How about renaming Prefs to Config?

So of the 5 official and maintained skins, Non-Vector, non-Timeless just leaves us with MonoBook that's affected, right? For me, it overflows no matter what, so dunno if that's a blocker. The long term fix for that is probably ticket #716.

Guess this will die on the vine unless someone else weighs in. Sounds like you're not in favor. Ah well.

NovemLinguae:
Hey I just had a good idea. How about we only have "Config" display for Vector and Timeless. Any objections to that?

NovemLinguae:
Hi @Amorymeltzer. Sorry to bug you again. Was just wondering if my two changes above assuage your concern? Thanks for your time.

Amorymeltzer:
Certainly better that way

PR #1615, opened 2022-07-06
morebits: Fix commentOutImage for multiple files with same params

Xi-Plus:
It's required to unbind comments immediately after the replacement. Otherwise the first file will be replaced multiple times.
Fix https://en.wikipedia.org/w/index.php?title=User:Xiplus/sandbox&diff=prev&oldid=1096698125

### DIFF
--- a/tests/morebits.wikitext.js
+++ b/tests/morebits.wikitext.js
@@ -151,8 +151,8 @@ describe('Morebits.wikitext', () => {
 			{
 				name: 'multiple',
 				method: 'commentOutImage',
-				input: 'O, [[File:Fee.svg]] she [[File:Fee.svg|doth|teach]] the [[File:Fee.svg|torches]] to burn bright!',
-				expected: 'O, <!-- [[File:Fee.svg]] --> she <!-- [[File:Fee.svg|doth|teach]] --> the <!-- [[File:Fee.svg|torches]] --> to burn bright!',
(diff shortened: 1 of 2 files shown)


NovemLinguae:
Hey there. I added the test case from your sandbox, to be sure we captured the bug in the tests. You OK with the change?

tests/morebits.wikitext.js
```diff
@@ -151,10 +151,17 @@ describe('Morebits.wikitext', () => {
 			{
 				name: 'multiple',
 				method: 'commentOutImage',
-				input: 'O, [[File:Fee.svg]] she [[File:Fee.svg|doth|teach]] the [[File:Fee.svg|torches]] to burn bright!',
-				expected: 'O, <!-- [[File:Fee.svg]] --> she <!-- [[File:Fee.svg|doth|teach]] --> the <!-- [[File:Fee.svg|torches]] --> to burn bright!',
+				input: 'O, [[File:Fee.svg]] she [[File:Fee.svg|doth|teach]] the [[File:Fee.svg|torches]] to burn bright! [[File:Fee.svg]]',
+				expected: 'O, <!-- [[File:Fee.svg]] --> she <!-- [[File:Fee.svg|doth|teach]] --> the <!-- [[File:Fee.svg|torches]] --> to burn bright! <!-- [[File:Fee.svg]] -->',
```
Xi-Plus, line 155: In fact, I've included the case for the bug in my patch.

PR #1619, opened 2022-07-09
Revert "morebits.date: Don't import methods that collide with PageTriage (T268513)"

siddharthvp:
This reverts commit c6f2bd4f3809e3f59124949d31c791923d4a93f1.

The linked phab task has been resolved now - PageTriage is no longer using date.js external which messes up `Date.prototype`.


NovemLinguae:
@siddharthvp. What's the easiest way to manually test this? Perhaps a test case that was broken before the old patch?

siddharthvp:
This is a no-op now since the MediaWiki changes have been merged. So this is just code simplification - so doesn't require testing per se (other than to check Morebits.date is not broken 😅).

PR #1632, opened 2022-08-13
config: Replace removed successbox classes

Xi-Plus:
Fixes for https://phabricator.wikimedia.org/T300314

Before:
image
After:
image


NovemLinguae:
Did Twinkle generate a similar green box inside the status box before the Phab change that broke this?

I wonder what the historical reasons are that this one page does this, and other pages and modules use text only for their status messages.

Will probably still accept this PR. But just wondering.

Xi-Plus:
> Did Twinkle generate a similar green box inside the status box before the Phab change that broke this?

Yes. There was a green box before.

Issue #1550, opened 2022-03-31
protect: when should noinclude be checked by default?


NovemLinguae:
In the protect module, there is code that determines if "Wrap protection template with <noinclude>" should be checked or not checked by default in two places. Each place has slightly different logic.

Place # 1: in the Morebits form initialization code:

![image](https://user-images.githubusercontent.com/79697282/160982155-83f90d1e-2015-439f-9030-5fbb92b262b7.png)

Place # 2: in the code that detects changes to the form:

![image](https://user-images.githubusercontent.com/79697282/160982220-66de1cff-a9aa-4f99-ac21-6fd913045925.png)

If you look closely, you'll notice the logic is different for each of these. I suggest that we decide in this ticket which logic to use, and then make it consistent in both places.

Logic # 1 summary: if template namespace OR AFD, mark checkbox. else, don't mark checkbox

Logic # 2 summary: if {{Pp-template}} selected OR AFD, mark checkbox. elseif not template namespace, unmark checkbox. else, do nothing, leave whatever the previous value was (this last case could occur if, for example, something in template namespace is not marked with {{Pp-template}}

PR #1637, opened 2022-08-18
protect: sanitized-css: don't tick <noinclude> check box

NovemLinguae:
Don't tick the "Wrap protection template with `<noinclude>`" check box by default if the content model is css, sanitized-css, or javascript.

image

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -534,7 +534,7 @@ Twinkle.protect.callback.changeAction = function twinkleprotectCallbackChangeAct
 						name: 'noinclude',
 						label: 'Wrap protection template with &lt;noinclude&gt;',
 						tooltip: 'Will wrap the protection template in &lt;noinclude&gt; tags, so that it won\'t transclude',
-						checked: (isTemplateNamespace || (isWikipediaNamespace && isArticlesForDeletion)) && !isCode
+						checked: (isTemplateNamespace || isAFD) && !isCode
 					}
 				]
 			});
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -1076,7 +1076,7 @@ Twinkle.protect.callback.changePreset = function twinkleprotectCallbackChangePre
 
 			// Default settings for adding <noinclude> tags to protection templates
 			var isTemplateEditorProtection = form.category.value === 'pp-template';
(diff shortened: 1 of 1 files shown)


modules/twinkleprotect.js
```diff
@@ -517,6 +517,11 @@ Twinkle.protect.callback.changeAction = function twinkleprotectCallbackChangeAct
 				list: Twinkle.protect.protectionTags,
 				event: Twinkle.protect.formevents.tagtype
 			});
+
+			var isTemplateNamespace = mw.config.get('wgNamespaceNumber') === 10;
+			var isWikipediaNamespace = mw.config.get('wgNamespaceNumber') === mw.config.get('wgNamespaceIds').project;
+			var isArticlesForDeletion = mw.config.get('wgTitle').indexOf('Articles for deletion/') === 0;
```
siddharthvp, line 523: I think we can combine these two in one line
```suggestion
			var isAFD = mw.config.get('wgPageName').indexOf('Wikipedia:Articles for deletion/') === 0;
```

modules/twinkleprotect.js
```diff
@@ -517,6 +517,11 @@ Twinkle.protect.callback.changeAction = function twinkleprotectCallbackChangeAct
...
+			var isCode = ['javascript', 'css', 'sanitized-css'].indexOf(mw.config.get('wgPageContentModel')) !== -1;
```
siddharthvp, line 524: shouldn't scribunto also come here?

NovemLinguae: From my testing, I think the "Tagging options" part of the window is hidden when protecting a module. I could not get it to appear. From my testing, the act of protecting also doesn't modify any wikicode (i.e. no {{Pp-semi}} template is written anywhere).

modules/twinkleprotect.js
siddharthvp: oops sorry, wgPageName encodes spaces as underscores. While we're at it, might as well use Morebits.pageNameNorm (which really does use spaces) and use String#startsWith()

modules/twinkleprotect.js
siddharthvp: Ah yeah, the tagging options are [indeed disabled](https://github.com/wikimedia-gadgets/twinkle/blob/30ec330be93e84dffebc286484094bc8e66421cb/modules/twinkleprotect.js#L70) for Scribunto.

NovemLinguae:
Related #1550

PR #1642, opened 2022-09-25
fluff: Fix rollback links for top rev in history

Xi-Plus:
https://en.wikipedia.org/w/index.php?title=Wikipedia&action=history

image

"First link" now wrapped by `<span>` with `.mw-firstlink` class on the first page of history. Use `a.mw-firstlink` to check clickable firstlink.

With this patch:
image


NovemLinguae:
Note to testers: you have to have this rollback preference enabled

image

PR #1651, opened 2022-10-21
morebits: fix handling of hard errors

siddharthvp:
In #1179, support for new error formats were added and used by default. However, the error handling code still assumes the legacy errorformat, which passes details of the first `error`, while in the newer ones we have an array `errors`.

Notes for testing:
On testwiki, add text "Trigger abuse filter warn" to any page (it triggers https://test.wikipedia.org/wiki/Special:AbuseFilter/256):
```js
var p = new Morebits.wiki.page('SD0001test', '');
p.setAppendText('trigger abuse filter warn');
p.setEditSummary('edit summary');
p.append();
```

Without this patch, we get:
Screenshot 2022-10-21 at 11 03 53 PM

The TypeError prevents any failure handling callback from running (one of the reasons why the callback approach is sub-optimal and we should be using promises everywhere).

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -4064,6 +4064,9 @@ Morebits.wiki.page = function(pageName, status) {
 
 		// hard error, give up
 		} else {
+			var response = ctx.saveApi.getResponse();
+			var errorData = response.error || // bc error format
+				response.errors[0].data; // html/wikitext/plaintext error format
 
 			switch (errorCode) {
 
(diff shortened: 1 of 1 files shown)


siddharthvp:
Originally reported by User:Sdkb at https://en.wikipedia.org/wiki/User_talk:SD0001/DYK-helper#Bug:_Loses_nomination_when_edit_filter_triggered (DYK-helper uses morebits)

siddharthvp:
Merged; clear-cut bug affecting users

Issue #1515, opened 2022-02-18
Linter set to ES5


NovemLinguae:
I notice our linter is set to enforce ES5 and reject any JavaScript features newer than that. This is a little inconvenient to the developer, who has to avoid newer language features such as match, let, class, await, startsWith, endsWith, includes, regex s flag, etc.

Was just wondering what the idea behind keeping the linter on ES5 is. I know [Twinkle has 45,000 users](https://en.wikipedia.org/wiki/Special:GadgetUsage) or so. I guess the assumption is that a small subset of these folks are using very old browsers that don't support ES6+?

Is it worth the tradeoff? Should we plan to allow ES6 sometime in the future?

siddharthvp:
The reason is just that MW gadgets don't yet support ES6. We need https://phabricator.wikimedia.org/T75714 to be resolved first one way or another. It is quite hard to resolve satisfactorily - as it requires introducing a ES6-capable validator. I have a [rather hacky patch](https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Gadgets/+/758086) filed as an interim solution.

NovemLinguae:
[This patch](https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Gadgets/+/758086/) was +2'd today, opening the door for gadgets to start using ES6. Next Thursday when this deploys, do we want to switch Twinkle over to ES6?

If so, here is our todo list:

[ ] Make a post at the interface administrator's noticeboard requesting this change
[ ] Reconfigure our linter to allow ES6

siddharthvp:
I am on board with this. If @MusikAnimal is also in agreement, @NovemLinguae can you please file the patch for updating the linter reconfiguration, [gadget definition file](https://github.com/wikimedia-gadgets/twinkle/blob/master/gadget.txt) and the README? On-wiki gadget definition change can be handled as part of deployment and IMO doesn't need an IANB post.

MusikAnimal:
100% in support! We may kill IE support in doing this, but nowadays that's the intention: https://www.mediawiki.org/wiki/Compatibility/IE11

I'm happy to do the deployment when the time comes. Just ping me.

NovemLinguae:
I see a bunch of JavaScript versions here. Are these all safe for Twinkle/gadget use now? Or do we need to look into having the linter continue enforcing a block of some of the newer ones?

[...]

PR #1652, opened 2022-10-26
configure eslint to use es6

NovemLinguae:
Related #1515

Bump eslint from ES5 to ES6

Stuff I tested, all looked good:
- test ES6 backticks, make sure they didn't work before and they work now
- test ES6 arrow function, make sure they didn't work before and they work now
- test ES2017 async/await, make sure they don't work now
- test ES2018 regex /s, make sure they don't work now

I deleted eslint-plugin-es5. I assume that doesn't need replacing.


siddharthvp (changes requested):
Line 11 in DEVELOPER.md needs an update. As does the gadget.txt file.

DEVELOPER.md
```diff
@@ -2,22 +2,22 @@
...
-- Certain positional jQuery selectors like `:first`, `:last`, and `:eq` were [deprecated in jQuery version 3.4.0](https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/) and should probably not be reintroduced.  Instead, use methods like `.first()`, `.last()`, or `.eq()`.
+- Twinkle is meant to run on the latest weekly version of MediaWiki as rolled out every Thursday on the English Wikipedia. Backwards compatibility is not guaranteed.
+- The goal is for Twinkle and Morebits to support the same [browsers that MediaWiki supports](https://www.mediawiki.org/wiki/Browser_compatibility). The Twinkle gadget on enwiki is configured so that we can use up to JavaScript version ES6. However, due to the MediaWiki minifier, we must not use keywords in ES2016 or later, such as async/await and RegEx /s flag. New functions in ES2016 or later, such as includes() and matchAll(), should be OK since these will not break the minifier.
```
siddharthvp, line 11:
```suggestion
- The goal is for Twinkle and Morebits to support the same [browsers for which MediaWiki provides Grade A support](https://www.mediawiki.org/wiki/Browser_compatibility), except IE 11. The Twinkle gadget on enwiki is configured so that we can use up to JavaScript version ES6. However, due to the MediaWiki minifier, we must not use keywords from ES2016 or later, such as async/await and RegEx /s flag. New functions from ES2016 or later, such as Array#includes() should be okay since these will not break the minifier.
```

Fixing grammar and dropping the mention of `matchAll()` since might be too new (supported only in Safari 13 while Grade A is purportedly there for Safari 9 as well -- although safari's MW support is rather sketchy).

PR #1658, opened 2022-10-28
twinkleblock: "Vandalism-only account": Link to policy rather than essay

ToBeFree:
"WP:Vandalism-only account" is an essay, "WP:DISRUPTONLY" is a policy section.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -1234,7 +1234,7 @@ Twinkle.block.blockPresetsInfo = {
 		nocreate: true,
 		pageParam: true,
 		reason: '{{uw-vaublock}} <!-- Username violation, vandalism-only account -->',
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
FWIW, see [[MediaWiki:Ipbreason-dropdown]]

ToBeFree:
Ah, it's in there too. Well, I rarely use the manual block form, and I never use it with this dropdown reason. I thus don't feel a need for me to change the system-wide MediaWiki builtin dropdown in this way. If Twinkle, on the other hand, had statistics about which block reason is used how often by whom, I'd probably appear in the top X for this one.

NovemLinguae:
I'll merge this if my code review suggestion above is addressed. Feel free to upload a new patchset.

ToBeFree:
I hope this addresses what you had noted; the note was lost in an attempt to clean up my followers list, which is only possible by blocking (and then unblocking) users on GitHub. This probably removed the review as well, though; sorry for that.

PR #1662, opened 2022-11-02
twinkle: notifications for CSD C1 should default to true

NovemLinguae:
Requested by DB1729 at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#%22Notify_page_creator_if_possible%22

For CSD C1, have "welcome page creator", "notify page creator when tagging", and "notify page creator when deleting" default to true instead of false.


NovemLinguae:
I messed up the rebase on this one. Will open a new PR.

Issue #1510, opened 2022-02-11
speedy: add {{Db-afc-move}}


NovemLinguae:
New template that @Primefac created. This issue is a note to self to add it if there's consensus, once it stabilizes, etc.

siddharthvp:
I think it's better to integrate this template into AFCH. If available in twinkle, it should probably be restricted to AFC reviewers, but then it isn't easy for twinkle to determine if the user is an AFC reviewer.

primefac:
I don't disagree, it's a good idea to chuck it into AFCH, but I fail to see the usefulness - AFCH is for accepting - if you're going to accept, why "accept" it twice (once to place the db- tag, once to move it)? I guess I'm just not seeing the workflow.

siddharthvp:
If the user is not an admin, AFCH could offer an option to {{db-afc-move}}-tag the target page since actually accepting is not possible.

NovemLinguae:
Tried to use {{Db-afc-move}} today. It's not in Twinkle or AFCH yet. First I forgot what the template name was so I had to spend a few minutes finding it. Then I forgot to put the required |page= parameter.

Putting this in Twinkle would be an efficiency for this workflow.

PR #1664, opened 2022-11-06
speedy: add {{db-afc-move}}

timuzhti:
Had to quote everything in the normalizeHash because of the hyphen in afc-move (unless I'm misunderstanding something).

Did a quick test with and without the page name for an actual draft (since I had one I wanted to move anyway). Worked.


timuzhti:
resolves #1510

NovemLinguae:
Any consensus objections to this? If not I plan to merge it after testing and reviewing the code. #1510 seems to have had some consensus objections, but there is some support onwiki to add this, for example in https://en.wikipedia.org/wiki/Wikipedia_talk:WikiProject_Articles_for_creation#Accept_Draft%3AGiuseppe_Dell'Anno_over_redirect%3F. And also I am in favor of it. I don't see why Twinkle and AFCH can't both possess this feature. And if Twinkle can get it out quicker than AFCH, then at least one script will have it.

PR #1667, opened 2022-11-15
tag: better detection of "not marked as reviewed"

### LINKED ISSUE #1573: tag: better detection of "not marked as reviewed"
Requested by IAmChaos at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Really_odd_question

Currently detects "marked as patrolled", which is slightly different than "marked as reviewed". Currently uses HTML detection.

`if (document.getElementsByClassName('patrollink').length) {`

HTML mark as patrolled detection is simple, but has at least 2 bugs:

* Detects unreviewed redirects created from page moves as reviewed. [Test page.](https://test.wikipedia.org/w/index.php?title=PageToMove3&redirect=no)
  * Steps to reproduce: Special:NewPagesFeed, filter by redirects, sort by oldest, find a page that has an {{R from page move}} tag. This is likely to fit the criteria.
* Detects pages that were reviewed and then unreviewed by clicking the "add to the new pages feed" link as reviewed. [Test page.](https://test.wikipedia.org/wiki/PageToMove4)

Will probably need to switch to an API call that detects "mark as reviewed". Example from one of my user scripts:

````
	async function isReviewed(pageID) {
		let api = new mw.Api();
		let response = await api.get( {
			action: 'pagetriagelist',
			format: 'json',
			page_id: pageID,
		} );
		// if old page
		if ( response.pagetriagelist.result !== 'success' || response.pagetriagelist.pages.length === 0 ) return true;
		// if new page that has been marked as reviewed
		if ( response.pagetriagelist.pages[0].patrol_status > 0 ) return true;
		// if new page that has not been marked as reviewed
		return false;
	}
````

NovemLinguae:
Fix #1573

<s>Not working yet. Because of async, form is rendered before the "Mark the page as patrolled/reviewed" checkbox is appended.

Perhaps the pagetriagelist API query needs to run before friendlytagCallback(). Any suggestions on how to set that up? I'm good at async/await, but not so good at promises and callbacks.</s>

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -59,7 +59,8 @@ Twinkle.tag.callback = function friendlytagCallback() {
 		}
 
 		if (!isReviewed) {
-			form.append({
+			// Quickform is probably already rendered. Instead of using form.append(), we need to make an element and then append it using JQuery.
+			var checkbox = new Morebits.quickForm.element({
 				type: 'checkbox',
 				list: [
 					{
@@ -70,6 +71,8 @@ Twinkle.tag.callback = function friendlytagCallback() {
 					}
 				]
 			});
+			var html = checkbox.render();
+			$('.quickform').prepend(html);
 		}
 	});
 
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -45,19 +45,33 @@ Twinkle.tag.callback = function friendlytagCallback() {
 
 	var form = new Morebits.quickForm(Twinkle.tag.callback.evaluate);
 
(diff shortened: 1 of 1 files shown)


siddharthvp:
I think the API call should not block loading of the form, to avoid visible delay before the form appears (which is more for people far away from the data centres, i.e. outside US). 

This checkbox can be added after the rest of the form has loaded by replacing `form.append` with `var checkbox = `, after which `checkbox.render()` gives the html which can be inserted into the DOM with jQuery.

PR #1676, opened 2022-12-14
twinkle: vector-2022: hovering should open menu

### LINKED ISSUE #1674: Vector 2022: Twinkle menu does not appear on hover
My apologies if this is a duplicate bug. I looked through the list but did not see this issue.

See https://phabricator.wikimedia.org/T325037 for details. There are two bugs here, possibly related:

1. The TW menu does not drop down when hovering the mouse over the menu. It works fine in Vector (legacy).

~~2. Click on the Twinkle (TW) menu, then Click on the More menu (the one with Move in it). The TW menu does not disappear when you click the More menu.~~ split into #1675

NovemLinguae:
Fix #1674

This has a side effect of also making the "More" menu stay open on hover in vector-2022. Not sure if this is desirable, or if we should change the CSS to target the TW menu only.

### DIFF
--- a/twinkle.css
+++ b/twinkle.css
@@ -11,8 +11,7 @@
  * In skin vector-2022, open the TW menu when mouse hovers over it. Credit to
  * [[en:User:Nardog]] for this fix.
  */
-#p-twinkle > .vector-menu-checkbox:hover ~ .vector-menu-content,
-#p-twinkle > .vector-menu-content:hover {
+#p-twinkle:hover > .vector-menu-content {
 	opacity: 1;
 	visibility: visible;
 	height: auto;
--- a/twinkle.css
+++ b/twinkle.css
@@ -11,8 +11,8 @@
  * In skin vector-2022, open the TW menu when mouse hovers over it. Credit to
  * [[en:User:Nardog]] for this fix.
  */
-.vector-menu-checkbox:hover ~ .vector-menu-content,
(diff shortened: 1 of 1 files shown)


nardog:
I don't think Twinkle should alter the UX of what's not Twinkle. Given the size of Twinkle users, many would mistake it's part of the skin or the site CSS, and misunderstandings among users and developers would ensue.

We *should* change the CSS to target the TW menu only if we are to hard-code it at all, but I think it's better to make this CSS a gadget on its own, thus giving users more choice and consistency. We already have a number of gadgets for "bring back X" anyway.

NovemLinguae:
Patch updated to only target the TW menu.

nardog:
Just `#p-twinkle:hover > .vector-menu-content` would do also.

NovemLinguae:
Patch is no longer working. @nardog do you have time to take a look?

NovemLinguae:
OK, I got this working but this branch/PR needs a rebase. I've never gotten a GitHub PR to rebase successfully though. If anyone has any tricks let me know. For now am going to try the following and maybe some other stuff:

```
VS Code -> checkout master
VS Code -> pull
VS Code -> checkout the PR (the VS Code GitHub app names the branch pr/NovemLinguae/1676)
git rebase origin/mastrer
git push --force origin pr/NovemLinguae/1676

Issue #1542, opened 2022-03-17
protect: rename {{pp-30-500}} to {{pp-extended}}


NovemLinguae:
https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Page_protection_template_pp-30-500

![2022-03-17_082641](https://user-images.githubusercontent.com/79697282/158834174-7f55ffc9-15b3-4e07-816f-258d38a80bac.png)

Might want to refactor some other spots too:

![image](https://user-images.githubusercontent.com/79697282/158835957-2a590481-9352-4ae7-adac-8b20c71fa6fc.png)

codesmith25103:
Sir, I want to work on this issue. Please assign me this

codesmith25103:
Sir, I had clone the git repository and run the command "npm install".  But i am not getting what to do next. Can you please guide me. I am new to open source.

NovemLinguae:
Assuming the `npm install` went smoothly, next do `npm start`, then add `mw.loader.load('http://127.0.0.1:5500')` to your User:yourUserName/common.js file on https://test.wikipedia.org/. Then make your code changes and test them out. Detailed directions at https://github.com/wikimedia-gadgets/twinkle/blob/master/CONTRIBUTING.md#getting-started

PR #1678, opened 2022-12-19
#1542 protect: rename {{pp-30-500}} to {{pp-extended}}

codesmith25103:
Sir, I have rename 'pp-30-500' to 'pp-extended'.

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -712,11 +712,11 @@ Twinkle.protect.protectionTypes = [
 	{
 		label: 'Extended confirmed protection',
 		list: [
-			{ label: 'Arbitration enforcement (ECP)', selected: true, value: 'pp-30-500-arb' },
-			{ label: 'Persistent vandalism (ECP)', value: 'pp-30-500-vandalism' },
-			{ label: 'Disruptive editing (ECP)', value: 'pp-30-500-disruptive' },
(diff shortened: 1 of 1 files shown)


azatoth (changes requested):
There's sadly a lot of unrelated changes like formatting that makes it difficult to to pinpoint what has actually been changed. If a formatting change is warranted in a PR I suggest place them in a separate commit so it can be separated.

Also I don't think the changes in `package-lock.json` is warranted.

NovemLinguae:
The package-lock.json changes are good, but should be a separate PR. I'll file a PR for that.

Did you run a code beautifier/code style fixer on these files? There are a bunch of changes unrelated to the original issue. Can you perhaps undo those changes and resubmit, to make this easier to evaluate?

codesmith25103:
Yes Sir, code formatter had given some extra spaces in code. I will try to undo it and then I will resubmit the code.

codesmith25103:
Sir, Please merge it

NovemLinguae:
Sorry this patch took so long to review.

After thinking about this, I think it'd be OK to continue using "30-500" in internal names. So I have undone that part of your patch. That should make your patch simpler/less risky/easier to review.

Looks good now. Merging. Thank you for contributing this patch.

If you do more patches for us in the future, you may want to use a git "branch". Using a master branch on a forked repo can cause problems if you submit multiple patches at the same time. So one custom branch per patch is usually a good idea.

PR #1708, opened 2023-01-26
tag: fix bug where "mark as reviewed" shows up for everyone

NovemLinguae:
PR #1667 that was just merged has a bug in it. I thought of the bug after merging it. The bug is that, in the tag module, "mark as reviewed" shows up for non-patrollers. The fix in this patch is...

```
let isPatroller = mw.config.get('wgUserGroups').some(r => ['patroller', 'sysop'].includes(r));`
if (isPatroller) {
    // do stuff
}

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -52,7 +52,8 @@ Twinkle.tag.callback = function friendlytagCallback() {
 			action: 'pagetriagelist',
 			format: 'json',
 			page_id: mw.config.get('wgArticleId')
-		}).done(function(response) {
+		}).then(function(response) {
+			// figure out whether the article is marked as reviewed in PageTriage
 			var isReviewed = false;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/friendlytag.js [hunk 1]
```diff
@@ -46,35 +46,38 @@ Twinkle.tag.callback = function friendlytagCallback() {
...
-		if (isOldPage || isMarkedAsReviewed) {
-			isReviewed = true;
-		}
+	let isPatroller = mw.config.get('wgUserGroups').some(r => ['patroller', 'sysop'].includes(r));
+	if (isPatroller) {
+		new mw.Api().get({
+			action: 'pagetriagelist',
+			format: 'json',
+			page_id: mw.config.get('wgArticleId')
+		}).done(function(response) {
```
siddharthvp, line 55: `.done` can be changed to `.then` which is more modern and compatible with ES6 promise spec.

PR #1710, opened 2023-01-26
eslint: prefer startsWith and includes

NovemLinguae:
This patch introduces eslint rules that warn for `.indexOf(a) !== -1` and `.indexOf(a) === 0` type antipatterns. These are hard to read.

Now that we're using requiresES6 in the gadget file, we should be able to use `.startsWith(a)` and `.includes(a)`. Includes is technically above ES6, but the MediaWiki minifier should be OK with it because it's not new syntax.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,6 +14,9 @@
 		"Map": "readonly"
 	},
 	"ignorePatterns": ["scripts/", "tests/", "lib/"],
+	"plugins": [
+		"lodash"
+	],
 	"rules": {
 		"no-console": "error",
 		"no-extra-parens": ["error", "all", { "nestedBinaryExpressions": false }],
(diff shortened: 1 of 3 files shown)


siddharthvp:
I like the idea of introducing an eslint rule, but here the warning message says `Prefer _.startsWith instead of comparing indexOf() to 0  lodash/prefer-startswith` which can be confusing to new contributors as we don't actually use lodash, so using [`_.startsWith`](https://docs-lodash.com/v4/starts-with/) isn't an option.

There does exist https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/prefer-string-starts-ends-with.md but that only works with TypeScript files.

NovemLinguae:
Hmm. So your concern is that the word "Iodash" appears in the name of the eslint rule? It was the easiest way to install this plugin, and in my opinion doesn't really detract from understanding the warning.

Open to alternative suggestions, but am a bit surprised at this feedback.

image

siddharthvp:
Well, it appearing in the name of the rule is not the problem. `_.startsWith` specifically refers to [the lodash function](https://docs-lodash.com/v4/starts-with/), not to the native String#startsWith(). Lodash is a pretty well-known library so an eslint rule which suggests using it where it isn't available is quite confusing.

NovemLinguae:
I don't have time to work on this one. Closing for now.

Issue #1719, opened 2023-02-13
speedy: remove F10


NovemLinguae:
Requested by Extraordinary Writ at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#F10_is_no_more

Remove CSD F10

Per RFC result

siddharthvp:
@Trialpearswiki you can write "Closes #1719" or "Fixes #1719" in the PR description or commit message so that the issue gets auto-closed when the PR is merged.

PR #1722, opened 2023-02-21
speedy: Remove speedy deletion criteria F10 

Trialpearswiki:
Removes speedy deletion criteria F10 per [RfC](https://en.wikipedia.org/wiki/Wikipedia_talk:Criteria_for_speedy_deletion#RfC:_Removing_F10._Useless_non-media_files) for issue #1719. I believe this is my first actually successful edit to twinkle, but it is tested and I could not find any other reference to F10 as well as this being basically identical to similar pull request #1043 which removed T2.


NovemLinguae:
Tested CSD tagging, works. Searched codebase for F10 and badfiletype, no results. LGTM. Thanks for the patch.

PR #1732, opened 2023-02-22
Remove A5

### LINKED ISSUE #1686: A5 criterion was repealed
Per https://en.wikipedia.org/wiki/Wikipedia_talk:Criteria_for_speedy_deletion#A5._Transwikied_articles_-_Wiktionary. Needs to be removed from the delete menu (and possibly other places)

Trialpearswiki:
Closes #1686. This is basically the same thing as the other two.


siddharthvp:
rebased to fix merge conflict

PR #1733, opened 2023-02-22
Add AI-generated tag

### LINKED ISSUE #1712: tag: add {{AI generated}}
Requested by Blaze Wolf at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Add_new_maintenance_template

Wait a few weeks to make sure it's popular enough, and to let the template name stabilize (has been moved a couple times). Then this can be added.

Trialpearswiki:
Closes #1712. The name has stabilized and the tag is enjoying significant usage in draft space so adding it definitely seems reasonable. [Tested](https://en.wikipedia.org/w/index.php?title=Draft:Test&diff=prev&oldid=1140931930) and working like expected.


NovemLinguae:
Quick tip: feel free to use https://test.wikipedia.org/ for Twinkle testing. I just keep `mw.loader.load('http://127.0.0.1:5500'); // twinkle dev` in my sandbox there permanently so I never need to change gadget settings, common.js, etc.

PR #1734, opened 2023-02-22
Fix double punctuation issues at AIV

### LINKED ISSUE #1714: arv: extra period in AIV reports
Requested by Blaze Wolf at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Period_after_diff_when_making_an_AIV_report

There's an extra period in the wikicode. Example diff: https://en.wikipedia.org/w/index.php?title=Wikipedia%3AAdministrator_intervention_against_vandalism&diff=1136480755&oldid=1136478574&diffmode=source

Does this happen when a parameter is left blank? Play around on testwiki and figure out the steps to reproduce, then fix.

Steps to reproduce
* 
* 
*

Trialpearswiki:
Closes #1714 which is about some aiv reports including ":.". This can be reproduced ([example](https://test.wikipedia.org/w/index.php?title=Wikipedia:Administrator_intervention_against_vandalism&diff=prev&oldid=559081)) by giving a diff and no type. [Here](https://test.wikipedia.org/w/index.php?title=Wikipedia:Administrator_intervention_against_vandalism&oldid=559097) are tests for all three cases, reason needing to be terminated, reason not needing to be terminated and reason being empty until the comment is added.

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -524,7 +524,11 @@ Twinkle.arv.callback.evaluate = function(e) {
 				reason += ' ' + types;
 			}
 			if (comment !== '') {
-				reason += (/([.?!;:]|^)$/.test(reason) ? '' : '.') + (reason === '' ? '' : ' ') + comment; // Ends sentence if needed, does nothing if empty string
+				var reasonNeedsPunctuation = /([.?!;:]|^)$/.test(reason);
+				reason += reasonNeedsPunctuation ? '' : '.';
+				var reasonIsBlank = reason === '';
+				reason += reasonIsBlank ? '' : ' ';
+				reason += comment;
 			}
 			reason = reason.trim();
 			if (!/[.?!;]$/.test(reason)) {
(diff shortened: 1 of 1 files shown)


modules/twinklearv.js
```diff
@@ -524,7 +524,7 @@ Twinkle.arv.callback.evaluate = function(e) {
 				reason += ' ' + types;
 			}
 			if (comment !== '') {
-				reason += (reason === '' ? '' : '. ') + comment;
+				reason += (/([.?!;:]|^)$/.test(reason) ? '' : '.') + (reason === '' ? '' : ' ') + comment; // Ends sentence if needed, does nothing if empty string
```
NovemLinguae, line 527: Two ternaries in the same sentence isn't super readable. Maybe extracting the ternaries here will make it more readable and self-documenting? Probably wouldn't need the comment after refactoring.

```js
var reasonNeedsPunctuation = /([.?!;:]|^)$/.test(reason);
reason += (reasonNeedsPunctuation ? '' : '.');
var reasonIsBlank = reason === '';
reason += (reasonIsBlank ? '' : ' ');
reason += comment;
```

NovemLinguae (changes requested):

PR #1735, opened 2023-02-24
Rmtr order

### LINKED ISSUE #1718: RM/TR addition order
As of a recent change to the addition of new requests to [WP:RM/TR](https://en.wikipedia.org/wiki/Wikipedia:Requested_moves/Technical_requests), new requests are now placed at the bottom (which makes much more sense), so <code>listAtRMTR</code> in <code>modules/twinklexfd.js</code> will need to be updated.

Trialpearswiki:
Closes  #1718. Simply a regex change to place it above the header below. If you change  Tested [here](https://test.wikipedia.org/w/index.php?title=Wikipedia:Requested_moves/Technical_requests&diff=prev&oldid=559331). The regex may seem more complicated than necessary, but it's required to avoid spacing issues if we have extra newlines which were tested in the two edits before the linked one.

### DIFF
--- a/modules/twinklexfd.js
+++ b/modules/twinklexfd.js
@@ -2016,8 +2016,8 @@ Twinkle.xfd.callbacks = {
 			var params = pageobj.getCallbackParameters();
 			var statelem = pageobj.getStatusElement();
 
-			var placementRE = /===== ?Requests to revert undiscussed moves ?=====/i;
-			var newtext = text.replace(placementRE, Twinkle.xfd.callbacks.getDiscussionWikitext('rm', params) + '\n$&');
+			var placementRE = /\n\n?(===== ?Requests to revert undiscussed moves ?=====)/i;
+			var newtext = text.replace(placementRE, Twinkle.xfd.callbacks.getDiscussionWikitext('rm', params) + '\n$1');
 			if (text === newtext) {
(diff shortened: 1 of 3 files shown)


siddharthvp:
The comments on RMTR still indicate that requests should be placed at the top of the section rather than bottom.
Screenshot 2023-04-29 at 5 04 56 PM

NovemLinguae:
* Re SD0001: I think the HTML comment is not clear about where it wants the new entry. I'd be OK with leaving the comment alone for now. I don't think it needs to be a blocker for this patch.
* I think this patch could benefit from adding a newline after new entries and before the "Controversial technical requests" header. Right now it squeezes everything together.
image
* I think any situation where you're using regexes or transforming wikicode are great candidates for a string in, string out function and unit tests. Saves a lot of manual testing.

I may code up bullets #2 and #3 if I get time today.

NovemLinguae:
I tried adding unit tests, looks hard. The code that makes the wikitext is tightly coupled to Morebits in a subfunction. Giving up on that for now.

Will revisit this later. If the \n issue (bullet number 2) can be easily solved, would appreciate a patchset for that.

NovemLinguae:
Hey @siddharthvp, can you take a look at the code in this one when you get a moment? I added some unit tests.

PR #1747, opened 2023-03-03
protect: clearer wording for edit and move protection

### LINKED ISSUE #1743: protect: clearer wording
image

I find "Edit protection: All" ambiguous. Does that mean the page is protected against all people editing it? It doesn't, it actually means the page can be edited by all.

Easy to clarify this. Just change the wording.

* Edit protection -> Who can edit
* Move protection -> Who can move

anupvs77:
Will resolve #1743


NovemLinguae:
Thanks for the patch!

Issue #1365, opened 2021-03-21
BUG FIX: Needtrans/Duflu merger needs reversion


HalHaynes:
In November 2020, @Amorymeltzer merged the "needs translation" and "needs translation cleanup" templates **needtrans** and **duflu**. I don't know what the impetus for it was, but the result is that the message being posted to the ["Pages needing translation into English"](https://en.wikipedia.org/wiki/Wikipedia:Pages_needing_translation_into_English) project page for articles written in other languages now reads the same as the message that's posted for articles in English that has been poorly translated from another language: "The initial language of this article was [language]." For articles that *need* translation, the message should be "The language of the article is [language]." The other message doesn't make sense in this context.

The commit in question is 555351cfa834c3ac7d8cced5183000ba16a92058.

siddharthvp:
@HalHaynes can you link a diff for the problematic edit(s)?

HalHaynes:
Hi, @siddharthvp . See the commit I'd provided, 555351cfa834c3ac7d8cced5183000ba16a92058, which consists of just a few deletions and additions, all within one file. Reverting it would have to be combined with reverting the Needtrans and Duflu templates. All of this was the outcome of a template discussion [here](https://en.wikipedia.org/wiki/Wikipedia:Templates_for_discussion/Log/2020_October_31#Template:Duflu), a very short one to which I contributed a peripheral comment but where I didn't pay attention to the foreseeable outcome, which is that an incorrect message is being displayed for articles needing translation.

siddharthvp:
Looking at the [template code](https://en.wikipedia.org/w/index.php?title=Template:Needtrans&action=edit) it does appear to support the intended wording, which suggests a code change here may be what's needed.

HalHaynes:
That template provides two options:

1.  If `orig` is provided, it produces, for example, "The original Portuguese article is [pt:Hello](https://pt.wikipedia.org/wiki/Hello)."
1. Otherwise, it produces, for example, "The initial language was Portuguese."

The latter is what appears now for *all* entries, though it's suited only for requests for clean-up to rough translations. The former would also apply only to clean-up to rough translations; by default, it's assumed that, given the initial language, the source article can be found among the article's interlanguage links but, if there is no link there, this is the way to display it in the  message.

[...]

PR #1752, opened 2023-03-05
BUG FIX: Use separate templates for two sections of WP:PNT

HalHaynes:
In [a discussion](https://en.wikipedia.org/wiki/Wikipedia:Templates_for_discussion/Log/2020_October_31#Template:Duflu) to which I actually contributed but, I think, without understand the full context, it was agreed a couple of years ago to merge the English Wikipedia templates Duflu and Needtrans. It seems the main issue was that there was some overlap between the Twinkle functionality used for the two main sections of [[WP:PNT]], the [project page on "Pages needing translation to English"](https://en.wikipedia.org/wiki/Wikipedia:Pages_needing_translation_into_English).

Twinkle can be used to add an article not in English to a section at [[WP:PNT]] devoted to listing such pages. When an article is added, the default text under it was supplied by substing {{needtran}}, which generated a line reading "The language is (language name/"uncertain")." Twinkle can also add an article to a listing of rough translations needing cleanup. The template used there was {{duflu}} ("dual fluency"), which, when substed, read "The language was (language name/"uncertain")." Nearly the same, but not exactly, and the difference is important.

The changes made two years ago involved changing {{needtrans}} to display "was" rather than "is", to redirect {{duflu}} to {{needtrans}}, and to have {{needtrans}} inserted in both sections of [[WP:PNT]]. The result is wrong, because now, when a page needs translation, the message recorded has "was" instead of "is".

This pull request is to restore the old separation between the two messages, so that {{duflu}}, not {{needtrans}}, is substed for rough translations. Once this goes live, the following steps will be taken on the English Wikipedia end:
- Needtrans will need its text reset on English Wikipedia to read "The language _is_ ...".
- Duflu will need to be restored to its old text rather than redirecting.

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1382,14 +1382,17 @@ Twinkle.tag.callbacks = {
 						var lang = params.translationLanguage;
 						var reason = params.translationComments;
 
-						var templateText = '{{subst:needtrans|pg=' + Morebits.pageNameNorm + '|Language=' +
-							(lang || 'uncertain') + '|Comments=' + reason.trim() + '}} ~~~~';
+						var templateText;
 
 						var text, summary;
 						if (params.tags.indexOf('Rough translation') !== -1) {
+							templateText = '{{subst:duflu|pg=' + Morebits.pageNameNorm + '|Language=' +
+							(lang || 'uncertain') + '|Comments=' + reason.trim() + '}} ~~~~';
 							text = old_text + '\n\n' + templateText;
 							summary = 'Translation cleanup requested on ';
 						} else {
+							templateText = '{{subst:needtrans|pg=' + Morebits.pageNameNorm + '|Language=' +
+							(lang || 'uncertain') + '|Comments=' + reason.trim() + '}} ~~~~';
 							text = old_text.replace(/\n+(==\s?Translated pages that could still use some cleanup\s?==)/,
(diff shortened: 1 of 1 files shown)


NovemLinguae:
Fix #1365

HalHaynes:
FWIW, for further background, this was discussed on the [[WP:PNT]] talk page: https://en.wikipedia.org/wiki/Wikipedia_talk:Pages_needing_translation_into_English#Proposal_to_undo_merger_of_Duflu_and_Needtrans

HalHaynes:
It's been four months since I posted this bug fix. Is there any way to get it moved up for consideration or is this just the way it is right now? Or is it stuck because I neglected to set something on my end?

NovemLinguae:
Sorry for the delay. I'm the main PR reviewer and I get busy sometimes. Will try to test and review today.

NovemLinguae:
+1. Tested, works.

@siddharthvp, this patch would basically reverse [this TFD that you created](https://en.wikipedia.org/wiki/Wikipedia:Templates_for_discussion/Log/2020_October_31#Template:Duflu) and that closed as merge. Any objections?

Fixing the template (and possibly patching Twinkle to pass a parameter to the fixed template) would be another option, but would probably be more brainpower than just hitting the merge button on this already written patch, then hitting the revert button on the two templates.

PR #1762, opened 2023-03-22
protect: add generic ECP option to protection module

### LINKED ISSUE #1635: Add "generic (ECP)" to protection module
Since pages sometimes get extended confirmed protected for reasons that don't fit the dropdown menu. The specific case that inspired this was Module:Split article, which I wanted to request ECP of to match the template it implements.

anupvs77:
Will resolve #1635

### DIFF
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -712,6 +712,7 @@ Twinkle.protect.protectionTypes = [
 	{
 		label: 'Extended confirmed protection',
 		list: [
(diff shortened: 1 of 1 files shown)


NovemLinguae:
Generic (semi), a similar preset, doesn't set any move protection. I'll upload a patch to remove setting move protection for Generic (ECP), then I think this will be good to merge.

NovemLinguae:
Actually, you are right, all the other ECPs set move protection. Never mind my last comment.

PR #1768, opened 2023-04-11
xfd: fix "undefined" error in userspace log XfD CfDS edit summary

NovemLinguae:
Reported by LaundryPizza03 at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Incorrect_link_in_CfDS_edit_summary

Fixes a regression introduced in #1684


NovemLinguae:
The above patch also fixes a bug with Requested Moves edit summaries.

PR #1779, opened 2023-05-20
Fix jest error message on localhost

NovemLinguae:
* Installs library jest-environment-jsdom
* Bumps Jest from v27 to v29

Before this patch, when I tried to run `npm test` on localhost, I was getting the error "Test environment jest-environment-jsdom cannot be found. Make sure the testEnvironment configuration option points to an existing node module."

Stack Overflow and the error message suggest that installing the node package "jest-environment-jsdom" is the fix.

After installing this, the tests run on localhost, but a bunch of them fail. I plan to address that in a follow up patch.

This will probably make all CI fail for all pull requests on this repo until #1780 is fixed. I think that's worth it, as it may motivate people to fix 1780, which I've tried a couple times to fix over the last year and gotten stuck on.

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -1082,6 +1082,10 @@ Twinkle.warn.messages = {
 			label: 'Posting at the top of talk pages',
 			summary: 'Notice: Posting at the top of talk pages'
 		},
+		'uw-translation': {
+			label: 'Adding translations without proper attribution',
+			summary: 'Notice: Attribution required when translating articles'
+		},
 		'uw-unattribcc': {
 			label: 'Copying from compatibly-licensed sources without attribution',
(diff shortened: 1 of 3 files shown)


NovemLinguae:
Changed my mind. We should probably wait to merge this one until we get it completely working. The list of things we still need to fix is at https://github.com/wikimedia-gadgets/twinkle/issues/1780#issuecomment-1556011455

siddharthvp:
mock-mediawiki is using MW js files from [Oct 2021](https://github.com/wikimedia/mediawiki/tree/7c895219365380e9a2c939c2c16348c8b60ac8ba) - we should start by updating those first.

NovemLinguae:
will combine into one patch where CI passes

Issue #1780, opened 2023-05-20
Many unit test failures on localhost


NovemLinguae:
Tests didn't run at all on Jest 27.

On Jest 29, there are a couple of failures and tons of notices. This ticket is to clean up all the failures and all the notices.

```
> test
> jest


  ●  Cannot log after tests are done. Did you forget to wait for something async in your test?
    Attempted to log "Use of "addOnloadHook" is deprecated. Use jQuery instead.".

      at console.warn (node_modules/@jest/console/build/BufferedConsole.js:191:10)
      at maybeLog (node_modules/mock-mediawiki/lib/mediawiki.base/log.js:52:11)
      at Window.get [as addOnloadHook] (node_modules/mock-mediawiki/lib/mediawiki.base/log.js:68:5)
          at Array.forEach (<anonymous>)


  ●  Cannot log after tests are done. Did you forget to wait for something async in your test?
    Attempted to log "Use of "loadedScripts" is deprecated. Use mw.loader instead.".

      at console.warn (node_modules/@jest/console/build/BufferedConsole.js:191:10)
      at maybeLog (node_modules/mock-mediawiki/lib/mediawiki.base/log.js:52:11)
      at Window.get [as loadedScripts] (node_modules/mock-mediawiki/lib/mediawiki.base/log.js:68:5)
          at Array.forEach (<anonymous>)


  ●  Cannot log after tests are done. Did you forget to wait for something async in your test?
    Attempted to log "Use of "importScriptURI" is deprecated. Use mw.loader instead.".

      at console.warn (node_modules/@jest/console/build/BufferedConsole.js:191:10)
      at maybeLog (node_modules/mock-mediawiki/lib/mediawiki.base/log.js:52:11)
      at Window.get [as importScriptURI] (node_modules/mock-mediawiki/lib/mediawiki.base/log.js:68:5)
          at Array.forEach (<anonymous>)


  ●  Cannot log after tests are done. Did you forget to wait for something async in your test?
    Attempted to log "Use of "importStylesheetURI" is deprecated. Use mw.loader instead.".

      at console.warn (node_modules/@jest/console/build/BufferedConsole.js:191:10)
      at maybeLog (node_modules/mock-mediawiki/lib/mediawiki.base/log.js:52:11)
      at Window.get [as importStylesheetURI] (node_modules/mock-mediawiki/lib/mediawiki.base/log.js:68:5)
          at Array.forEach (<anonymous>)

[...]

PR #1783, opened 2023-05-21
jest: fix snapshot tests

NovemLinguae:
Related #1780

This patch should be merged after #1779. 1779 bumps the version of jest, creating the snapshot error. This patch fixes the snapshot error.


NovemLinguae:
will combine into one patch where CI passes

PR #1797, opened 2023-06-12
added {{uw-blar}} and {{uw-articletodraft}}

Monik2002:
added {{uw-blar}} and {{uw-articletodraft}}

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -943,6 +943,10 @@ Twinkle.warn.messages = {
 			label: 'Bad AIV report',
 			summary: 'Notice: Bad AIV report'
 		},
+		'uw-articletodraft': {
+			label: 'Article moved to draftspace',
+			summary: 'Notice: Article moved to draftspace'
+		},
 		'uw-autobiography': {
 			label: 'Creating autobiographies',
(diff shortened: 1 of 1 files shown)


modules/twinklewarn.js
```diff
@@ -1226,7 +1226,9 @@ Twinkle.warn.messages = {
 		'uw-userpage': {
 			label: 'Userpage or subpage is against policy',
 			summary: 'Warning: Userpage or subpage is against policy'
-		}
+		},
+		'uw-blar': {},
```
NovemLinguae, line 1230: Let's also fill in the braces with at least label and summary. For label, please read the templates onwiki (e.g. https://en.wikipedia.org/wiki/Template:Uw-blar ) to see what a good summary is. For summary, please do the same and also copy the style of neighboring code (e.g. start the summary with "Notice:"

NovemLinguae (changes requested):

NovemLinguae:
I fixed the two issues above, but it looks like both these templates also have a mandatory second parameter. Will see if I can add code to collect and insert that.

image

NovemLinguae:
I decided to fix the templates instead. They shouldn't be outputting `[[{{{2}}}]]` like that.

PR #1818, opened 2023-07-12
warn: allow user to preview/change edit summary before warning

### LINKED ISSUE #1803: warn: make edit summary modifiable via a text box
Requested by Sheep8144402 at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Custom_edit_summaries

When warning a user, edit summaries are currently hard coded. The idea of this ticket is to put these in a text box in the form window, and populate it with the default edit summary, then give the user a chance to modify it before pressing "Submit".

m-etroo:
Upon opening the warning dialog, the user has the option to modify the edit summary that will eventually be used.

Currently, the edit summary (including any edits by the user) will be replaced entirely when a new warning is selected or when the linked page is changed. Not sure if this needs to be changed down the line to try to preserve the user's edits.

Resolves #1803 

image

### DIFF
--- a/modules/twinklewarn.js
+++ b/modules/twinklewarn.js
@@ -106,7 +106,7 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 		label: 'Linked page',
 		value: mw.util.getParamValue('vanarticle') || '',
 		tooltip: 'A page can be linked within the notice, perhaps because it was a revert to said page that dispatched this notice. Leave empty for no page to be linked.',
-		event: Twinkle.warn.callback.set_edit_summary
+		event: Twinkle.warn.callback.set_tentative_edit_summary
 	});
 
 	form.append({
@@ -122,8 +122,8 @@ Twinkle.warn.callback = function twinklewarnCallback() {
 		type: 'input',
 		label: 'Edit summary:',
 		name: 'summary',
-		size: 60, tooltip:
-		'Customize the warning edit\'s edit summary. This will be reset to the default when any warning options are changed.'
+		size: 60,
+		tooltip: 'Customize the warning edit\'s edit summary. This will be reset to the default when any warning options are changed.'
 	});
 
 	more.append({ type: 'textarea', label: 'Optional message:', name: 'reason', tooltip: 'Perhaps a reason, or that a more detailed notice must be appended' });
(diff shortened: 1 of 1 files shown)


NovemLinguae (changes requested):
A pretty complex first patch. Nice work.

Can we delete the `<legend>Warning information</legend>`? I don't think it makes much sense to have linked page outside of it, and edit summary, optional message, and preview inside of it.

I think we can also move "edit summary" to above "linked page". Visually I'd like linked page and optional message grouped, since those both get turned off sometimes. Would be weird to have edit summary in the middle of that.

modules/twinklewarn.js
```diff
@@ -1479,6 +1488,79 @@ Twinkle.warn.callback.postCategoryCleanup = function twinklewarnCallbackPostCate
 	}
 };
 
+// Build the expected edit summary to allow user to make edits through a text box
+Twinkle.warn.callback.set_tentative_edit_summary = function twinkleWarnCallbackSetTentativeEditSummary(e) {
+	var selected_main_group = e.target.form.main_group.value;
+	var selected_template = e.target.form.sub_group.value;
+	var messageData = $(e.target.form.sub_group).find('option[value="' + $(e.target.form.sub_group).val() + '"]').data('messageData');
```
siddharthvp, line 1495: There has to be a less error-prone (what if the value contains special chars?) and more elegant way to do this. 

Maybe something like `$(e.target.form.sub_group).find('option:selected')` ?

siddharthvp (commented):
I'd put the edit summary at the bottom of the form, since that seems the least important input here.

NovemLinguae:
Hi @m-etroo. We have a couple comments above, and this needs a rebase due to merge conflict. If you have time, would be great if you could take a look. I'm trying to clear the PR queue this week :)

siddharthvp:
Is there value in this feature? Edit summary is not important for talk page messages. If there is something you want to customize, it's probably to better to do it in the message itself than in the summary.  

This adds clutter to the dialog for something which I think is unlikely at all.

siddharthvp:
I don't think anyone has actually requested this. #1803 cites a thread where someone who said "I don't use TW much" was just asking if it's possible.

NovemLinguae:
There were some objections to this in https://github.com/wikimedia-gadgets/twinkle/pull/1818#issuecomment-2480943626, so I am closing as wontfix for now. Anyone can reopen if additional people would like this feature.

PR #1836, opened 2023-07-26
twinkle.js: group all portlet code in a class

NovemLinguae:
Moved away from storing skin variables in the preferences system, which is confusing, makes them seem like they're configurable at Twinkle/Preferences, but they are not.

Also renamed a variable, and deleted some variables that weren't used.

Later, will refactor that class a bit more so that we can add some unit tests. For example, `document` should probably be passed in through the constructor so it can be mocked.

What is `$.collapsibleTabs`? Can that be deleted? Doesn't look like it's ever set.

### DIFF
--- a/modules/friendlyshared.js
+++ b/modules/friendlyshared.js
@@ -15,7 +15,7 @@
 Twinkle.shared = function friendlyshared() {
 	if (mw.config.get('wgNamespaceNumber') === 3 && mw.util.isIPAddress(mw.config.get('wgTitle'))) {
 		var username = mw.config.get('wgRelevantUserName');
-		Twinkle.addPortletLink(function() {
+		Twinkle.HTMLGenerator.addPortletLink(function() {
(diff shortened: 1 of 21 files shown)


NovemLinguae:
My searches indicate that 6 users are using the hidden skin preferences, one of which is probably a developer's test account. Not ideal to disrupt user's preferences, but I think it's acceptable since this patch will enable further refactoring that will result in the portlet code being much more maintainable.

NovemLinguae:
This patch is kind of big, and I have some reservations about introducing ES6 classes when the rest of the code is still on ES5 style classes. I think I'll abandon this patch and then submit a smaller patch that just eliminates portlet preferences.

siddharthvp:
> ... I have some reservations about introducing ES6 classes when the rest of the code is still on ES5 style classes.

I think that's fine, as we have to start somewhere. Can't refactor the whole codebase to ES6 in one go.

PR #1878, opened 2023-10-26
tag: {{R to scientific name}}: add support for picking type

### LINKED ISSUE #1593: tag: {{R to scientific name}}: add support for picking type
This template accepts a first parameter that specifies the type of organism. For example {{R to scientific name|plant}}

List of types: https://en.wikipedia.org/wiki/Template:R_to_scientific_name#Usage

Could implement this with a combo box

anupvs77:
Will resolve #1593

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1042,8 +1042,28 @@ Twinkle.tag.redirectList = {
 			{ tag: 'R from gene symbol', description: 'redirect from a Human Genome Organisation (HUGO) symbol for a gene to an article about the gene', restriction: 'insideMainspaceOnly' }
 		],
 		Organisms: [
-			{ tag: 'R to scientific name', description: 'redirect from the common name to the scientific name', restriction: 'insideMainspaceOnly' },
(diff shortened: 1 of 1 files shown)


modules/friendlytag.js
```diff
@@ -1042,8 +1042,28 @@ Twinkle.tag.redirectList = {
...
 		],
 		Organisms: [
-			{ tag: 'R to scientific name', description: 'redirect from the common name to the scientific name', restriction: 'insideMainspaceOnly' },
-			{ tag: 'R from scientific name', description: 'redirect from the scientific name to the common name', restriction: 'insideMainspaceOnly' },
+			{
+				tag: 'R to scientific name',
+				description: 'redirect from the common name to the scientific name',
+				subgroup: {
+					name: 'toScientificName',
+					type: 'input',
```
NovemLinguae, line 1050: could we do this as a drop-down list instead? https://en.wikipedia.org/wiki/Drop-down_list

the morebits library probably has something like this, although it may be called something else.

can the drop-down list preload these values? https://en.wikipedia.org/wiki/Template:R_to_scientific_name#Usage

NovemLinguae (changes requested):

Issue #1835, opened 2023-07-26
twinkle: switch to mw.util.addPortlet() function


NovemLinguae:
When `mw.util.addPortlet()` is finished and deployed, replace a lot of our `Twinkle.addPortlet()` code with the new `mw.util.addPortlet()`.

This will pay off some technical debt and make Twinkle less likely to break when changes are made to Vector 2022.

https://phabricator.wikimedia.org/T303488

## Skins we need to support

Twinkle has special cases in the code for vector-2022, vector, modern, and timeless. Twinkle probably also needs to support monobook. Twinkle does not need to support Minerva (there is a different Twinkle for mobile users).

## Existing bugs in the `Twinkle.addPortlet()` code

Twinkle's current `Twinkle.addPortlet()` code is not perfect. For example, the portlet code generated for Timeless doesn't match other nearby portlet code, suggesting we are using the wrong HTML here.

<img width="581" alt="image" src="https://github.com/wikimedia-gadgets/twinkle/assets/79697282/dbf22e68-f5b2-4df6-a792-7bb38f085f15">

## Complexity

The `Twinkle.addPortlet()` code at first glance looks like it lets the user adjust a bunch of settings, because it takes 5 parameters, and the parameters come from `getPref()`...

```js
/**
 * @param {String} navigation id of the target navigation area (skin dependant, on vector either of "left-navigation", "right-navigation", or "mw-panel")
 * @param {String} id id of the portlet menu to create, preferably start with "p-".
 * @param {String} text name of the portlet menu to create. Visibility depends on the class used.
 * @param {String} type type of portlet. Currently only used for the vector non-sidebar portlets, pass "menu" to make this portlet a drop down menu.
 * @param {Node} nextnodeid the id of the node before which the new item should be added, should be another item in the same list, or undefined to place it at the end.
 */
Twinkle.addPortlet(
    Twinkle.getPref('portletArea'),
    Twinkle.getPref('portletId'),
    Twinkle.getPref('portletName'),
    Twinkle.getPref('portletType'),
    Twinkle.getPref('portletNext')
);
```

But in reality these are hard-coded in twinkle.js. I don't see the word "portlet" or "skin" anywhere on the [Twinkle preferences page](https://en.wikipedia.org/wiki/Wikipedia:Twinkle/Preferences).

[...]

PR #1881, opened 2023-11-02
Twinkle uses new mw.util.addPortlet API

jdlrobson:
Fixes: #1835


twinkle.js
```diff
@@ -262,122 +262,10 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
...
-		var innerDiv = document.createElement('div');
-		innerDiv.className = innerDivClass;
-		innerDiv.appendChild(ul);
-		outerNav.appendChild(innerDiv);
-	} else {
-		outerNav.appendChild(ul);
-	}
-
-	return outerNav;
+	return mw.util.addPortlet(id, text, '#' + nextnodeid);
```
NovemLinguae, line 268: With this patch, the Twinkle menu has moved from the right side of the Tools/More menu, to the left side. Is that an easy fix to make it show up on the right side?

Old:
image

New:
image

NovemLinguae (changes requested):

NovemLinguae:
Manual testing was successful on all the skins Twinkle supports: vector, vector-2022, timeless, modern, monobook, cologneblue. Can probably merge this after the one code review comment above is addressed.

twinkle.js
jdlrobson: This is one of the constraints the Vector API has - for dropdowns you can only inject in front of the #p-cactions element - it's how we signal you want a dropdown.  In Vector 2022 it was also intentional given the pinnable functionality - the dropdown for more menu should be the furthest right since it can be moved to the right hand side. 

You could add additional code to move the portlet after it's inserted if you are okay with this complexity like so:
```
const portlet = mw.util.addPortlet(id, text, '#' + nextnodeid);
$(portlet).insertAfter('#p-cactions')
```

NovemLinguae: I just tested it out. `$(portlet).insertAfter('#p-cactions')` only works in Vector 2010. Would need separate code for V22 (`$(portlet).insertAfter('.vector-page-tools-landmark')` maybe), which combined with the conditional logic introduces a lot of complexity to this otherwise simple patch. Let's just keep it on the left for now and hopefully users are OK with it.

NovemLinguae (approved):

PR #1887, opened 2023-11-04
twinkle: Put TW menu back on the right side of the More menu

### LINKED ISSUE #1886: Put TW menu back on the right side of the More menu
In #1881 where we did some code simplification, the TW menu moved from the right of the More/Tools menu to the left.

At the time I decided it wasn't worth the code complexity to put it on the right.

I think I have changed my mind. Because now I am noticing that putting it on the left causes the More menu to shift about one second after page loading.

There's also muscle memory issues of people used to TW being on the right. I encountered this myself when I went to click on something.

NovemLinguae:
Fixes #1886
Related #1881

Menus are jumping around too much during page load with TW menu on the left. Need to move it back to its spot on the right.

The spot on the right is reserved for it in https://en.wikipedia.org/wiki/MediaWiki:Gadget-Twinkle.css

### DIFF
--- a/twinkle.js
+++ b/twinkle.js
@@ -271,6 +271,7 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
 	var portlet = mw.util.addPortlet(id, text, '#' + nextnodeid);
 
 	if (mw.config.get('skin') === 'vector') {
+		// The Twinkle dropdown menu has been added to the left of p-cactions. Move it to the right.
 		$('#p-twinkle').insertAfter('#p-cactions');
 	} else if (mw.config.get('skin') === 'vector-2022') {
 		$('#p-twinkle-dropdown').insertAfter('.vector-page-tools-landmark');
--- a/twinkle.js
+++ b/twinkle.js
@@ -274,10 +274,7 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
 		// The Twinkle dropdown menu has been added to the left of p-cactions. Move it to the right.
 		$('#p-twinkle').insertAfter('#p-cactions');
 	} else if (mw.config.get('skin') === 'vector-2022') {
-		$('#p-twinkle-dropdown').insertAfter('.vector-page-tools-landmark');
-		// this creates two #p-twinkle-dropdowns for some reason, in different spots on the DOM. maybe some JS in vector 2022? delete the bad one:
-		$('.vector-column-end > #p-twinkle-dropdown').remove();
-		portlet = $('#p-twinkle-dropdown');
+		return $('#p-twinkle-dropdown').detach().insertAfter('.vector-page-tools-landmark');
 	}
 
 	return portlet;
--- a/twinkle.js
+++ b/twinkle.js
@@ -268,7 +268,7 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
 		nextnodeid = 'p-cactions';
 	}
 
-	let portlet = mw.util.addPortlet(id, text, '#' + nextnodeid);
+	var portlet = mw.util.addPortlet(id, text, '#' + nextnodeid);
 
 	if (mw.config.get('skin') === 'vector') {
 		$('#p-twinkle').insertAfter('#p-cactions');
--- a/twinkle.js
+++ b/twinkle.js
@@ -262,11 +262,24 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
 		}
 		return null;
 	}
+
 	if (type === 'menu') {
 		// In order to get mw.util.addPortlet to generate a dropdown menu in vector and vector-2022, the nextnodeid must be p-cactions. Any other nextnodeid will generate a non-dropdown portlet instead.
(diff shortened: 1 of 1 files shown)


NovemLinguae:
@jdlrobson. Can you code review this one?

twinkle.js
```diff
@@ -262,11 +262,24 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
...
 		return null;
 	}
+
 	if (type === 'menu') {
 		// In order to get mw.util.addPortlet to generate a dropdown menu in vector and vector-2022, the nextnodeid must be p-cactions. Any other nextnodeid will generate a non-dropdown portlet instead.
 		nextnodeid = 'p-cactions';
 	}
-	return mw.util.addPortlet(id, text, '#' + nextnodeid);
+
+	let portlet = mw.util.addPortlet(id, text, '#' + nextnodeid);
```
jdlrobson, line 271: Should this be var?

Does this code get transpiled before it's deployed to wikis?

I saw you are using let but not template syntax  (`#{nextnodeid}` ) and given the rest of the file uses `var` I'm not sure which syntax this should be written in.

twinkle.js
```diff
@@ -262,11 +262,24 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
...
+
+	if (mw.config.get('skin') === 'vector') {
+		$('#p-twinkle').insertAfter('#p-cactions');
+	} else if (mw.config.get('skin') === 'vector-2022') {
+		$('#p-twinkle-dropdown').insertAfter('.vector-page-tools-landmark');
```
jdlrobson, line 276: `vector-page-tools-landmark` is not stable so may change at any point. If that's the case I think jQuery will quietly ignore this code but not move it. With that in mind would it make sense to check and log that situation to aid debuggin?

```
if ( !document.querySelector( '.vector-page-tools-landmark' ) ) {
  mw.log.warn('Unexpected change in DOM' );
}

File: twinkle.js [hunk 3]
```diff
@@ -262,11 +262,24 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
...
+		// this creates two #p-twinkle-dropdowns for some reason, in different spots on the DOM. maybe some JS in vector 2022? delete the bad one:
+		$('.vector-column-end > #p-twinkle-dropdown').remove();
```
jdlrobson, line 278: .vector-column-end is also not considered stable. Ideally we'd be able to avoid this line if we can understand it better.

twinkle.js
```diff
@@ -262,11 +262,24 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
...
 	}
-	return mw.util.addPortlet(id, text, '#' + nextnodeid);
+
+	let portlet = mw.util.addPortlet(id, text, '#' + nextnodeid);
+
+	if (mw.config.get('skin') === 'vector') {
+		$('#p-twinkle').insertAfter('#p-cactions');
+	} else if (mw.config.get('skin') === 'vector-2022') {
+		$('#p-twinkle-dropdown').insertAfter('.vector-page-tools-landmark');
+		// this creates two #p-twinkle-dropdowns for some reason, in different spots on the DOM. maybe some JS in vector 2022? delete the bad one:
```
jdlrobson, line 277: I'm curious about why you are getting two p-twinkle-dropdowns  - could you provide some replication instructions? That doesn't seem expected.

Perhaps try detaching the node first before appending to avoid the need for this?

twinkle.js
```diff
@@ -262,11 +262,24 @@ Twinkle.addPortlet = function(navigation, id, text, type, nextnodeid) {
...
 	if (type === 'menu') {
 		// In order to get mw.util.addPortlet to generate a dropdown menu in vector and vector-2022, the nextnodeid must be p-cactions. Any other nextnodeid will generate a non-dropdown portlet instead.
 		nextnodeid = 'p-cactions';
 	}
-	return mw.util.addPortlet(id, text, '#' + nextnodeid);
+
+	let portlet = mw.util.addPortlet(id, text, '#' + nextnodeid);
+
+	if (mw.config.get('skin') === 'vector') {
+		$('#p-twinkle').insertAfter('#p-cactions');
```
jdlrobson, line 274: Suggest adding a comment to explain what's happening here
e.g.
```
// twinkle has been added /before/ p-cactions, reinsert it after the node.

**Review by jdlrobson — COMMENTED** [review]
Is the plan to merge this before the deployment or is the plan to see if people care about this enough first? 

Another solution that did occur to me is a CSS only solution like so:
```
#p-twinkle { order: 2; } 
```
This might be easier to maintain?

twinkle.js
NovemLinguae: We don't transpile. Went ahead and changed to var for consistency.

twinkle.js
NovemLinguae: Detach fixed it. Thank you for that tip!

NovemLinguae: Just kidding, detach didn't fix it. Hmm.

NovemLinguae: OK, fixed it by switching `insertAfter()` to `appendTo()`

PR #1888, opened 2023-11-11
Support batch requested moves

### LINKED ISSUE #1681: Support multi-page requested moves
https://en.wikipedia.org/wiki/Wikipedia:Requested_moves#Requesting_multiple_page_moves
https://en.wikipedia.org/wiki/Template:Requested_move#Multiple_related_move_requests

The interface for ARV -> Sockpuppeteer is probably a good reference.

nardog:
Fixes #1681

### DIFF
--- a/morebits.js
+++ b/morebits.js
@@ -353,7 +353,7 @@ Morebits.quickForm.prototype.append = function QuickFormAppend(data) {
  *  - `number`: A number input box.
  *      - Attributes: Everything the text `input` has, as well as: min, max, step, list
  *  - `dyninput`: A set of text boxes with "Remove" buttons and an "Add" button.
- *      - Attributes: name, label, min, max, sublabel, value, size, maxlength, event
+ *      - Attributes: name, label, min, max, inputs, sublabel, value, size, maxlength, event
  *  - `hidden`: An invisible form field.
  *      - Attributes: name, value
  *  - `header`: A level 5 header.
@@ -371,7 +371,7 @@ Morebits.quickForm.prototype.append = function QuickFormAppend(data) {
  * There is some difference on how types handle the `label` attribute:
  * - `div`, `select`, `field`, `checkbox`/`radio`, `input`, `textarea`, `header`, and `dyninput` can accept an array of items,
  * and the label item(s) can be `Element`s.
- * - `option`, `optgroup`, `_dyninput_element`, `submit`, and `button` accept only a single string.
+ * - `option`, `optgroup`, `_dyninput_cell`, `submit`, and `button` accept only a single string.
  *
  * @memberof Morebits.quickForm
  * @class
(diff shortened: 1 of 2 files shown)


NovemLinguae:
Before:


After:

morebits.js
```diff
@@ -371,7 +371,7 @@ Morebits.quickForm.prototype.append = function QuickFormAppend(data) {
  * There is some difference on how types handle the `label` attribute:
  * - `div`, `select`, `field`, `checkbox`/`radio`, `input`, `textarea`, `header`, and `dyninput` can accept an array of items,
  * and the label item(s) can be `Element`s.
- * - `option`, `optgroup`, `_dyninput_element`, `submit`, and `button` accept only a single string.
+ * - `option`, `optgroup`, `_dyninput_cell`, `submit`, and `button` accept only a single string.
```
NovemLinguae, line 374: For my understanding, what does _dyninput_cell do and why did its name change?

morebits.js
```diff
@@ -730,14 +730,17 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
...
-				name: data.name,
-				value: data.value,
-				size: data.size,
+				type: '_dyninput_row',
 				remove: false,
 				maxlength: data.maxlength,
-				event: data.event
+				event: data.event,
+				inputs: data.inputs || [{
+					// compatibility
```
NovemLinguae, line 738: For my understanding, what's going on with the compatibility sub-section here?

morebits.js
```diff
@@ -800,6 +785,41 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 				removeButton.morebutton = data.morebutton;
 			}
 			break;
+		case '_dyninput_cell': // Private, similar to normal input
```
NovemLinguae, line 788: for my understanding, why did this case move down?

NovemLinguae:
Hi @nardog. I made a couple comments above. If you have time, would be great if you could take a look. I'm trying to clear the PR queue this week :)

morebits.js
nardog: So other scripts using `dyninput` will continue to work as they did. We don't have to do that for `_dyninput_element`, which is private.

morebits.js
nardog: It didn't move down, just renamed.

morebits.js
nardog: `_dyninput_element` was an internal type used by `dyninput`, and is now renamed `_dyninput_cell` because there's now `_dyninput_row`.

morebits.js
nardog: `_dyninput_element` was an internal type used by `dyninput`. So in the code `dyninput` was followed by `_dyninput_element`. Now `_dyninput_cell` is an internal type used by `_dyninput_row`, which is in turn used by `dyninput`. So in the code `dyninput` is followed by `_dyninput_row`, which is followed by `_dyninput_cell`.

NovemLinguae:
Note to self: the Morebits changes are so that dyninput can support more than 1 text box per row.

I found a behavior change between the old patch and the new patch. Before if you did an RM on a talk page, it'd assume you were talking about the non-talk page it is paired with. Now it assumes you are talking about the talk page. I may push a commit to this PR to change this back.

nardog:
> I found a behavior change between the old patch and the new patch. Before if you did an RM on a talk page, it'd assume you were talking about the non-talk page it is paired with. Now it assumes you are talking about the talk page. I may push a commit to this PR to change this back.

With the status quo you couldn't ever request a move of a talk page (even for RMTR). This patch fixes this, but I wouldn't object to defaulting to the subject page.

PR #1900, opened 2023-12-02
Remove IE 11 compatibility code

siddharthvp:
IE 11 is no longer supported, by Twinkle or by MediaWiki, for a while now.

### DIFF
--- a/modules/twinklefluff.js
+++ b/modules/twinklefluff.js
@@ -348,7 +348,6 @@ Twinkle.fluff.addLinks = {
 };
 
 Twinkle.fluff.disableLinks = function disablelinks(parentNode) {
-	// Array.from not available in IE11 :(
 	$(parentNode).children().each(function(_ix, node) {
 		node.innerHTML = node.textContent; // Feels like cheating
 		$(node).css('font-weight', 'normal').css('color', 'darkgray');
--- a/morebits.js
+++ b/morebits.js
@@ -2124,7 +2124,7 @@ Morebits.date.prototype = {
 		level = parseInt(level, 10);
 		level = isNaN(level) ? 2 : level;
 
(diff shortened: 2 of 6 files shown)


modules/twinklefluff.js
```diff
@@ -348,8 +348,7 @@ Twinkle.fluff.addLinks = {
 };
 
 Twinkle.fluff.disableLinks = function disablelinks(parentNode) {
-	// Array.from not available in IE11 :(
-	$(parentNode).children().each(function(_ix, node) {
+	Array.from($(parentNode).children()).forEach(function (node) {
```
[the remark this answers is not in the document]
siddharthvp: I think the intent was to avoid jQuery each() which is confusing as it takes the index before the element. Revereted it now as I see there are lots of jQuery each() uses in the codebase.

NovemLinguae:
@siddharthvp , did you do any manual testing when you wrote this code? Will determine how much testing I do on my end.

siddharthvp:
Yes, I tested this.

morebits.js
```diff
@@ -2124,7 +2124,7 @@ Morebits.date.prototype = {
 		level = parseInt(level, 10);
 		level = isNaN(level) ? 2 : level;
 
-		var header = Array(level + 1).join('='); // String.prototype.repeat not supported in IE 11
+		var header = '='.repeat(level);
```
NovemLinguae, line 2127: Did the +1 get dropped here by mistake?

NovemLinguae (changes requested):

morebits.js
siddharthvp: no, it was just hacky code initially: to create a string with n `=` signs, you need an array of n+1 undefined elements and join them with `=`.

PR #1906, opened 2023-12-12
Update twinkle.css for Vector Zebra CSS refactor

jandre3000:
Removes hotfix for T337893. 
These extra styles for Vector are currently not necessary and cause excess padding.

Bug: T353214


NovemLinguae:
Thank you very much for the quick patch!

Issue #1907, opened 2023-12-12
PR labeler is failing CI


NovemLinguae:
Perhaps due to breaking changes upstream in the format of the .github/labels.yml file. Sounds like they recently switched to version 5.

https://github.com/actions/labeler#pull-request-labeler

https://github.com/actions/labeler/issues/712

PR #1908, opened 2023-12-13
fix PR labeler

NovemLinguae:
fix PR labeler is failing CI #1907

- upgrade to version 5 format
- start labeling .github/* files with the "tests" label, because it makes sense, and so I can test this


NovemLinguae:
@MaksimZhukov, any chance you can help me spot the bug in this one? Thanks for your time!

NovemLinguae:
I think the only way to test this is to merge it. If it doesn't work I'll revert it.

Issue #1863, opened 2023-09-07
tag: Twinkle doesn't detect {{More footnotes needed}}


NovemLinguae:
Steps to reproduce
* https://en.wikipedia.org/w/index.php?title=Montgomery_Riverfront_brawl&oldid=1174308453 (which includes the wikicode `{{More footnotes needed|date=September 2023}}`)
* TW->Tag
* uncheck {{More footnotes}}
* type "sufficient citations" for the reason
* click "submit query"

What happens?
<img width="726" alt="image" src="https://github.com/wikimedia-gadgets/twinkle/assets/79697282/81ce96e9-2bbb-4368-905f-1dfa1ebd030a">

9382:
I'll also note I've just seen #1618 which appears to be the exact same bug

PR #1960, opened 2024-04-15
tag: Fix removing tags for maintenance templates where |name= points to a redirect

### LINKED ISSUE #1618: tag: Twinkle doesn't detect {{Self-contradictory}}
Reported by LaundryPizza03 at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Self-contradictory_tag

Normally if you load a page with maintenance tags, Twinkle will pre-check the maintenance tags when you load the tag module. For {{self-contradictory}}, it does not.

This may make sense since self-contradictory is not in Twinkle's dictionary. But even after adding it to the dictionary, I was still not able to get Twinkle to detect it on this page. This could indicate some bugs within the tag module that need investigation.

https://test.wikipedia.org/w/index.php?title=Alert,_Nunavut&oldid=516643

9382:
For #1863/#1618

If a maintenance template had been moved but its `|name=` wasn't updated, the code would pick up every version of that template *except* for the current name (since obviously that won't appear in a linkshere for itself and it tries to use the `|name=` for its current name). This fixes that by ensuring that whatever page linkshere resolves to is also included in the templates it attempts to remove

Credit to #1618 for the solution, I just tested it to make sure it worked

(Testing) Fixes #1618

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -1545,7 +1545,7 @@ Twinkle.tag.callbacks = {
 				});
 				pages.forEach(function(page) {
 					var removed = false;
-					page.linkshere.forEach(function(el) {
+					page.linkshere.concat({title: page.title}).forEach(function(el) {
 						var tag = el.title.slice(9);
 						var tag_re = new RegExp('\\{\\{' + Morebits.pageNameRegex(tag) + '\\s*(\\|[^}]*)?\\}\\}\\n?');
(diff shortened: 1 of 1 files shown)


NovemLinguae:
I can't get #1618 to link. Note to self: when merging this, manually close #1618.

9382:
Managed to get it to link by editing my original comment to include some keywords

PR #1965, opened 2024-04-22
eslint: switch from custom rules to eslint-config-wikimedia

NovemLinguae:
https://github.com/wikimedia/eslint-config-wikimedia

Why switch?
- it's standard for wikimedia projects
- it has some cool features such as jquery deprecation warnings, and enforcing a specific version of JavaScript
- I believe it has many more rules than our previous configuration, which will help with identifying areas of the code to modernize

What this patch does
- delete all old linter rules, except for these rules I carried over:
	- `no-nested-ternary`
	- `no-restricted-syntax` for `.done()` and `.fail()`
- add eslint-config-wikimedia and its default rules
- disable controversial rules such as
	- whitespace in parentheses (`space-in-parens`, `space-before-function-paren`, `array-bracket-spacing`, `computed-property-spacing`, `object-curly-spacing`)
	- max line length 100 (`max-len`)
- disable some other rules
	- `indent` because it would involve touching every line, messing up git blame
	- couple others
- set it to ES6 since wikimedia grade A support is now on ES6
- warn/error for syntax or internal functions higher than ES6 (more thoroughly than the old >ES6 detection we were using)
- apply all autofixes
	- converting var to let/const (`no-var`)
	- `wrap-iife`
	- `no-multiple-empty-lines`
	- `no-multi-spaces`
	- `quote-props`
	- `unicorn/prefer-string-slice`
	- `no-unneeded-ternary`
	- `dot-location`
	- `jsdoc/require-asterisk-prefix`
	- couple others
- set all remaining detected violations to "warn" for now, until they can be fixed in future patches

All changes in this patch are related to installing and configuring eslint-config-wikimedia, or applying autofixes. No manual changes in this patch.


NovemLinguae:
I've decided I don't like the substring() -> slice(Math.max) conversion. Too verbose. Note to self: rollback those 9 auto fixes, and set `unicorn/prefer-string-slice` to off

NovemLinguae:
Note to self: I could do the "indent" autofix and add that commit to a .git-blame-ignore-revs file, to avoid spamming git blame.

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

NovemLinguae:
This patch is too big. I'm going to close and resubmit it in favor of a smaller patch, so I can do this more incrementally.

Issue #1418, opened 2021-06-17
Pre-empt edit conflicts


siddharthvp:
If the page is edited (or deleted, or in case of ARV/Block if the user is blocked, etc) while the user has a Twinkle dialog open, they should be notified of that in the UI itself as soon as possible. 

Possible approaches: 
- Poll the relevant API every 30 seconds?
- Use [EventStreams](https://wikitech.wikimedia.org/wiki/Event_Platform/EventStreams)? Probably no. This is a real-time solution but the bandwidth usage may be unacceptable as there's no server-side filtering (every edit made on any WMF wiki is streamed). Also, the [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) API isn't available in IE.

siddharthvp:
I think polling the API every 5 seconds or so is a reasonably good approach.

PR #1967, opened 2024-05-03
Show notification on new changes to pre-empt edit conflicts

siddharthvp:
Solves #1418.


twinkle.js
```diff
@@ -467,6 +467,38 @@ Twinkle.makeFindSourcesDiv = function makeSourcesDiv(divID) {
...
+		let api = new mw.Api();
+		api.get({
+			action: 'query',
+			format: 'json',
+			prop: 'revisions',
+			titles: Morebits.pageNameNorm,
+			formatversion: '2',
+			rvprop: 'ids',
+			rvslots: 'main',
+			rvstartid: mw.config.get('wgCurRevisionId') + 1,
```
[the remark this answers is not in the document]
siddharthvp: Fixed using revision timestamp instead

twinkle.js
```diff
@@ -467,6 +467,58 @@ Twinkle.makeFindSourcesDiv = function makeSourcesDiv(divID) {
...
+					rvprop: 'ids',
+					rvslots: 'main',
+					rvstart: new Date(new Date(timestamp).getTime() + 1000).toISOString(),
+					rvdir: 'newer'
+				}).then(function (response) {
+					if (checkIfChanges(response)) {
+						clearInterval(interval);
+					}
+				});
+			}, 5000);
```
NovemLinguae, line 517: Will this run every 5 seconds that the page is open, indefinitely?

If the page loses focus, does this still run?

If so, this could really hammer the servers. Sometimes I'll keep dozens of tabs open overnight. If needed, we should consider writing a back-off algorithm. For example, change the check rate to every hour after the first 5 minutes. Or only run when the tab is focused.

NovemLinguae: @siddharthvp , any thoughts on this one? I want to clear the PR queue this week :)

siddharthvp: This needs to be updated to stop polling once the Twinkle dialog has been closed. And we can include a document.hasFocus() check as well. I'll likely be able to get to this only next week :(

I think adding backoff is overcomplicating things, as people are presumably not going to keep Twinkle open for hours. Even if a few folks do, 1 read request every 5 seconds is nothing – the servers can handle a lot more than that. Besides, most browsers tend to throttle or suspend JS execution in inactive tabs.

siddharthvp: > 1 read request every 5 seconds is nothing – the servers can handle a lot more than that.

Some pages like watchlist and RecentChanges in core already do this kind of polling. So does DiscussionTools I think to show new comments on talk page? (not sure about DT as I use CD instead - which also does it btw).

NovemLinguae:
Closing as >1 year stale. Trying to clean out the PR queue. Feel free to re-open if you get time to work on it.

PR #1976, opened 2024-06-22
https://github.com/wikimedia/eslint-config-wikimedia

NovemLinguae:
Why switch?
- it's standard for wikimedia projects
- it has some cool features such as jquery deprecation warnings, and enforcing a specific version of JavaScript
- I believe it has many more rules than our previous configuration, which will help with identifying areas of the code to modernize

What this patch does
- delete all old linter rules, except for these rules I carried over:
	- `no-nested-ternary`
	- `no-restricted-syntax` for `.done()` and `.fail()`
- upgrade eslint a couple minor versions
- add eslint-config-wikimedia and its default rules
- disable controversial rules such as
	- whitespace in parentheses (`space-in-parens`, `space-before-function-paren`, `array-bracket-spacing`, `computed-property-spacing`, `object-curly-spacing`)
	- max line length 100 (`max-len`)
- disable some rules
	- `indent` because it would involve touching almost every line, messing up git blame (although I may autofix this in the future and set a .git-blame-ignore-revs to reduce git blame spam)
	- `wrap-iife` to reduce VS Code yellow underline spam (will autofix this in a future patch)
	- couple others that I don't like and have disabled in my other projects
- set it to ES6 since wikimedia "modern layer" support is now on ES6
- warn/error for syntax or internal functions higher than ES6 (this is more thorough than the old >ES6 detection we were using)
- set all remaining detected violations to "warn" for now, so that CI passes
- will apply autofixes for all the rules that "warn" in a future patch

The rules are in two groups, with an extra line break in the middle separating them.
- The top group are rules that should not be fixed
- The bottom group are rules that should be fixed over time via autofixes and manual fixing. Once everything for that rule is fixed, that rule can be deleted, so that the linter starts enforcing it in CI.

Over time, rules will be fixed and removed from the config, and the code will be incrementally modernized, until it arrives at the Wikimedia standard.


package.json
```diff
@@ -14,7 +14,8 @@
 		"deployall": "perl scripts/sync.pl --mode=deploy --all"
 	},
 	"devDependencies": {
-		"eslint": "^8.54.0",
+		"eslint": "^8.57.0",
```
NovemLinguae, line 17: Can probably just delete this. I think it's included in eslint-config-wikimedia

NovemLinguae:
I may merge this one when the PR queue is mostly empty. The follow up patches with autofoxes would otherwise create a lot of merge conflicts

PR #1985, opened 2024-07-27
Create npm-publish-github-packages.yml

JdaWstripe3:
[fafa0cb.diff.zip](https://github.com/user-attachments/files/16400870/fafa0cb.diff.zip)
[view-source_https___raw.githubusercontent.com_jaseemabid_mit-scheme_47c68ef5fa41821c3f52f886ef69f99d232d10bb_v7_src_runtime_rep.scm.txt](https://github.com/user-attachments/files/16400871/view-source_https___raw.githubusercontent.com_jaseemabid_mit-scheme_47c68ef5fa41821c3f52f886ef69f99d232d10bb_v7_src_runtime_rep.scm.txt)
            - name: Download a Build Artifact
  uses: actions/download-artifact@v4.1.8
  with:
    # Name of the artifact to download. If unspecified, all artifacts for the run are downloaded.
    name: # optional
    # Destination path. Supports basic tilde expansion. Defaults to $GITHUB_WORKSPACE
    path: # optional
    # A glob pattern matching the artifacts that should be downloaded. Ignored if name is specified.
    pattern: # optional
    # When multiple artifacts are matched, this changes the behavior of the destination directories. If true, the downloaded artifacts will be in the same directory specified by path. If false, the downloaded artifacts will be extracted into individual named directories within the specified path.
    merge-multiple: # optional, default is false
    # The GitHub token used to authenticate with the GitHub API. This is required when downloading artifacts from a different repository or from a different workflow run. If this is not specified, the action will attempt to download artifacts from the current repository and the current workflow run.
    github-token: # optional
    # The repository owner and the repository name joined together by "/". If github-token is specified, this is the repository that artifacts will be downloaded from.
    repository: # optional, default is ${{ github.repository }}
    # The id of the workflow run where the desired download artifact was uploaded from. If github-token is specified, this is the run that artifacts will be downloaded from.
    run-id: # optional, default is ${{ github.run_id }}

### DIFF
--- a/.github/workflows/npm-publish-github-packages.yml
+++ b/.github/workflows/npm-publish-github-packages.yml
@@ -1,0 +1,96 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: npm ci
+      - run: npm test
+
+  publish-gpr:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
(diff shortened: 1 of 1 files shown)


NovemLinguae:
Closing this because linking a zip file looks suspicious for viruses/phishing. If this is legit, please comment with a detailed explanation of what you're trying to do with this patch.

PR #2003, opened 2024-09-05
tag: fix multiple issues grouping

### LINKED ISSUE #1997: tag: when combining tags into {{Multiple issues}}, ignore all tags below the first ==Heading==
Reported by Fram at https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&curid=9241161&diff=1242911714&oldid=1242599754

Steps to reproduce:
* make a test page such as
```
== Heading ==
{{one source|section|date=August 2021}}
```
* TW -> Tag
* tick "use multiple issues box"
* add a tag

What happens?
* section tag is moved to top and placed in multiple issues
* Example diff showing the bug: https://en.wikipedia.org/w/index.php?title=Ceremonial_magic&diff=1242909182&oldid=1242907058

What should happen instead?
* section tag is not moved to top

TollensWP:
Fix #1997

### DIFF
--- a/modules/friendlytag.js
+++ b/modules/friendlytag.js
@@ -315,7 +315,7 @@ Twinkle.tag.callback = function friendlytagCallback() {
 
 				// break out on encountering the first heading, which means we are no
 				// longer in the lead section
-				if (e.tagName === 'H2') {
+				if (e.classList.contains('mw-heading')) {
 					return false;
 				}
 


siddharthvp:
Can you elaborate how this fixes the issue?

TollensWP:
Sure - the code stops looking for new issue tags to include in the group after it reaches the first section heading, which were previously represented in the HTML as H2 tags at the same level as the rest of the page text. At some point this was changed and the actual header tag is wrapped in other elements, so because the parsing code is only looking at direct children of the container holding the page contents, it never saw any H2 tags and assumed the whole page was one big lead section. The outermost element that is still part of a section heading is tagged with the class .mw-heading, so looking for elements with that class accomplishes the same thing as looking for H2 tags used to do.

PR #2004, opened 2024-09-05
speedy: replace db-templatecat with C4

### LINKED ISSUE #1994: Add new C4 criterion
Per https://en.wikipedia.org/wiki/Wikipedia_talk:Criteria_for_speedy_deletion/Archive_89#C4_%E2%80%93_unused_maintenance_categories

This replaces {{db-templatecat}}

TollensWP:
Resolves #1994

Also removes "G8: Redirects to non-existent targets" criterion from category CSD list, since that list doesn't display on redirects. The redirect CSD list does display on category redirects, and includes G8.

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -38,7 +38,7 @@ Twinkle.config.commonSets = {
 		a1: 'A1', a2: 'A2', a3: 'A3', a7: 'A7', a9: 'A9', a10: 'A10', a11: 'A11',
 		u1: 'U1', u2: 'U2', u5: 'U5',
 		f1: 'F1', f2: 'F2', f3: 'F3', f7: 'F7', f8: 'F8', f9: 'F9',
-		c1: 'C1',
+		c1: 'C1', c4: 'C4',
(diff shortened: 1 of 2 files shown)


TollensWP:
I should also note that I've excluded C4 from the set of templates which allow sending a notification, similar to G6. There shouldn't really be any situation where the creator of the category doesn't agree with deletion when the page is being tagged, because the category will never be used, unlike, for example, C1, where it is just currently unused. If that shouldn't be how this one works we'll need the corresponding templates to be created (Template:db-c4-notice and Template:db-c4-deleted).

NovemLinguae:
Tested, works.

Nice patch with no changes needed. Thank you :)

Notes from researching:
* C4 is replacing templatecat (G8) and is a superset of it. Per the WT:CSD discussion, C4 includes like 3 things including templatecat.
* The special "G8: Redirects to non-existent targets" should indeed not be displayed at the top of Twinkle's CSD options for non-redirect categories since they are not redirects. In my testing, I confirmed that the special "G8: Redirects to non-existent targets" is still displayed for both {{Redirect category}} and #REDIRECT categories.

PR #2014, opened 2024-09-19
get ready for mw-message-box deprecation

### LINKED ISSUE #2012: config: change mw-message-box to something else
The HTML class mw-message-box is deprecated and will be removed in a week or two.

It is only used in one spot in the Twinkle code, in twinkleconfig.js:

image

Here is a patch where mw-message-box is changed to Codex:

https://gerrit.wikimedia.org/r/c/mediawiki/extensions/CentralNotice/+/1057938

It looks like this (green box):

image

NovemLinguae:
fixes #2012

Was unable to use Codex, because Codex is usually only loaded on certain pages by certain extensions that use Codex. There is no guarantee it would be loaded and available on the page [[Wikipedia:Twinkle/Preferences]].

The strategy in this patch is to copy paste the necessary CSS from MediaWiki to Morebits.

### DIFF
--- a/modules/twinkleconfig.js
+++ b/modules/twinkleconfig.js
@@ -1742,7 +1742,7 @@ Twinkle.config.saveSuccess = function twinkleconfigSaveSuccess(pageobj) {
 	pageobj.getStatusElement().info('successful');
 
 	var noticebox = document.createElement('div');
-	noticebox.className = 'mw-message-box mw-message-box-success';
+	noticebox.className = 'morebits-message-box morebits-message-box-success';
 	noticebox.style.fontSize = '100%';
 	noticebox.style.marginTop = '2em';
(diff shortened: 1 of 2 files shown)


jdlrobson:
> Was unable to use Codex, because Codex is usually only loaded on certain pages by certain extensions that use Codex.
Codex is opt in so if you are using the markup and you need to add the styles you  should add mediawiki.htmlform.codex.styles as well and then you can use them.

Note, on skins which use Codex e.g. Vector 2022 and Minerva Codex so the module should only be needed on skins where it is not, so if you wanted to optimize you could limit where it's loaded.

NovemLinguae:
Thanks @jdlrobson . So you're saying that one approach to writing this patch would be to do something like `mw.loader.using( [ 'mediawiki.htmlform.codex.styles' ]`, to ensure that Codex is loaded and available on skins such as Vector 2010?

jdlrobson:
Yes! If you are not ready to change the markup just yet, you can also add mediawiki.legacy.messageBox for now.

NovemLinguae:
I decided to use Codex instead of mediawiki.legacy.messageBox, because the latter emitted a deprecation warning in my JavaScript console.

Before:
image

After:
image

PR #2023, opened 2024-10-17
Add dark mode compatibility to Twinkle

Matr1x-101:
I have tested locally, it seems to work.


morebits.js
```diff
@@ -459,7 +459,6 @@ Morebits.quickForm.element.prototype.compute = function QuickFormElementCompute(
 			node = document.createDocumentFragment();
 			// fragments can't have any attributes, so just return it straight away
 			return [ node, node ];
-		// Sometimes Twinkle uses fancy searchable "select" elements. This is powered by the third party library "select2". Activate it by creating a Morebits "select" element, then call `$('select[name=sub_group]').select2({});` or similar towards the end of your main code.
```
NovemLinguae, line 462: Can we keep this comment please? Seems helpful.

modules/twinklewarn.js.orig
```diff
@@ -0,0 +1,1978 @@
+// <nowiki>
```
NovemLinguae, line 1: What is a .orig file? Was this supposed to be named something else?

NovemLinguae (changes requested):

modules/twinklewarn.js.orig
Matr1x-101: Re-added.

morebits.js
Matr1x-101: Re-added

modules/twinklewarn.js.orig
Matr1x-101: Sorry, this is just me being sloppy that file shouldn't exist.

morebits.css
```diff
@@ -304,3 +349,39 @@ th.morebits-mbox-text, td.morebits-mbox-text {
...
+	color: inherit;
+	padding: 0.5em;
+}
+
+.twinkle-config-helptip {
+	font-size: 90%;
+	color: var(--color-subtle, #54595d);
+}
+
+/* Override select2 silliness */
```
siddharthvp, line 380: select2 dropdowns look broken:

Screenshot 2024-10-20 at 1 12 42 AM

morebits.css
```diff
@@ -8,22 +8,60 @@
...
+
+@media screen {
+	html.skin-theme-clientpref-night {
+		--morebits-color-status-status: #A6C4DE;
+		--morebits-color-legend: #6B9DCC;
+		--morebits-color-tooltip: #7A7AFF;
+		--morebits-color-info: #6BDB6B;
+		--morebits-color-warning: #FFC7B3;
+		--morebits-color-titlebar-links: #8BADDF;
+		--morebits-bgcolor-dialog: #17244A;
```
siddharthvp, line 31: This is subjective, but I think the blue on black looks terrible. Might want to use a shade of grey instead.

morebits.css
Matr1x-101: I think we might need to fork select2, and make our own custom css styles for dark mode. Applying this css hotfix to morebits.css is only likely to cause problems down the road.

morebits.css
```diff
@@ -304,3 +349,39 @@ th.morebits-mbox-text, td.morebits-mbox-text {
     padding: 0.25em 0.9em;
     width: 100%;
 }
+
+/* Twinkle config */
```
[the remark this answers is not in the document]
Matr1x-101: Thanks, done

morebits.css
Matr1x-101: Sorry, which color are you talking about?

siddharthvp: morebits-bgcolor-dialog

morebits.css
```diff
@@ -10,58 +10,58 @@
...
-	--morebits-bgcolor-dialog: #F0F8FF;
-	--morebits-bgcolor-titlebar: #BCCADF;
+	--twinkle-color-status-status: #4682B4;
+	--twinkle-color-legend: #31628F;
+	--twinkle-color-tooltip: #0000cd;
+	--twinkle-color-info: #228B22;
+	--twinkle-color-warning: #FF4500;
+	--twinkle-color-titlebar-links: #3062AD;
+	--twinkle-bgcolor-dialog: #F0F8FF;
+	--twinkle-bgcolor-titlebar: #BCCADF;
```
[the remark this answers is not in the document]
Matr1x-101: I thought it would be good to standardise the css variables by only having one variable for one color, but now I think having two css variables for the same color is better, depending on if twinkle or morebits is loaded. I've therefore updated it.

twinkle.css
```diff
@@ -42,3 +63,40 @@
...
+
+.twinkle-config-helptip {
+	font-size: 90%;
+	color: var(--color-subtle, #54595d);
+}
+
+@media screen {
+	html.skin-theme-clientpref-night #twinkle-config-headerbox {
+		background: #532B18;
+}
```
siddharthvp, line 97: A closing brace is missing here. Same issue in the other media query block as well.

morebits.css
Matr1x-101: I experimented with some other colours, and I think this might be better (screenshot attached):

- bgcolor-dialog: #1b1b30
- bgcolor-titlebar: #1c2a52

image

morebits.css
Matr1x-101: My main concern with this solution is because it's not specific to Twinkle, it may affect styles for content outside of Twinkle. That's why I suggested changing the select2 library's css directly.

twinkle.css
Matr1x-101: Fixed

morebits.css
siddharthvp: All select2 CSS we have today are added via JS in the module code ([example](https://github.com/wikimedia-gadgets/twinkle/blob/34259325a9043b246b7d8cd5ae338d09ee244415/modules/twinklewarn.js#L1499)) so that they are applied only when that Twinkle dialog is opened. We could do the same here, though it will still affect non-Twinkle content when these modules (warn/xfd/block) are opened. I think it's a reasonable trade-off.

There's no way around that short of creating a [custom select2 theme](https://select2.org/appearance#themes), which involves duplicating a lot of CSS of the default theme. 

Maintaining a fork of the entire library is not something we have bandwidth for. Even if we had, select2 fits in poorly with MediaWiki's interfaces and we should ideally move away from it (in favour of integrations with native libraries like Codex or OOUI), not deepen our involvement with it.

siddharthvp: I found a way to avoid impacting non-Twinkle content. Turns out we can use the `theme` field to hack in an extra class on the select2-container. By setting `theme: 'default select2-morebits'`, the default theme styles get applied, and `select2-morebits` class also appears on the container using which we can target the styles: 
```css
.select2-morebits,
.select2-morebits .select2-results,
.select2-morebits .select2-results__option[aria-selected=true],
.select2-morebits .select2-selection__rendered,
.select2-morebits .select2-selection--multiple,
.select2-morebits .select2-selection__choice {
    background: var(--background-color-base, #fff) !important;
    color: inherit !important;
}
```

morebits.css
siddharthvp: I think it's better to reuse `--background-color-neutral` or `--background-color-neutral-subtle` for the dialog background. Titlebar colour seems okay.

Matr1x-101: I disagree: background-color-neutral-subtle (and the other one) don't fit the blue aesthetic of twinkle. There needs to be a touch of blue, otherwise it doesn't feel like Twinkle and creates a weird contrast between the titlebar and dialog.
image

morebits.css
Matr1x-101: I'm not that familiar with select2, where do I apply the `theme` field, so it applies to all select2 objects? I tried doing so by changing the default in lib/select2.min.js, but that didn't work.

morebits.css
siddharthvp: Choice of colors is dependent on the theme. Pale blue dialog against a white background looks nothing like this gawdy dark blue dialog on a black background. If the contrast with the titlebar is a problem, that can be changed as well.

Matr1x-101: We still need some kind of accent colour to differentiate Twinkle from other content. Maybe a more subtle dark blue is better?

- bgcolor-dialog: #141c26
- bgcolor-titlebar: #1c2a52

image

morebits.css
Matr1x-101: Actually I found a better strategy:

```css
.morebits-dialog ~ .select2-container:first-of-type > .select2-dropdown {
    background: var(--background-color-base, #fff) !important;
    color: inherit !important;
}
```

This CSS only matches the first select2-container after morebits-dialog, therefore only matching select2 twinkle content.

siddharthvp: > I'm not that familiar with select2, where do I apply the `theme` field

In the `.select2({...})` initialization at each place where select2 is used (in twinklewarn.js, twinkleblock.js and twinklexfd.js).

> so it applies to all select2 objects?

Even if it was possible to apply it in one place, that would mean any non-Twinkle uses also get the styles, which is what you set out to avoid in the first place?

> Actually I found a better strategy:

That's a pretty complex selector. I think using `.select2-morebits` in place of `.morebits-dialog ~ .select2-container:first-of-type` is cleaner and less error-prone.

morebits.css
siddharthvp: Yes, this looks okay to me.

morebits.css
Matr1x-101: Okay, that should be done now.

modules/twinkleblock.js
```diff
@@ -649,6 +649,7 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 
 
 		$form.find('[name=pagerestrictions]').select2({
+			theme: 'default select2-morebits',
```
siddharthvp, line 652: Need to do this for `$form.find('[name=namespacerestrictions]')` as well down below.

morebits.css
```diff
@@ -8,22 +8,60 @@
...
+	--morebits-bgcolor-titlebar: #BCCADF;
+}
+
+@media screen {
+	html.skin-theme-clientpref-night {
+		--morebits-color-status-status: #A6C4DE;
+		--morebits-color-legend: #6B9DCC;
+		--morebits-color-tooltip: #7A7AFF;
+		--morebits-color-info: #6BDB6B;
+		--morebits-color-warning: #FFC7B3;
```
siddharthvp, line 29: Looks too whitish red. Not attention-grabbing as this color is supposed to be. 

(To test this on testwiki, in XFD module try to list a page at RM/TR - it fails and an error with this color is shown.)

siddharthvp (commented):
Two final issues (other than the unresolved comment above regarding the close button being nearly invisible). Everything else looks good to me.

morebits.css
Matr1x-101: #FF6D38 seemed to work better.

morebits.css
```diff
@@ -216,7 +258,8 @@ div.morebits-usertext {
 
 body .ui-dialog.morebits-dialog .ui-dialog-titlebar {
```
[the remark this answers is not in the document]
Matr1x-101: I just added skin-invert class, which seems to do the trick.

Matr1x-101:
Anything else, or can this be merged @NovemLinguae?

NovemLinguae:
@siddharthvp is the better person to review/merge this one since he has been following more closely.

morebits.js
```diff
@@ -5868,6 +5861,9 @@ Morebits.simpleWindow = function SimpleWindow(width, height) {
 
 	// resize the scrollbox with the dialog, if one is present
 	$widget.resizable('option', 'alsoResize', '#' + this.content.id + ' .morebits-scrollbox, #' + this.content.id);
+
+	// add skin-invert to "close" button
+	$('.morebits-dialog .ui-dialog-titlebar-close').addClass('skin-invert');
```
siddharthvp, line 5866: This feels like a hack, but looks like there's no other way to do it - as the 'x' is an icon rather than text.

Matr1x-101: I agree it's a hack. Long term we should probably move away from jQuery to ES6.

siddharthvp:
Looks good to me now. Thanks @Matr1x-101 for persisting through this!

PR #2031, opened 2024-10-20
arv: preview arv reports

### LINKED ISSUE #671: ARV missing previews
Conspicuously absent. Would be mainly helpful for SPI and AN3 reports.

TollensWP:
Resolves #671.

Sorry that this diff is hard to read, this change includes a fairly substantial refactor so that the creation of the report wikitext is done on its own rather than coupled with report posting logic. Aside from the obvious, I don't think there are any changes to functionality, other than that curly braces in edit summaries will now be escaped to prevent templates from getting rendered in AN3 reports given that they don't render in edit summaries.

Because of that refactor, it's highly possible this has broken something somewhere (especially considering there's now promise chains involved, which I can only barely wrap my head around): I've done quite a lot of testing to make sure that didn't happen but would appreciate someone double-checking my work, especially with AN3 reporting.

One thing that did come up is that in AN3 reports, the logic for filling the |orig= parameter in the report template seems to be broken (both before my change and after, I think) - I would have fixed it but frankly I don't really understand what should go into that parameter in the first place. The API call is not parsed properly (the following is from the current version):

https://github.com/wikimedia-gadgets/twinkle/blob/34259325a9043b246b7d8cd5ae338d09ee244415/modules/twinklearv.js#L854-L894

Seems like the `data` parameter of the callback is trying to be read as an array, but the API call always returns an object. Adding `data = data.query.pages[data.query.pageids[0]].revisions;` at the start of that callback would convert the object to the array it thinks it is, but I don't know enough about what the |orig= parameter is even for to be able to decide whether that actually fixes the problem. I can't even really create a new ticket because I don't know what the expected behavior is, but if someone is able to explain I can probably fix it pretty easily.

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -1073,7 +1073,7 @@ Twinkle.arv.callback.getAn3ReportData = function(input) {
 			confirmations: data.confirmations
 		};
 	}).catch((errorData) => {
-		if (typeof errorData !== 'object' || Object.prototype.hasOwnProperty.call(errorData, 'message')) {
+		if (typeof errorData !== 'object' || !Object.prototype.hasOwnProperty.call(errorData, 'message')) {
 			return Promise.reject({ message: 'Unknown error: ' + errorData });
 		}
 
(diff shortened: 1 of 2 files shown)


TollensWP:
Well that's nifty! Never seen that before.

NovemLinguae:
Thanks for your work on this!

Sorry, this is a big ask... but any chance I can get you to split this into lots of little patches instead? I see you did the "extract method" refactoring several times. Maybe one patch for each extraction? Then maybe a patch for the Morebits modifications. And finally, a patch for adding preview to ARV? The diffs should be a lot cleaner and easier to review if we do it this way.

TollensWP:
Yep - that's totally fine, I expected that would probably be necessary after seeing the diff (multiple patches was my original plan too but I got a bit carried away).

PR #2041, opened 2024-11-12
tag: add {{Maybe free media}} and {{Fake SVG}} for files

### LINKED ISSUE #1844: tag: add {{Maybe free media}} and {{Fake SVG}} for files only
Requested by Jonteemil at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Add_two_file_templates_to_the_standard_installations

fredster33:
Fixes #1844


NovemLinguae:
The form validation code around line 2114 was pretty bad before your patch, and I was having trouble even reading what was going on there, so I refactored it in #2047 and then adjusted your patch.

PR #2045, opened 2024-11-12
scripts: delete unused

### LINKED ISSUE #2044: delete unused scripts
* [ ] delete 4
* [ ] rename 2
    * [ ] update deploy documentation

-----

There's 6 scripts in the /scripts/ folder. I've only ever used 2 of them...

* dev-server.js (via `npm start`)
* sync.pl (the deploy script)

This suggests to me that, unless CI uses them or something, the other 4 can be deleted.

* draftRelease.sh
* morebits-test.js
* patch-test.js
* server.js - this one in particular, since it is a partial duplicate of dev-server.js and these are at risk of drifting out of sync with each other

After these 4 scripts are deleted (assuming nothing uses them), let's also rename the two remaining scripts...

* dev-server.js -> server.js
* sync.pl -> deploy.pl

NovemLinguae:
related #2044 delete unused scripts

note: test and make sure `npm start` still works


NovemLinguae:
@siddharthvp, have you used patch-test-loader.js in the last couple years, or do you think it's abandoned?

siddharthvp:
Haven't used it for a long time. Should be okay to delete. 

It's an alternative way of testing Twinkle while the onwiki version is also running – hence quite error-prone, but more or less the standard approach before the much more reliable dev-server.js script was written.

PR #2056, opened 2024-11-21
rename twinklefluff.js to twinklerollback.js

### LINKED ISSUE #1975: rename "twinklefluff.js" to "twinklerollback.js"
Unlike the other module names, "fluff" is not descriptive of what the module does. The word is meaningless to new devs

Thanks to @nardog for the idea.

NovemLinguae:
fixes #1975

Summary
- Did an all repo, case sensitive search and replace of Twinkle.fluff -> Twinkle.rollback
- Did an all repo search for "fluff" and changed remaining results to "rollback"

Note
- This will cause [937 people](https://en.wikipedia.org/w/index.php?search=%22confirmOnFluff%22&title=Special:Search&profile=advanced&fulltext=1&ns0=1&ns1=1&ns2=1&ns3=1&ns4=1&ns5=1&ns6=1&ns7=1&ns8=1&ns9=1&ns10=1&ns11=1&ns12=1&ns13=1&ns14=1&ns15=1&ns100=1&ns101=1&ns118=1&ns119=1&ns126=1&ns127=1&ns710=1&ns711=1&ns828=1&ns829=1)'s fluff preferences to get reset back to default, because some of these preferences were renamed.

Deployment checklist
- Rename the "Module: fluff" tag in issues and PRs to "Module: rollback"
- Modify MediaWiki:Gadget-definition
- Move the page onwiki to prevent cut and paste move by deploy script


NovemLinguae:
No objections in https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Renaming_fluff_to_rollback, I assume this is OK to proceed.

siddharthvp:
I ... am not sure if no objections in 24 hours is a a reason to break preferences of 937 users who are soon going to find out that their rollback buttons now immediately rollback instead of triggering a confirm prompt, even on mobile devices. We should have back-compat here to support the old preferences. 

Changes to preferences format over the years were always back-compatible. We [still support](https://github.com/wikimedia-gadgets/twinkle/blob/6f795abacb1ee0f998cf20d4b51a27621f93dcea/twinkle.js#L331) the pre-2019 preferences format in which friendly and twinkle were separate blocks in twinkleoptions.js, as well as [the pre-2011 preferences](https://github.com/wikimedia-gadgets/twinkle/blob/6f795abacb1ee0f998cf20d4b51a27621f93dcea/twinkle.js#L182) that were set directly in user common.js.

NovemLinguae:
Copy that. Will test and deploy your patch right now.

PR #2057, opened 2024-11-21
eslint: apply autofixes

NovemLinguae:
I made it one autofix per commit, to make the diffs easier to read.

`npx eslint --no-eslintrc --fix --rule '{wrap-iife: error}' --env 'es6' .`

I have found eslint autofixes to be extremely reliable. I have never encountered a bug with them that has broken the program. I have seen them mess with code formatting in unexpected ways though, so for that reason I am still spot checking each diff.

This will probably be the first of several patches. Doing all the autofixes in one patch would make a revert a pain.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+scripts/
+tests/
+lib/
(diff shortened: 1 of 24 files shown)


NovemLinguae:
Fixed merge conflict in #2058

PR #2058, opened 2024-11-22
eslint: apply autofixes 1

NovemLinguae:
* move ignorePatterns to .eslintignore, makes bash commands easier
* wrap-iife
* no-var
* prefer-const
* prefer-arrow-callback
* arrow-parens
* arrow-body-style
* manual fixes, to deal with comments that confused the autofix

I made it one autofix per commit, to make the diffs easier to read.

`npx eslint --no-eslintrc --fix --rule '{wrap-iife: error}' --env 'es6' .`

I have found eslint autofixes to be extremely reliable. I have never encountered a bug with them that has broken the program. I have seen them mess with code formatting in unexpected ways though, so for that reason I am still spot checking each diff.

This will probably be the first of several patches. Doing all the autofixes in one patch would make a revert a pain.

### DIFF
--- a/modules/twinklebatchprotect.js
+++ b/modules/twinklebatchprotect.js
@@ -82,10 +82,7 @@ Twinkle.batchprotect.callback = function twinklebatchprotectCallback() {
 		name: 'movelevel',
 		label: 'Move protection:',
 		event: Twinkle.protect.formevents.movelevel,
-		list: Twinkle.protect.protectionLevels.filter((level) => 
-			// Autoconfirmed is required for a move, redundant
-			 level.value !== 'autoconfirmed'
-		)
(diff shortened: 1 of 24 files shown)


modules/twinklebatchprotect.js
```diff
@@ -82,10 +82,7 @@ Twinkle.batchprotect.callback = function twinklebatchprotectCallback() {
 		name: 'movelevel',
 		label: 'Move protection:',
 		event: Twinkle.protect.formevents.movelevel,
-		list: Twinkle.protect.protectionLevels.filter(function(level) {
-			// Autoconfirmed is required for a move, redundant
-			return level.value !== 'autoconfirmed';
-		})
+		list: Twinkle.protect.protectionLevels.filter((level) => /* Autoconfirmed is required for a move, redundant */ level.value !== 'autoconfirmed')
```
siddharthvp, line 85: line too long and awkward

PR #2059, opened 2024-11-22
rollback: support older fluff preferences

siddharthvp:
Continue honoring the confirmOnFluff and confirmOnMobileFluff preferences (renamed in #2056) which are in use by 900+ users.

When a user changes any of their Twinkle preferences, these two preferences if in use will get automatically renamed. When this has occurred for most users, support for them can be dropped.

### DIFF
--- a/modules/twinklerollback.js
+++ b/modules/twinklerollback.js
@@ -673,9 +673,12 @@ Twinkle.rollback.callbacks = {
 				break;
 		}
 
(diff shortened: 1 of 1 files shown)


NovemLinguae:
Without additional backwards compatibility code, Twinkle will also delete the old fluff preferences when a user saves their config file. We may want to do a follow up patch to handle that.

PR #2060, opened 2024-11-22
arv: refactor AIV report generation

TollensWP:
This is part of a set of patches to implement the preview functionality described in #671.

No functionality has changed here - the only change is a refactor of the ANI reporting logic to pull out two methods responsible for generating the reports from the code that posts the reports.

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -494,7 +494,7 @@ Twinkle.arv.callback.evaluate = function(e) {
 				alert('You must specify some reason');
 				return;
 			}
-			
+
 			Morebits.simpleWindow.setButtonsEnabled(false);
 			Morebits.status.init(form);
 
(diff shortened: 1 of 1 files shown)


modules/twinklearv.js
```diff
@@ -812,6 +768,68 @@ Twinkle.arv.callback.evaluate = function(e) {
 	}
 };
 
+Twinkle.arv.callback.getAivReasonOnlyWikitext = function(input) {
```
NovemLinguae, line 771: Consider renaming to getAivReasonWikitext, or a different name, or adding a comment. I don't understand the "only" part.

TollensWP: Thanks - I didn't like that name and thought it was too long but somehow it didn't click that "only" was completely redundant.

PR #2061, opened 2024-11-22
config: don't ignore old fluff preferences

NovemLinguae:
- fixes an issue where if a user still had the old confirmOnFluff or confirmOnRollback preferences, and visited the preferences page and saved, their preferences were ignored and silently deleted
- refactor complicated conditional in twinklerollback.js module
- reverts #2059 because we accomplish the same thing by putting the code in twinkle.js -> twinkleGetPref()

### DIFF
--- a/modules/twinklerollback.js
+++ b/modules/twinklerollback.js
@@ -679,7 +679,7 @@ Twinkle.rollback.callbacks = {
 				(
 					Twinkle.getPref('confirmOnMobileRollback') &&
 					// Mobile user agent taken from [[en:MediaWiki:Gadget-confirmationRollback-mobile.js]]
-					/Android|webOS|iPhone|iPad|iPod|BlackBerry|Mobile|Opera Mini/i.test(navigator.userAgent);
+					/Android|webOS|iPhone|iPad|iPod|BlackBerry|Mobile|Opera Mini/i.test(navigator.userAgent)
 				)
 			) &&
 			!userHasAlreadyConfirmedAction;
--- a/modules/twinklerollback.js
+++ b/modules/twinklerollback.js
@@ -673,14 +673,13 @@ Twinkle.rollback.callbacks = {
 				break;
 		}
 
-		// Mobile user agent taken from [[en:MediaWiki:Gadget-confirmationRollback-mobile.js]]
-		const isMobileUser = /Android|webOS|iPhone|iPad|iPod|BlackBerry|Mobile|Opera Mini/i.test(navigator.userAgent);
 		const needToDisplayConfirmation =
 			(
 				Twinkle.getPref('confirmOnRollback') ||
 				(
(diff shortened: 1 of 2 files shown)


modules/twinklerollback.js
```diff
@@ -673,13 +673,19 @@ Twinkle.rollback.callbacks = {
...
 
-		// Preferences confirmOnRollback and confirmOnMobileRollback were formerly named
-		// confirmOnFluff and confirmOnMobileFluff, respectively. Check both variations for now.
-		if ((Twinkle.getPref('confirmOnRollback') || Twinkle.getPref('confirmOnFluff') ||
-			((Twinkle.getPref('confirmOnMobileRollback') || Twinkle.getPref('confirmOnMobileFluff')) &&
-				// Mobile user agent taken from [[en:MediaWiki:Gadget-confirmationRollback-mobile.js]]
-				/Android|webOS|iPhone|iPad|iPod|BlackBerry|Mobile|Opera Mini/i.test(navigator.userAgent))) &&
-			!userHasAlreadyConfirmedAction && !confirm('Reverting page: are you sure?')) {
+		// Mobile user agent taken from [[en:MediaWiki:Gadget-confirmationRollback-mobile.js]]
+		const isMobileUser = /Android|webOS|iPhone|iPad|iPod|BlackBerry|Mobile|Opera Mini/i.test(navigator.userAgent);
```
siddharthvp, line 677: This doesn't need to be evaluated unless confirmOnMobileRollback is true. Don't think it's worth refactoring it out.

twinkle.js
```diff
@@ -179,13 +179,22 @@ Twinkle.getPref = function twinkleGetPref(name) {
...
 		return window.TwinkleConfig[name];
 	}
 	if (typeof window.FriendlyConfig === 'object' && window.FriendlyConfig[name] !== undefined) {
 		return window.FriendlyConfig[name];
 	}
+
+	// Backwards compatibility code because we renamed confirmOnFluff to confirmOnRollback, and confirmOnMobileFluff to confirmOnMobileRollback
+	if (name === 'confirmOnRollback' && Twinkle.prefs.confirmOnFluff !== undefined) {
+		return Twinkle.prefs.confirmOnFluff;
+	} else if (name === 'confirmOnMobileRollback' && Twinkle.prefs.confirmOnFluff !== undefined) {
```
siddharthvp, line 194:
```suggestion
	} else if (name === 'confirmOnMobileRollback' && Twinkle.prefs.confirmOnMobileFluff !== undefined) {
```

siddharthvp (commented):
Ah yeah, I missed testing the config update flow earlier.

modules/twinklerollback.js
NovemLinguae: My clean code book says that readability is usually more important than performance. But not worth holding up this patch over. Done.

PR #2063, opened 2024-11-23
speedy: add {{Db-x3}}

### LINKED ISSUE #1970: speedy: add {{Db-x3}}
Requested by Gonnym at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Template%3ADb-x3

fredster33:
Fixes #1970


NovemLinguae:
Usually we add a bunch of preferences when adding a CSD.

image

Although there are some exceptions, such as G5.

Of the 4 types of preferences above, do you think we should add any preferences for X3?

fredster33:
I think that all 4 are relevant, thanks for flagging! Added.

NovemLinguae:
I think we need to create {{Db-x3-notice}} and {{Db-x3-deleted}} on enwiki in order for the various talk page notifications to work. Can you please create those then let me know in this ticket? You can look at similar templates for other CSDs and copy/paste then adapt those.

NovemLinguae:
Do we want any of these notifications to be on by default? If so, they'll need to be added to twinkle.js -> Twinkle.defaultConfig.

PR #2077, opened 2024-11-25
split twinkle.js into two files

NovemLinguae:
One file (twinkleutil.js, new file) for functions, one file (twinkle.js, original file) for everything else.

For unit test reasons. The non-function code will get in the way if we want to write unit tests for the functions.

Refactoring. No-op.

-----

Todo
* [ ] resolve merge conflicts. one of them will involve copying these docblock changes over: https://github.com/wikimedia-gadgets/twinkle/pull/2076/files
* [ ] Investigate "console.log
https://github.com/wikimedia-gadgets/twinkle/actions/runs/12001912518/job/33453089724?pr=2077#step:4:39 Could not load your Twinkle preferences, resorting to default preferences"
* [ ] Does the load order matter? twinkle.js declares the Twinkle global variable so that may need to come first. But twinkle.js also calls functions declared in twinkleutil.js, so maybe twinkleutil.js needs to come first?
* [ ] Update documentation pages that list every file
* [ ] Update deploy script
* [ ] Create the page onwiki, if the deploy script gets confused.
* [ ] Add to CI labeler
* [ ] Create separate labels in issues and PR, or reuse Module: twinkle label?
* [ ] Anything else?


siddharthvp:
The load order matters - the file defining `Twinkle` global needs to be first. It's okay for this file to call functions declared in the other, as long as these calls are in async callbacks rather than at the top level (0 indent level).

siddharthvp:
As the intent is to increase coverage, I'd suggest putting all top-level code into functions and continuing to call the file `twinkle.js`. The function which starts the loading can be invoked from a new file,  say `twinkleload.js`, which would be only one line and doesn't contain anything else. Only twinkleload.js would need to be skipped from coverage reporting.

PR #2091, opened 2024-12-01
welcome: change {{welcomeunsourced}} to {{welcome-unsourced}} as result of page move

rustykitty:
https://en.wikipedia.org/w/index.php?title=Wikipedia_talk:Twinkle&oldid=1260049504#Moving_Template:Welcomeunsourced_to_Template:Welcome-unsourced

### DIFF
--- a/modules/twinklewelcome.js
+++ b/modules/twinklewelcome.js
@@ -323,10 +323,10 @@ Twinkle.welcome.templates = {
 				linkedArticle: true,
 				syntax: '{{subst:welcomenpov|$ARTICLE$|$USERNAME$}} ~~~~'
 			},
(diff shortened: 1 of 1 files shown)


modules/twinklewelcome.js
```diff
@@ -323,10 +323,10 @@ Twinkle.welcome.templates = {
 				linkedArticle: true,
 				syntax: '{{subst:welcomenpov|$ARTICLE$|$USERNAME$}} ~~~~'
 			},
-			welcomeunsourced: {
+			'welcome-unsourced': {
```
NovemLinguae, line 326: Can we alphabetize this please? We can do this by cutting and pasting this block of code a little higher.

PR #2097, opened 2024-12-04
speedy: add T5

### LINKED ISSUE #2094: speedy: add CSD T5
requested by Gonnym at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Template%3ADb-t5

https://en.wikipedia.org/wiki/Template:Db-t5

fredster33:
Fixes #2094

### DIFF
--- a/modules/twinklespeedy.js
+++ b/modules/twinklespeedy.js
@@ -302,6 +302,13 @@ Twinkle.speedy.callback.modeChanged = function twinklespeedyCallbackModeChanged(
 				}
 				break;
 
+			case 10: // template
+			case 11: // template talk
+			case 828: // module
+			case 829: // module talk
+				appendList('Templates and modules', Twinkle.speedy.templateList);
+				break;
+
 			case 14: // category
 			case 15: // category talk
 				appendList('Categories', Twinkle.speedy.categoryList);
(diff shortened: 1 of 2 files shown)


modules/twinklespeedy.js
```diff
@@ -302,6 +302,13 @@ Twinkle.speedy.callback.modeChanged = function twinklespeedyCallbackModeChanged(
 				}
 				break;
 
+			case 10: // template
+			case 11: // template talk
+			case 828: // module
+			case 829: // module talk
+				appendList('Templates and modules', Twinkle.speedy.templateList);
```
NovemLinguae, line 309: Can you please walk me through what this code is doing?

NovemLinguae:
I think we need to create {{Db-t5-notice}} and {{Db-t5-deleted}} on enwiki in order for the various talk page notifications to work. Can you please create those then let me know in this ticket? You can look at similar templates for other CSDs and copy/paste then adapt those.

NovemLinguae:
Do we want any of these notifications to be on by default? If so, they'll need to be added to twinkle.js -> Twinkle.defaultConfig.

modules/twinklespeedy.js
fredster33: This enables T5 for the template and module namespaces, as well as their talk pages

ToThAc:
There's no need for any talk page notice, since the criterion is usually for non-controversial maintenance.

PR #2098, opened 2024-12-04
eslint: fix no-jquery/no-constructor-attributes

NovemLinguae:
manual fixes

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -369,18 +369,19 @@ Twinkle.arv.callback.changeCategory = function (e) {
 							const pageid = data.query.pageids[0];
 							const page = data.query.pages[pageid];
 							if (!page.revisions) {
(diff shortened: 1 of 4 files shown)


modules/twinklearv.js
```diff
@@ -369,18 +369,19 @@ Twinkle.arv.callback.changeCategory = function (e) {
...
-										type: 'checkbox',
-										name: 's_' + field,
-										value: rev.revid
-									});
+									const $entry = $('<div>')
+										.addClass('entry');
+									const $input = $('<input>')
+										.attr('type', 'checkbox')
+										.attr('name', 's_' + field)
+										.attr('value', rev.revid);
```
siddharthvp, line 384: It's also possible to use the .attr({ field1: val1, field2: val2 }) form to be more concise. 

Fine either way though.

PR #2101, opened 2024-12-04
eslint: fix es-x/no-array-prototype-includes

NovemLinguae:
manual fixes

apparently Array.includes() is >ES6, so to keep compatibility with old browsers it should be removed for now

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -41,7 +41,6 @@
 			}
 		],
 
-		"es-x/no-array-prototype-includes": "warn",
 		"es-x/no-object-values": "warn",
 		"mediawiki/class-doc": "warn",
 		"new-cap": "warn",
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -537,7 +537,7 @@ Twinkle.arv.callback.evaluate = function(e) {
 
 		// Report inappropriate username
 		case 'username':
-			var censorUsername = input.arvtype.includes('offensive'); // check if the username is marked offensive
+			var censorUsername = input.arvtype.indexOf('offensive') !== -1; // check if the username is marked offensive
 
 			reason = Twinkle.arv.callback.getUsernameReportWikitext(input);
 
--- a/modules/twinkleprotect.js
+++ b/modules/twinkleprotect.js
@@ -508,7 +508,7 @@ Twinkle.protect.callback.changeAction = function twinkleprotectCallbackChangeAct
 
 			var isTemplateNamespace = mw.config.get('wgNamespaceNumber') === 10;
 			var isAFD = Morebits.pageNameNorm.startsWith('Wikipedia:Articles for deletion/');
-			var isCode = ['javascript', 'css', 'sanitized-css'].includes(mw.config.get('wgPageContentModel'));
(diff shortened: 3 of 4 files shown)


NovemLinguae:
The modern layer is still on ES6. https://www.mediawiki.org/wiki/Compatibility#General_information

Until Wikimedia devs decide to raise that, eslint-config-wikimedia's default settings, and all code written in Wikimedia repos, is expected to not be above ES6.

I assume the reasoning is a combination of supporting older browser versions, and maintaining compatibility with ResourceLoader tools such as the minifier. In this particular case, there isn't a unique syntax, so the minifier probably isn't affected, but in other cases such as async/await, it is.

nardog:
Again, Promise.prototype.finally() *is* >ES6 and yet required by MW. This seems pointless and counterproductive.

NovemLinguae:
Where are you seeing `Promise.finally()`?

As you can probably tell, I'd be in favor of a strategy of trusting the eslint-config-wikimedia rule writers and slowly reducing our turned off rules until we're down to as few as possible.

@siddharthvp, any thoughts?

NovemLinguae:
Upstream, I've filed https://phabricator.wikimedia.org/T381537 to centralize discussion of increasing Grade A / modern layer from ES6 to ES7.

nardog:
> Where are you seeing `Promise.finally()`?

In the [feature test](https://github.com/wikimedia/mediawiki/blob/master/resources/src/startup/startup.js#L9) linked in your link.

> Upstream, I've filed https://phabricator.wikimedia.org/T381537 to centralize discussion of increasing Grade A / modern layer from ES6 to ES7.

But Promise.prototype.finally() is from ES2018 (ES9).

NovemLinguae:
> In the [feature test](https://github.com/wikimedia/mediawiki/blob/master/resources/src/startup/startup.js#L9) linked in your link.

Odd. I'll see if I can get more info about that in the Phab ticket.

> But Promise.prototype.finally() is from ES2018 (ES9).

I understand, but Array.includes() (what this PR is about) is ES7, and an ES7 ticket is a good incremental step towards ES9.

nardog:
Not only that, the RL module `es6-polyfills` was removed when MW upped the requirement to ES6, even though it included the polyfill for Array.prototype.includes(). TheDJ [said](https://phabricator.wikimedia.org/T178356#8771731):

> We target ES6 + Promise finally (ES 2018, but was implemented pretty early by most browsers) + Array includes (ES 2016, but present in most ES6 browsers) + Arrow function support + Regexp flags support.

NovemLinguae:
> We target ES6 + Promise finally (ES 2018, but was implemented pretty early by most browsers) + Array includes (ES 2016, but present in most ES6 browsers) + Arrow function support + Regexp flags support.

That might be a minority opinion since the official linter is still blocking most of that. Let's dig into it more in the phab ticket.

siddharthvp:
> Where are you seeing `Promise.finally()`?
> 
> As you can probably tell, I'd be in favor of a strategy of trusting the eslint-config-wikimedia rule writers and slowly reducing our turned off rules until we're down to as few turned off as possible.
> 
> @siddharthvp, any thoughts?

I support the use of Array.includes(). We have been discussing this for twinkle since at least 2019 and IE 11 was the only reason we held off from using it. Following the update of Grade A to exclude IE, all browsers support it (https://caniuse.com/?search=Array.includes).

PR #2122, opened 2024-12-09
tests: ignore twinkle.js in code coverage report

NovemLinguae:
Was showing like 35% covered. But this was just from scaffolding code or something. Actual # of unit tests is 0.


NovemLinguae:
Rewrite this patch to just remove it from jest.config.js instead.

PR #2124, opened 2024-12-09
move .js and .css files out of the root

### LINKED ISSUE #2078: move all client-side javascript into modules folder, and adjust eslint config accordingly
Why
- shouldn't have non-config files in the root
- right now the root contains both node files (server-side JS code) and twinkle.js/morebits.js (client-side JS code), which makes it hard to set up eslint to lint both types of files

What
- move twinkle.js and morebits.js into modules/ folder
- fix everything that breaks (deploy script, jest, documentation, etc.)
- eslint config
    - use wikimedia/server for the root/whole project (.eslintrc.json)
    - move client-side JS config settings to the eslint config file in the modules folder (modules/.eslintrc.json)
    - remove scripts/ and tests/ from .eslintignore. should work now with wikimedia/server set up in the root config

Follow up patch
- apply eslint autofixes to .js files in scripts/ and tests/ folders

NovemLinguae:
Move twinkle.js, morebits.js, CSS files, etc. out of the root. The root directory should be config files and readmes only.

Wikimedia also does this. JS files are always all placed in a directory called /modules/ (or /resources/). But never in the root.

Fixes #2078

To test:
- npm start
- perl deploy.pl - test by deploying to testwiki

Maybe:
- put the old modules in a subdirectory? modules/menuItems/*.js or something?


NovemLinguae:
After thinking about this a bit more, I think I do want to have a subfolder for the Twinkle modules, but I also want to avoid an ugly `modules/modules`. Perhaps `src/modules` would be a good way to do it?

```
src/
   |--morebits.css
   |--morebits.js
   |--twinkle.css
   |--twinkle.js
   |--twinkle-pagestyles.css
   |--modules/
      |--twinklearv.js
      |--twinklebatchdelete.js
      |--twinklebatchprotect.js
      |--etc.
```

@siddharthvp, can I get a second opinion on that?

siddharthvp:
Putting everything under `src/` looks good to me and is consistent with modern JS projects. In fact, I had a branch locally that made the same change but didn't get around to pushing it.

PR #2151, opened 2025-01-02
Link to the deleted page in batch deletions

SilverLocust:
Change the edit summaries for batch deletions to link to the deleted page. This will make it noticeable in the deletion log when the other page has been recreated or undeleted, such that the batch-deleted redirect or talk page is no longer orphaned and could be undeleted.


siddharthvp:
This seems like a good idea and I don't see the need to limit this to the batchdelete module. For consistency, you might want to make the same change in speedy and deprod modules as well.

SilverLocust:
Oh, you're correct. I was under a misimpression that "batchdelete" just referred to deleting associated redirects and talk pages. I'm not the most code-savvy Wikipedian.

siddharthvp:
Thanks!

PR #2180, opened 2025-05-27
block: soft redirect to Special:Block for multi-blocked targets

### LINKED ISSUE #2178: block: add support for multiblocks
[Multiblocks](https://www.mediawiki.org/wiki/Help:Manage_blocks) is coming to enwiki [soon](https://en.wikipedia.org/wiki/Special:Permalink/1290621024#Deployment_of_Multiblocks_on_this_wiki_on_June_2-4). See [mw:Help:Manage blocks/Developers](https://www.mediawiki.org/wiki/Help:Manage_blocks/Developers) for more info.

For targets with exactly 0 or 1 active blocks, Twinkle should continue to work just fine.

When reblocking a multi-blocked target, we'll need to pass in an `id` of the block we want to override, or pass in `newblock=1` to specify we want a new block. For the latter, we could have a new checkbox, "Issue a new block instead of re-blocking".

The tougher problem I think is what to do when first opening Twinkle on a multi-blocked target. We need a UI to list the active blocks, allowing the admin to decide whether to edit an existing block or add a new additional block.

As a short-term solution, I propose simply disabling the form and showing a link to Special:Block/Username when the Twinkle block module is opened on a multi-blocked user.

I just finished 6 months of working on multiblocks, so needless to say I'm a bit burnt out. However as the original author of the Twinkle block module, I feel inclined to keep it functioning. I can take care of the quick fix, and maybe I can guide someone willing through implementing the full multiblocks support?

MusikAnimal:
Rename the Window variable to blockWindow, since Window is a global function.

Closes #2178

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -3,7 +3,7 @@
 (function() {
 
 const api = new mw.Api();
-let relevantUserName, blockedUserName;
+let relevantUserName, blockedUserName, blockWindow;
 const menuFormattedNamespaces = $.extend({}, mw.config.get('wgFormattedNamespaces'));
 menuFormattedNamespaces[0] = '(Article)';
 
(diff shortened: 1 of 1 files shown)


MusikAnimal:
Can be tested on testwiki, i.e. https://test.wikipedia.org/wiki/Special:Contribs/MusikAnimal

NovemLinguae:
Will try to take a look at this soon. If I drop the ball, feel free to self-merge and self-deploy.

Issue #2189, opened 2025-07-02
deploy.js: gadgets-definition warning/diff not working correctly


NovemLinguae:
**Steps to reproduce**
* deploy to enwiki using deploy.js

**What happens?**
* warning about MediaWiki:Gadgets-definition and GitHub gadget.txt having diverged
<img width="952" alt="Image" src="https://github.com/user-attachments/assets/eddee2f2-fe50-4dd4-9184-40a11eff7dbd" />

**What should happen instead?**
* no warning, either because we fix the diff-checking algorithm, or we delete this check entirely

PR #2193, opened 2025-07-02
deploy.js: fix gadget definition diffing

siddharthvp:
Instead of hardcoding 5 as the number of lines of the wiki gadget definition to diff against, check N lines where N = number of (non-empty) lines in the local gadget definition file.

Solves #2189


NovemLinguae:
Tested deploying with the definitions in sync, and with them [not in sync](https://en.wikipedia.org/w/index.php?title=MediaWiki:Gadgets-definition&diff=prev&oldid=1299174566). Both cases worked. Thanks!

siddharthvp:
> P.S. I think if you use "fix #2189" or "fixes #2189" instead of "Solves #2189", GitHub will auto-link the task.

ah ok, I used to use "Closes" which also worked the same way. At some point I switched to "Solves" without realizing it doesn't work :)

PR #2198, opened 2025-08-08
Add db-gs (G5 variant) to Twinkle's CSD menu

IlyasLebleu:
{{[db-gs](https://en.wikipedia.org/wiki/Template:Db-gs)}} is currently not accessible from Twinkle, and uses different parameters than the ones in regular {{db-g5}}.

Tested on test.wikipedia.org (https://test.wikipedia.org/w/index.php?title=User:Chaotic_Enby/sandbox&diff=prev&oldid=669279)


IlyasLebleu:
@NovemLinguae It's been reworked with the new JSON code, and seems to work on test.wiki, please tell me if it's good to merge!

PR #2209, opened 2025-09-22
Add search aliases for warning templates

IlyasLebleu:
Adds an option for searching warning templates through aliases, for users unsure about the exact name of the template. For example, typing "llm" in the search field returns the {{uw-ai}} template series.

Currently only used for {{uw-ai}}, although written such that other aliases can easily be added if needed.


NovemLinguae (changes requested):
This patch breaks the warn -> select2 element in manual testing. Makes it fall back to a regular dropdown. Please fix :)

image

image

PR #2214, opened 2025-10-05
Implement CTOP protection logging

IlyasLebleu:
Implement logging of CTOP page protections per ArbCom decision [here](https://en.wikipedia.org/wiki/Wikipedia:Arbitration_Committee/Noticeboard#Arbitration_motion_regarding_Wikipedia:Arbitration_Committee/Procedures), Twinkle implementation requested [here](https://en.wikipedia.org/wiki/Special:GoToComment/c-Curbon7-20250926011500-Arbitration_motion_regarding_Wikipedia:Arbitration_Committee/Procedures). Tested on test.wikipedia [as usual](https://test.wikipedia.org/w/index.php?title=Special:Log&logid=436171).

image


modules/twinkleprotect.js
```diff
@@ -993,6 +1065,70 @@ Twinkle.protect.protectionTags = [
...
+		if (item.aliases) {
+			item.label = `${item.value}, ${item.aliases.join(', ')} – ${item.label}`;
+		} else {
+			item.label = `${item.value} – ${item.label}`;
+		}
+	}
+});
+
+Twinkle.protect.gsCodes = [
+	{ value: '', label: 'Specify a code...' },
```
NovemLinguae: Same as above comment. Let's pull these from on-wiki if we can. Maybe create a new .json page at https://en.wikipedia.org/wiki/Template:Gs/topics.json ?

NovemLinguae (changes requested):

NovemLinguae (dismissed):
You mention "logging". Does this need to write the protection to a logging page? Or does it just put it in the edit summary so that a bot can do the logging?

As the patch currently is, it does not appear to write to any log pages.

image

IlyasLebleu:
Regarding logging, it's being done by L235's bot, which only needs this to be in the edit summary for it. I'll take care of implementing on-wiki logging, thanks a lot!

modules/twinkleprotect.js
```diff
@@ -993,6 +1065,70 @@ Twinkle.protect.protectionTags = [
 // Filter FlaggedRevs
 .filter((type) => hasFlaggedRevs || type.label !== 'Pending changes templates');
 
+Twinkle.protect.ctopCodes = [
```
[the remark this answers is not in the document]
IlyasLebleu: One small issue with Ds/topics.json: I'm using IMH and SASG as separate codes as they often are for CTOP protection purposes (as they are officially distinct), but this is not reflected in the JSON, and I don't know if adding them might be an issue for the block module. I'm going to add them manually, but let me know if you have any better option.

siddharthvp (dismissed):

modules/twinkleprotect.js
```diff
@@ -993,6 +1069,58 @@ Twinkle.protect.protectionTags = [
...
+		}
+	} else {
+		val.label = 'Specify a code...';
+	}
+});
+
+return codes;
+});
+
+Twinkle.protect.gsCodes = Morebits.wiki.getCachedJson('Template:Gs/topics.json');
```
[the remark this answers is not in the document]
IlyasLebleu: Also fixed!

src/modules/twinkleprotect.js
```diff
@@ -998,6 +1073,63 @@ Twinkle.protect.protectionTags = [
...
+			}
+		} else {
+			val.label = 'Specify a code...';
+		}
+	});
+
+	return codes;
+	});
+
+	gsCodes = Morebits.wiki.getCachedJson('Template:Gs/topics.json');
```
NovemLinguae, line 1107: Looks like there's a bug around here somewhere. Despite https://test.wikipedia.org/wiki/Template:Gs/topics.json having content, I'm getting this during testing:

Image

NovemLinguae (changes requested):

src/modules/twinkleprotect.js
```diff
@@ -998,6 +1073,63 @@ Twinkle.protect.protectionTags = [
 // Filter FlaggedRevs
 .filter((type) => hasFlaggedRevs || type.label !== 'Pending changes templates');
 
+Twinkle.protect.fetchTopics = async function twinkleprotectFetchTopics() {
+	let ctopCodes, gsCodes;
+	ctopCodes = Morebits.wiki.getCachedJson('Template:Ds/topics.json');
```
NovemLinguae, line 1078: https://test.wikipedia.org/w/index.php?title=Template%3ADs%2Ftopics.json&diff=748318&oldid=684451

I wasn't able to craft a perfect XSS, but I think it might be craftable, based on the below screenshot. @siddharthvp , do you have an opinion on this? Is this code vulnerable to XSS? 

Image

NovemLinguae (changes requested):

src/modules/twinkleprotect.js
siddharthvp: Looks safe to me. The html generation happens in Morebits.QuickForm which mostly uses setAttribute() to set the labels and values, which is XSS-safe.

PR #2222, opened 2025-11-04
DEVELOPER.md: rewrite "Adding a CSD" section

NovemLinguae:
- readability refactor
- add updating snapshot tests
- add updating Template:Db-notice-multiple and Template:Db-deleted-multiple, to prevent bugs such as https://en.wikipedia.org/wiki/Wikipedia_talk:WikiProject_Articles_for_creation#G15_speedy_notices_on_user_talk_pages

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -147,6 +147,8 @@ Here is a checklist for writing a patch to add a speedy deletion criteria (CSD).
     * modules/twinklespeedy.js -> Twinkle.speedy.data
 * If Template:Db-XX takes any parameters, make the following code changes:
     * modules/twinklespeedy.js -> Twinkle.speedy.getParameters
+* If Template:Db-XX-notice or Template:Db-XX-deleted take any parameters, make the following code changes:
+    * modules/twinklespeedy.js -> Twinkle.speedy.getUserTalkParameters
 * Update snapshot tests using `npx jest -u`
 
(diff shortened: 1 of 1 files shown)


NovemLinguae:
Thanks. Change made.

IlyasLebleu:
Speaking of, should this be cross-referenced with [Wikipedia:Speedy deletion/Creating a new criterion](https://en.wikipedia.org/wiki/Wikipedia:Speedy_deletion/Creating_a_new_criterion)? There's some overlapping guidance (for example in the templates to create), so we could link this at the end instead of the (outdated) base Twinkle repo link and have "if you are coming from Wikipedia:Speedy deletion/Creating a new criterion, please skip to X".

NovemLinguae:
Nice. I didn't know about that page. Added it to my watchlist just now. I think it's probably OK as is. They have a checklist for all their basics that includes informing Twinkle, and we have this internal checklist for the Twinkle stuff. Having the two lists overlap might get confusing or complicated (e.g. we'd have to try to keep them in sync with each other).

IlyasLebleu:
Yep, it's just that the "create the following pages onwiki" that we have in DEVELOPER.md already causes a bit of overlap, although I don't think that's too much of a problem.

NovemLinguae:
I suppose it would be OK to add the templates to the CSD work instruction onwiki. If we can get folks to create those templates for us, it will make writing and reviewing Twinkle patches easier. Would be worth the overlap. Will go ahead and add it.

PR #2229, opened 2025-11-04
twinklearv.js: Add notice for temporary accounts

IlyasLebleu:
Exactly what it says on the tin. Wording of the notice by Tamzin, tested on test.wikipedia. Notice color set to `--morebits-color-info` but can be freely changed if needed for readability.

### DIFF
--- a/modules/twinklearv.js
+++ b/modules/twinklearv.js
@@ -81,7 +81,7 @@ Twinkle.arv.callback = function (uid, isIP) {
 	});
 
 	// Temporary account notice
-	if (mw.config.get('wgRelevantUserName') && mw.util.isTemporaryUser(mw.config.get('wgRelevantUserName'))) {
+	if (mw.config.get('wgRelevantUserName') && mw.util.isTemporaryUser(mw.config.get('wgRelevantUserName')) && !mw.config.get('wgCheckUserTemporaryAccountIPRevealAllowed')) {
 		const temporaryAccountNotice = form.append({
 			type: 'field',
 			label: 'Temporary account notice',


modules/twinklearv.js
```diff
@@ -80,6 +82,21 @@ Twinkle.arv.callback = function (uid, isIP) {
 		id: 'twinkle-arv-blockwarning'
 	});
 
+	// Temporary account notice
+	if (mw.config.get('wgRelevantUserName') && /^~2/.test(mw.config.get('wgRelevantUserName'))) {
```
NovemLinguae: can we use `mw.util.isTemporaryUser(mw.config.get('wgRelevantUserName'))` or similar here instead?

NovemLinguae (changes requested):

modules/twinklearv.js
```diff
@@ -28,6 +28,8 @@ Twinkle.arv = function twinklearv() {
 	}, 'ARV', 'tw-arv', 'Report ' + userType + ' to administrators');
 };
 
+Twinkle.arv.temporaryAccountNotice = 'You are reporting a new temporary account. These accounts, with usernames like ~2025-12345-67, have replaced IP addresses. Before reporting any temporary accounts, please familiarize yourself with [[Wikipedia:Temporary accounts|how they work]].';
```
NovemLinguae: I'm not sure we need a global variable for this. Any objection to inlining this on line 96?

modules/twinklearv.js
```diff
@@ -80,6 +80,21 @@ Twinkle.arv.callback = function (uid, isIP) {
...
+		const temporaryAccountNotice = form.append({
+			type: 'field',
+			label: 'Temporary account notice',
+			name: 'ta_notice',
+			style: 'color: var(--morebits-color-warning, #FF4500)'
+		});
+
+		temporaryAccountNotice.append({
+			type: 'div',
+			label: 'You are reporting a new temporary account. These accounts, with usernames like ~2025-12345-67, have replaced IP addresses. Before reporting any temporary accounts, please familiarize yourself with [[Wikipedia:Temporary accounts|how they work]].'
```
NovemLinguae: Sorry, one more comment. This message is kind of vague. Is there a specific incident or common misunderstanding that was the motivation for adding this? Like a certain kind of incorrect AIV report? Maybe it'd be better to give specific advice about whatever that is, rather than just linking to [[WP:TA]].

PR #2236, opened 2025-11-17
twinkleblock.js: add {{uw-tempevadeblock}} (with autoblock) and forTempAccountsOnly

IlyasLebleu:
Splits the previous {{uw-ipevadeblock}} (previously used for both IPs and TAs) in two. [Tested on test.wikipedia](https://test.wikipedia.org/wiki/User_talk:~2024-1604), the two warning options show only for IPs and TAs respectively.

### DIFF
--- a/modules/twinkleblock.js
+++ b/modules/twinkleblock.js
@@ -1092,6 +1092,7 @@ Twinkle.block.blockPresetsInfo = {
 	},
 	'uw-ipevadeblock': {
 		forIPsOnly: true,
+		expiry: '1 week',
 		nocreate: true,
 		reason: '[[WP:Blocking policy#Evasion of blocks|Block evasion]]',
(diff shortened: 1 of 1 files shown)


modules/twinkleblock.js
```diff
@@ -1381,7 +1388,8 @@ Twinkle.block.blockGroups = [
 		list: [
 			{ label: 'Advertising', value: 'uw-adblock' },
 			{ label: 'Arbitration enforcement', value: 'uw-aeblock' },
-			{ label: 'Block evasion', value: 'uw-ipevadeblock' },
+			{ label: 'Block evasion - IP', value: 'uw-ipevadeblock' },
+			{ label: 'Block evasion - TA', value: 'uw-tempevadeblock' },
```
NovemLinguae: Can we change TA to Temporary Account? Less jargony. I think folks know the acronym IP, but are less likely to know TA.

NovemLinguae (changes requested):

PR #2248, opened 2025-11-21
speedy: add admin-only "Delete" and "Decline" buttons for quick parsing

IlyasLebleu:
Technically, there is already a delete button, but it is nearly invisible, doesn't delete talk pages and redirects, doesn't log, and only takes you to the (non-Twinkle) deletion interface. This is a much quicker and more visible delete button, plus an option to decline (removing the db template). Tested [here](https://test.wikipedia.org/w/index.php?title=NewlyCreatedPage&action=history).

image

### DIFF
--- a/src/modules/twinklespeedy.js
+++ b/src/modules/twinklespeedy.js
@@ -1098,56 +1098,56 @@ Twinkle.speedy.instantDecline.main = function twinklespeedyInstantDeclineMain(pa
 	pageObj.save(window.location.reload.bind(window.location));
 };
 
+Twinkle.speedy.createCodexButton = function twinkleSpeedyCreateCodexButton({ label, action = 'neutral', weight = 'normal', size = 'medium', onClick }) {
+	const button = document.createElement('button');
+	button.classList.add('cdx-button');
+
+	// Action classes
+	if (action !== 'neutral') {
+		button.classList.add(`cdx-button--action-${action}`);
+	}
+
+	// Weight classes
+	if (weight !== 'normal') {
+		button.classList.add(`cdx-button--weight-${weight}`);
+	}
+
+	// Size classes
+	if (size !== 'medium') {
+		button.classList.add(`cdx-button--size-${size}`);
+	}
+
+	button.textContent = label;
+
+	if (typeof onClick === 'function') {
+		button.addEventListener('click', onClick);
+	}
+
+	return button;
+};
+
 // Adds a quick "Delete" button to a speedy deletion template
 Twinkle.speedy.addButton = function twinklespeedyAddButton() {
 	if (!Morebits.userIsSysop) {
 		return;
 	}
 
-	function createCodexButton({ label, action = 'neutral', weight = 'normal', size = 'medium', onClick }) {
-		const button = document.createElement('button');
-		button.classList.add('cdx-button');
-
-		// Action classes
-		if (action !== 'neutral') {
-			button.classList.add(`cdx-button--action-${action}`);
-		}
-
-		// Weight classes
-		if (weight !== 'normal') {
-			button.classList.add(`cdx-button--weight-${weight}`);
-		}
-
-		// Size classes
-		if (size !== 'medium') {
-			button.classList.add(`cdx-button--size-${size}`);
-		}
-
-		button.textContent = label;
-
-		if (typeof onClick === 'function') {
-			button.addEventListener('click', onClick);
-		}
-
-		return button;
-	}
-
 	const buttonNodes = {
-		delete: createCodexButton( { label: 'Delete', action: 'destructive', weight: 'primary', size: 'medium', onClick: Twinkle.speedy.instantDelete } ),
-		decline: createCodexButton( { label: 'Decline', action: 'neutral', weight: 'normal', size: 'medium', onClick: Twinkle.speedy.instantDecline } )
+		delete: Twinkle.speedy.createCodexButton( { label: 'Delete', action: 'destructive', weight: 'primary', size: 'medium', onClick: Twinkle.speedy.instantDelete } ),
+		decline: Twinkle.speedy.createCodexButton( { label: 'Decline', action: 'neutral', weight: 'normal', size: 'medium', onClick: Twinkle.speedy.instantDecline } )
 	};
 
 	const buttonContainer = document.createElement('div');
-	buttonContainer.classList.add('cdx-button-group'); // Codex-style grouping
-	buttonContainer.style.marginLeft = 'auto';
-	buttonContainer.style.marginRight = 'auto';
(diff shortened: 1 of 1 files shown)


modules/twinklespeedy.js
```diff
@@ -1010,6 +1013,156 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
...
+	pageObj.save(window.location.reload.bind(window.location));
+};
+
+// Adds a quick "Delete" button to a speedy deletion template
+Twinkle.speedy.addButton = function twinklespeedyAddButton() {
+	if (!Morebits.userIsSysop) {
+		return;
+	}
+
+	function createCodexButton({ label, action = 'neutral', weight = 'normal', size = 'medium', onClick }) {
```
NovemLinguae: Can we move this function up a level? I'm not a big fan of nested functions. Maybe `Twinkle.speedy.createCodexButton`?

modules/twinklespeedy.js
```diff
@@ -1010,6 +1013,156 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
...
+	buttonContainer.classList.add('cdx-button-group'); // Codex-style grouping
+	buttonContainer.style.marginLeft = 'auto';
+	buttonContainer.style.marginRight = 'auto';
+
+	// Append the buttons to the container
+	buttonContainer.appendChild(buttonNodes.delete);
+	buttonContainer.appendChild(buttonNodes.decline);
+
+	const contestButton = document.querySelector(
+	'table.ambox-speedy form[name="commentbox"].mw-inputbox-form-inline'
```
NovemLinguae: Can we add another tab character here?

NovemLinguae (changes requested):

modules/twinklespeedy.js
IlyasLebleu: Done!

modules/twinklespeedy.js
```diff
@@ -1010,6 +1013,156 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
...
+		// Size classes
+		if (size !== 'medium') {
+			button.classList.add(`cdx-button--size-${size}`);
+		}
+
+		button.textContent = label;
+
+		if (typeof onClick === 'function') {
+			button.addEventListener('click', onClick);
+		}
```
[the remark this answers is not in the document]
NovemLinguae: Up to you. Your way sounds like it might be a more proper solution, but I'm not very good at Codex, so not sure.

IlyasLebleu:
Small mistake, accidentally pushed the reload timer on this branch rather than on its own separate branch. Forced pushed it back, so it should be all good now.

modules/twinklespeedy.js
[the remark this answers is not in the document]
IlyasLebleu: I've gone with a flex with some spacing for now! Less Codex-y, but the aesthetics work better than the red and grey borders clashing.

NovemLinguae:
This is a fairly visible change for users, so I am soliciting user feedback at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#speedy%3A_add_admin-only_%22Delete%22_and_%22Decline%22_buttons_for_quick_parsing

modules/twinklespeedy.js
```diff
@@ -1010,6 +1013,156 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
...
+	buttonContainer.style.display = 'flex';
+	buttonContainer.style.gap = '1em';
+	buttonContainer.style.justifyContent = 'center';
+
+	// Append the buttons to the container
+	buttonContainer.appendChild(buttonNodes.delete);
+	buttonContainer.appendChild(buttonNodes.decline);
+
+	const contestButton = document.querySelector(
+		'table.ambox-speedy form[name="commentbox"].mw-inputbox-form-inline'
```
siddharthvp: Please add a class or id to the button and avoid such complex and fragile selectors

modules/twinklespeedy.js
```diff
@@ -24,7 +24,10 @@ Twinkle.speedy = function twinklespeedy() {
 		return;
 	}
 
+	mw.loader.load('codex-styles');
```
siddharthvp: This should be loaded only if needed (inside createCodexButton).

modules/twinklespeedy.js
```diff
@@ -1010,6 +1013,156 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
...
+		welcomeuser: welcomeuser,
+		promptForSummary: promptForSummary,
+		templateParams: templateParams
+	};
+
+	Twinkle.speedy.callbacks.sysop.main(params);
+};
+
+Twinkle.speedy.instantDecline = function twinklespeedyInstantDecline() {
+	const wikipediaPage = new Morebits.wiki.Page(mw.config.get('wgPageName'), 'Tagging page');
```
siddharthvp: "Tagging page" is not what this actually does

modules/twinklespeedy.js
```diff
@@ -1010,6 +1013,156 @@ Twinkle.speedy.initDialog = function twinklespeedyInitDialog(callbackfunc) {
...
+	if (splitTag in Twinkle.speedy.normalizeHash) {
+		if (Twinkle.speedy.normalizeHash[splitTag] === 'db') {
+				pageObj.setEditSummary('Declining speedy deletion (no code provided).');
+		} else {
+				pageObj.setEditSummary('Declining speedy deletion ([[WP:CSD#' + Twinkle.speedy.normalizeHash[splitTag].toUpperCase() + '|CSD ' + Twinkle.speedy.normalizeHash[splitTag].toUpperCase() + ']]).');
+		}
+	} else {
+		pageObj.setEditSummary('Declining speedy deletion (multiple criteria).');
+	}
+	pageObj.setChangeTags(Twinkle.changeTags);
```
siddharthvp: This doesn't handle watchlisting, which should be based on `watchSpeedyPages` pref, if not a new pref.

siddharthvp:
> We should decide if we want to only have a "Delete" button (2 clicks), only want to have a "Quick Delete" button (1 click), or have both.

I would suggest having only "Delete". The "Decline" button currently is really a "Quick Decline", but once a `prompt()` for reason is added, that would stop being the case.

IlyasLebleu:
Working on all the changes. Regarding "Delete" vs "Quick Delete", I don't see the point in making it two clicks instead of one. It makes sense for "Decline" since you'd want to input a reason, but "Delete" fundamentally means you agree with the speedy deletion tag, and adding an extra click would just make it more cumbersome (and defeat the original purpose of a shortcut button).

PR #2251, opened 2025-12-02
Support recipient-level optout for Twinkle notices

### LINKED ISSUE #2367: "Do Not Notify" registry of user talk pages
Compose a local blacklist of users to ignore when notifying the creator of a page about certain actions done with Twinkle. For the XfD tool, I'd point specifically to Hyacinth (who is deceased) and AHI-3000 (who is topic banned from categories and recently asked on their own talk page that they not be notified of CfDs involving categories they created).

siddharthvp:
Allows users to opt out from CSD/PROD/XFD notices by placing {{Twinkle optout}} on their talk page. The template creates an invisible external link to `https://optout.twinkle` with a `types` query param being the list of opted-out notice types. Opting out of specific CSD criteria notices is left for later, but opting out from specific XfD venues is supported. See https://en.wikipedia.org/wiki/Template:Twinkle_optout for usage details.

In Twinkle, we use a prop=extlinks API call to check for external links to that domain and parse the `types` parameter. An external link is used because it's the only way to create metadata that can be modified in wikitext and also accessed via the API.

This is a popular request, the most recent version being https://en.wikipedia.org/wiki/Wikipedia:Village_pump_(technical)#Mute_feature_not_working.

### DIFF
--- a/src/modules/twinkleprod.js
+++ b/src/modules/twinkleprod.js
@@ -347,7 +347,7 @@ Twinkle.prod.callbacks = {
 		return def;
 	},
 
-	notifyAuthor: function twinkleprodNotifyAuthor() {
+	notifyAuthor: async function twinkleprodNotifyAuthor() {
 		const def = $.Deferred();
 
 		if (!params.blp && !params.usertalk) {
@@ -359,6 +359,11 @@ Twinkle.prod.callbacks = {
 			Morebits.Status.info('Notifying creator', 'You (' + params.initialContrib + ') created this page; skipping user notification');
 			return def.resolve();
 		}
+
+		if (await Twinkle.hasUserOptedOutOfNotice(params.initialContrib, ['prod'])) {
+			return def.resolve();
+		}
+
 		// [[Template:Proposed deletion notify]] supports File namespace
 		let notifyTemplate;
 		if (params.blp) {
--- a/src/modules/twinklespeedy.js
+++ b/src/modules/twinklespeedy.js
@@ -1329,7 +1329,7 @@ Twinkle.speedy.callbacks = {
 		api.post();
 	},
 
-	noteToCreator: function(pageobj) {
+	noteToCreator: async function(pageobj) {
 		const params = pageobj.getCallbackParameters();
 		let initialContrib = pageobj.getCreator();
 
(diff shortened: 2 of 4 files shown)


NovemLinguae:
1) Is there enough consensus for this? The linked section is just one user (Dr. Blofeld) requesting it, I think.

2) The situation in the linked section was Dr. Blofeld wanted to stop JoeNMLC from leaving Twinkle messages. I don't think he said he wanted all Twinkle messages stopped. Makes me wonder if the template should support blocking notifications from a particular user as one of its params.

3) Even though you put in the docs that this shouldn't be included in templates such as {{Not around}} and {{Deceased}}, somehow I have a feeling this might eventually make its way in there, placed by someone who didn't read the docs.

NovemLinguae:
Some additional thoughts about the consensus part of this:

https://github.com/wikimedia-gadgets/twinkle/issues/2367#issuecomment-4476663884

> People ask this frequently, but people also oppose this frequently. Here's some past WT:TWINKLE discussions about it (not a complete list):
> 
> https://en.wikipedia.org/w/index.php?search=deceased&prefix=Wikipedia+talk%3ATwinkle%2F&title=Special%3ASearch&profile=advanced&fulltext=1&ns0=1
> 
> This might need an RFC to move forward. It is both frequently requested and frequently opposed.

NovemLinguae:
Related: https://en.wikipedia.org/wiki/Wikipedia:Village_pump_(proposals)/Archive_225#Have_Twinkle_stop_notifying_blocked_users_of_deletion_nominations%3F

siddharthvp:
I think this is a feature which is often mis-requested in more general forms like "opt out all blocked users" or "opt out all deceased users" which indeed don't have consensus. Allowing individual opt-outs specifically has not come up frequently, but when it has come up I have not heard anyone opposing it. The template prevents itself from being included in other templates [due to the namespace check](https://en.wikipedia.org/wiki/Module:Twinkle_optout#L-18). It's still possible to include the external link manually, but I don't think that's going to happen.

> The situation in the linked section was Dr. Blofeld wanted to stop JoeNMLC from leaving Twinkle messages. I don't think he said he wanted all Twinkle messages stopped. Makes me wonder if the template should support blocking notifications from a particular user as one of its params.

I believe he wants to opt out of all automated messages altogether, not just Twinkle ones. But sure, we can add a param to specify usernames to block.

PR #2256, opened 2025-12-04
xfd: avoid using mw.Title getNameText()

### LINKED ISSUE #2162: xfd: cfd rename garbles "target category" when there's periods in the target category title
Reported by Paradoctor at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Target_field_not_correctly_processed_in_CfD_request

Target category `Category:Road to ... (Family Guy)` gets garbled by Twinkle to `Road to ..`

Image

siddharthvp:
~~mw.Title.newFromUserInput uses unusual sanitization rules such as removing everything after `..` if it appears in a title. Use mw.Title.newFromText instead. This function also returns null instead of throwing on invalid input, so works fine here.~~

getNameText() returns the page name without the extension which creates problems with non-file page titles that have dot in them. Use getMainText() instead.

Closes #2162

### DIFF
--- a/src/modules/twinklexfd.js
+++ b/src/modules/twinklexfd.js
@@ -69,7 +69,7 @@ const utils = {
 		if (!title_obj) {
 			return title; // user entered invalid input; do nothing
 		}
-		return title_obj.getNameText();
+		return title_obj.getMainText();
 	},
 
 	/**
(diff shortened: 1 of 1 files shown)


siddharthvp:
Working fine now. The issue was actually due to `getNameText`, not `newFromUserInput`.

PR #2263, opened 2025-12-06
move .md files in the root to /docs/ folder

NovemLinguae:
De-clutters the root folder. Similar goal as moving all our .js files to the /src/ folder.

I think GitHub is smart enough to auto detect these. Another repo has these in the /docs/ folder, and the README and all the tabs still display just fine on the repo homepage.


siddharthvp:
I'd suggest not doing this. It does look like GitHub detects them fine even when under docs/. But humans might still look for them in root as that's the popular convention. Also it might confuse other tools that expect them in root. For example, some IDEs like IntelliJ have a GUI for cloning repos, they automatically open the root README.md after cloning completes.

For decluttering the root folder further, there are other things that could be done instead:
* Removing .gitattributes as it does nothing
* Merging .eslintignore into package.json

PR #2266, opened 2025-12-07
Make linter pass without warnings

siddharthvp:
`npm run lint` currently raises over 800 warnings, which makes it hard to find any newly introduced errors, or even newly introduced warnings. I doubt anyone is paying attention to warnings when there are so many of them.

While we should probably fix these rule violations some day, for now I think it's better to set them to `off` instead of `warn`. I fixed the jsdoc lint rules not mentioned in the conf file which were also raising warnings.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -30,25 +30,25 @@
 
 		// non-default rules we do like
 		"no-nested-ternary": "error",
-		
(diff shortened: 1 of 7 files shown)


siddharthvp:
Makes sense, done.

PR #2269, opened 2025-12-07
morebits: use change tags on action=pagetriageaction requests

### LINKED ISSUE #1944: When marking an article as reviewed, Twinkle should use the "tags" param of the pagetriageaction API to apply a Twinkle tag
That way Special:Log entries are tagged with Twinkle. This would be in addition to the PageTriage tag or instead of it (either is fine, I haven't tested it yet).

image

image

siddharthvp:
Supported in MediaWiki since November 2022 (https://phabricator.wikimedia.org/T252980)

Closes #1944

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -34,7 +34,6 @@
 		// rules we should probably fix someday
 		"camelcase": ["off", {"properties": "never"}],
 		"mediawiki/class-doc": "off",
-		"new-cap": "off",
 		"no-jquery/no-done-fail": "off",
 		"no-jquery/no-each-util": "off",
 		"no-jquery/no-extend": "off",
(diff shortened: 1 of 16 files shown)


src/modules/twinkleprod.js
```diff
@@ -311,18 +311,18 @@ Twinkle.prod.callbacks = {
...
 			pageobj.setEditSummary(summaryText);
 			pageobj.setChangeTags(Twinkle.changeTags);
 			pageobj.setWatchlist(Twinkle.getPref('watchProdPages'));
 			pageobj.setCreateOption('nocreate');
 			pageobj.save(def.resolve, def.reject);
 
+			// curate/patrol the page
+			if (Twinkle.getPref('markProdPagesAsPatrolled')) {
+				pageobj.triage();
+			}
```
siddharthvp: This block is moved down so that by this time `.setChangeTags()` has been called on `pageobj`.

PR #2275, opened 2025-12-11
Add {{uw-gender}}

owenblacker:
Added new template {{uw-gender}} to gentle warn users about misgendering BLP subjects.

Summary wording chosen to avoid the word "misgendering", which might feel accusatory or controversial to some editors, so a more neutral suggestion to read MOS:GENDERID is used instead.

### DIFF
--- a/src/modules/twinklewarn.js
+++ b/src/modules/twinklewarn.js
@@ -1138,6 +1138,11 @@ Twinkle.warn.messages = {
 			label: 'Not communicating in English',
 			summary: 'Notice: Not communicating in English'
 		},
+		'uw-gender': {
+			label: 'Misgendering BLPs',
+			summary: 'Notice: Please read the Manual of Style section on gender identity',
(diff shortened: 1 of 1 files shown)


NovemLinguae:
Probably not used enough to be included in Twinkle's default installation. However, it is easy for an individual user to add custom templates just for themselves:

https://en.wikipedia.org/wiki/Wikipedia:Twinkle/Preferences

image

@siddharthvp , perhaps we should decide on a heuristic/rule for what templates qualify for inclusion in Twinkle, then document it in one of the readme files. For example, >25 or >50 uses of a subst'd template.

siddharthvp:
> @siddharthvp , perhaps we should decide on a heuristic/rule for what templates qualify for inclusion in Twinkle, then document it in one of the readme files. For example, >25 or >50 uses of a subst'd template.

#533 was created to discuss that, but received no inputs.

PR #2288, opened 2026-01-01
arv: Remove colon causing multi-colon escape lint errors when reporting to AIV

TenshiSWR:
When a user specifies a page in the "Primary linked page" field, the code adds an extra : unnecessarily which {{No redirect}} will filter out and link to the page anyway (but still causes a lint error), this leads to multi-colon escape lint errors being reported across AIV's transclusions (which is around 200 pages or so).

See also [here](https://en.wikipedia.org/wiki/Wikipedia_talk:Linter#No_redirect) where it was reported on WT:Linter.


siddharthvp:
Makes sense. The colon seems redundant as {{no redirect}} handles files/categories fine even without it.

PR #2290, opened 2026-01-03
tag: add parameters to {{R from sort name}}

### LINKED ISSUE #2281: tag: add parameters to {{R from sort name}}
For {{R from other language}} and for {{R avoided double redirect}}, when selected Twinkle will provide a box to fill in the parameter (or parameters, for alt lang). Could we get the same option for {{R from sort name}}, which is supposed to pass two parameters? {{{1}}} being the first letter of the redirect, and {{{2}}} being the first letter of the target.

I copied this from WT:TW, but this is my first time ever using github, so I hope I'm doing it right!

User-IAmChaos:
I still don't know how to use github and I definitely messed up at least something. I think I ended up copying the entire Twinkle code to my own account? I definitely didnt mean to and I hope I didnt break copyright law or anything. Please check my work. I think this addresses Issue 2281, per my comments on that issue.


NovemLinguae:
Looks like you got the GitHub side of things correct. "Forking" a repo to make your changes is the way you're supposed to do it and that's what you did. The idea is that you don't have edit access to this repo, but can edit your fork, so the fork is where you store your git branch. Copyright isn't a problem since all the code is open source.

You can edit your original post to end in "fixes #2281" to automatically link this PR to that issue. Would recommend you do that.

Did you get a chance to load up your patch in localhost, do `npm start`, then test it on test.wikipedia.org? That step is helpful to make sure your patch is bug free and works: https://github.com/wikimedia-gadgets/twinkle/blob/master/CONTRIBUTING.md#testing-your-code

NovemLinguae (changes requested):

src/modules/twinkletag.js
```diff
@@ -1022,7 +1022,27 @@ Twinkle.tag.redirectList = {
 			{ tag: 'R from more specific name', description: 'redirect from a more specific title to a less specific, more general one' },
 			{ tag: 'R from non-neutral name', description: 'redirect from a title that contains a non-neutral, pejorative, controversial, or offensive word, phrase, or name' },
 			{ tag: 'R from short name', description: 'redirect from a title that is a shortened form of a person\'s full name, a book title, or other more complete title' },
-			{ tag: 'R from sort name', description: 'redirect from the target\'s sort name, such as beginning with their surname rather than given name', restriction: 'insideMainspaceOnly' },
+			{ tag: 'R from sort name', description: 'redirect from the target\'s sort name, such as beginning with their surname rather than given name', restriction: 'insideMainspaceOnly' ,
+				subgroup: [
```
NovemLinguae, line 1025: Let's get tag, description, and restriction on their own lines, since subgroup is on its own line.

NovemLinguae:
Manual testing of this patch doesn't work. I don't see the 3 elements you added.

image

src/modules/twinkletag.js
```diff
@@ -1021,7 +1021,27 @@ Twinkle.tag.redirectList = {
...
+						label: 'Page Title Initial',
+						tooltip: 'Enter the first letter of the target page.'
+					},
+					{
+						name: 'SortName Info',
+						type: 'div',
+						label: $.parseHTML('<p>Please provide the required parameters for sort name redirects. See <a href="/wiki/Template:R_from_sort_name#Parameters">the template page for more info</a></p>')
+					}
+				]
+			},
```
NovemLinguae, line 1044: I don't see any code that will change the wikitext once the form information is collected. You'll probably need to pair these form changes with some more code changes farther down, and the farther down code changes will change the output of this tag's wikitext.

NovemLinguae (changes requested):

Issue #2300, opened 2026-01-22
Remove jquery.ui from Twinkle


siddharthvp:
jQuery UI is deprecated, and at [74 kb](https://en.wikipedia.org/w/load.php?modules=jquery.ui), it is a bit heavy to load on each page view.

We use it for dialogs and tooltips – neither of which are that hard to replicate from scratch. Especially as web standards have progressed: the [`<dialog>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) has been baseline since 2022. The main selling points of jquery.ui dialogs are draggability and resizability (neither of which are available in OOUI, by the way). Native [CSS support for resizing](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/resize) might also be ready for wide usage.

siddharthvp:
Done in #2309 and #2310. HTML `<dialog>` was not used. CSS `resize` was initially used but I had to abandon it as it was creating more trouble than it's worth. Dialogs and resizing are both implemented from scratch in JS.

PR #2304, opened 2026-01-27
Format and lint CSS files using Stylelint

siddharthvp:
Use stylelint with stylelint-config-wikimedia, specifically the support-modern preset which targets grade A browsers. This is in preparation for #2300 which will introduce some involved CSS.

Rule customisations used:
* @stylistic/*-space-inside: "never", per repo code style of not putting spaces in parens
* selector-attribute-quotes: "never".
* selector-max-id: 1. MediaWiki uses 0 which disallows ids in CSS completely! That leads to a lot of errors for us.
* declaration-no-important: null. Disabled temporarily. Ideally, we should remove this.

All CSS changes are no-op, mainly:
* Changed hex codes to use lowercase letters instead of uppercase
* Consistently used tabs
* Consistently put opening brace on same line as the selector
* The two blocks for .morebits-ui-tooltip in morebits.css are combined
* prefers-color-scheme: dark isn't supported in Safari 11 which still happens to be Grade A, so we have to use per-line stylelint-disable. In practise, Safari 11 isn't much used, and per CSS spec unsupported rules and media queries are just ignored - they don't cause the stylesheet to fail.
* stylelint-config-wikimedia disallows px units in font-size and line-height for accessibility reasons, which is probably something we should also fix. Temporarily disabled inline as fixes won't be no-op.
* stylelint-config-wikimedia disallows `border: none`. It is replaced with `border: 0` which has the same effect.

### DIFF
--- a/src/modules/twinklearv.js
+++ b/src/modules/twinklearv.js
@@ -86,7 +86,7 @@ Twinkle.arv.callback = function (uid, isIP) {
 			type: 'field',
 			label: 'Temporary account notice',
 			name: 'ta_notice',
-			style: 'color: var(--morebits-color-warning, #FF4500)'
+			style: 'color: var(--morebits-color-warning)'
 		});
 
 		temporaryAccountNotice.append({
--- a/src/morebits.css
+++ b/src/morebits.css
@@ -50,19 +50,19 @@ html {
 /* Morebits.Status */
 
 .morebits_status_status {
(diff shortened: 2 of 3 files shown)


NovemLinguae:
Oh duh. `"lint": "eslint . --max-warnings=0 && stylelint src/*.css",`

PR #2309, opened 2026-01-31
morebits: port SimpleWindow from jQuery UI to plain JS 

siddharthvp:
Looks and behaves exactly like the existing UI. There is also not much difference in the number of lines of code. It turns out that using jquery.ui and applying our tweaks over it wasn't really making the code lesser than re-implementing the dialog from scratch.

It took more than a week to get all the rough edges straightened out. I think the result is cleaner and more maintainable than the previous version. In particular, the CSS no longer needs to use `!important` or use over-specific selectors or `inherit` values to override jquery.ui behaviour.

The only visible changes are:
* Resizing is only supported from the bottom-right corner (similar to CSS native resize). Earlier, it was supported from all four corners and four edges. We could do that if there's demand.
* The close button looks a bit different (but behaves similarly).

Any other differences from the previous UI should be reported as a bug.

Partially solves #2300.

### DIFF
--- a/src/morebits.css
+++ b/src/morebits.css
@@ -69,10 +69,20 @@ html {
 /* Morebits.QuickForm */
 
 form.quickform {
-	width: 96%;
-	margin: auto;
+	box-sizing: border-box;
+	width: 98%;
+	height: 100%;
+	min-height: 0;
+	margin: 0 auto;
 	padding: 0.5em;
 	color: var(--color-base, #202122);
+}
+
+/* use flex layout for quickform to make the scrollbox resize with
+   the dialog, and to keep it from obscuring elements below it */
+form.quickform.has-scrollbox {
+	display: flex;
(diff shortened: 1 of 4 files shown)


src/morebits.css
```diff
@@ -220,11 +232,28 @@ div.morebits-usertext {
...
 	background-color: var(--morebits-bgcolor-dialog);
 	color: inherit;
 	background-image: none;
+
+	/* Make it a flex box (covering titlebar, content, buttonpane) */
+	display: flex;
+	flex-direction: column;
+
+	/* Make resizable and scrollable */
+	resize: both; /* stylelint-disable-line plugin/no-unsupported-browser-features */
```
siddharthvp: Safari on macOS doesn't show the slanted lines at the bottom-right. But it works alright if you know it's clickable.

PR #2311, opened 2026-02-08
block: rewrite user type options for more modularity + adjust some

IlyasLebleu:
Replaced the `forIPsOnly`, `forTempAccountsOnly` and `forRegisteredOnly` by a single `userTypes` array that can take `'ip'`, `'ta'` and `'named'` as values, to cover more options without multiplying variables. Allowed me to also fix a few that couldn't be treated previously (such as username blocks not showing for TAs anymore). Tested on test.wikipedia.org, block options menu works exactly as intended.

### DIFF
--- a/src/modules/twinkleblock.js
+++ b/src/modules/twinkleblock.js
@@ -1180,7 +1180,7 @@ Twinkle.block.blockPresetsInfo = {
 		summary: 'You have been blocked from editing for using user and/or article pages as a [[WP:NOTMYSPACE|blog, web host, social networking site or forum]]'
 	},
 	'uw-sockblock': {
-		userTypes: ['ta', 'named'],
+		userTypes: ['ip', 'ta', 'named'],
 		autoblock: true,
 		nocreate: true,
 		reason: 'Abusing [[WP:Sock puppetry|multiple accounts]]',
--- a/src/modules/twinkleblock.js
+++ b/src/modules/twinkleblock.js
@@ -836,9 +836,6 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
  *   autoblock: <autoblock any IP addresses used (for registered users only)>
  *   disabletalk: <disable user from editing their own talk page while blocked>
  *   expiry: <string - expiry timestamp, can include relative times like "5 months", "2 weeks" etc>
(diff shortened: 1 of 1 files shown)


src/modules/twinkleblock.js
```diff
@@ -854,69 +851,71 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
...
 	colocationwebhost: {
+		userTypes: ['ip'],
 		expiry: '1 year',
-		forIPsOnly: true,
 		nonstandard: true,
 		reason: '{{colocationwebhost}}',
 		sig: null
 	},
 	oversightblock: {
+		userTypes: ['ip', 'ta', 'named'],
```
siddharthvp: I think we can treat this as the default. No need to write it out for every preset.

src/modules/twinkleblock.js
```diff
@@ -1099,128 +1109,137 @@ Twinkle.block.blockPresetsInfo = {
...
 		autoblock: true,
-		forRegisteredOnly: true,
 		nocreate: true,
 		reason: 'Abusing [[WP:Sock puppetry|multiple accounts]]',
 		summary: 'You have been blocked from editing for abusing [[WP:SOCK|multiple accounts]]'
 	},
 	'uw-softerblock': {
+		userTypes: ['named'],
 		expiry: 'infinity',
 		forRegisteredOnly: true,
```
NovemLinguae, line 1192: Can remove `forRegisteredOnly: true` here, I think.

src/modules/twinkleblock.js
```diff
@@ -1030,32 +1034,33 @@ Twinkle.block.blockPresetsInfo = {
...
 		expiry: 'infinity',
-		forRegisteredOnly: true,
 		reason: '{{uw-causeblock}} <!-- Username represents a non-profit, soft block -->',
 		summary: 'You have been indefinitely blocked from editing because your [[WP:U|username]] gives the impression that the account represents a group, organization or website'
 	},
 	'uw-compblock': {
+		userTypes: ['named'],
 		autoblock: true,
 		expiry: 'infinity',
 		forRegisteredOnly: true,
```
NovemLinguae, line 1066: Can remove `forRegisteredOnly: true` here, I think.

NovemLinguae (changes requested):

PR #2313, opened 2026-02-08
block: turn templating on by default only for registered accounts

IlyasLebleu:
Usually, IPs don't need a block template since it won't be visible, although the option is left available in case an administrator might want to leave a notice there for future use. Works as expected on test.wikipedia.org


NovemLinguae (changes requested):
Marking as "requested changes" since there's an unresolved code review comment.

PR #2315, opened 2026-02-11
block: add preset for LTA

IlyasLebleu:
Adds a quick LTA preset (no template by default, everything disabled), tested at https://test.wikipedia.org/wiki/Special:Contributions/~2024-1844

### DIFF
--- a/src/modules/twinkleblock.js
+++ b/src/modules/twinkleblock.js
@@ -845,6 +845,7 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
  *   prependReason: <string - prepends the value of 'reason' to the end of the existing reason, namely for when revoking talk page access>
  *   nocreate: <block account creation from the user's IP (for unregistered users only)>
  *   nonstandard: <template does not conform to stewardship of WikiProject User Warnings and may not accept standard parameters>
+ *   notemplate: <do not display a template by default>
  *   reason: <string - block rationale, as would appear in the block log,
  *            and the edit summary for when adding block template, unless 'summary' is set>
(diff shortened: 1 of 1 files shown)


src/modules/twinkleblock.js
```diff
@@ -1359,6 +1360,19 @@ Twinkle.block.blockPresetsInfo = {
...
+		autoblock: true,
+		disabletalk: true,
+		expiry: 'infinity',
+		hardblock: true,
+		nocreate: true,
+		noemail: true,
+		notemplate: true,
+		pageParam: false,
+		reason: 'Long-term abuse',
+		templateName: 'uw-blocknotalk'
```
NovemLinguae, line 1375: Could we code this so that, instead of having `notemplate: true`, that we instead leave templateName blank or null or undefined, and the system detects a falsey templateName, and in that situation doesn't leave a user talk template?

NovemLinguae (changes requested):

NovemLinguae:
I asked the talk page WT:TWINKLE to double check the consensus and details of this idea: https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Admins%2C_interested_in_an_%22LTA%22_block_preset%3F

NovemLinguae:
> I asked the talk page WT:TWINKLE to double check the consensus and details of this idea: https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Admins%2C_interested_in_an_%22LTA%22_block_preset%3F

Consensus check looks good. No problems there. Next step is to resolve the code review comment.

PR #2316, opened 2026-02-11
block: additional options for partial blocks

IlyasLebleu:
Adds options to block users from creating pages, moving pages, thanking other users or uploading/overwriting pages when selecting a partial block. Tested on https://test.wikipedia.org/wiki/Special:Contributions/~2024-2520

### DIFF
--- a/src/modules/twinkleblock.js
+++ b/src/modules/twinkleblock.js
@@ -424,6 +424,42 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 					ns.append({ type: 'option', label: name, value: number });
 				}
 			});
+			const partialblockoptions = [{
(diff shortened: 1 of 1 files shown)


src/modules/twinkleblock.js
```diff
@@ -424,6 +424,44 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 					ns.append({ type: 'option', label: name, value: number });
 				}
 			});
+			const partialblockoptions = [{
```
siddharthvp: variable can be inlined

NovemLinguae (changes requested):
Marking as "changes requested" since there are unresolved code review comments.

Issue #2335, opened 2026-04-04
xfd: add support for preferred outcome (delete, merge, redirect, draftify) at AFD


NovemLinguae:
Requested by Voorts at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Updates_to_AfD_backend

Mostly copy/pasted from https://github.com/wikimedia-gadgets/twinkle/issues/2335#issuecomment-4188615635, with thanks to FaviFake:

* [x] Update XFD -> AFD to add a dropdown that lets the user select their preferred outcome, and also do some additional things based on what is selected:
  * **Delete**
  * **Merge**
    1. On the nominated page:
       * `{{Article for deletion/dated |outcome=merging |...}}`
  * **Redirect**
    1. On the nominated page:
       * `{{Article for deletion/dated |outcome=redirecting |...}}`
  * **Draftify**
    1. On the nominated page:
       * `{{Article for deletion/dated |outcome=draftification |...}}`
* [x] **Draftify**
    1. On the author's user talk page, place this:
       * `{{subst:Afd notice |1=SourcePage |outcome=draftification}}`
    2. Pre-fill some draftify text in the AFD reason textarea that is published on the AFD page, i.e. "Propose '''draftifying'''
* [x] **Redirect**
    1. Collect "TargetPage" in a text box
    2. On the nominated page:
       * `{{Article for deletion/dated |target=TargetPage |...}}`
    3. On the author's user talk page:
       * if "TargetPage" was given, place this:
         * `{{subst:Afd notice |1=SourcePage |target=TargetPage |outcome=redirecting}}`
       * if "TargetPage" was not given, place this:
         * `{{subst:Afd notice |1=SourcePage |outcome=redirecting}}`
    4. Pre-fill some redirect text in the AFD reason textarea that is published on the AFD page, i.e. "Propose '''redirecting''' [[]] to [[]]"
* [x] **Merge**
    1. Collect "TargetPage" in a text box
    2. On the nominated page:
       * `{{Article for deletion/dated |target=TargetPage |...}}`
    3. if "TargetPage" was given, place this on the target page (always include NominationName):
        * `{{Merge from |1=SourcePage |target=TargetPage |afd=NominationName |date ={{subst:CURRENTMONTHNAME}} {{subst:CURRENTYEAR}} }}`
    4. On the author's user talk page:
       * if "Target" was given, place this:
         * `{{subst:Afd notice |1=SourcePage |target=TargetPage |outcome=merging}}`
[...]

PR #2338, opened 2026-04-11
xfd: add "desired outcome" dropdown

NovemLinguae:
Why
- a recent RFC decided to bring the merging process into the AFD process

What
- add dropdown called "Desired outcome:" with options Delete, Merge, Redirect, and Draftify
- add |outcome= to tagging wikicode when merge, redirect, or draftify are selected
- refactor tagging wikicode to not use ternaries (more readable)

Note that Twinkle uses {{subst:Afd}} and {{subst:Afdx}}, not {{subst:Article for deletion/dated}} directly. Those two templates will need to be modified with template editor edit requests. That doesn't need to hold up this patch though.

Example wikicode outputted by this patch:

Delete:
- `{{subst:afd|help=off}}\n`
- `{{subst:afdx|18th|help=off}}\n`

Merge:
- `{{subst:afd|help=off|outcome=merging}}\n`
- `{{subst:afdx|18th|help=off|outcome=merging}}\n`

If you want the params in a different order, let me know.

Related #2335

### DIFF
--- a/src/modules/twinklexfd.js
+++ b/src/modules/twinklexfd.js
@@ -302,7 +302,6 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 
 	form.previewer.closePreview();
 
-	let outcome;
 	switch (value) {
 		case 'afd':
 			work_area = new Morebits.QuickForm.Element({
@@ -318,31 +317,32 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 				style: 'margin-bottom: 5px; margin-top: -5px;'
 			});
 
-			outcome = work_area.append({
+			work_area.append({
 				type: 'select',
 				name: 'outcome',
-				label: 'Desired outcome:'
-			});
-
-			outcome.append({
-				type: 'option',
-				label: 'Delete',
-				value: 'deletion'
-			});
-			outcome.append({
-				type: 'option',
-				label: 'Merge',
-				value: 'merging'
-			});
-			outcome.append({
-				type: 'option',
-				label: 'Redirect',
-				value: 'redirecting'
-			});
-			outcome.append({
-				type: 'option',
-				label: 'Draftify',
-				value: 'draftification'
+				label: 'Desired outcome:',
+				list: [
+					{
+						type: 'option',
+						label: 'Delete',
+						value: 'deletion'
+					},
+					{
+						type: 'option',
+						label: 'Merge',
+						value: 'merging'
+					},
+					{
+						type: 'option',
+						label: 'Redirect',
+						value: 'redirecting'
+					},
+					{
+						type: 'option',
+						label: 'Draftify',
+						value: 'draftification'
+					}
+				]
 			});
 
 			work_area.append({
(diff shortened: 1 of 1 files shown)


NovemLinguae (commented):
@FaviFake

- please let me know if this patch looks OK
- can you please handle the necessary code changes and template editor edit requests to Template:Afd and Template:Afdx?

FaviFake:
Yes, this looks good as a first step! (Maybe the outcome parameter could come first to help distinguish the proposal more quickly? But that's an incredibly minor change.)

I wasn't aware of the fact that Twinkle used 2 other templates which need to be updated, so I've pinged Chaotic Enby to see if they're willing to help as they did with the /dated template. Btw, why doesn't Twinkle use the /dated version? Wouldn't it be easier to maintain in the future (e.g., if another process is merged) if everything relied on a single template?

NovemLinguae:
No idea of the coding decision history here.

I suspect updating those two templates to pass through the |outcome= parameter would be trivial though. Would just need some code such as `|outcome={{{outcome}}}`, and maybe some code to not write that if {{{outcome}}} is empty.

FaviFake:
@NovemLinguae [Aaaaand that's done too!](https://en.wikipedia.org/wiki/Wikipedia_talk:Articles_for_deletion#c-Chaotic_Enby-20260411162500-FaviFake-20260411103000) You should be able to publish this update :)

NovemLinguae:
Nice. Looks good. I see Twinkle code generating merge templates now in testing.

Will leave this open for another day in case @siddharthvp wants to comment.

image

src/modules/twinklexfd.js
```diff
@@ -317,6 +318,33 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 				style: 'margin-bottom: 5px; margin-top: -5px;'
 			});
 
+			outcome = work_area.append({
+				type: 'select',
+				name: 'outcome',
+				label: 'Desired outcome:'
+			});
+
+			outcome.append({
```
siddharthvp: `list` param can be used with type: select to add all options together. No need to add each one separately.

src/modules/twinklexfd.js
```diff
@@ -1179,8 +1207,27 @@ Twinkle.xfd.callbacks = {
 				Twinkle.xfd.callbacks.addToLog(params, null);
 			}
 
-			params.tagText = (params.noinclude ? '<noinclude>{{' : '{{') + (params.number === '' ? 'subst:afd|help=off' : 'subst:afdx|' +
-					params.number + '|help=off') + (params.noinclude ? '}}</noinclude>\n' : '}}\n');
+			let noIncludeStart = '';
+			let noIncludeEnd = '';
+			if ( params.noinclude ) {
```
siddharthvp: extra spaces in parens

src/modules/twinklexfd.js
```diff
@@ -318,31 +317,32 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
-				value: 'redirecting'
-			});
-			outcome.append({
-				type: 'option',
-				label: 'Draftify',
-				value: 'draftification'
+				label: 'Desired outcome:',
+				list: [
+					{
+						type: 'option',
```
siddharthvp: type: option is redundant. List items in `select` are always options.

siddharthvp:
There seems to be a lot of duplicate code in #2342 and #2345. Can we merge them all into a single PR? They can still remain separate commits.

NovemLinguae:
Unless I'm missing something, GitHub doesn't support patch chains nearly as well as Gerrit. So you basically have to review and merge the first PR in the chain, then rebase the second one. On rebase, the duplication goes away. Then repeat that process for subsequent PRs.

If you prefer to review everything at once, you can close PRs 1 and 2, and just review patch the latest PR (currently PR 3). PR 3 contains all changes between 1 and 3.

NovemLinguae:
Combining 3 patches into one at #2345

PR #2363, opened 2026-05-13
welcome: remove duplicate header for non-english welcome

### LINKED ISSUE #2358: welcome: {{welcomeen-nl}} adds ==Welcome== twice
Steps to reproduce:
- navigate to a user talk page (seen [here](https://en.wikipedia.org/wiki/User_talk:Suzannewelp) as an example)
- open TW --> welcome
- navigate to non-english welcomes
- click on {{welcomeen-nl}} and hit submit

What happens:
- message is pasted on talk page with two Level 2 headings with Welcome, and a Level 3 heading for Welcome in Dutch
Image

What should happen:
- presumably not this, either there should only be one Level 2 heading or two separate English Welcome headers; one in Level 2 and one in Level 3

Notes:
- not sure whether other non-english welcome templates have it, might do some testing later if needed.

rustykitty:
Fixes #2358 and fixes the rest of the foreign language welcome templates too.

Testing: https://en.wikipedia.org/w/index.php?title=User_talk:Sandbox_for_user_warnings&diff=prev&oldid=1353891806

### DIFF
--- a/src/modules/twinklewelcome.js
+++ b/src/modules/twinklewelcome.js
@@ -528,131 +528,131 @@ Twinkle.welcome.templates = {
 			},
 			'welcomeen-ar': {
 				description: 'welcome for users whose first language appears to be Arabic',
-				syntax: '== Welcome! ==\n{{subst:welcomeen-ar}}'
+				syntax: '{{subst:welcomeen-ar}}'
 			},
 			'welcomeen-sq': {
 				description: 'welcome for users whose first language appears to be Albanian',
-				syntax: '== Welcome! ==\n{{subst:welcomeen-sq}}'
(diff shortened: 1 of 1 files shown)


src/modules/twinklewelcome.js
```diff
@@ -528,131 +528,131 @@ Twinkle.welcome.templates = {
 			},
 			'welcomeen-ar': {
 				description: 'welcome for users whose first language appears to be Arabic',
-				syntax: '== Welcome! ==\n{{subst:welcomeen-ar}}'
```
siddharthvp, line 531: This and welcomeen-it are using a level-3 header on-wiki. Might need an on-wiki fix.

Issue #2370, opened 2026-05-18
Tag menu fails to parse tags with dashes correctly


IlyasLebleu:
After tagging a page with {{AI-generated}}, the tag menu instead retrieves it as {{AI generated}} in the "tags already present" list. This also seems to affect tag removal, as unchecking the tag fails to remove it.

<img width="804" height="488" alt="Image" src="https://github.com/user-attachments/assets/4a49a6bc-20ad-43c1-bae6-16d13bc90c81" />

NovemLinguae:
When debugging this, keep in mind that Twinkle checks the HTML ID or class being emitted by the template in order to identify templates presently in the article. This is counter-intuitive but could be involved here.

PR #2371, opened 2026-05-18
Tag: Add parameter fields for {{AI-generated}} tag

IlyasLebleu:
Cf. [on-wiki discussion](https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Feature_request:_additional_parameters_for_tagging_as_AI-generated), currently going for the layout that is more in-line with what is already being done (e.g. short reason field for tag + talk page section name) instead of having a free-form field for a talk page addition. That separate, more experimental layout is also available in the intermediate commit being pushed for reference.

Tested on testwiki as usual, [here](https://test.wikipedia.org/w/index.php?title=Meow_Meow_Meow&diff=prev&oldid=743063) (but note #2370 which affects both this branch and the main production branch).


NovemLinguae (commented):
Tested, works. Thanks!

PR #2374, opened 2026-05-25
prod: Add "Llm Prod" function to PRODing

### LINKED ISSUE #2362: prod: Add "Llm Prod" function to PRODing
Recently it was decided that a separate PROD template specifically for articles created entirely with LLMs and AI-generated content should be created. It can be found [here](https://en.wikipedia.org/wiki/Template:Prod_llm). I would like for this template, and the functions that go with it, to be added to the PROD function of Twinkle, similarly to how BLP-PRODs are their own separate function in Twinkle's PROD menu.

xzkdeng:
This addes LLMPROD function to PROD (fixes #2362).
- Add radio buttons with options about why the article meets [WP:LLMPROD#Requirements 1](https://en.wikipedia.org/wiki/Wikipedia:Presumptive_removal_of_AI-generated_content#Requirements), and a textbox for additional reasons.
- Sometimes add a input "at" to link to discussion about the presumptive removal such as on WP:AINB or WP:ANI.
- The "at" input and option from the radio button are then used to generate a string like "There is consensus at [[WP:ANIB]] that presumptive removal is appropriate.", which together with additional reasons from the user, becomes the "|concern=" param of the {{Prod llm}} template.
- Things like log and notify are also updated.

I don't know how others think of the "semi-automatic" reason string. Should there be just a single textbox like the normal prod instead?

### DIFF
--- a/src/modules/twinkleprod.js
+++ b/src/modules/twinkleprod.js
@@ -45,6 +45,7 @@ Twinkle.prod.callback = function twinkleprodCallback() {
 	if (namespace === 'article') {
 		Window.addFooterLink('Proposed deletion policy', 'WP:PROD');
 		Window.addFooterLink('BLP PROD policy', 'WP:BLPPROD');
+		Window.addFooterLink('LLM PROD policy', 'WP:LLMPROD');
 	} else { // if file
 		Window.addFooterLink('Proposed deletion policy', 'WP:PROD');
 	}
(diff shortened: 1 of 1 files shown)


jhd0919:
> I don't know how others think of the "semi-automatic" reason string. Should there be just a single textbox like the normal prod instead?

As some unfamiliar with said string and how it works, I'm deferring to others on this. What do others think?

Kowal2701:
happy to help answer questions/offer feedback if needed. I currently use https://en.wikipedia.org/wiki/User:Kowal2701/LLMPROD to do this

PR #2376, opened 2026-05-31
block: add parameter for COIVRT tickets in uw-upeblock

IlyasLebleu:
Requested by Asilvering. Logic taken from the DS reason form, [tested here](https://test.wikipedia.org/wiki/User_talk:Lawful_Binary). Helpful for VRT blocks as the parameter can't be filled from Twinkle otherwise. Also syncs edit summary.

### DIFF
--- a/src/modules/twinkleblock.js
+++ b/src/modules/twinkleblock.js
@@ -352,6 +352,7 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 	// field will linger. No need to keep the old value around, so just
 	// remove it; saves trouble when hiding/evaluating
 	$form.find('[name=dstopic]').parent().remove();
+	$form.find('[name=ticket]').parent().remove();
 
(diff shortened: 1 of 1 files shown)


src/modules/twinkleblock.js
```diff
@@ -564,6 +565,17 @@ Twinkle.block.callback.change_action = function twinkleblockCallbackChangeAction
 		tooltip: 'If selected, it will inform the template and may be added to the blocking message',
 		event: Twinkle.block.callback.toggle_ds_reason
 	};
+
+	// DS selection visible in either the template field set or preset,
```
NovemLinguae, line 569: Please change DS to VRT.

Sorry, tried to edit this myself since it's a small change, but my VS Code is acting up.

NovemLinguae (changes requested):
This is tested and good to go after the comment is fixed.

PR #2382, opened 2026-06-23
xfd: avoid overwriting reason when afd outcome is toggled

siddharthvp:
When AFD outcome dropdown is switched, reset the reason text only if it ends in "because", the same heuristic used when modifying the AFD target.

Switch from jQuery to plain JS usage.

### DIFF
--- a/src/modules/twinklexfd.js
+++ b/src/modules/twinklexfd.js
@@ -775,46 +775,52 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
 
 Twinkle.xfd.callbacks = {
 	/** If the user hasn't modified the reason much, modify the reason to include the target article. If the user has modified the reason box with a custom reason, do nothing, since we don't want to blank their work. */
-	changeAfdTarget: function() {
-		const $afdTarget = $('[name="afdtarget"]');
-		const $reason = $('[name="reason"]');
-		const $outcome = $('[name="outcome"]');
-		if ($reason.val().endsWith('because ')) {
+	changeAfdTarget: function(e) {
+		const form = e.target.form;
+		const afdTarget = form.afdtarget;
+		const reason = form.reason;
+		const outcome = form.outcome;
+		if (reason.value.endsWith('because ')) {
 			// Target has something typed in
-			if ($afdTarget.val()) {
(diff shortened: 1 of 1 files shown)


siddharthvp:
Was raised at https://en.wikipedia.org/wiki/Wikipedia_talk:Twinkle#Loss_of_data_when_switching_in_AfD

src/modules/twinklexfd.js
```diff
@@ -775,46 +775,52 @@ Twinkle.xfd.callback.change_category = function twinklexfdCallbackChangeCategory
...
+		const reason = form.reason;
+		const afdTarget = form.afdtarget;
+		afdTarget.value = '';
+		if (outcome.value === 'redirecting' || outcome.value === 'merging') {
+			$(afdTarget).parent().show();
+		} else {
+			$(afdTarget).parent().hide();
+		}
+		// Reset reason if the user hasn't modified it much
+		if (reason.value.endsWith('because ')) {
```
NovemLinguae: In my testing, doesn't appear to handle the blank case. That is, if a user opens the XFD -> AFD form, then selects Merging, with this patch loaded it currently fills in nothing, instead of filling in "I propose merging because". Looks like a regression that will lead to none of the reasons ever being loaded unless the user types "because " into the reason field before making their selection.

NovemLinguae (changes requested):

src/modules/twinklexfd.js
siddharthvp: Fixed

NovemLinguae: Same bug still appears to be present in my testing. As a recap, pre-patch the behavior is to display no default reason when switching to "delete" reason, and to display a default reason when switching to "merge", "redirect", "draftify". Currently this patch never displays anything, unless the user types "because " into the reason box, then switches.

siddharthvp: it works for me. Can you try in safe mode?

NovemLinguae: Like &safemode=1? Won't that disable Twinkle?

FaviFake:
In addition to the complaints at WT:TW, I've seen a couple users complain about this issue in their AfD nomination statements.

PR #2383, opened 2026-06-25
Fix alignment with Twinkle menu

jdlrobson:
We made some changes to Vector 2022's menu system so Twinkle's styles need to be updated to be compatible with it.

I already applied this on wiki but adding here for completeness https://en.wikipedia.org/w/index.php?title=MediaWiki%3AGadget-Twinkle.css&diff=1361137310&oldid=1339045356

### DIFF
--- a/src/twinkle.css
+++ b/src/twinkle.css
@@ -106,7 +106,7 @@ html {
 
 #p-twinkle-dropdown {
 	display: flex;
-    min-width: 40px;
+	min-width: 40px;
 }
 
 @media screen {
--- a/src/twinkle.css
+++ b/src/twinkle.css
@@ -104,6 +104,11 @@ html {
 	border: 1px solid #f28500;
 }
 
+#p-twinkle-dropdown {
+	display: flex;
+    min-width: 40px;
+}
+
 @media screen {
 	html.skin-theme-clientpref-night #twinkle-config-headerbox {
 		background: #532b18;


jdlrobson:
Here's the bug for context (without this CSS):

Screenshot 2026-06-25 at 3 43 32 PM

NovemLinguae:
Thank you for filing this as a PR so that your onwiki hotfix doesn't get overwritten. Good thinking!

src/twinkle.css
```diff
@@ -104,6 +104,11 @@ html {
 	border: 1px solid #f28500;
 }
 
+#p-twinkle-dropdown {
+	display: flex;
+	min-width: 40px;
+}
```
Emojigit, line 110:
```suggestion
.skin-vector-2022 #p-twinkle {
	display: flex;
	min-width: 40px;
}
```

1. Only Vector 2022 is affected;
2. The ID on Chinese Wikipedia does not have the `-dropdown` suffix.

Also: I am not sure why `min-width: 40px` is needed here. The layout works just fine without it, and we have been using Twinkle without this line.

Fix applied on-wiki: https://zh.wikipedia.org/w/index.php?title=MediaWiki:Gadget-Twinkle.css&diff=93217427

Emojigit: Oopsie, merged. Will open a new PR.

Emojigit: and I went to the wrong repo, sorry

PR #2391, opened 2026-07-22
welcome: change welcome-uk to welcome-ua

gomalley411:
change `welcome-uk` to `welcome-ua` per discussion on wiki: https://en.wikipedia.org/wiki/Wikipedia:Redirects_for_discussion/Log/2026_July_15#Template:Welcome-uk


src/modules/twinklewelcome.js
```diff
@@ -504,7 +504,7 @@ Twinkle.welcome.templates = {
 			},
 			'welcome-uk': {
```
siddharthvp, line 505: Shouldn't this be changed as well?

